PR #29, opened 2020-01-09
Fixed useWindowSize not being reactive

romansp:
Fixed `useWindowSize` not being reactive. I believe this mistake have sneaked during recent refactor.

Speaking of testing I couldn't come up with better option then spying on `window.addEventListener`. Let me know if you want that to be added to this PR.

```js
it('sets handler for window "resize" event', () => {
  const windowAddEventListener = jest.spyOn(window, 'addEventListener');

  renderHook(() => {
    const { width, height } = useWindowSize(100, 200)
    return { width, height }
  })

  expect(windowAddEventListener).toHaveBeenCalledWith('resize', expect.anything(), undefined)
})
```

### DIFF
--- a/packages/core/useWindowSize/index.test.ts
+++ b/packages/core/useWindowSize/index.test.ts
@@ -19,4 +19,15 @@ describe('useWindowSize', () => {
     expect(wrapper.vm.width).toBe(window.innerWidth)
     expect(wrapper.vm.height).toBe(window.innerHeight)
   })
+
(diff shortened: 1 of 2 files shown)


antfu:
I must be too careless. Thanks for the fix!

Testing for the browser API is pretty hard IMO. I think your idea is great, please go head for it. Thanks!

romansp:
PR updated - test added.

antfu:
Thank you!

PR #30, opened 2020-01-09
Added module Share [ Browser ]

joaoeudes7:
I added a module of Share in Core, I think so is useful.


packages/core/useShare/index.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+type ShareOptions = {
+  title?: string
+  files?: File[]
+  text?: string
+  url?: string
+}
+
+const _navigator = (window.navigator as NavigatorWithShare)
```
antfu: Please move this line inside the function.

packages/core/useShare/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export async function useShare<T>(shareOpts: ShareOptions) {
```
antfu: Look like the `<T>` is not used anywhere. Can you remove it?

packages/core/useShare/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (_navigator && _navigator.share) {
+    let granted = true
+
+    if (shareOpts.files && _navigator.canShare)
+      granted = _navigator.canShare({ files: shareOpts.files })
+
+    if (granted) {
+      _navigator.share(shareOpts)
+        .then(() => console.log('Successful share'))
+        .catch(err => console.log(err.message))
```
antfu: Using `console.log` seems not a good practice for a util function.

Maybe you can simply return the promise and let users to handle the callbacks
```js 
return _navigator.share(shareOpts)
```

packages/core/useShare/index.ts
joaoeudes7: I forgot that hehe thanks

Issue #60, opened 2020-03-21
Adding utility to test viewport visibility


scottbedard:
First off, awesome repo you have here!

Would you accept a utility for determining a component's position within the viewport? This sort of thing can be handy when trying to defer expensive operations like data fetching or complex rendering. Here is a brief example of what I am proposing...

```vue
<template>
  <div>
    {{ elementIsVisible ? "I'm visible in the viewport" : "You can't see me" }}
  </div>
</template>

<script>
export default {
  setup() {
    const { elementIsVisible } = useElementVisibility();

    return { elementIsVisible };
  },
};
</script>
```

If you're open to this util, I'll put together a PR. And just for fun, here is a more realistic example showing how this kind of thing could be used to do infinite-scroll pagination 🤷‍♂️

```js
import { useElementVisibility } from '@vueuse/core'

export default {
  setup() {
    const { elementIsVisible } = useElementVisibility();

    return { elementIsVisible };
  },
  watch: {
    elementIsVisible(visible) {
      if (visible) this.$emit('enter-viewport');
    },
  },
};
```

```vue
<template>
  <div>
    <Whatever
      v-for="thing in things"
      :thing="thing"
      :key="thing.id"
    />
    <InfiniteScroll @enter-viewport="fetchMoreArticles" />
  </div>
</template>
```

scottbedard:
Quick update, I was bored so I threw together a basic implementation. Let me know if there is anything you want to tweak. One thing that might be nice to add is a "threshold", so we can treat something as on-screen once it gets within 200px or whatever.

Also if you decide you don't want this util, just go ahead and close the issue/pr 👍

antfu:
It looks great, thank you for the proposal and implementation! I have left some comments in the PR, please have a look :)

For the threshold, I am not sure what use cases this utility would be. A basic one comes in my mind would be returning an `invisible | partial | visible` state. Another way with better controllability (but more complex and costly) might be returning an object with `visibleArea` `visbileWidth` `visibleHeight` `visableState` etc.

What do you think?

scottbedard:
My use case for a threshold is [this page](https://speedcube.site/records). I'd like to avoid rendering 3D content that is off the screen, but it might be nice to render it when it is soon to be on screen to avoid a flicker. This could be over-optimizing though, so why don't we keep it simple to start and return a boolean. If it turns out this is not enough, it'll be an easy option to add and won't be a breaking change.

[...]

PR #61, opened 2020-03-21
add basic implementaton of useElementVisibility

scottbedard:
Initial swing at #60


antfu:
Looks great! Thank you! Some thoughts though:

- Since this function returns basically a boolean state, would it better to return a single `ref` instead of an object? `element` is already provided in the parameters I didn't see any good reason to return it again.

meaning:
```ts
const elementIsVisible = useElementVisibility()

const orWhateverYouWant = useElementVisibility()
```

- It would be great if you can run the linter so the tests can be passed. But if not, I can do that after merging.

Thanks!

antfu:
LGTM, thanks!

Issue #62, opened 2020-03-21
State easing


scottbedard:
While I'm in here, would you mind if I added one more util to the library? This one would be for transitioning state along [cubic bezier curves](https://easings.net/). Below is an example of what I'm proposing...

```vue
<template>
  <div>
    <input v-model.number="baseNumber" /> {{ number }}
  </div>
</template>

<script>
export default {
  setup() {
    const baseNumber = ref(0)

    const number = useStateTransition(baseNumber, {
      duration: 1000,
      steps: 30, // probably a reasonable default, but allows for custom framerates
      transition: [0.77, 0, 0.175, 1] // maybe have aliases like 'easeInOutQuart'?
    })

    return { baseNumber, number }
  }
}
</script>
```

scottbedard:
And just for fun, an example of using the util to transition colors, and maybe some shorthand?

```vue
<template>
  <div>
    <input v-model.number="red" type="range" min="0" max="255" />
    <input v-model.number="green" type="range" min="0" max="255" />
    <input v-model.number="blue" type="range" min="0" max="255" />

    <h1 :style="{ color }">Hello world</h1>
  </div>
</template>

<script>
export default {
  setup() {
    const baseRed = ref(0)
    const baseGreen = ref(0)
    const baseBlue = ref(0)

    const [red, green, blue] = useStateTransition([baseRed, baseGreen, baseBlue], {
      duration: 1000,
      transition: 'easeInOutCubic'
    })

    const color = computed(() => `#${red}${green}${blue}`)

    return {
      baseRed,
      baseGreen,
      baseBlue,
      color
    }
  }
}
</script>
```

antfu:
This looks very useful, but my concern is that this can be over complicated to implement well and may surpass the goal of VueUse. Libraries like [tween.js](https://github.com/tweenjs/tween.js/) are much more mature and flexible to this, maybe we could add a wrapper addon for that?

scottbedard:
I don't think it's necessary to pull in additional libraries. I've thrown together a proof of concept that I can push up in just a few minutes, just cleaning some things up right now. If you look at it and decide you'd rather use a library like Tween, no worries, just go ahead and close the PR when I open it 🙂

scottbedard:
[...]

PR #63, opened 2020-03-22
initial swing at state easing

scottbedard:
#62

### DIFF
--- a/packages/core/useStateTransition/index.md
+++ b/packages/core/useStateTransition/index.md
@@ -1,6 +1,6 @@
 # useStateTransition
 
-Transition state between values.
+> Transition between values
 
 ## Usage
 
--- a/packages/core/useStateTransition/index.md
+++ b/packages/core/useStateTransition/index.md
@@ -1,0 +1,3 @@
+# useStateTransition
+
+Soon...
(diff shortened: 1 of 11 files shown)


antfu:
Wow! looks very promising. Looking forward for that, thank you!

packages/core/useStateTransition/index.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+          <p>Transitioned number: <b>{Math.round(this.number)}</b></p>
+        </div>
+        {Docs}
+      </div>
+    )
+  },
+})
+
+storiesOf('State', module)
+  .add('useStateTransition', () => Demo as any)
```
antfu: Maybe simply `useTransition` under `Animation` category would be more straightforward. What do you think?

packages/core/useStateTransition/index.md
```diff
@@ -0,0 +1,42 @@
+# useStateTransition
+
+Transition state between values.
```
antfu:
```suggestion
> Transition between values
```

packages/core/useStateTransition/index.stories.tsx
scottbedard: I like that better, I'll go ahead and make the. change 👍

antfu:
Great, thanks!

Issue #67, opened 2020-03-23
Page visibility util


scottbedard:
Hi, sorry I know I've been spamming this repo a bit lately, but I've got just one more util if you're open to it. I'd like to add a util to [detect if the window is active / minimized](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API). This can be useful for things like auto-pausing media, updating the document title, etc...

```js
// hidden and visibility would be reactive aliases of
// document.hidden and document.visibilityState

const { hidden, visibility } = usePageVisibility({
  onChange() {
    // ...
  },
  onHide() {
    // ...
  },
  onShow() {
    // ...
  }
});
```

antfu:
Contributions are always welcome! Thanks for great work.

Your proposal is definitely good to have, however, I think the callback functions are kinda unnecessary. Users can simply use `watch` to do the job and I think it is most of the functions in this lib are for.

```js
const visibility = usePageVisibility()

watch(visibility, () => { /*...*/ })
```

This may be a reference for your implementation: https://github.com/kripod/react-hooks/tree/master/packages/web-api-hooks#usedocumentvisibility

Thanks

scottbedard:
Ok I've pushed up the utility function, let me know if there is anything else you need from me 🙂

PR #68, opened 2020-03-24
useDocumentVisibility

scottbedard:
#67


antfu:
Great. I like how you present the demo :) Thanks

PR #71, opened 2020-03-30
New Hook: useWebWorkerFn

Donskelle:
Hey, i would like to contribute a port of https://github.com/alewin/useworker.

This is just a work in progress branch.

There is still some work to do:
- Implement more Demos
- Better Documentation
- Tests

[Preview Url](https://deploy-preview-71--vueuse.netlify.com/?path=/story/misc--usewebworkerfn)


packages/core/useWorkerWebFunction/index.ts
```diff
@@ -0,0 +1,112 @@
...
+      timeoutId.value = window.setTimeout(() => {
+        killWorker(WORKER_STATUS.TIMEOUT_EXPIRED)
+      }, timeout)
+    }
+    return newWorker
+  }
+
+  const callWorker = (...fnArgs: Parameters<T>) => new Promise<ReturnType<T>>((resolve, reject) => {
+    promise.value = {
+      [PROMISE_RESOLVE]: resolve,
```
antfu: Is there any particular reason for this approach?

packages/core/useWorkerWebFunction/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  onUnmounted(() => {
+    killWorker()
+  })
+
+  const generateWorker = () => {
+    const {
+      dependencies = DEFAULT_OPTIONS.dependencies,
+      timeout = DEFAULT_OPTIONS.timeout,
+    } = options
+    const blobUrl = createWorkerBlobUrl(fn, dependencies!)
```
antfu:
```suggestion
    const {
      dependencies = DEFAULT_OPTIONS.dependencies || [],
      timeout = DEFAULT_OPTIONS.timeout,
    } = options
    const blobUrl = createWorkerBlobUrl(fn, dependencies)
```

I would not suggest using null assertion, maybe this would be better.

packages/core/useWorkerWebFunction/lib/jobRunner.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param {Function} userFunc {Function} fn the function to run with web worker
+ *
+ * @returns {Function} returns a function that accepts the parameters
+ * to be passed to the "userFunc" function
+ */
+const jobRunner = (userFunc: Function) => (e: MessageEvent) => {
+  const userFuncArgs = e.data[0]
+
+  return Promise.resolve(userFunc(...userFuncArgs))
```
antfu:
```suggestion
  // eslint-disable-next-line prefer-spread
  return Promise.resolve(userFunc.apply(undefined, userFuncArgs))
```

I think this can solve the `_toConsumableArray is not defined` problem. I think this may vary in different babel settings, but for this, I think it's not necessary to use spread since it may have a higher chance to introduce errors.

packages/core/useWorkerWebFunction/lib/jobRunner.ts
```diff
@@ -0,0 +1,29 @@
...
+ * This function can be used as a wrapper, only inside a Worker
+ * because it depends by "postMessage".
+ *
+ * @param {Function} userFunc {Function} fn the function to run with web worker
+ *
+ * @returns {Function} returns a function that accepts the parameters
+ * to be passed to the "userFunc" function
+ */
+const jobRunner = (userFunc: Function) => (e: MessageEvent) => {
+  const userFuncArgs = e.data[0]
```
antfu: Should this be `e.data`?

packages/core/useWorkerWebFunction/index.ts
Donskelle: No, it was just the copy of the original source code. I dont see the reason as well. I rewrote it

packages/core/useWorkerWebFunction/lib/jobRunner.ts
Donskelle: Ya that fixed it. But this means, you need to write es5 code to be secure the code works fine at least in this setup.

This limits the ability of the hook. Maybe you also dont wanna add it, because it can lead to frustrating bugs, if you dont read the docu enough

Donskelle:
There is some 

> Thanks for the great work!
> 
> I did a quick review and please check it out :)
> 
> I am a little confused about the demo, I suppose it's just not finished yet.
> 
> And for this function, I think maybe we would need to add some unit tests in case it breaking in the future API changes.
> 
> Thanks again, indeed great util!

I think, it cant be tested using jest. There is an error
`TypeError: URL.createObjectURL is not a function`
because of [this issue](https://github.com/jsdom/jsdom/issues/1721). Its not supported in jsdom.

packages/core/useWorkerWebFunction/lib/jobRunner.ts
antfu: I think this may depend on the userland babel/ts config. There are tons of limitations in web worker such as dom access as well, I think adding a notice in the docs would be fine for this. (and maybe jsdoc as well, so people can see the note in there editor)

antfu:
OK let's forget about the tests for now. Maybe I could find some browser testing later.

About the demo, I have tried the demo of useWorker. Now I got the spinning it to demonstration UI blocking. I think it's good for react's logo but not ours. A simpler solution may displaying the current time(precise to ms), and when the clock stops, the UI blocking starts.

Sorry I am a little picky, maybe `useWebWorkerFn` would better aligned with other functions.

Thanks!

Donskelle:
OK i'll test and work on documentation in the next days.

I like cypress a lot. What do you have in mind?

antfu:
I have used cypress for my other projects, and it's awesome. I think it's a good idea. Go ahead for it if you would like to set it up, and it's ok if you don't. I can do that later.

And if you don't mind, I can push some commits directly, to adjust the docs and demo a bit.

Thanks :)

Donskelle:
Ofc you can. I things to do in the next days. 
Lets see if i have some free time at weekend

antfu:
I have updated the demo and docs. And the returning signature to `{ workerFn, workerStatus, workerTerminate }`.

Let me know what you think about it :)

Donskelle:
In React i see a lot of array return values from Hooks. Like [here](https://github.com/alewin/useWorker/blob/develop/src/useWorker.ts#L107). Its unpopular in Vue right? Can you explain why?

Because ref/reactive dont use arrays and useState does?

antfu:
I think it's just about preference. I am personally not a big fan of array spreading. You have to know the one you want on which index and give it a name. And something you don't want all of the returning values. you have to write like `const [value, _, _, something] = useWhatever()` which is kinda annoying.

Another iron fact is that I think Vue prefers object spreading: https://vue-composition-api-rfc.netlify.com/#code-organization

And yeah, let's get it merged. Thanks for your work!

PR #86, opened 2020-05-26
Add support for custom transition functions

scottbedard:
The other day I needed to get a bit fancier with state transitions, this PR allows for custom functions to be provided instead of just cubic bezier curves. I didn't add any presets, but now we're free to specify weird transitions like these 😅 

[easings](https://easings.net/#easeInOutElastic)


antfu:
Cool, thanks a lot! Can you help with the failing tests?

scottbedard:
@antfu Sorry about that! I didn't notice that there was a real error beyond just forgetting to run the linter. Let me know if there is anything else you need, and thanks again for this awesome library!

PR #98, opened 2020-07-19
fix: avoid warning - onMounted is called when...

songlairui:
> using createGlobalState and useStorage will cause a warning

image

` createGlobalState(() => useStorage(''))`   

--> 

`[Vue warn]: onMounted is called when there is no active component instance to be associated with. Lifecycle injection APIs can only be used during execution of setup(). If you are using async setup(), make sure to register lifecycle hooks before the first await statement.`

make the `useStorage`( in createGlobalState factory ) executed at the first time to use would avoid this warning.


antfu:
Thanks!

PR #105, opened 2020-08-01
feat: Add useReducer

Donskelle:
This Hook is inspired by Reacts buildin useReducer Hook. Its an easy way to centralize logic and use it in sub components without needing reference to the function.

The Name Reducer is a little misleading because its not a pure function to update state. Its mutating the reactive vue variable.

Maybe the name useActionState would be better? What do you think about this in general?

I had some issue with ts

### DIFF
--- a/packages/core/useReducer/ComplexDemo.tsx
+++ b/packages/core/useReducer/ComplexDemo.tsx
@@ -1,0 +1,152 @@
+import "vue-tsx-support/enable-check";
+import Vue from "vue";
+import { defineComponent, provide, inject } from "vue-demi";
+import { useReducer, Action } from ".";
+
+const initialState = {
+  loading: false,
+  data: [],
+};
+
+const fetchMock = () =>
+  new Promise((res, rej) => {
+    window.setTimeout(() => {
+      if (Math.random() > 0.5) {
+        res([1, 2, 3, 4, 5]);
+      } else {
+        rej("Request Error");
+      }
+    }, 1500);
+  });
+
+function reducer(state: any, action: any) {
+  switch (action.type) {
+    case "loadData":
+      state.error = undefined;
+      state.loading = true;
+      fetchMock()
+        .then((data) => (state.data = data))
+        .catch((e) => (state.error = e))
+        .finally(() => (state.loading = false));
+      return;
+    case "addItem":
+      state.data.push(action.payload);
+      return;
+    case "removeItem":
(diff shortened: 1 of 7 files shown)


antfu:
Thanks for the PR!

I am personally not a big fan of reducers as they make the types harder. But I am happy to have this included, as long as we can have a proper way to either auto infer the types or let users able to define the types.

I am fine with both `useReducer` and `useActionState`. If we go with `useReducer`, please add a line in docs addressing the difference.

BTW, can you run the linter to fix the style error? Thanks :)

Donskelle:
Ok, ya linting is no problem.
TS I need some days to research

packages/core/useReducer/index.ts
```diff
@@ -0,0 +1,16 @@
+import { reactive, UnwrapRef } from 'vue-demi'
+
+export interface Action {
+  type: string
+  [propName: string]: any
+}
```
Donskelle, line 6: Is this okay? Or do you think it should be more flexible?

packages/core/useReducer/ComplexDemo.tsx
```diff
@@ -0,0 +1,148 @@
...
+    data: {
+      type: Array,
+      default: () => [],
+    },
+  },
+  render(this: Vue) {
+    return (
+      <ul>
+        {this.$props.data.map((ele: number) => (
+          <ListItem id={ele} />
```
Donskelle, line 73: `##[error]packages/core/useReducer/ComplexDemo.tsx(73,21): error TS2769: No overload matches this call.`
I dont understand how i can solve this.

packages/core/useReducer/index.ts
antfu: Just checked react's approach https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bbc8c7c7b7e92ba094ea349d56977e00f6f7f42d/types/react/index.d.ts#L927-L1006
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bbc8c7c7b7e92ba094ea349d56977e00f6f7f42d/types/react/index.d.ts#L871-L880

Which I think you can refer to :)

Donskelle: Sorry, wasnt very motivated to do this in the last month and still havent figured out how i wanna do the reset. I'll close this for now, sorry

Donskelle: Enjoy holiday and thanks for your awesome open source work !

PR #133, opened 2020-09-26
feat: add asyncComputed

### LINKED ISSUE #128: Async Computed Values
I've been wanting something like [`vue-async-computed`](https://github.com/foxbenjaminfox/vue-async-computed) for Vue's composition API, but couldn't find anything so I created it myself. I made a gist with an explainer and an initial implementation [here](https://gist.github.com/loilo/fbe3124108a46ff50ab1b867bb5b4bf9).

Because of its generic nature and broad use cases, the `useAsyncComputed` function may be a nice counterpart to the existing `useAsyncState`.

What do you think?

loilo:
Okay, this took me quite a while. Mostly due to Storybook — I had to locally downgrade PostCSS to v7 because of an incompatibility between PostCSS 8 and Tailwind — but also due to tests. I haven't added any of those, even though I would've loved to, but I'd need refs for tests and `vue-demi` smacks me with "No vue dependency found" errors. Ideas?

(Resolves #128)

### DIFF
--- a/packages/core/asyncComputed/index.ts
+++ b/packages/core/asyncComputed/index.ts
@@ -1,0 +1,57 @@
+import { ref, watchEffect, Ref } from 'vue-demi'
+
+/**
+ * Handle overlapping async evaluations
+ *
+ * @param cancelCallback The provided callback is invoked when a re-evaluation of the computed value is triggered before the previous one finished
+ */
+export type AsyncComputedOnCancel = (cancelCallback: () => void) => void
+
+/**
+ * A two-item tuple with the first item being a ref to the computed value and the second item holding a boolean ref, indicating whether the async computed value is currently (re-)evaluated
+ */
+export type AsyncComputedResult<T> = [Ref<T>, Ref<boolean>]
+
+/**
+ * Create an asynchronous computed dependency
+ *
+ * @param evaluationCallback     The promise-returning callback which generates the computed value
+ * @param defaultValue           A default value, used until the first evaluation finishes
+ */
+export function asyncComputed<T>(
+  evaluationCallback: (onCancel: AsyncComputedOnCancel) => T | Promise<T>,
+  defaultValue?: T,
+): AsyncComputedResult<T> {
+  let counter = 0
+  const current = ref(defaultValue) as Ref<T>
+  const evaluating = ref<boolean>(false)
+
+  watchEffect(async(onInvalidate) => {
+    counter++
+    const counterAtBeginning = counter
+    let hasFinished = false
+
+    try {
+      // Defer initial setting of `evaluating` ref
+      // to avoid having it as a dependency
+      Promise.resolve().then(() => {
+        evaluating.value = true
+      })
+
+      const result = await evaluationCallback((cancelCallback) => {
+        onInvalidate(() => {
+          evaluating.value = false
+          if (!hasFinished) cancelCallback()
+        })
+      })
+
+      if (counterAtBeginning === counter) current.value = result
+    }
+    finally {
+      evaluating.value = false
+      hasFinished = true
+    }
+  })
+
+  return [current, evaluating]
+}
--- a/packages/core/asyncComputed/index.md
+++ b/packages/core/asyncComputed/index.md
@@ -1,0 +1,41 @@
+# asyncComputed
+
+> Like [computed refs](https://composition-api.vuejs.org/api.html#computed), but with support for promises.
+
+## Usage
+
+```jsx
+import { ref } from 'vue-demi'
+import { asyncComputed } from '@vueuse/core'
+
+export default {
+  template: `<p>
+    Package: <input v-model="packageName"><br>
+    Weekly Downloads: {{ isFetchingDownloads ? 'fetching...' : downloads }}
+  </p>`,
+  setup() {
+    const packageName = ref('vueuse')
+    const [downloads, isFetchingDownloads] = asyncComputed((onCancel) => {
+      const abortController = new AbortController()
+
+      onCancel(() => {
+        abortController.abort()
+      })
+
+      return fetch(`https://api.npmjs.org/downloads/point/last-week/${packageName.value}`, {
+        signal: abortController.signal,
+      })
+        .then(response => response.json())
+        .then(result => result.downloads)
+    })
+
+    return { packageName, downloads, isFetchingDownloads }
+  },
+}
+```
+
+## Caveats
+- Just like Vue's built-in `computed` function, `useAsyncComputed` does dependency tracking and is automatically re-evaluated when dependencies change.
+
+  Note however that only dependencies referenced in the first call stack are considered for this. In other words: Dependencies which are accessed asynchronously are not triggering re-evaluation of the async computed value.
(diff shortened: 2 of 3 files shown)


antfu:
Sorry, I forget to checkout storybook on the last deps bumps. Thanks for catching this, and fix in the latest commit.

About the testing, you can take a reference with https://github.com/antfu/vueuse/blob/914b6e056ea35ace93000d192434271ddf3502de/packages/core/createGlobalState/index.test.ts. As it's kinda straightforward, I am OK if you don't want to write the tests.

`No vue dependency found` should not happen, maybe try to reinstall the dependency?

Overall looks good to me. Thanks for working on this and the detailed jsdoc. One thing tho is that in VueUse, most of the functions prefer returning object for deconstructing instead of an array. Just feeling a bit inconsistent for users here, but as we discussed before, I think it's reasonable with the current design/implementation.

No future changes are required, the rest is up to you. Just let me know whether you still want to make changes or I can merge it now. Thanks a lot!

packages/core/asyncComputed/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  const current = ref(defaultValue) as Ref<T>
+  const evaluating = ref<boolean>(false)
+
+  watchEffect(async(onInvalidate) => {
+    counter++
+    const counterAtBeginning = counter
+    let hasFinished = false
+
+    try {
+      // Defer initial setting of `evaluating` ref
```
Demivan, line 35: Consider adding parameter `reset` (maybe different name) and adding:
```js
if (reset) {
  current.value = defaultValue
}
```
It would be useful for uses with `vue-router` to reset page to default loading state when page props change. Instead of page content staying the same until new data is loaded.

Demivan:
Why not make it return object with refs, same as `useAsyncState`?
Implementation would be exactly the same just different return.

loilo:
@Demivan My main objection against the object-returning approach is that it complicates naming which usually makes code less legible.

Assuming `asyncComputed` returns a `{ state, ready }` object (just like `useAsyncState`), there's essentially four ways to consume the result, and I dislike each of them in a certain way:

---

**1. No destructuring**
```js
const myResult = asyncComputed(/*...*/)
```
**Objection:** Having to always access the state via `myResult.state` is rather inconvenient, which is even more annoying when all other properties on the `myResult` object can be ignored (`myResult.ready` should be needed way less often).

---

**2. No destructuring, properly named variables**
```js
const myResult = asyncComputed(/*...*/)
const myState = myResult.state
```
**Objection:** We create a barely used "ghost variable" (`myResult`) and need multiple lines for something that should be doable in one.

---

**3. Naive destructuring**
```js
const { state } = asyncComputed(/*...*/)
```
**Objection:** `state` (as well as `ready`) is a terrible variable name in a real-world scenario since it's not descriptive at all.

---

**4. Elaborate destructuring**
```js
const { state: myState } = asyncComputed(/*...*/)
```
**Objection:** While this is closest to being acceptable, renaming in object destructuring always is a last resort for me (e.g. to avoid naming conflicts) because it creates a lot of noise and thus hurts readability.

This becomes particularly obvious when using multiple async computed values:
```js
const { state: myState } = asyncComputed(/*...*/)
const { state: anotherState } = asyncComputed(/*...*/)
const { state: moreState } = asyncComputed(/*...*/)
```

---

This kind of annoyance can not always be avoided, but IMO array destructuring is a good alternative when there is a clear order of preference for the resulting values (i.e. there is virtually no imaginable case where a user wants the `ready` value but not the `state`).

In this setting, array destructuring allows for a very un-noisy way for the most popular use case:
```js
const [myState] = asyncComputed(/*...*/)
```

packages/core/asyncComputed/index.ts
loilo: I considered this use case, but decided against it because [boolean parameters are undesirable](https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/) and this case can dealt with pretty easily through an additional `computed`:

```js
function resettableAsyncComputed(evaluationCallback, defaultValue) {
  const [result, evaluating] = asyncComptued(evaluationCallback, defaultValue)
  return computed(() => evaluating.value ? defaultValue : result.value)
}
```

Demivan: Thanks for explanation. I really like this solution. This also allows for initial value and evaluating value to be different.

Demivan:
My issue with this approach is that it is different from `useAsyncState`. It is really similar to `asyncComputed` as users will mostly just use `state` and using `ready` without `state` does not make sense.
@antfu Maybe `useAsyncState` can be updated to match `asyncComputed`?

packages/core/asyncComputed/index.ts
Demivan: Maybe this example could be added to docs? IMO this is pretty common use case.

Demivan:
What about supporting both approaches?
```js
const result = [state, ready]
result.state = state
result.ready = ready
```
Then you can destructure it as an array or as an object:
```js
const [state, ready] = result
const { state, ready } = result
```

antfu:
TBH, I am not a fan of array destructuring. The main reason is that it uses "magical indexes". From the user side, you get no idea of which is for what without looking into docs. Especially when you are returning more than two values. For example [this function](https://vueuse.js.org/?path=/story/sensors--usemouseinelement). It would also be terrible when you only want to use only some of them you would need to use `const [,,, elementX, elementY] = useXXX()`.

Object destructuring on the other hand, always explicitly named things which could be self-explaining. Also, it's flexible to be used as your 1234 examples by users' preferences. Using arrays means you would always need to destructure and thinking names by yourself, or `myResult[0]` against `myResult.state`.

I just said in the previous comment, it would be more consistent with other functions if you would change to object returns. But as I am not strongly against the current implementation in our discussion phrase, I think it does have some pros with returning array, changes are up to you and I am not against the current implementation.

And yes as @Demivan said, it's possible to support both at the same time and I would love to! I am actually doing experiments on it and writing [a blog post](https://antfu.me/posts/destructuring-with-object-or-array/) (wip, not yet published) with some of my thoughts and solutions. I can do that later after this get merged, or feel free to try if you want.

The implementation @Demivan provides comes with a problem that

```ts
const { state, ...rest } = result

console.log(rest) // { 0: state, 1: ready, ready: ready }
```

Thanks again for working on it! I will keep you informed after I finished my post.

loilo:
Maybe I need to clarify my point: *Of course* destructuring large arrays is a terrible idea. Destructuring arrays can be a good idea in only a very narrow niche, which is two items with the first one always and te second one occasionally being needed.
I don't really share your point about users having to name things themselves (because I think that burden *should* be on users), but apart from that I can definitely relate to your objections against destructuring arrays @antfu. I don't particularly like it either, but in some situations I consider it the lesser evil.

Straight from the gut I cannot tell what my feelings are about mixing both formats. It might be beneficial from a dev experience point of view, but I generally dislike offering two ways to do a thing. Not sure if it's worth it.

antfu (approved):
LGTM, I will think about a proper way to support both ways of destructuring later on if it's possible.

loilo:
Okay. As far as I'm concerned, this can be merged. I would've loved to write some unit tests for the function, but the error message still gets the better of me. 🤷‍♂️

antfu:
Thanks. I will improve the tests soon.

PR #135, opened 2020-09-28
fix: rollup typescript config overwrite

aa900031:
1. Change usage of `rollup-plugin-typescript2` opts.tsconfigOverride
2. Remove unnecessary `tsconfig.rollup.json`


antfu:
Thanks

PR #141, opened 2020-10-07
refactor!: useRefHistory api improves

### LINKED ISSUE #137: RFC: possible useRefHistory API improvements
Hey @antfu! Thanks for all the work with the lib ❤️ 

Great to see you added a composable for ref history! We [had to build one](https://github.com/Leniolabs/layoutit-grid/blob/main/src/composables/useRefHistory.js) because the `useUndo` from vue-composable was not flexible enough for our use case. I was planning to submit a proposal for a similar one at one point. The API is not the same though, so I may release it as a separate lib since it has different tradeoffs. There are some API tweaks that could be useful in the current on in vueuse. If you like any of these changes, I could create PRs for them.

1. Add option to avoid `flush: 'sync'`
I think the default should be for it not to be 'sync'. We had several actions that were generating spurious history points because of this. An example could be to in a handler to change a size that is defined as {value,unit}, and you need to touch both. We could do this using pause/resume but it will be quite error prone in some apps.
2. Rename `history, redoHistory` to `prev, next`
This is the way it is in vue-composable `useUndo`. I think they are more in line with the rest of the API (undo, redo, clear) and it reads better when you use an object for the composable return:
```
const history = useRefHistory(aRef)
if( history.prev.length > 0 ) { ... } // instead of history.history
if( history.next.length > 0 ) { ... } // instead of history.redoHistory
```
3. Expose `batch(), commit()` functions, that complements `pause(), resume()` (it can be discussed if we should have both pairs after adding batch and commit, IMO they both have different use cases)
When doing real time changes, like dragging or playing with a slider, we want only one history point for the whole interaction. The code for this would be
```
history.batch()
// change state, across several user events (pointermove for example)
```
and when the interaction ends
```
history.commit()
```
If we want to do the same with `pause(), resume()`, we would need to add another set to the ref to trigger the history point after it is resumed. 
I also think that `batch(), commit()` is better suited for this cases because we can add a `cancel()` function, that will go back to the last history point, finish the batch and resume history (it can also be called `revert()` if we want to follow a git analogy).
`commit()` can also let us build a manual history composable, with the same API, where the trigger is not automatic but the user is the one in charge of deciding when there is a history point. It also serves as a way to add a sync history point in case the user decides not to use `flush: 'sync'` but still needs it in a particular case.

I think the other differences with what I had in mind are more about extra features and configuration, and it is not clear to me that the gained flexibility will be enough to justify a more complex API in vueuse. This is why I think there could be space for another lib, with different tradeoffs (more options, but more time to learn it, and more verbose config). But the above proposals could be good for vueuse, and later if people migrate from vue-composable or to other libs, they would have a closer API.

antfu:
close #137 

Changed `next` and `prev` to `undoStack` and `redoStack`, also providing an alias `history`

### DIFF
--- a/packages/core/useRefHistory/index.md
+++ b/packages/core/useRefHistory/index.md
@@ -9,7 +9,7 @@ import { ref } from 'vue'
 import { useRefHistory } from '@vueuse/core'
 
 const counter = ref(0)
-const { prev, undo, redo } = useRefHistory(counter, {
+const { history, undo, redo } = useRefHistory(counter, {
   deep: false,
   clone: false,
   capacity: 15, // limit to n history records, default to unlimited
@@ -18,7 +18,7 @@ const { prev, undo, redo } = useRefHistory(counter, {
 counter.value += 1
 counter.value = 10
 
-console.log(prev) // [{ value: 10, timestamp: 1601912898062 }, { value: 1, timestamp: 1601912898061 }]
+console.log(history) // [{ value: 10, timestamp: 1601912898062 }, { value: 1, timestamp: 1601912898061 }]
 
 console.log(counter.value) // 10
 undo()
--- a/packages/core/useRefHistory/index.md
+++ b/packages/core/useRefHistory/index.md
@@ -23,7 +23,7 @@ console.log(counter.value) // 1
 
 ### Objects / arrays
 
-When working with objects or arrays, since changing their attributes does not change it's reference, all the value in the history will point to a same reference. If you want to keep the state changes, you would need to pass `clone: true`. It will create clones for each history records.
+When working with objects or arrays, since changing their attributes does not change the reference, it will not trigger the committing. To track attribute changes, you would need to pass `deep: true`. It will create clones for each history record.
 
 ```ts
 const state = ref({
@@ -33,7 +33,6 @@ const state = ref({
 
 const { history, undo, redo } = useRefHistory(state, {
   deep: true,
-  clone: true,
 })
 
 state.value.foo = 2
@@ -45,12 +44,35 @@ console.log(history.value)
 ] */
 ```
 
-### History Capcity
+#### Custom Clone Function
+
+`useRefHistory` only embeds the minimal clone function `x => JSON.parse(JSON.stringify(x))`. To use a full featured or custom clone function, you can set up via the `dump` options.
+
+For example, using [lodash's `cloneDeep`](https://lodash.com/docs/4.17.15#cloneDeep):
+
+```ts
+import { cloneDeep } from 'lodash-es'
+import { useRefHistory } from '@vueuse/core'
+
+const refHistory = useRefHistory(target, { dump: cloneDeep })
+```
+
+Or a more lightweight [`klona`](https://github.com/lukeed/klona):
+
+```ts
+import { klona } from 'klona'
+import { useRefHistory } from '@vueuse/core'
+
+const refHistory = useRefHistory(target, { dump: klona })
+```
+
+
+### History Capacity
 
 We will keep all the history by default (unlimited) until you explicitly clean them up, you can set the maximal amount of history to be kept by `capacity` options.
 
 ```ts
-const refHistory = useRefHistory(ref, {
+const refHistory = useRefHistory(target, {
   capacity: 15 // limit to 15 history records
 })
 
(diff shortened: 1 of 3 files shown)


antfu:
@matias-capeletto I have made the changes addressing our previous discussions. Can you help to review a bit to see if it meets your usages? Thanks.

matias-capeletto:
Looks good @antfu, the only thing that I would change is the default for array and objects. I think it will cause a lot of confusion because people will forget to pass `{ clone: true }` and then the history will not be maintained. 
What about checking the type of the reference and have good defaults for?
- If there is a custom dump and parse, that is what it is used 
- else 
  - If it is a primitive, you can bypass
  - else, you clone using JSON.stringify+JSON.parse

For me, using { clone: false } should be an implementation detail if possible and not something that the user needs to think about. Maybe the default could also be an implementation similar https://lodash.com/docs/4.17.15#cloneDeep, that takes into account primitives types, and is also more efficient because doesn't stringify but directly traverse the object cloning the nested values.

antfu:
That's a fair point. It seems no strong usages for using objects without clone set to true.

However, I feel varing default value can be confusing and for users using javascript might use refs mixing objects and primitives, this behaviour could be tricky. So I feel explicitly setting would be better and clear.

I do agree clone could be easy forgot, so I am think - about we removing the clone option and combine it with in `deep`? Currently, deep only present "watch deeply" which is carried from the internal watch's concept, but we can change to to "track deep changes" to be the concept of useRefHistory and will enable clone automatically.

What do you think?

antfu:
There are a lot different cloneDeep covering different eage cases. But as VueUse is trying to be dependencies free, it's not ideal to include them but offering users to custom their clone function via dump options.

Probably good to include an example in the docs.

matias-capeletto:
> about we removing the clone option and combine it with in deep?

I think this is a really good idea. And it will be easy to explain if you can match how watch works. An looking from that point of view, I now see that there may be at the end usages for avoiding clone. An app could have immutable arrays or objects to track that are only modified by assignment, and `deep: false` makes sense in that case. But yes, what was confusing is using `{ deep: true, clone: false }` together.
I also like it because we do not need to pick what should be the default, as it should follow what watch does by default. So deep false is the default, and deep true means deep tracking (and deep cloning to make history useful).

> There are a lot different cloneDeep covering different eage cases. But as VueUse is trying to be dependencies free, it's not ideal to include them but offering users to custom their clone function via dump options.

Totally agree, I was thinking about adding a cloneDeep utility in the lib, not adding a dependency. I think this may not be the last time you would need it. I understand that to simplify you may want to use JSON.stringify+parse as the current implementation of this utility. The only issue with it is that it may be surprising for users of Vue 3. Now that we have support for Map, Set, WeakMap and WeakSet in the reactivity API, it would be great if the default cloneDeep implementation would also supports them. Maybe this can be added later though, or as some kind of config option to avoid paying for the extra 0.5kb of this function if you do not use Map et al in your apps. 

In case you want to add an utility like this, I think klona is a good place to grab a good implementation.
Here is the version that supports everything we need: https://unpkg.com/klona@2.0.4/dist/index.mjs @ 451 bytes

antfu:
klone looks good! Since we are tree-shakeable, including some utils should be fine for users to opt-in. For example:

```ts
import { useRefHistory, cloneDeep } from '@vueuse/core'

useRefHistory(r, { dump: cloneDeep })
```

This does add some convenience for developers. However, after a second thought, I doubt if we should port those libraries into ours. Mainly for maintaining sake, if we port them, we would need to either maintain those ourselves or keep the code up to date to the upstream. Neither sounds good to me. This also obeys the spirit of having modules to be reused.

I would think it's better to guide users about such usages like

```ts
import { klona } from 'klona'
import { useRefHistory } from '@vueuse/core'

useRefHistory(r, { dump: klona })
```

```ts
import cloneDeep from 'clone-deep'
import { useRefHistory } from '@vueuse/core'

useRefHistory(r, { dump: cloneDeep })
```

matias-capeletto:
Totally agree. I only suggested it in case the default deepClone implementation would be changed to support Map et al. If the user still needs to opt in when using the composables, then it is better to let the user decide which lib it wants to use.
I think with time there will be several cases like this one in vueuse, where you need a general utility but you do not want to decide for the user because of the tradeoffs (debounce may be another example). It would be great if these general utilities could be configured for all the composables in vueuse. So people can still have the clean API, and decide globally for the app what they want to use. Probably not a discussion for this PR, though. 

LGTM, I'll try it out once is published in our app 🙌

antfu:
I am merging this and preparing for the release now. Thanks a lot for the feedback!

Issue #148, opened 2020-10-15
feat: add useCookies with reactive methods


crutch12:
Just like https://github.com/reactivestack/cookies/tree/master/packages/react-cookie but in Vue style

I already wrote a composition-api wrapper for [universal-cookie module](https://github.com/reactivestack/cookies/tree/master/packages/universal-cookie) in my work project and want to add pull request to vueuse with this wrapper

But it would require `universal-cookie ` as peerDependency

@antfu what do you think about third party library?

antfu:
While I stil what to keep the main library dependencies free, but I do understand something it's better to introduce some dependencies that can handle the jobs better. So I guess we can have another package, maybe `@vueuse/libs` or `@vueuse/integrations` to put popular packages wrapper that users can opt-in. What do you think? If it's good, I can make the setup so you can better contribute. Thanks!

crutch12:
@antfu 
I think this naming won't be intuitive for other developers, who, for example look for in google: "vue composition-api cookies".
I think it should be separated package like a `@vueuse/cookie` or `@vueuse/universal-cookie`

But I checked `@vueuse/firebase` and `@vueuse/rxjs` and there are no package.json file (where we could add peerDependencies) and it confuses me

crutch12:
Is it okay if I will provide PR for `@vueuse/cookie`?

antfu:
Well, it's probably okay. But my main concern is that cookie is kinda a small utility, making it a separate package meaning we are likely to have a lot of @vueuse/xxx for just simple wrappers. My idea of `@vueuse/integrations` is like a collection of libraries wrappers (with optional deps and tree-shaking) and add some guide to install corresponding libraries (`universal-cookie` in this case) if they would need some of the wrappers. Or think it of `@vueuse/core` but with deps version.

crutch12:
Ok, let's try `@vueuse/integrations`, could you setup it?

antfu:
Sure

antfu:
Done. And I think I should switch to monorepo instead of this magic things, I will do that later :p

crutch12:
solved by https://github.com/antfu/vueuse/pull/150

PR #150, opened 2020-10-18
feat: add useCookies (#148)

crutch12:
solves #148


packages/integrations/useCookies/index.ts
```diff
@@ -0,0 +1,103 @@
+import { ref, getCurrentInstance, watchEffect } from 'vue-demi'
+import Cookie from 'universal-cookie'
+import type { IncomingMessage } from 'http'
+
+type RawCookies = {
+  [name: string]: string
+}
+
+let universalCookie: Cookie | undefined
```
antfu, line 9: I am not sure if we should keep the global variable. I think it's better to hand over the controls to users. They can use the hooks globally to have a global instance or use in components locally which will be disposed when the component unmounted.

packages/integrations/useCookies/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Adds reactivity to returning methods
+   */
+  const touches = ref(0)
+
+  /**
+   * Calls inside setup() function only
+   */
+  if (getCurrentInstance()) {
```
antfu, line 40: I think it's ok to also be called outside of `setup()`

packages/integrations/useCookies/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  return universalCookie
+}
+
+/**
+ * Reactive methods to work with cookies (firstly call {@link createCookies} method if using SSR)
+ * @param {string[]|null|undefined} dependencies - array of watching cookie's names. Pass empty array if don't want to watch cookies changes.
+ * @param {boolean} [doNotParse=false] - don't try parse value as JSON
+ */
+export const useCookies = (dependencies?: string[] | null, doNotParse = false) => {
```
antfu, line 27: For exported functions, I would prefer `function` over arrow functions.

packages/integrations/useCookies/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    watchEffect(() => {
+      cookies.addChangeListener(onChange)
+
+      return () => {
+        cookies.removeChangeListener(onChange)
+      }
+    })
+  }
+
+  return {
```
antfu, line 65: Would it be nice to also export a reactive cookies state?

Thinking:

```ts
const cookies = computed<T>(() => {
  touches.value
  return _cookies.getAll<T>({ doNotParse })
})
```

antfu (commented):
Thanks a lot and it's looking great! However, I get a few comments and would love to know what you think.

packages/integrations/useCookies/index.ts
crutch12: I think (1)
```vue
{{ cookies.get('locale') }}
```
is better than (2)
```vue
{{ cookies.getAll().locale }}
or
{{ cookies.value.locale }}
```

Because honestly you will never use `cookies.getAll()` inside your application (only for debugging purposes)
And (1) example ... em ... reminds (?) you don't forget provide dependencies array :)

I also think about automatically adding new cookie names to dependencies when someone tried to get its value. With your example it won't be possible.

packages/integrations/useCookies/index.ts
crutch12: For example use global `universalCookie` if it was provided before

packages/integrations/useCookies/index.ts
```diff
@@ -0,0 +1,103 @@
...
+          previousCookies,
+        )
+      )
+        touches.value++
+
+      previousCookies = newCookies
+    }
+
+    watchEffect(() => {
+      cookies.addChangeListener(onChange)
```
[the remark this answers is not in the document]
crutch12: @antfu

packages/integrations/useCookies/index.ts
antfu: You can use `tryOnMounted` and `tryOnUnmounted` in utils instead. It should be fine to initialize the hook outside setup() and be reused across multiple components.

packages/integrations/useCookies/index.ts
antfu: Mostly for tree-shaking sake. Variables without `/*#__PURE__*/` will be treat as side effects and not tree-shakable by default. Although bundles should be intelligent enough to tell it's an arrow function, using the `function` keyword feels more explicit to me.

packages/integrations/useCookies/index.ts
antfu: I would prefer those utility functions to be side effects free if possible. Global instance sharing is implicit to users which might cause some confusion.

`createCookies` should not rely on the global side effects as well I think. I am thinking something like this could be better.

#### User defined global state

```ts
// global-utilties.ts
export const cookies = useCookies()

// xxx.vue
import { cookies } from './global-utilties'
```

#### Basic use in components

```ts
setup() {
  const cookies = useCookies() // should always create new UniversalCookie instance
  return { cookies }
}
```

#### Shared instances

`createCookies` returns a new `useCookies` hook that shares the instance.

```ts
export const useMyCookies = createCookies()

// xxx.vue
import { useMyCookies } from './utils'

{
  setup() {
    const cookies = useMyCookies() // should always create new UniversalCookie instance
    return { cookies }
  }
}
```

packages/integrations/useCookies/index.ts
crutch12: Could you add a use case for this? `onMounted` and `onUnmounted` will work only in `setup()`, so why do I need call these methods in other places? I don't understand

crutch12: Ok, I imagined one use case:
```ts
const cookies = useCookies();

watch(() => cookies.get('locale'), () => {
  alert('locale changed', cookies.get('locale'))
})
```
But you can't unsubscribe from `universalCookie` this way, because onMounted/onUnmounted won't work here.

So I would prefer this way:
```ts
const cookies = useCookies()
const locale = cookies.get('locale')

const onChange = () => {
  if (locale !== cookies.get('locale')) {
    locale = cookies.get('locale')
    alert('locale changed', locale)
  }
}

cookies.addChangeListener(onChange)
```

Or return additional method `unmount/deactivate` (for example) from `useCookies()`

antfu: https://github.com/antfu/vueuse/blob/d13172237d42981a1ccfe760fa7d0c969152c182/packages/core/useEventListener/index.ts#L12-L18

When you use it globally, most of the time you don't need to dispose them.

One other use case is to share the state globally. You can see this comment: https://github.com/antfu/vueuse/pull/150#discussion_r508320224

antfu:
Great, thanks a lot!

PR #153, opened 2020-10-21
fix: useIntersectionObserver stopping

sp1ker:
- Fix `stoping` function 
- Add cases for story of `useIntersectionObserver`
- Add description of `stopping` function


antfu:
Thanks!

PR #155, opened 2020-10-23
feat(useAxios): added request cancelation

lstoeferle:
Hey, it would be nice to have axios request cancelation out of the box.


antfu:
Thanks

PR #158, opened 2020-11-01
feat: new function - useFavicon

anteriovieira:


packages/core/useFavicon/index.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import 'vue-tsx-support/enable-check'
+import Vue from 'vue'
+import { storiesOf } from '@storybook/vue'
+import { defineComponent } from 'vue-demi'
+import { ShowDocs } from '../../_docs/showdocs'
+
+const Demo = defineComponent({
```
antfu, line 7: You can use the new style to define the stories. Refer to `packages/core/_template/index.stories.tsx`

antfu:
This is great. Thanks! However, I am thinking the API could be more Vue / reactive-ness. Something like (supporting the two usages at the same time):

```js
const icon = useFavicon()

icon.value = 'dark.ico'
```

```js
const favicon = computed(() => isDark.value ? 'dark.ico' : 'light.icon')

useFavicon(favicon, { ...options })

// const icon = useFavicon(favicon)
// `icon` will be same as `favicon`
```

What do you think?

anteriovieira:
I liked the idea, I will try this approach

antfu:
This implementation might do some help
https://github.com/antfu/vueuse/blob/master/packages/core/useTitle/index.ts

anteriovieira:
Thanks!

packages/core/useFavicon/index.ts
```diff
@@ -0,0 +1,29 @@
+import { isString } from '@vueuse/shared'
+import { ref, watch, Ref, ComputedRef, isRef } from 'vue-demi'
+
+export function useFavicon(
+  newIcon: Ref<string> | ComputedRef<string> | string | null = null,
+  options: { baseUrl?: string; rel?: string } = {},
+) {
+  const {
+    baseUrl = '/',
```
antfu, line 9:
```suggestion
    baseUrl = '',
```

Maybe it's better to be empty by default. Users can set `icon` to `/favicon.ico` more naturally as they will do in html.

packages/core/useFavicon/index.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+or
+
+```jsx
+import { computed } from 'vue'
+import { useFavicon, usePreferredDark } from '@vueuse/core'
+
+const isDark = usePreferredDark()
+const favicon = computed(() => isDark.value ? 'dark.ico' : 'light.icon')
```
antfu, line 22:
```suggestion
const favicon = computed(() => isDark.value ? 'dark.ico' : 'light.ico')
```

packages/core/useFavicon/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+```jsx
+import { useFavicon } from '@vueuse/core'
+
+const icon = useFavicon()
+
+icon.value = 'dark.ico' // change current icon
+```
+
+or
```
antfu, line 15:
```suggestion
or passing a `ref` to it
```

packages/core/useFavicon/index.md
```diff
@@ -0,0 +1,28 @@
+# useFavicon
+
+> Changes the current favicon
```
antfu, line 3:
```suggestion
> Reactive favicon
```

antfu:
Thanks a lot!

PR #160, opened 2020-11-02
feat: new function - useQRCode

anteriovieira:


antfu:
Looking good! One little thing is that it's better to be named as `useQRCode`.

> PS. updating the readme link is not required, I will handle that. If you want to do it, don't do it by hand, just `yarn prepare` ;)

antfu:
Never mind, I will do that! Thanks!

antfu:
Changes:
- Renamed to `useQRCode`
- Accept a second arg for options passed to QRCode

Released in `beta.32`.

Thanks!

PR #168, opened 2020-11-04
chore(useDeviceOrientation): adds event target

anteriovieira:
@antfu what do you think about it?

image

the demo currently does not display the values because it does not have context access

### DIFF
--- a/packages/core/useDeviceOrientation/index.ts
+++ b/packages/core/useDeviceOrientation/index.ts
@@ -4,7 +4,7 @@ import { isClient } from '@vueuse/shared'
 import { ref, Ref } from 'vue-demi'
 import { useEventListener } from '../useEventListener'
 
-export function useDeviceOrientation(target: Window = window) {
+export function useDeviceOrientation(target: Window | null = isClient ? window : null) {
   const eventTarget = isClient ? target : null
   const isAbsolute = ref(false)
   const alpha: Ref<number | null> = ref(0)
--- a/packages/core/useDeviceOrientation/index.ts
+++ b/packages/core/useDeviceOrientation/index.ts
@@ -1,20 +1,24 @@
 /* this implementation is original ported from https://github.com/logaretm/vue-use-web by Abdelrahman Awad */
 
+import { isClient } from '@vueuse/shared'
 import { ref, Ref } from 'vue-demi'
 import { useEventListener } from '../useEventListener'
 
-export function useDeviceOrientation() {
+export function useDeviceOrientation(target: Window = window) {
+  const eventTarget = isClient ? target : null
   const isAbsolute = ref(false)
   const alpha: Ref<number | null> = ref(0)
   const beta: Ref<number | null> = ref(0)
   const gamma: Ref<number | null> = ref(0)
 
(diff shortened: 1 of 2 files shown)


packages/core/useDeviceOrientation/index.ts
```diff
@@ -1,20 +1,24 @@
 /* this implementation is original ported from https://github.com/logaretm/vue-use-web by Abdelrahman Awad */
 
+import { isClient } from '@vueuse/shared'
 import { ref, Ref } from 'vue-demi'
 import { useEventListener } from '../useEventListener'
 
-export function useDeviceOrientation() {
+export function useDeviceOrientation(target: Window = window) {
```
antfu, line 7:
```suggestion
export function useDeviceOrientation(target: Window = isClient ? window : null) {
```

SSR will only work this way

packages/core/useDeviceOrientation/index.ts
```diff
@@ -1,20 +1,24 @@
 /* this implementation is original ported from https://github.com/logaretm/vue-use-web by Abdelrahman Awad */
 
+import { isClient } from '@vueuse/shared'
 import { ref, Ref } from 'vue-demi'
 import { useEventListener } from '../useEventListener'
 
-export function useDeviceOrientation() {
+export function useDeviceOrientation(target: Window | null = isClient ? window : null) {
+  const eventTarget = isClient ? target : null
```
antfu, line 8: Let's just use target

anteriovieira: I was analyzing here, we have many places that would need this implementation. Maybe not the best way. And I will think about it better.

anteriovieira: For example: `useDeviceMotion`, `useDeviceLIght`

antfu: Yeah, it's better to have an option for it. Will also be more extensible for us to add more options in the future.

I think we can make a shared interface with shared js-docs (providing more details/explication for users) like

```ts
export interface ConfigurableWindow {
  /*
   * Specify a custom `window` instance, e.g. working with iframes or in testing envroiments.
   * ...(more)
   */
  window?: Window
}

export interface ConfigurableDocument {
  /*
   * Specify a custom `document` instance, e.g. working with iframes or in testing envroiments.
   * ...(more)
   */
  document?: Document
}
```

```ts
interface DeviceMotionOptions extends ConfigurableWindow {
  throttleMs: number
}

export function useDeviceMotion(options: DeviceMotionOptions = { throttleMs: 10 }) {
  // options.window
  /* ... */
}
```

antfu: `navigator`, `location` as well. Any global reference in functions should be configurable then I think.

anteriovieira: I agree with you, I was looking for a solution as a global configuration, but I think your idea will give more flexibility for the use of the functions.

antfu: I will do the refactor 👍

PR #174, opened 2020-11-05
refactor!(useWebWorkerFn): export WebWorkerStatus as type instead of enum

### LINKED ISSUE #172: TypeScript error with --isolatedModules flag
The [TypeScript compiler option](https://www.typescriptlang.org/docs/handbook/compiler-options.html) `--isolatedModules` reports an error:
```
node_modules/@vueuse/core/dist/index.d.ts:492:278 - error TS2748: Cannot access ambient const enums when the '--isolatedModules' flag is provided.
```
It would be nice if vueuse would not prevent the use of this compiler option.

neelance:
This is to allow use of the TypeScript compiler option --isolatedModules.

Fixes #172

### DIFF
--- a/packages/core/useWebWorkerFn/index.md
+++ b/packages/core/useWebWorkerFn/index.md
@@ -23,9 +23,9 @@ import { useWebWorkerFn } from '@vueuse/core'
 const fn = dates => dates.sort(dateFns.compareAsc)
 
 const { workerFn, workerStatus, workerTerminate } = useWebWorkerFn(fn, {
-  timeout: 50000, // 5 seconds
+  timeout: 5_000, // 5 seconds
   dependencies: [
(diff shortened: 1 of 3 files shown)


antfu:
The usage of `const enum` is because `enum` generated code wouldn't be tree-shakeable. Refer https://github.com/microsoft/TypeScript/issues/27604

I don't think it's a good trade-off for this PR.

Maybe we can simply do this:

```ts
export type WorkerStatus = 'PENDING' | 'SUCCESS' | 'RUNNING' | 'ERROR' | 'TIMEOUT_EXPIRED'
```

WDYT?

neelance:
`export type WorkerStatus` sounds good!

antfu:
Would you mind changes the pr addressing that? Thanks

neelance:
Done.

neelance:
Thanks!

Issue #176, opened 2020-11-05
Add isSupported for useShare function


anteriovieira:
@antfu what do you think about adding isSupported to useShare? This will break the current implementation.

```js
export async function useShare(shareOpts: ShareOptions) {
  const _navigator = (window.navigator as NavigatorWithShare)
  const isSupported = _navigator && _navigator.share

  const share = () => {
    if (isSupported) {
      let granted = true

      if (shareOpts.files && _navigator.canShare)
        granted = _navigator.canShare({ files: shareOpts.files })

      if (granted)
        return _navigator.share(shareOpts)
    }
  }

  return {
    isSupported,
    share,
  }
}

```
Demo
```html
<script>
import { useShare } from '@vueuse/core'

setup() {
   const { isSupported, share } = useShare({ title: 'Hello', text: 'Hello my friend!', url: location.href })
   return {
     isSupported
     share,
   }
},
</script>

<div>
    <button @click="share" :disabled="!isSupported">Share</button>
</div>
```

antfu:
Sure! You can make a PR on dev branch which we are aiming to introduce breaking changes altogether.

Issue #166, opened 2020-11-04
Roll out from beta


antfu:
Since Vue 3 has officially released for a while now. I think it's time for us to roll out from beta aim for a stable release.

Checklist

- [x] Review functions, frozen the APIs (consistency, flexibility, simplicity)
	- It should be our last chance (at this phrase) to introduce breaking changes to the existing functions. Breaking changes should be working on the dev branch.
    - [x] Make global reference configurable #184 
	- [x] Docs should not contain JSX, `@vue/composition-api`, `vue-demi`, etc.
    - [x] JSX docs #183
    - [x] Filters system #192  
- [x] Unit-tests for utilties as possible
	- It's ok for not having tests for simple web API wrappers.
- [x] Setup bundle size GitHub Actions via [export-size](https://github.com/antfu/export-size)
- [x] Demo rewrites working by @anteriovieira

## Functions

- [x] **useStorage**: debounce
- [x] **useShare**: returns a function instead of call it immediately, #176 

## Docs 

Preview: https://dev--vueuse.netlify.app/

- [x] Path redirecting, support short link like [vueuse.js.org/useMouse](https://dev--vueuse.netlify.app/useMouse)
- [x] Write a doc about how to use the functions #191 
	- [x] reactive to unwrap the refs
    - [x] configurable window, document
    - [x] timing
    - [x] filter system
## Preview Opt-in (Nightly Builds)

[npm package](https://www.npmjs.com/package/@knightly/vueuse/v/dev), powered by [![Nightly Build](https://github.com/antfu/knightly/blob/main/res/badge.svg?raw=true)](https://github.com/antfu/knightly)

To replace the package with nightly build:

via npm

```
npm i @vueuse/core@npm:@knightly/vueuse@dev
```

or edit your package.json

```json
"dependencies": {
  "@vueuse/core": "npm:@knightly/vueuse@dev"
}
```

## New Features

- New @vueuse/router package

- Event filter system

- `pausableWatch`, `debouncedWatch`, `thottledWatch`, `watchWithFilter`

### BREAKING CHANGE

<details>
<summary>
<b><code>useTransition</code></b> - named transition presets are removed. Import them from <code>TransitionPresets</code> intead. #190 
</summary>

<br>

```ts
import { useTransition } from '@vueuse/core'

useTransition(baseNumber, {
  duration: 1000,
  transition: 'easeInOutCubic',
})
```

Changed to
 
```ts
import { useTransition, TransitionPresets } from '@vueuse/core'

[...]

PR #180, opened 2020-11-06
feat(useShare): adds isSupported

anteriovieira:
ref: #166
closes: #176


antfu:
I think the implementation was ported from some other libs. The current API was not matched with VCA's style and I think your approach is definitely in a good direction.

A few thoughts,
- I think we should make the share options to the `share()` instead of `useShare()`, or maybe can allow options from `share()` override the initial options from `useShare()`.
- Since it's related to permissions, maybe we can utilize our `usePermission` to handle that? And do we miss the `grant()` feature for `usePermission`?

anteriovieira:
yes, sounds good! I will update this pull request soon.

anteriovieira:
Hi @antfu 

> I think we should make the share options to the share() instead of useShare(), or maybe can allow options from share() override the initial options from useShare().

I followed your second idea, I found it more convenient to use it this way, what do you think?

```js
    setup() {
      const { share, isSupported } = useShare()

      return {
        startShare: () => share({
          title: 'Hello',
          text: 'Hello my friend!',
          url: location.href,
        }),
        isSupported,
      }
    }
```

> Since it's related to permissions, maybe we can utilize our usePermission to handle that? 

I'm checking what kind of access I should grant. I didn't find anything related on the internet, they usually use  `'canShare ' in navigator` to see if it is supported.

> And do we miss the grant() feature for usePermission?

I'm testing this too, I'll send it in another pull request

packages/core/useShare/index.md
```diff
@@ -2,13 +2,17 @@
...
 
 ```js
 import { useShare } from '@vueuse/core'
 
-function onShare() {
-  useShare({ title: 'Hello', text: 'Hello my friend!', url: location.href })
+const { share, isSupported } = useShare({ title: 'Hello', text: 'Hello my friend!', url: location.href })
+
+function startShare() {
+  if (isSupported) share()
```
antfu, line 15: The usage needs to be updated. And I think in the example, we don't need to guide users to check `isSupported` again, `isSupported === false` is essentially make `share` a `noop`. `isSupported` could be used in UI guidances or something.

packages/core/useShare/index.ts
```diff
@@ -10,16 +10,24 @@ interface NavigatorWithShare {
   canShare?: (data: ShareOptions) => boolean
 }
 
-export async function useShare(shareOpts: ShareOptions) {
+export function useShare() {
```
antfu, line 13: Oh I am thinking this, maybe we can allow users to pass a ref or reactive data to the constructor and then calling share to use the current state, what do you think?

```ts
const options = ref<ShareOptions>({ url: '...' })

return { options, ...useShare(options) }
```

```html
<input v-model='options.url'/>
<button @click='share()'>Share</button>
````

Comment (anteriovieira) [reply packages/core/useShare/index.ts:13]: Good idea!

Comment (anteriovieira) [reply packages/core/useShare/index.ts:13]: 🤔    but thinking again hehe,  the users may also want to use it more programmatically, perhaps using it a little more complex. For example:

```js
const shareContentX =() = share({ ... })
const shareContentZ =() = share({ ... })
const shareContentZ = () = share({ ... })

// or

const shareContent = payload => share(payload)

if (...) shareContent(...)
else if (...) shareContent(...)
else shareContent(...)

// ...
```

Changing to the constructor he will have to manage the shared data in a single place (ref).

I haven't used this feature in my projects yet, so my thinking may be wrong  😬

antfu: I am thinking like this, the returned share can accept an argument that overrides the "defaults"

```ts
const options = ref<ShareOptions>({ foo: 'bar' })

const { share } = useShare(options)

share() // { foo: 'bar' }

// change the source to change the payload. In the real-world, this can be a computed what dynamics on user content
options.value = { hello: 'world' }

share() // { hello: 'world' }

// temporary override payload
share({ url: 'github.com', text: 'hello world' }) // { hello: 'world', url: 'github.com', text: 'hello world' }

share() // { hello: 'world' }
```

This might cover most of the use cases I guess?

anteriovieira: Yes, it is a good solution and it might cover most use cases. Thanks!

anteriovieira:
useShare

packages/core/useShare/index.ts
anteriovieira: @antfu , I tried using permissions but was unsuccessful. It looks like we don't have support yet.

anteriovieira:
@antfu , I tried using permissions but was unsuccessful. It looks like we don't have support yet.

antfu:
Thanks a lot. Did a few changes, supported using ref as options which is useful for pass computed or refs.

PR #182, opened 2020-11-06
BREAKING CHANGE: refactor!(useRefHistory): flush 'pre' as default for useRefHistory

matias-capeletto:
As discussed in https://github.com/antfu/vueuse/issues/181, this PR aligns useRefHistory with watch/watchEffect using flush: 'pre' as the default. As stated in the Vue docs, flush: 'sync' should only be used in rare cases. After our experience working with this composable, I think it is worth the effort to try to find a way to use 'pre'

In the current PR, there are new tests for the new default case, and the old tests are also working when using flush: 'sync'. I tried to follow the async pattern in other composable.

batch only makes sense when using 'sync'

Supported operations in sync when using 'pre'
  undo, undo
  undo, modify
  modify, undo (but the modification is lost)
  modify, commit, modify, commit, undo
  undo, modify, commit

We could also directly define that undo/redo and other history operations can not be chained with other sync modifications. But I tried to support them here.

### DIFF
--- a/packages/core/useRefHistory/index.md
+++ b/packages/core/useRefHistory/index.md
@@ -1,24 +1,36 @@
 # useRefHistory
 
-> Track the change history of a ref, also provide undo and redo functionality
+> Track the change history of a ref, also provides undo and redo functionality
 
 ## Usage
 
 ```ts
-import { ref } from 'vue' 
+import { ref } from 'vue'
 import { useRefHistory } from '@vueuse/core'
 
 const counter = ref(0)
 const { history, undo, redo } = useRefHistory(counter)
+```
 
+Internally, `watch` is used to trigger a history point when the ref value is modified. This means that history points are triggered asynchronously batching modifications in the same "tick".
+
+```ts
 counter.value += 1
-counter.value = 10
 
-console.log(history.value) // [{ value: 10, timestamp: 1601912898062 }, { value: 1, timestamp: 1601912898061 }]
+await nextTick()
+console.log(history.value)
+/* [
+  { value: 1, timestamp: 1601912898062 }, 
+  { value: 0, timestamp: 1601912898061 }
+] */
+```
 
-console.log(counter.value) // 10
+You can use `undo` to reset the ref value to the last history point.
+
+```ts
+console.log(counter.value) // 1
 undo()
-console.log(counter.value) // 1 
+console.log(counter.value) // 0
 ```
 
 ### Objects / arrays
@@ -28,7 +40,7 @@ When working with objects or arrays, since changing their attributes does not ch
 ```ts
 const state = ref({
   foo: 1,
-  bar: 'bar'
+  bar: "bar",
 })
 
 const { history, undo, redo } = useRefHistory(state, {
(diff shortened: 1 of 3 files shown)


matias-capeletto:
Something else, I think it is a good idea to avoid modifying the tracking ref to ignore history operations because we are exposing it in the API and users may be watching this ref to modify the UI (like adding a red/green dot somewhere in case history is on/off). So it would be better to only change it as a result of pause/resume operations

antfu:
I made some refactoring during the reviewing & understanding, hope you won't mind :P 

The implementation is interesting and good work! But it took me a while to get a rough understanding where it still not sure about how to explain the `syncCounter` 😅 (but I trust you). It would be great if you can add some more comments to the code for the future contributor better understand it, which is up to you.

One thing tho is that I think it's better to **keep the flush default to `sync`**. `sync` is something that "just works" and it's easier for newcomers to get hands-on easier. `pre` can be an advanced usage which requires users to have a good understanding of updating cycles, `nextTick` and `watch`'s `flush`.

```ts
const counter = ref(0)

const { history } = useRefHistory(counter)

console.log(history.length) // 1

counter.value = 3

console.log(history.length) // 1 // WTF?
```

WDYT?

matias-capeletto:
Not at all, thanks for changing the nextTick one! There are other tests in vueuse using instance.$nextTick where I copied that from and I did not like it either.

I add some comments, I hope they help to explain what each counter is tracking.

About the default, I really think that for Vue, the simplest one is 'pre'. From the vue docs https://v3.vuejs.org/guide/reactivity-computed-watchers.html#effect-flush-timing

_Vue's reactivity system buffers invalidated effects and flushes them asynchronously to avoid unnecessary duplicate invocation when there are many state mutations happening in the same "tick". 
The flush option also accepts 'sync', which forces the effect to always trigger synchronously. This is however inefficient and should be rarely needed_

The docs are encouraging users to embrace the way 'pre' works and I think it is the right approach. It is quite hard to debug bugs that arise when there is a missing batch, the error appears when using the history (undo, redo) and seeing that there is an invariant broken (two arrays that needed to be in sync for example, that is normal to do in Vue code).
The issue with array.splice() is also really nasty. I do not know if that is a problem with Vue itself or if it is expected that one array operation will trigger two history points (with an invariant broken for the middle one probably). In our code base right now we have a batch() that is wrapping a single splice(). I actually think that we will need to explain even more gotchas with sync that with pre, and mainly we can not say: "it works exactly like watch, same default, same timing, etc".

antfu:
OK. I am convinced. Could you update the docs about it and adding a Caveats section? Thanks a lot.

> There are other tests in vueuse using instance.$nextTick

Oops, they were written before `@vue/composition-api` implements `nextTick` API. That was... long ago XD. Just updated.

matias-capeletto:
Updated the docs, I had to change several sections because of this change. Feel free to modify my changes 👍

antfu:
@matias-capeletto I have invited you to be the collaborator of this project. Thanks so much for the discussions, people like you really push the design and implementations better!

matias-capeletto:
Thanks! 💚 

Should I create a branch directly in antfu/vueuse for the next PR (that we delete later) or keep using my fork of the lib to push PRs?

antfu:
Both are fine :)

PR #185, opened 2020-11-06
fix(useRafFn): incorrect lifecycle hook

scottbedard:
It looks like [in this commit](https://github.com/antfu/vueuse/commit/eda2fd4b7fb7072fd7bb56fbbe886114fec661fa#diff-80fba29722597fce2f73a9b693b85a627c4c4192a581f0af7cd3f3ce0e68ebdeL28-R28) the lifecycle binding was accidentally changed from `onUnmounted` to `onMounted`.

This caused [`useRafFn`](https://vueuse.js.org/?path=/story/animation--useraffn) to immediately stop it's animation loop, and had the cascading effect of breaking [`useTransition`](https://vueuse.js.org/?path=/story/animation--usetransition) which relies on that loop.

### DIFF
--- a/packages/core/useRafFn/index.ts
+++ b/packages/core/useRafFn/index.ts
@@ -1,4 +1,4 @@
-import { tryOnMounted } from '@vueuse/shared'
+import { tryOnUnmounted } from '@vueuse/shared'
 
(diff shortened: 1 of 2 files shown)


antfu:
Thanks!

PR #186, opened 2020-11-07
chore: types

maxma241:
## Fix `useDebounceFn` return type

The type of this composition should be `() => void` when the argument `delay` is greater than 0.
But, if `delay` is a negative number, the type is wrong with the logic now.

In addition, we usually use `useDebounceFn` with a natural number to the delay time.
So, maybe the type is fine when delay time is a negative number. 😨 

FYI,
thanks.

----
About the Typescript issue
Typescript could not define type with range of number yet. 😞 
https://github.com/microsoft/TypeScript/issues/15480


antfu:
I am not sure what I was thinking, it should always be `T` regardless of the timeout. Fixed. Thanks

PR #188, opened 2020-11-07
Clean up useTransition function

scottbedard:
Closes https://github.com/antfu/vueuse/issues/187

@antfu The only significant change here was https://github.com/antfu/vueuse/pull/188/commits/68c82eb986ea7cc587b5c3bf6b018ce960a7dd31

Previously, `useRafFn` was being called every time the base number changed. This caused a new raf loop to be created every time, and bound a lifecycle event on `onUnmounted` to kill the loop when the component is destroyed. This might have actually been a bug though, because I believe calling `onUnmounted` outside the context of `setup` has no effect. Regardless though, now `useRafFn` will be called only once, and there are no behavior changes.

### DIFF
--- a/packages/core/useTransition/index.md
+++ b/packages/core/useTransition/index.md
@@ -9,7 +9,7 @@ import { useTransition } from '@vueuse/core'
 
 useTransition(baseNumber, {
   duration: 1000,
-  transition: 'easeInOutCubic',
+  transition: [0.75, 0, 0.25, 1],
 })
 ```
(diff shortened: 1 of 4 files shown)


antfu:
Great job!

This seems to be a breaking change, could you rebase your code on top of `dev` branch? thanks!

scottbedard:
Closing in favor of https://github.com/antfu/vueuse/pull/190/

PR #200, opened 2020-11-11
feat(watchWithFilter): 4 new watch utilities

antfu:
`watchWithFilter` -> `throttledWatch`, `debouncedWatch`, `pausableWatch`

Example:

```ts
import { pausableWatch } from '@vueuse/core'
import { ref, nextTick } from 'vue'

const source = ref('foo')

const { stop, pause, resume } = pausableWatch(source, (v) => console.log(`Changed to ${v}!`))

source.value = 'bar'
await nextTick() // Changed to bar!

pause()

source.value = 'foobar'
await nextTick() // (nothing happend)

resume()

source.value = 'hello'
await nextTick() // Changed to hello!
```

@matias-capeletto What do you think?


matias-capeletto:
Fancy! I especially like the pausableWatch, and vueuse can offer these goodies without a lot of maintenance overhead 🎉

antfu:
@matias-capeletto FYI, I am trying to offer an overall usage guide to VueUse in [this doc](https://github.com/antfu/vueuse/blob/0339e816fe1e0672e32a18c54dbe542d71c386b8/docs/Guide.md), which will cover features like EventFilters or reactive timing. It would be great if you can help to review it a bit or even replenish some. Thanks!

PR #204, opened 2020-11-12
BREAKING CHANGE: refactor(useRefHistory): API changes

matias-capeletto:
As discussed in https://github.com/antfu/vueuse/issues/203, added a `last` ref to `useRefHistory`

Added some tests, but docs are still missing. I'll add them later in this PR or in another one if you would like to merge this one as is.

I am returning a record in last `{ value, timestamp }`. I wonder if we should change the name of value as we have to access the serialized value using `last.value.value` right now. I can not think of a better name though.


antfu:
That would be great if you can update the docs as well. But maybe the jsdoc is enough I guess?

> `last.value.value`

Emmm, yeah. It does feel weird. I guess it's better to avoid using `value` in objects in cause confusion with refs. Maybe `snapshot`? And with the `destructing` section in the guide, we can have:

```ts
const history = reactive(useRefHistory(source))

history.last.snapshot
```

&

```ts
const { last } = useRefHistory(source)

last.value.snapshot
```

And if we do, we should change the history stacks as well (breaking change).

matias-capeletto:
I like `snapshot` a lot. Updated the PR with the breaking change. 

I found that there may be an issue with the way we are handling the undoStack. We already push a value to the undoStack at the beginning, but I think this is not right. Since in apps for example, you are received with the possibility to undo (and you didn't do anything yet).

I think that now that we added `last`, this can be fixed because we can have:

full history: `[...undoStack, last, ...redoStack]`

So `last` is not undoStack[0], but it is the last value used to update `current`. Please leave this PR open and I will review this better.

I see that for using storage, it is also broken now, because undoStack[0] will be undefined if you undo all the way to the beginning. So looks like `last` may not be an alias but something more important to the way the composable works at the end

antfu:
> [...undoStack, last, ...redoStack]

I think you are right. It's better to do this way. And if so, we should reconsider the behaviour of the functions we have.

- `commit` - push `last` to undoStack and make clone the current state to `last`
- `reset` - set `last` to the source ref, no affecting stacks
- `undo`/`redo` - how should they behave? we are having two situations here, `last` is aligned to current or not

matias-capeletto:
`reset` was using `undoStack[0]` before so it was not working when undoStack was empty. Now that we added `last`, I was able to fix this. 

About starting with `undoStack = [last]`, maybe we can discuss that in another issue, PR. It doesn't change the implementation it seems at the end. We just need to start with `undoStack = []`, and update the tests to reflect that at the beginning you have empty history. I think we should do the change, even if it is not intuitive that history will be empty at the beginning because if not we will need to ask users to use `canUndo = () => undoStack.length > 1` and `canRedo = () => redoStack.length > 0` if they want to avoid that extra state. What do you think? I can do a PR for this separately and we discuss there if you want.

antfu:
I think we can discuss and make changes in this PR

matias-capeletto:
I am doing the changes and `history` now doesn't look as good as before just being an alias for undoStack. 
Maybe we should change it to be:
```
history = [ last, ...undoStack ]
```
I do not how costly it is to add these extra arrays, refs. But I think this is a good change, as users can see the history until the moment, that is different from what they can undo/redo.
Almost all the tests can pass without being changed in that way also.

What do you think?

antfu:
Cool, that makes sense. Use computed for history won't be costly. computed is lazy and only reevaluate on accessing, there shouldn't be much cost, especially for users who don't need it.

matias-capeletto:
I see now that in my old implementation, I used only one history array and an index pointer to the current history point. That is why this issue didn't appear there. 

We now have full  = ... undo values, last, ... redo values. And I think maybe one extra renaming should be done. `last` is really the `current` history value. For example, if we do an undo, the "prev" values becomes the "last" value with the naming so far, and this may be a bit confusing. When you reset, you set the source to the current value. 
Should we rename them?

  current -> source  ( forwarded source ref )
  last -> current ( so history.value.current )

What do you think? I am fine either way.

matias-capeletto:
One more thing, that I could add to this PR. I was thinking that having `canUndo` and `canRedo` computed refs could remove complexity in the API. So we have:
```
<button :disabled="canUndo" @click="undo">

**Comment by matias-capeletto** [comment]
undo and redo implementations ended up being a mirror of each other
```
  const undo = () => {
    const state = undoStack.value.shift()

    if (state) {
      redoStack.value.unshift(last.value)
      _setCurrentValue(state)
    }
  }

  const redo = () => {
    const state = redoStack.value.shift()

    if (state) {
      undoStack.value.unshift(last.value)
      _setCurrentValue(state)
    }
  }
```

**Comment by antfu** [comment]
> current -> source ( forwarded source ref )

Sounds good!

> last -> current ( so history.value.current )

Well, I am not very sure about this since it's not strictly `current`. There will be some cases it will not align with `source`. However, I can't come up with a better name for it 😅. Maybe let's keep as `last` for now?

> canUndo / canRedo

👍

**Comment by matias-capeletto** [comment]
Done, I think we can merge the PR now 😄

**Comment by antfu** [comment]
Great, thanks a lot!

**Comment by antfu** [comment]
FYI, I have scheduled a build for on dev including this PR, which should be live in a few minutes. You can opt-in for your apps can try them out. [progress](https://github.com/knightlyjs/tasks/actions/runs/362909865)

[npm](https://www.npmjs.com/package/@knightly/vueuse/v/dev)

```json
"dependencies": {
  "@vueuse/core": "npm:@knightly/vueuse@dev"
}
```

matias-capeletto:
Thanks a lot for that, I was actually going to ask if it was possible so we can try these changes in a real app 🙌

antfu:
🤫 I am using this a lot to dogfooding myself in my 100 day project.

matias-capeletto:
Everything working as expected 🎉 

Changeset migrating to `flush: 'pre'`, removing `batch` guards in several places, using `last` for the storage watch and using `canUndo` `canRedo` instead of the stacks for the disabled checks:
https://github.com/Leniolabs/layoutit-grid/commit/724632c6e12c374b77c2ef230e5f52fcbccf1886

PR #206, opened 2020-11-13
chore: add 'position: relative;' for markdown-body

sibbng:
Links and headings and not clickable/selectable on https://vueuse.js.org/?path=/story/sensors--usewindowscroll due to absolute div. I believe this change doesn't have any side effect on other docs.


antfu:
Nice catch, thanks!

PR #209, opened 2020-11-14
feat(ignorableWatch): new function

matias-capeletto:
Extracted a new composable: ignorableWatch. It extends watch allowing to ignore selected updates

```
const { stop, ignoredUpdate, reset } = ignorableWatch(source, cb, options)

ignoredUpdate( ()=>{
  // All sync updates to source are ignored
})
``` 

It is implemented using the double watch tick for flush pre and post, and a simple watch similar to pausableWatch for sync case.

This greatly reduced the complexity of the useRefHistory implementation

@antfu maybe this is a good composable to share. For the moment, I left it as an internal helper in useRefHistory. Do you think this can be useful for other use cases?

In the second commit, I implemented the pausable feature of useRefHistory using the new pausableFilter.


antfu:
I was thinking about the exact same thing when we are working on the double watch things, but I was not sure if it's useful for other usages. After seeing your example, I think it's kinda useful, and let's have it. One thing tho is that I am not sure about the naming, but can't come up with a better one for now. Maybe `ignoredUpdate` -> `ignoreUpdates` could be more clear?

matias-capeletto:
`ignoreUpdates` reads better, I'll change it.

I was thinking in `ignoreAsyncUpdatesTilNow()` for `reset()` as that is what is really doing, but I could come up with a short name for it

Do you prefer to extract the composable in a separate PR, or this one open til is standalone?

antfu:
Let's make it in this PR since without it there is no meaning to refactor useRefHistory anyway.

matias-capeletto:
Extracted `ignorableWatch` and added tests, docs, and a story.
It is now in the watch section in the docs, following alphabetical order.

Ended up with the names: `ignoreUpdates(updater)` and `ignorePrevAsyncUpdates()` for the API for the moment.

antfu:
Looking great, thanks a lot!

PR #210, opened 2020-11-14
docs(usePermission): remove undefined variable

sibbng:
[usePermission](https://vueuse.js.org/?path=/story/browser--usepermission) document is broken due to this undefined variable.

First I was trying to implement that permission. Looks like it is renamed to 'speaker-selection' from 'speaker' on latest draft. That require me to polyfill types like how you did for clipboard permission. Then I realized no one of existing browsers support that permission. And probably that's why this variable was removed.


anteriovieira:
Thanks!

PR #213, opened 2020-11-15
docs(useWebWorkerFn): improve demo

sibbng:
Current worker demo has some issues.

Normal sort block UI only on first run. Result of `sortNumbers(numbers)` cached by browser because of numbers are the same.
Also worker sort is blocks UI because `data.slice(0, 10)` is still a heavy task.

So I extracted array generation, sorting and slicing into a single function for keep heavy task in one place.
I printed result on screen to make the demo more optimistic.

I removed spread operator and filled Array with undefined because of how [Babel bundles that code](https://github.com/babel/babel/issues/7332). It was failing because Babel's `_toConsumableArray` function was not available in **worker**.

### DIFF
--- a/packages/core/useWebWorkerFn/index.stories.tsx
+++ b/packages/core/useWebWorkerFn/index.stories.tsx
@@ -1,11 +1,14 @@
 import dayjs from 'dayjs'
 import { defineDemo, html } from '../../_docs'
-import { defineComponent, computed } from 'vue-demi'
+import { defineComponent, computed, ref } from 'vue-demi'
 import { useWebWorkerFn, WorkerStatus } from '.'
 import { useNow } from '../useNow'
 
-const numbers: number[] = [...Array(5000000)].map(_ => ~~(Math.random() * 1000000))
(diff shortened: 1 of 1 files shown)


packages/core/useWebWorkerFn/index.stories.tsx
```diff
@@ -1,11 +1,14 @@
...
-import { defineComponent, computed } from 'vue-demi'
+import { defineComponent, computed, ref } from 'vue-demi'
 import { useWebWorkerFn, WorkerStatus } from '.'
 import { useNow } from '../useNow'
 
-const numbers: number[] = [...Array(5000000)].map(_ => ~~(Math.random() * 1000000))
-const sortNumbers = (nums: number[]): number[] => nums.sort()
+const heavyTask = () => {
+  const randomNumber = () => ~~(Math.random() * 5000000)
+  const numbers: number[] = Array(randomNumber()).fill(undefined).map(randomNumber)
```
antfu, line 9: Should we make the Array size fixed to the blocking time could be more consistent?

antfu (approved):
Great, thanks!

Issue #184, opened 2020-11-06
Review: Make global dependencies configurable


antfu:
Global dependencies such as `window`, `document`, `navigator` should be configurable. #173

## Checklist 

This is a checklist of all the functions of VueUse. It's is useful for review or making changes the go through every function.

### shared (21)
  - [ ] biSyncRef
  - [ ] controlledComputed
  - [ ] debouncedWatch
  - [ ] extendRef
  - [ ] makeDestructurable
  - [ ] pausableWatch
  - [ ] syncRef
  - [ ] throttledWatch
  - [ ] tryOnMounted
  - [ ] tryOnUnmounted
  - [ ] useDebounce
  - [ ] useDebounceFn
  - [ ] useInterval
  - [ ] useIntervalFn
  - [ ] useThrottle
  - [ ] useThrottleFn
  - [ ] useTimeout
  - [ ] useTimeoutFn
  - [ ] utils
  - [ ] watchWithFilter
  - [ ] when

### core (51)
  - [x] asyncComputed
  - [x] createGlobalState
  - [x] onStartTyping
  - [x] useAsyncState
  - [x] useBattery
  - [x] useBrowserLocation
  - [x] useClipboard
  - [x] useCounter
  - [x] useCssVar
  - [x] useDeviceLight
  - [x] useDeviceMotion
  - [x] useDeviceOrientation
  - [x] useDevicePixelRatio
  - [x] useDocumentVisibility
  - [x] useElementVisibility
  - [ ] useEventListener
  - [ ] useEventSource
  - [x] useFavicon
  - [x] useFullscreen
  - [ ] useGeolocation
  - [ ] useIdle
  - [ ] useIntersectionObserver
  - [ ] useLocalStorage
  - [ ] useMediaQuery
  - [ ] useMouse
  - [ ] useMouseInElement
  - [ ] useMutationObserver
  - [ ] useNetwork
  - [ ] useOnline
  - [ ] usePageLeave
  - [ ] useParallax
  - [ ] usePermission
  - [ ] usePreferredColorScheme
  - [ ] usePreferredDark
  - [ ] usePreferredLanguages
  - [ ] useRafFn
  - [ ] useRefHistory
  - [ ] useResizeObserver
  - [ ] useSessionStorage
  - [ ] useShare
  - [ ] useSpeechRecognition
  - [ ] useStorage
  - [ ] useTimestamp
  - [ ] useTitle
  - [ ] useTransition
  - [ ] useVModel
[...]

PR #214, opened 2020-11-15
chore: make global dependencies configurable

anteriovieira:
ref: #184

### DIFF
--- a/packages/core/useGeolocation/index.ts
+++ b/packages/core/useGeolocation/index.ts
@@ -2,12 +2,15 @@
 
 import { ref, Ref } from 'vue-demi'
 import { tryOnMounted, tryOnUnmounted } from '@vueuse/shared'
+import { ConfigurableNavigator, defaultNavigator } from '../_configurable'
 
-export function useGeolocation(options: PositionOptions = {
+export function useGeolocation(positionOptions: PositionOptions = {
   enableHighAccuracy: true,
   maximumAge: 30000,
   timeout: 27000,
-}) {
+}, { navigator = defaultNavigator }: ConfigurableNavigator = {}) {
+  const isSupported = navigator && 'geolocation' in navigator
+
   const locatedAt: Ref<number | null> = ref(null)
   const error = ref<PositionError | null>(null)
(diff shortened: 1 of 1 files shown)


packages/core/useGeolocation/index.ts
```diff
@@ -2,12 +2,15 @@
 
 import { ref, Ref } from 'vue-demi'
 import { tryOnMounted, tryOnUnmounted } from '@vueuse/shared'
+import { ConfigurableNavigator, defaultNavigator } from '../_configurable'
 
-export function useGeolocation(options: PositionOptions = {
+export function useGeolocation(positionOptions: PositionOptions = {
```
antfu, line 7: We can merge this with ConfigurableNavigator

packages/core/useGeolocation/index.ts
```diff
@@ -4,11 +4,15 @@ import { ref, Ref } from 'vue-demi'
 import { tryOnMounted, tryOnUnmounted } from '@vueuse/shared'
 import { ConfigurableNavigator, defaultNavigator } from '../_configurable'
 
-export function useGeolocation(positionOptions: PositionOptions = {
+interface PositionOpts extends PositionOptions, ConfigurableNavigator {}
```
antfu, line 7:
```suggestion
export interface GeolocationOptions extends PositionOptions, ConfigurableNavigator {}
```

packages/core/useGeolocation/index.ts
```diff
@@ -4,11 +4,15 @@ import { ref, Ref } from 'vue-demi'
...
+
+export function useGeolocation(positionOptions: PositionOpts = {
   enableHighAccuracy: true,
   maximumAge: 30000,
   timeout: 27000,
-}, { navigator = defaultNavigator }: ConfigurableNavigator = {}) {
+  navigator: defaultNavigator,
+}) {
```
antfu, line 14: The default value was not correct, partial options will override the entire default value. Should change to destructuring default assignments. This was a mistake, it would be nice if you can fix it alongside. Thanks

antfu:
I have made the change, thansk!

PR #215, opened 2020-11-15
docs: fix story links

peter50216:
The links in the @vueuse/firebase section has one extra `@` in the path argument, fix them.


antfu:
Thanks for spotting that. However, the links are auto-generated and I have updated scripts to fix that.

PR #219, opened 2020-11-18
chore: remove relative

sibbng:
I had issues with that when I was working on a modal demo for an utility function I'm currently working on.

I fixed select/click problem by pushing absolute div to outside of view. Also moved it up to make the texts selectable on fixed div.


antfu:
Thanks!

sibbng:
Oh no, I mistakenly moved fixed div instead of absolute one. 🤦‍♂️

antfu:
Haha, np. Maybe send another one? Thanks a lot!

PR #221, opened 2020-11-18
feat: useClickOutside

sibbng:
I'm not sure is this best way to do it but seems fine for me.

Alternatively we can provide a `useClickOutsideFn` utility for be able to use it with `@click="..."`. For example:

```js
const clickOutside = useClickOutsideFn(handler)
```
```html
<div @click="clickOutside">test</div>
```


packages/core/useClickOutside/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    listeners = []
+  }
+
+  tryOnUnmounted(stopListeners)
+
+  watch(
+    targetRef,
+    () => {
+      if (!targetRef.value) {
+        stopListeners()
```
antfu, line 45:
```suggestion
      stopListeners()
      if (!targetRef.value) {
```

Should we unbind the last element when the element ref changed?

packages/core/useClickOutside/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const listener = (event: MouseEvent | TouchEvent) => {
+    const elements = event.composedPath()
+    if (targetRef.value === event.target || elements.includes(targetRef.value!))
+      return
+
+    handler(event)
+  }
+
+  let listeners: Fn[] = []
```
antfu, line 32: A bit confusing with the `listener` function, maybe `disposables`?

packages/core/useClickOutside/index.ts
```diff
@@ -0,0 +1,55 @@
...
+        return
+      }
+
+      events.forEach((event) => {
+        listeners.push(useEventListener(event, listener, { passive: true }))
+      })
+    },
+    { immediate: true },
+  )
+}
```
antfu, line 55: We could export a stop handler for stoping the events and the watch.

packages/core/useClickOutside/index.ts
antfu: Wait, since you are listening on window not the element, maybe there is no need to register events inside the watch, just do it outside will be enough.

packages/core/useClickOutside/index.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Listen for clicks outside of an element.
+ *
+ * @see   {@link https://vueuse.js.org/useClickOutside}
+ * @param target
+ * @param handler
+ */
+
+const events = ['mousedown', 'touchstart'] as const
+
+export function useClickOutside(
```
antfu, line 15: Maybe `onClickOutside` to align with `onStartTyping`?

packages/core/useClickOutside/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  target: MaybeRef<Element | null | undefined>,
+  handler: EventListener,
```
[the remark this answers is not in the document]
sibbng: Can you check last commit?

packages/core/onClickOutside/index.ts
```diff
@@ -0,0 +1,54 @@
...
+      return
+
+    handler(event)
+  }
+
+  let disposables: Fn[] = []
+
+  events.forEach((event) => {
+    disposables.push(useEventListener(window, event, listener, { passive: true }))
+  })
```
antfu, line 44:
```suggestion
  let disposables: Fn[] = events.map((event) => useEventListener(window, event, listener, { passive: true })))
```

antfu:
Thanks a lot! Well done and nice demo 👍

sibbng:
> Thanks a lot! Well done and nice demo 👍

Thank you too. Learned a lot 😊

PR #222, opened 2020-11-18
docs(useRefHistory): add warning about splice in flush sync

matias-capeletto:
I [asked today in VueLand](https://discord.com/channels/325477692906536972/713147202188607558/778634528172015636) and at least there, they think that splice generating three flushes for a single operation is not a bug in Vue3 reactivity. For the moment, this PR adds a small note about this for useRefHistory


antfu:
Thanks a lot for looking into it!

PR #225, opened 2020-11-19
feat: adds useNProgress function

anteriovieira:
I would like to know from the community if it makes sense to have this function here.

> Reactive wrapper for [`nprogress`](https://github.com/rstacruz/nprogress)

```js {6}
import { useNProgress } from '@vueuse/integrations'

const { isLoading } = useNProgress()

function toggle() {
  isLoading.value = !isLoading.value
}
```

useProgress


packages/integrations/useProgressBar/index.md
```diff
@@ -0,0 +1,34 @@
...
+
+const loading = ref(false)
+
+useProgressBar(loading)
+
+function toggle() {
+  loading.value = !loading.value
+}
+```
+
```
antfu, line 33: Maybe we should have some statements showing that the existence the `options` arg and redirect users to check `nprogress`'s documentation.

packages/integrations/useProgressBar/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Reactive progress bar.
+ *
+ * @see   {@link https://vueuse.js.org/useProgressBar}
+ * @param newIcon
+ * @param options
+ */
+export function useProgressBar(
+  current: MaybeRef<boolean | null | undefined> = null,
```
antfu, line 13: Thinking we should use numbers to set the progress.

https://github.com/rstacruz/nprogress#advanced-usage

packages/integrations/useProgressBar/index.ts
```diff
@@ -0,0 +1,28 @@
...
+    : ref<boolean | null>(current)
+
+  if (options)
+    progress.configure(options)
+
+  watch(loading, active => active ? progress.start() : progress.done())
+
+  tryOnUnmounted(progress.remove)
+
+  return loading
```
antfu, line 27: Maybe it's better to return an object giving more controls to users.

```ts
{
  progress: Ref<number>,
  isLoading: Ref<boolean>,
  remove: Fn,
  done: Fn,
  /* ... */
}
```

matias-capeletto:
Nice! Maybe it could be called `useNProgress()` so it is clear that this is a wrapper for NProgress?

antfu:
Agree with `useNProgress` tho it reads weird but make more sense 👍

packages/integrations/useProgressBar/index.md
anteriovieira:
```js {13}
export function useNProgress(
  currentProgress: MaybeRef<number | null | undefined> = null,
  options?: NProgressOptions | undefined,
) {
  const progress = isRef(currentProgress) ? currentProgress : ref(null)
  const isLoading = ref(false)

  if (options)
    nprogress.configure(options)

  watch(isLoading, active => active ? nprogress.start() : nprogress.done())

  watch(progress, (n) => {
    if (typeof n === 'number')
      nprogress.set(n)
  })

  tryOnUnmounted(nprogress.remove)

  return {
    isLoading,
    progress,
    start: nprogress.start,
    done: nprogress.done,
    remove: nprogress.remove,
  }
}
```

What do you think?

packages/integrations/useProgressBar/index.ts
anteriovieira:
```js {13}
export function useNProgress(
  currentProgress: MaybeRef<number | null | undefined> = null,
  options?: NProgressOptions | undefined,
) {
  const progress = isRef(currentProgress) ? currentProgress : ref(null)
  const isLoading = ref(false)

  if (options)
    nprogress.configure(options)

  watch(isLoading, active => active ? nprogress.start() : nprogress.done())

  watch(progress, (n) => {
    if (typeof n === 'number')
      nprogress.set(n)
  })

  tryOnUnmounted(nprogress.remove)

  return {
    isLoading,
    progress,
    start: nprogress.start,
    done: nprogress.done,
    remove: nprogress.remove,
  }
}
```

What do you think?

packages/integrations/useProgressBar/index.ts
anteriovieira: I think we have a problem, when I use the library functions to change the status I don't have any events that can notify me to update the refs

packages/integrations/useNProgress/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Passing a source ref
+
+You can pass a `ref` to it, changes from of the source ref will be reflected to your loading progress.
+
+```js {4}
+import { ref } from 'vue'
+import { useNProgress } from '@vueuse/integrations'
+
+const isLoading = ref(false)
```
antfu, line 25: Do we need to update the docs? I think we are accepting numbers now, or do you think we should support boolean as well?

anteriovieira: no, I don't think we need to support boolean here

PR #227, opened 2020-11-20
chore(onClickOutside): listen default window events

sibbng:


antfu:
I see, I run into a similar situation when working with usePallex. But instead of modifying the function to get around with this specific situation, maybe we can provided a mocked window that affect to the both windows.

If you get my points feel free to give a try, or I can do that later ;) (currently outside and don't bring my laptop with me)

sibbng:
> we can provided a mocked window that affect to the both windows

If we are exporting that mocked window from `@vueuse/core` as a user, I would say this is not a good idea.

Instead I would allow sending **array of targets** on `useEventListener`, I believe this will bring bunch of new use cases possible.
Also `ConfigurableWindow` will allow array of `Window`'s. Then users can send custom windows in an array for their specific use cases where our mocked window doesn't allow them to do, such as registering event listener on multiple iframes.

Examples
```js
{ window: window.parent }
{ window: [window.parent, window] }
{ window: [window, iframe.contentWindow] }
```

antfu:
> If we are importing that mocked window from @vueuse/core as a user

Well, no. I mean only the doc side. No user code will be affected. Since we already made event target configurable, this should already provide freedom to pass a duck window for users to do custom stuff.

About array of targets for useEventListener, in some cases, it does make sense. But most of time you would like to know which one triggered the event. For example, passing two windows to `useMouse` will just make it broken.

sibbng:
I got it. You right. Most of things I said about arrays can be done with `[].forEach(...)`, but I thought it would be nice providing this functionality in core, but useMouse example makes sense.

I'm curious about how that mocked window should look like. I would like to work on it, but I'm not sure where should I start.

antfu:
Updated according to #229. Thanks!

PR #229, opened 2020-11-20
chore(docs): mock window

sibbng:
Related #227

~~Hi @antfu, sorry for this ugly work 😅~~
~~Just asking for clarify, am I on right direction?~~


antfu:
Moved to docs code instead of the core lib. Thanks a lot!

PR #238, opened 2020-12-04
feat(useToggle): new function

sxzz:


antfu:
Can you share some use cases on this? This might be too simple that can be done without it.

sxzz:
Like this [useToggle](https://github.com/streamich/react-use/blob/master/docs/useToggle.md). It can be used on button switches. And just for convenience.

antfu:
I prefer the interface like `const [on, toggle] = useToggle(true)` just like react-use. And since it's pretty easy to get/set in Vue, I think just having the toggle function is quite enough, get() / set() feels a bit overdo.

antfu:
Yeah, but I am thinking that `useCounter` is more like a demo utils rather than something that will be used in the real world 😅

antfu:
I'd prefer `useToggle` as it gives the sense that it's used for creating toggle functions easily. What do you think?

sxzz:
OK, useBoolean is a bit strange 😂

antfu:
Thanks!

PR #239, opened 2020-12-04
feat(useElementSize): new function

sxzz:


antfu:
Maybe `useElementSize`? `useSize` feels a bit too general

sxzz:
OK. Already edited.

packages/core/useElementSize/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface Size {
+  width: number
+  height: number
+}
+
+/**
+ * Reactive size of an HTML element.
+ *
+ * @see   {@link https://vueuse.js.org/useResizeObserver}
```
antfu, line 13:
```suggestion
 * @see   {@link https://vueuse.js.org/useElementSize}
```

packages/core/useElementSize/index.ts
```diff
@@ -0,0 +1,39 @@
+import { MaybeRef } from '@vueuse/shared'
+import { ref } from 'vue-demi'
+import { ResizeObserverOptions, useResizeObserver } from '../useResizeObserver'
+
+export interface Size {
```
antfu, line 5:
```suggestion
export interface ElementSize {
```

packages/core/useElementSize/index.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Reactive size of an HTML element.
+ *
+ * @see   {@link https://vueuse.js.org/useElementSize}
+ * @param target
+ * @param callback
+ * @param options
+ */
+export function useElementSize(
+  target: MaybeRef<Element | null | undefined>,
+  initialSize: Size = { width: Infinity, height: Infinity },
```
antfu, line 20:
```suggestion
  initialSize: ElementSize = { width: 0, height: 0 },
```

Making the default size to zero makes more sense to me. (just like an empty <div/>)

sxzz:
Ok thanks, it has been edited.

antfu:
I have rethought it a bit, do you think it's better to make it expose more properties from an element?

Something like 

```js
const { x, y, width, height } = useElementBounding(el)
```

https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect

antfu:
And so that `useMouseInElement` can even reuse it to be better abstracted.

sxzz:
I don’t think it is necessary. This function is just to simplify getting the element size.
See [useSize](https://github.com/streamich/react-use/blob/master/docs/useSize.md)

sxzz:
Maybe can add another `useElementBounding` function

antfu:
Ok fair enough, would you like to add it as well? Thanks

antfu (approved):

sxzz:
Yes, maybe later.

antfu:
@sxzz Planned to do it now as I am going to refactor useMouseInElement, let me know if you are already working on it in case of duplicated works. Thanks.

sxzz:
No, I haven’t started do it yet, so please~😅

PR #244, opened 2020-12-10
feat: useManualRefHistory

matias-capeletto:
While [writing about useRefHistory](https://patak.dev/vue/history-and-persistence.html) and how to manage history across events, I got back to a suggestion in https://github.com/antfu/vueuse/issues/181#issuecomment-723137261

@antfu wrote:
> In addition, maybe we can have a manual option, and users would need to call commit() to create history points explicitly.

I think that if most operations end up needing pause/resume, then it makes sense for the app to use a manual strategy directly. 

This PR adds a new composable `useManualRefHistory` that presents the same API as `useRefHistory` but works with manual triggering by the user calling `commit()`. I added a new composable instead of using an option because the implementation is greatly simplified if we do not need to deal with tracking (no ignorableWath, no pausableFilter, etc). So users opting for this strategy doesn't need to pay for the extra kB added by automatic tracking.

The second commit in the PR uses `useManualRefHistory` to implement `useRefHistory`, ensuring that both composables work in the same way and simplifying a bit more useRefHistory that now only deals with automatic tracking logic and delegates the history stacks management to the manual composable,


antfu:
Looks great, thank you!

PR #247, opened 2020-12-13
feat: add lazy option to asyncComputed

### LINKED ISSUE #248: asyncComputed is not lazy like computed
Not sure if this should be a default but, I think, there should be option to make `asyncComputed` lazy like vue `computed`. Right now even if you use `v-if` to not render computed data it is still being retrieved.

#247 - Example of  failing test

Demivan:
closes #248

### DIFF
--- a/packages/core/asyncComputed/index.test.ts
+++ b/packages/core/asyncComputed/index.test.ts
@@ -41,7 +41,7 @@ describe('asyncComputed', () => {
     const func = jest.fn(() => Promise.resolve('data'))
 
     const instance = renderHook(() => {
-      const data = asyncComputed(func, undefined, undefined, { lazy: true })
+      const data = asyncComputed(func, undefined, { lazy: true })
 
       return {
         data,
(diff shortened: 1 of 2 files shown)


antfu:
Can you help to resolve the conflicts? Thanks

matias-capeletto:
About the new options params, given that the other two are optional, maybe at one point the three params could be in a single options object:
```ts
const asyncComputed( ()=> { ... }, { initial, evaluating, lazy } )
```
So users are not forced to add `undefined` twice if they want the default values for that params.

One question about the trick using `computed()` to defer the evaluation. In the lazy case, the internal ref and watchEffect will end up being created after the component is mounted, no? That means that they will not be in the same effect scope, and there will be a leak on unmount as they will not be stoped. Do this pattern also needs the effectScope PR from @antfu ? Maybe I am misunderstanding how this works, just checking.

Demivan:
@antfu Fixed merge conflict.

@matias-capeletto I don't like putting initial value into options object. I think this looks better:
```ts
const asyncComputed( ()=> { ... }, initial, { evaluating, lazy })
```
And it should be possible to check if third parameter is ref and support previous api.
```ts
const asyncComputed( ()=> { ... }, initial, evaluating)
```
I'm actually not sure about effect scope. Will try to write test for this.

antfu:
@Demivan the API you proposed looks good to me. Can you update this PR to it? Thanks.

The timing problem @matias-capeletto mentioned may cause some edge cases with `effectScope`, I think maybe we can have a guard flag for `watchEffect` and make it try on the computed get trigger. As it might be out of the scope of this PR, feel free to do so or I can take care of it later.

Demivan:
@antfu Updated PR with new API.
Not sure how to deal with described edge case so leaving it to you.

antfu:
Thanks! See 425a8e0

PR #249, opened 2020-12-14
test: add basic asyncComputed tests

matias-capeletto:
Some tests for `asyncComputed` that I used while checking https://github.com/antfu/vueuse/issues/248
The first two are modified versions of the ones included in https://github.com/antfu/vueuse/pull/247, checking that the callback is called immediately and not lazily.

### DIFF
--- a/packages/core/asyncComputed/index.test.ts
+++ b/packages/core/asyncComputed/index.test.ts
@@ -22,66 +22,75 @@ describe('asyncComputed', () => {
   test('it is not lazy', async() => {
     const func = jest.fn(() => Promise.resolve('data'))
 
-    const data = asyncComputed(func)
+    const instance = renderHook(() => {
+      const data = asyncComputed(func)
+      return { data }
+    }).vm
 
     expect(func).toBeCalledTimes(1)
 
-    expect(data.value).toBeUndefined()
+    expect(instance.data).toBeUndefined()
 
     await nextTick()
     await nextTick()
 
-    expect(data.value).toBe('data')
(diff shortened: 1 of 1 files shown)


packages/core/asyncComputed/index.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  })
+
+  test('re-computes when dependency changes', async() => {
+    const counter = ref(1)
+    const double = asyncComputed(() => {
+      const result = counter.value * 2
+      return Promise.resolve(result)
+    })
+
+    expect(double.value).toBeUndefined()
```
antfu, line 43: Might need to wrapper it with `renderHook` I think

matias-capeletto: It may have been a timing issue with Promise.resolve and nextTick, as it was working for me before. Looks like it is resolved by always waiting two ticks.
Should `renderHook` be used when possible in tests? I do not know exactly when I should include it.

antfu: I think we should always use `renderHook` if possible. Unlike Vue 3 that the APIs can be used standalone, Vue 2 is a bit tricky to do that (which is what we are using for testing, likely to move to Vue 3 or isomorphic in the future). Can you try with that and see if it still needs to invoking nextTick twice?

matias-capeletto: Looks like it is still needed, it was failing locally without it after adding render hooks

PR #252, opened 2020-12-14
feat: add option to use custom scrollTarget with `useElementVisibiity`

### LINKED ISSUE #251: feat: useElementVisibility with custom scroll target .
I happened to have a use case when the scrolling in my page wasn't on the document object and thus `useElementVisibilty` never changes since underneath it uses `useWindowScroll`, I could submit a PR if that's acceptable. 

for now I am using a custom version of `useElementVisibilty` that uses `useEventListener` to attach a scroll event on my scrolling element.

amrbashir:
Closes #251


antfu:
Thanks

PR #260, opened 2020-12-20
feat(useUrlSearchParams): new function

lstoeferle:
Hi there,
I worked on a useUrlSearchParams composable, which can handle both history and hash routing. I thought it's maybe useful for others.

Example use-case **Vue Micro-Frontends**:
In a micro-frontend setup the child web-components often do not have access to the vue-router of the hosting application. This composable enables the web-components to *get* and *set* the UrlSearchParams in a reactive way, without having the vue-router.


packages/core/useUrlSearchParams/index.ts
```diff
@@ -0,0 +1,47 @@
+import { computed, Ref, ref, watch } from 'vue-demi'
+import { useBrowserLocation } from '..'
+import { useEventListener } from '../useEventListener'
+
+export function useUrlSearchParams(method: 'history'|'hash'): Ref<URLSearchParams> {
+  const location = useBrowserLocation()
```
antfu, line 6: I think we don't need to use this hook, just read them from `window.location` should be good.

antfu (commented):
Looking good! Can you add the corresponding `index.md` and `index.stories.tsx` for the documentation? Thanks!

lstoeferle:
Yes for sure 😉

antfu:
@lstoeferle I would like to make some changes directly to the PR but it seems that I don't have permission to push to your branch. Can you check out if the ["Allow edits from maintainers"](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) checked? Thanks.

lstoeferle:
> @lstoeferle I would like to make some changes directly to the PR but it seems that I don't have permission to push to your branch. Can you check out if the ["Allow edits from maintainers"](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) checked? Thanks.

@antfu You should be able to access my fork. Strange... I've unchecked and checked the checkbox, can you check again?

antfu:
Thanks. Pushed. I made a few changes to make the API more consistent with the other ones. Although I am not quite sure if it's a good idea to directly return a ref with the URLQueryParams object. `params.value.set('foo', 'bar')` looks a bit weird to me. I am wondering if it's possible to have something like `params.value.foo = 'bar'` without too much overhead.

lstoeferle:
Good point, let me think about that 😉 I'll try to find a solution tomorrow 💪

lstoeferle:
@antfu I've pushed some changes, please have a look.

packages/core/useUrlSearchParams/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    const query = empty
+      ? hashWithoutParams.value
+      : (mode === 'hash')
+        ? `${hashWithoutParams.value}?${params}`
+        : `?${params}${hashWithoutParams.value}`
+
+    window.history.replaceState({}, '', window.location.pathname + query)
+  }
+
+  const params = ref(read()) as Ref<URLSearchParams>
```
antfu, line 51:
```suggestion
  const params = ref(read()) as Ref<URLSearchParams>
```

Since we don't expose this variable, I think can leave it as a plain variable.

antfu, line 51: Since we don't expose this variable, I think can leave it as a plain variable.

packages/core/useUrlSearchParams/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  const paramsMap: UrlParams = reactive<UrlParams>({})
+
+  const updateParamsMap = () => {
+    Object.keys(paramsMap).forEach(key => delete paramsMap[key])
+    for (const key of params.value.keys()) {
+      const paramsForKey = params.value.getAll(key)
+      paramsMap[key] = paramsForKey.length > 1 ? paramsForKey : (params.value.get(key) || '')
```
antfu, line 58: The writing to `paramsMap` here would trigger the watch I think. We can use `pauseableWatch` here.

packages/core/useUrlSearchParams/index.ts
```diff
@@ -0,0 +1,81 @@
...
+export type UrlParams = {[key: string]: string[] | string}
+
+/**
+ * Reactive URLSearchParams
+ *
+ * @see   {@link https://vueuse.js.org/useUrlSearchParams}
+ * @param mode
+ * @param options
+ */
+export function useUrlSearchParams(
```
antfu, line 14: Here we can have a generic for users to customize their expected type. You can leave this to me if you are not sure about how to do it.

packages/core/useUrlSearchParams/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  mode: 'history'|'hash' = 'history',
+  options: ConfigurableWindow = {},
+) {
+  const { window = defaultWindow } = options
+
+  if (!window)
+    return {}
```
antfu, line 21:
```suggestion
    return reactive({})
```

packages/core/useUrlSearchParams/index.ts
```diff
@@ -0,0 +1,81 @@
...
+        ? `${hashWithoutParams.value}?${params}`
+        : `?${params}${hashWithoutParams.value}`
+
+    window.history.replaceState({}, '', window.location.pathname + query)
+  }
+
+  const params = ref(read()) as Ref<URLSearchParams>
+  const paramsMap: UrlParams = reactive<UrlParams>({})
+
+  const updateParamsMap = () => {
```
antfu, line 54: I think we can have this function inlined in `write()`

packages/core/useUrlSearchParams/index.ts
lstoeferle: I've tried to change the composable to have a generic , but I failed resolving a TS error for accessing the object values by string index keys. Maybe you can help me there, i've tried several solutions, but none of them worked.

lstoeferle: I managed to fix the TS error by creating a `writeToPramsMap` function. Are you fine with 
```javascript
useUrlSearchParams<T extends Record<string, any> = UrlParams>
```

antfu:
Thanks!

mskr:
Hi, thanks for this function! I want to add more modes, but first it would help to understand the meaning of the `history` mode. Can you explain where the naming comes from?

PR #264, opened 2020-12-23
feat(useTransition): support for reactive durations

scottbedard:
This PR turns the `duration` property of `useTransition` from a `number` to a `MaybeRef<number>`. Duration changes have no effect on transitions already in progress, and only effect subsequent transitions. I also refactored the unit tests from relying on Jest's `done` to async / await. The nested `setTimeout` calls get difficult to read at even just a layer or two deep.

Also, congrats on tagging 4.x! Thank you for all your work building and maintaining such an awesome library!

### DIFF
--- a/packages/core/useTransition/index.ts
+++ b/packages/core/useTransition/index.ts
@@ -1,13 +1,13 @@
 import { useRafFn } from '../useRafFn'
-import { Ref, ref, watch } from 'vue-demi'
-import { clamp, isFunction } from '@vueuse/shared'
+import { Ref, ref, unref, watch } from 'vue-demi'
+import { clamp, isFunction, MaybeRef } from '@vueuse/shared'
 
 type CubicBezierPoints = [number, number, number, number]
 
 type EasingFunction = (x: number) => number
 
 interface TransitionOptions {
-  duration?: number
+  duration?: MaybeRef<number>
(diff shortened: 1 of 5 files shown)


packages/core/useTransition/index.ts
```diff
@@ -90,14 +90,15 @@ export function useTransition(source: Ref<number>, options: TransitionOptions =
     ? transition
     : createEasingFunction(transition)
 
+  let currentDuration = 0
```
antfu, line 93:
```suggestion
  let currentDuration = unref(duration)
```

Do we need this here?

scottbedard: @antfu This would be an unnecessary call to `unref`, and would have no effect on behavior. All of these variables are assigned their start values in the `watch` callback. I'm initializing them `0` just for the sake of casting to a number, we could do something like this if we wanted though...

```ts
let currentDuration: number
let diff: number
let endAt: number
let startAt: number
let startValue: number
```

But then they would all briefly have values of `undefined`. This wouldn't matter though since they all _will_ have values by the time the raf loop starts 🤷‍♂️ 

Whichever way you want to go here is fine with me, but I would recommend we call `unref` only once at the start of a transition.

antfu: That makes sense to me, thanks!

PR #274, opened 2021-01-02
feat(useScriptTag): new function

Tahul:
Hello, I needed to use an external <script> tag in one of my apps, and I love using your library.

I think <script> tag handling is a general use case in most apps that uses widgets and analytics tools.

I only find Vue plugins that handles this at application level, and I think handling it directly from the component that interacts with the script tag is more convenient in terms of performances.

This PR is a work in progress proposal for a useScriptTag hook.

Before writing the test suite, I would like to know if you would consider adding this feature to vueuse, or if I should make my own package with this feature.

Thanks for your time and amazing work on vueuse :)

PS: I searched for your Twitch account for the example @antfu, but couldn't find it, consider replacing mine by the one of your choice!


antfu:
Sounds great, please go ahead 👍

Tahul:
Hello @antfu ;

As described earlier in this PR, I implemented the useScriptTag hook.

I just finished writing the test suite for it, I think it needs your approval now.

I used this hook implementation in one of my apps and it went well so far.

If you find edge cases that you need me to fix, or if this implementation seem irrelevant to you, I would be glad to take all of your comments in consideration on further commits.

Thank you again for your amazing work and vueuse and all the time you'll be giving on reviewing the code of this PR.

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,126 @@
+import { isString, MaybeRef, tryOnMounted, tryOnUnmounted } from '@vueuse/shared'
+import { ref } from 'vue-demi'
+import { ConfigurableDocument, defaultDocument } from '../_configurable'
+
+export interface UseScriptTagOptions {
```
antfu, line 5: We can extend `ConfigurableDocument` here to have a single options object for configuration.

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  tryOnMounted(async() => {
+    if (loadOnMounted) await loadScript()
+  })
+
+  tryOnUnmounted(async() => {
+    await unloadScript()
+  })
+
+  return [scriptTag, loadScript, unloadScript] as const
```
antfu, line 125: VueUse prefers object returns whenever possible. 

Also, thinking we can expose the loading promise instance for users to use it with await

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,126 @@
+import { isString, MaybeRef, tryOnMounted, tryOnUnmounted } from '@vueuse/shared'
+import { ref } from 'vue-demi'
+import { ConfigurableDocument, defaultDocument } from '../_configurable'
+
+export interface UseScriptTagOptions {
+  loadOnMounted?: boolean
```
antfu, line 6:
```suggestion
  immediate?: boolean
```

Prefer this naming to align with Vue's options and other functions

packages/core/useScriptTag/index.ts
Tahul: Hello @antfu,

I updated the return type to an object as described in your comment.

What do you mean by exposing the loading promise instance ?

If users want to control the loading of the script themselves, they have to turn `immediate` parameter to false and then await on loadScript method (which is a Promise) anywhere in their code to trigger the script tag injection.

I don't see how I could expose the "tryOnMounted" promise to users, if you could explain me how you could achieve that I would be glad to implement it inside my PR.

Thank for your review comments.

antfu: Ah, never mind. I think people can use `await loadScript()` to do that 👍

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Load the script specified via `src`.
+   *
+   * @param waitForScriptLoad Whether if the Promise should resolve once the "load" event is emitted by the <script> attribute, or right after appending it to the DOM.
+   * @returns Promise<HTMLScriptElement>
+   */
+  const loadScript = (waitForScriptLoad = true): Promise<HTMLScriptElement> =>
+    new Promise((resolve, reject) => {
+      if (!document) {
```
antfu, line 33: We can just `resolve(false)` if `document` is underfined(SSR) instead of throwing errors

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,124 @@
...
+        reject(new Error('No document found!'))
+        return
+      }
+
+      if (!src) {
```
antfu, line 38: I'd leave this check to TypeScript, no need for this runtime error

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,124 @@
...
+      let shouldAppend = false
+
+      let el: HTMLScriptElement = <HTMLScriptElement>(
+        document.querySelector(`script[src="${src}"]`)
+      )
+
+      if (!el) {
+        el = document.createElement('script')
+        el.type = 'text/javascript'
+        el.async = true
```
antfu, line 52: Should we make this configurable? Also for `crossorigin` `defer` `referrerpolicy` etc.

https://www.w3schools.com/tags/tag_script.asp

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,124 @@
...
+        scriptTag.value = el
+        resolve(el)
+        return el
+      }
+
+      el.addEventListener('error', () =>
+        reject(new Error('Error while loading script!')),
+      )
+      el.addEventListener('abort', () =>
+        reject(new Error('Script load aborted!')),
```
antfu, line 66: We can bypass the event errors directly

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  tryOnMounted(async() => {
+    if (immediate) await loadScript()
+  })
+
+  tryOnUnmounted(async() => {
+    await unloadScript()
+  })
+
+  return { scriptTag, loadScript, unloadScript } as const
```
antfu, line 123:
```suggestion
  return { scriptTag, load, unload }
```

Preferring this to be more clean looking

packages/core/useScriptTag/index.ts
Tahul:
```
  async?: boolean
  type?: string
  crossOrigin?: 'anonymous' | 'use-credentials'
  referrerPolicy?: 'no-referrer' | 'no-referrer-when-downgrade' | 'origin' | 'origin-when-cross-origin' | 'same-origin' | 'strict-origin' | 'strict-origin-when-cross-origin' | 'unsafe-url'
  noModule?: boolean
  defer?: boolean
  ```
  
  I've added all these parameters, I think you're right we should make everything configurable.

Comment (antfu) [reply packages/core/useScriptTag/index.ts]: Great! Thanks

File: packages/core/useScriptTag/index.md [hunk 9]
```diff
@@ -0,0 +1,90 @@
...
+> Async script tag loading.
+
+## Usage
+
+```js
+import { useScriptTag } from '@vueuse/core'
+
+export default defineComponent({
+  setup() {
+    const [twitchScriptTag] = useScriptTag(
```
antfu, line 12: Might need to update the docs as well.

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      // Reset the <script> tag reference.
+      scriptTag.value = undefined
+
+      resolve(true)
+    })
+
+  // Try to load the script onMounted.
+  tryOnMounted(async() => {
+    if (immediate) await load()
+  })
```
antfu, line 134:
```suggestion
  if (immediate)
    tryOnMounted(load())
```

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,142 @@
...
+) {
+  const scriptTag = ref<HTMLScriptElement>()
+
+  /**
+   * Load the script specified via `src`.
+   *
+   * @param waitForScriptLoad Whether if the Promise should resolve once the "load" event is emitted by the <script> attribute, or right after appending it to the DOM.
+   * @returns Promise<HTMLScriptElement>
+   */
+  const load = (waitForScriptLoad = true): Promise<HTMLScriptElement|boolean> =>
```
antfu, line 43: We should have an internal promise instance here (singleton) to prevent calling load() twice.

packages/core/useScriptTag/index.ts
Tahul:
```
  // Try to load the script onMounted.
  if (immediate) tryOnMounted(async() => load())

  // Try to unload the script onUnMounted.
  tryOnUnmounted(async() => unload())
```

I came up with this, using "load()" as tryOnMounted parameter does not match the function signature.

I think we could update the "Fn" type to also allow promises, or create a new type instead.

This should be the subject of another PR / issue though.

Thanks again for your review.

packages/core/useScriptTag/index.ts
Tahul: I suggest you to review my updates in that commit: https://github.com/antfu/vueuse/pull/274/commits/fdee84562eb65bd20294d2d918501368be69615c

Thanks again for your review.

Tahul:
(I used force push to remove a commit that was not needed, sorry if that does not follow the contribution guidelines)

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,153 @@
...
+      document.querySelector(`script[src="${src}"]`)
+    )
+
+    // Script tag found, preparing the element for appending
+    if (!el) {
+      // Create element and set required attributes
+      el = document.createElement('script')
+      el.type = type
+      el.async = async
+      el.src = isString(src) ? src : src.value
```
antfu, line 72:
```suggestion
      el.src = unref(src)
```

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Async script tag loading.
+ *
+ * @see   {@link https://vueuse.js.org/useScriptTag}
+ * @param src
+ */
+export function useScriptTag(
+  src: MaybeRef<string>,
+  onLoaded: Function = () => {},
```
antfu, line 23:
```suggestion
  onLoaded: Fn = noop,
```

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,153 @@
...
+      resolveWithElement(el)
+    }
+
+    // Event listeners
+    el.addEventListener('error', event => reject(event))
+    el.addEventListener('abort', event => reject(event))
+    el.addEventListener('load', () => {
+      el.setAttribute('data-loaded', 'true')
+
+      if (onLoaded) onLoaded(el)
```
antfu, line 92:
```suggestion
      onLoaded(el)
```

Since we already set the default to `onLoaded` there is no need for this check

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,153 @@
...
+   * @returns Promise<boolean>
+   */
+  const unload = (): Promise<boolean> =>
+    new Promise((resolve) => {
+      // Check if document actually exists, otherwise reject the Promise.
+      if (!document) {
+        resolve(false)
+        return
+      }
+
```
antfu, line 130:
```suggestion

      loadingPromise = null

```

packages/core/useScriptTag/index.ts
```diff
@@ -0,0 +1,153 @@
...
+    async = true,
+    document = defaultDocument,
+    type = 'text/javascript',
+    crossOrigin,
+    referrerPolicy,
+    noModule,
+    defer,
+  }: UseScriptTagOptions = {},
+) {
+  const scriptTag = ref<HTMLScriptElement>()
```
antfu, line 35:
```suggestion
  const scriptTag = ref<HTMLScriptElement | null>(null)
```

I think it's better to have `null` instead of `undefined` here

antfu:
Thanks!

PR #275, opened 2021-01-04
chore: fix docs scrollbars

matias-capeletto:
scrollbars

1. style scrollbars for browsers that are not as nice as safari (maybe the non-hover color could be more prominent)
2. hide scrollbars in the code line highlight overlay (this is a bug in vitepress default theme)
3. there was a horizontal scrollbar in the body that I couldn't fix, but I think it is safe to just hide it.

scrollbars-after


antfu:
That was crazy on Windows 😅 Thanks a lot for the fix!

PR #279, opened 2021-01-05
feat: new functions - get, set, reactifyObject

antfu:
Probably went a bit too far with `reactify` 😅

Example

```ts
import { Math, JSON, console } from '@vueuse/reactified'

const { pow, sum, sqrt } = Math

//          _________
// for c = √ a² + b²
const a = ref(3)
const b = ref(4)
const c = sqrt(sum(pow(a, 2), pow(b, 2))) // 5

console.log(c) // no need for `.value`, and will log on changing
```


antfu:
After some consideration, I made this package a separate repo, with a better interface.

https://github.com/antfu/vueuse-reactified

PR #288, opened 2021-01-13
feat(templateRef): new function

aa900031:
Make Vue 2 use reactive $refs easy like in Vue 3


antfu:
I don't quite understand the context here, can you explain a bit? Thanks

aa900031:
For example have a project dependency are:
- vue: 2.6
- @vue/composition-api

Write component with JSX and `useResizeObserver`, in here have wrong to set <div> ref with `$message`, @vue/composition-api doesn't support that 
```ts
import { ref, defineComponent } from '@vue/composition-api'
import { useResizeObserver } from '@vueuse/core'

export default defineComponent({
  setup() {
    const $message = ref<HTMLElement | null>(null)
    useResizeObserver($message, (entries) => {
      const entry = entries[0]
      console.log('changed!');
    })
    
    return () => (
      <div class="container">
        // Not Work!
        <div ref="$message">
          Hi
        </div>
      </div>
    )
  },
})
```

antfu:
Maybe we can have the API looks like this

```js
import { defineComponent } from '@vue/composition-api'
import { templateRef } from '@vueuse/core'

export default defineComponent({
  setup() {
    const message = templateRef('message', null) // Ref<HTMLElement | null>

    useResizeObserver(message, () => { /* ... */ })
    
    return () => <div ref="message">
  },
})
```

WDYT?

aa900031:
It looks great, but that will change implement, that me think how to do it

antfu:
Also would like to have some unit-tests for this as it's really hard to just eye-bowing it. Let me know if you need any help. Thanks

aa900031:
@antfu It's all done, please review thanks

antfu:
Can you check the "Allow edit from maintainers" for this PR? Thanks.

aa900031:
OK 
截圖 2021-01-14 下午11 55 56

antfu:
I have updated the implementation, can you give it a try in your app to see if that matches what you expected? Thanks.

I will update the demo and docs later.

aa900031:
The implementation with `customRef` is amazing, Thanks.
And sorry for my bad demo and docs

antfu:
Not at all, that's a very good contribution (just me being a bit nitpicking about the overall consistency) Thank you 👍

aa900031:
It's right to overall consistency.
Thank you for the code, I learn a lot

antfu:
Thanks!

PR #293, opened 2021-01-18
useFullscreen demo: replace toogle with enter

mseele:
after the first "run" toggle will only work every second click. Using enter makes more sense.
Steps to reproduce toggle bug:
- Click "Go Fullscreen"
- Fullscreen is enabled
- Click ESC
- Fullscreen is disabled
- Click "Go Fullscreen"
- Nothing happens (toggle set's value to false)
- Click "Go Fullscreen"
- Fullscreen is enabled
- Click ESC
- ...

### DIFF
--- a/packages/core/useFullscreen/demo.vue
+++ b/packages/core/useFullscreen/demo.vue
@@ -3,7 +3,7 @@ import { ref } from 'vue-demi'
 import { useFullscreen } from '.'
 
 const el = ref(null)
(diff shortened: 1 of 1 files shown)


antfu:
This is more like a bug of the function itself as the ref should always reflecting the state of fullscreen. Will have a look later and I'd be grateful if someone would like to help out :)

PR #296, opened 2021-01-22
feat(useTimeoutFn): add callback args & state isPending

liaoliao666:


packages/shared/useTimeoutFn/index.ts
```diff
@@ -10,11 +10,12 @@ import { isClient } from '../utils'
  * @param immediate
  */
 export function useTimeoutFn(
-  cb: () => any,
+  cb: (...args: unknown[]) => any,
   interval?: number,
   immediate = true,
 ) {
   const isActive = ref(false)
+  const isPending = ref(false)
```
antfu, line 18: What is this for?

packages/shared/useTimeoutFn/index.ts
```diff
@@ -27,15 +28,19 @@ export function useTimeoutFn(
...
-  function start() {
+  function start(...args: unknown[]) {
     clear()
     isActive.value = true
+    isPending.value = true
     timer = setTimeout(() => {
+      isPending.value = false
       timer = null
-      cb()
+      // eslint-disable-next-line
```
antfu, line 42: Can you make this more specific by only disabling the rule that throws?

packages/shared/useTimeoutFn/index.ts
liaoliao666: sometimes i want to know it  whether in pending and i think the isActive is useless. maybe u could intead isPending of isActive
for example
```vue
<template>
  <button :disabled="isPending">
    <slot />
    <Spinner v-if="isPending" />
  </button>
</template>
```

packages/shared/useTimeoutFn/index.ts
liaoliao666: okay

packages/shared/useTimeoutFn/index.ts
antfu: I see, but looks like you forget to export it? Yeah, let's remove `isActive` in favor of `isPending`.

antfu (approved):

PR #299, opened 2021-01-22
feat(useJwt): new function

### LINKED ISSUE #269: Add useJwt function
I was thinking about a reactive way of handling a [JWT](https://jwt.io/introduction), since the token often expires/changes after a few minutes. The function should decode the `payload` part of the token and return it as a reactive object, providing the basic claims, which can be extended using generics.

I already implemented some parts of the described functionality in a non reactive way. It would be nice to be able to contribute a reactive solution to this project. What do you think @antfu ?

lstoeferle:
Hello there,

as discussed in [here](https://github.com/vueuse/vueuse/issues/269), I worked on a `useJwt` composable ✨ , which I would like to share. 

Closes #269 

Sorry for bad commit history... I definitely need to clean up my fork 😇


antfu:
Could you resolve the conflict and make the tests passing? Thanks.

lstoeferle:
Hey @antfu,
it seems there are some dependency issues. I could not figure out the root cause, it would be nice if you can assist me there, since I'm not that familiar with the new vitepress setup.

antfu:
Can you try to override the yarn.lock with the latest one on the `main` branch and do the installation again? Looks like your vote version does not align with what we have in the lock file (where I need to apply the patch to).

lstoeferle:
@antfu thanks for the hint, it seems this resolved the issue.

packages/integrations/useJwt/index.md
```diff
@@ -0,0 +1,61 @@
...
+
+wrapper for [`jwt-decode`](https://github.com/auth0/jwt-decode)
+
+## Install
+
+```bash
+# npm project
+npm install jwt-decode
+# yarn project
+yarn install jwt-decode
```
antfu, line 15:
```suggestion
npm install jwt-decode
```

Let's keep it simple

packages/integrations/useJwt/index.ts
```diff
@@ -0,0 +1,28 @@
+import jwt_decode, { JwtDecodeOptions, JwtHeader, JwtPayload } from 'jwt-decode'
+import { computed, Ref } from 'vue-demi'
+
+/**
+ * Reactive decoced jwt token.
+ *
+ * @see {@link https://vueuse.js.org/integrations/useJwt/}
+ * @param jwt
+ */
+export function useJwt<Payload extends object = JwtPayload, Header extends object = JwtHeader>(encodedJwt: Ref<string>) {
```
antfu, line 10:
```suggestion
export function useJwt<Payload extends object = JwtPayload, Header extends object = JwtHeader>(encodedJwt: MaybeRef<string>) {
```

packages/integrations/useJwt/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  const decodeWithFallback = <T extends object>(encodedJwt: string, options?: JwtDecodeOptions) => {
+    try {
+      return jwt_decode<T>(encodedJwt, options)
+    }
+    catch (err) {
```
antfu, line 15: Not sure about this, is the error doesn't matter, or should we let users to capture this error if need?

packages/integrations/useJwt/index.ts
```diff
@@ -0,0 +1,28 @@
...
+      return Object.assign({})
```
antfu, line 16:
```suggestion
      return {}
```

Any reason for this?

packages/integrations/useJwt/index.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+  }
+
+  const header = computed((): Header => decodeWithFallback<Header>(encodedJwt.value, { header: true }))
```
antfu, line 20:
```suggestion
  const header = computed(() => decodeWithFallback<Header>(encodedJwt.value, { header: true }))
```

Wouldn't the type auto inferred?

antfu (changes requested):

packages/integrations/useJwt/index.ts
lstoeferle: We could also return an the error beside payload and header.

packages/integrations/useJwt/index.ts
lstoeferle: I am facing type issues when returning a simple `{}`.

image

lstoeferle: I am facing type issues when returning `{}`. E.g. 
```javascript
// Property 'alg' does not exist on type '{}'.
header.value.alg
```

packages/integrations/useJwt/index.ts
lstoeferle: I would keep the fallback and maybe add an optional callback function, which enables users to react on errors. IMO an additional `error` return param does not make sense, since checking this param requires more effort than passing a simple callback. What do you think?

antfu: Sounds good :)

packages/integrations/useJwt/index.ts
antfu: Could just use `as` in TypeScript which is compile time. `Object.assign` is runtime.

lstoeferle:
Found a typo in the docs, will fix it today

antfu:
@lstoeferle Can you check the "allow edit from maintainers"? I'd like to push some commits

lstoeferle:
> @lstoeferle Can you check the "allow edit from maintainers"? I'd like to push some commits

Should be enabled

antfu:
Made a few changes, are they make sense to you @lstoeferle?

lstoeferle:
> Made a few changes, are they make sense to you @lstoeferle?

Looks fine to me. Only the `null` check looks confusing, since the fallback is `{}` instead of `null` 🤔

antfu:
Changed to `null` intentionally as I think users should better aware of when it's failed, while you can always set the fallback to `{}` if you want. Or do you have any particular reason to use `{}` over `null`?

lstoeferle:
Im fine with `null` 😉👍

antfu (approved):

PR #307, opened 2021-01-27
fix: bug in useMediaQuery demo causing demo to not work properly

jacobclevenger:
There currently exists a bug in the [`useMediaQuery` demo](https://vueuse.js.org/core/useMediaQuery/) that causes the demo to not work properly.

This pull request fixes the bug by changing the media query from `(min-width = 1024px)` to `(min-width: 1024px)` and `(prefers-color-scheme = dark)` to `(prefers-color-scheme: dark)`. It also fixes a typo in the `useMediaQuery` documentation that causes the link to the `Media Query` MDN documentation link to not work.

### DIFF
--- a/packages/core/useMediaQuery/demo.vue
+++ b/packages/core/useMediaQuery/demo.vue
@@ -3,8 +3,8 @@ import { computed, reactive } from 'vue-demi'
 import { useMediaQuery } from '.'
 import YAML from 'js-yaml'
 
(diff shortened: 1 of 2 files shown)


antfu:
That was probably introduced by my regex rewrite from tsx to Vue 😅, thanks a lot for the recent PRs!

antfu (approved):

jacobclevenger:
No problem, glad I could help 😄

PR #310, opened 2021-01-28
feat(controlledRef): new function

antfu:
[Rendered Docs](https://github.com/vueuse/vueuse/blob/feat/controlledRef/packages/shared/controlledRef/index.md)

/cc @matias-capeletto @Tahul love to hear how would you think about this, thanks!


matias-capeletto:
Looks useful! 👍🏼 

About the API, I think it is good to have the boolean option in case the trigger/track needs to be defined dynamically. But maybe we could also provide explicit methods for other cases?

```js
counter.untrackedGet()
counter.untriggeredSet(3)
```

A little bit verbose, but some people may prefer this. I like `untrackedGet`. For `untriggeredSet`, I think that it is a bit long, maybe `ignoredSet` following the `ignorableWath` naming could be an option.

antfu:
(I am thinking `peek` for `untriggeredSet` 😄

`ignoredSet` sounds good

matias-capeletto:
😄 I also thought of `peek`, but for `untrackedGet`, really tempting to name it like that!
And maybe `lay` for `ignoredSet`... but my being explicit hat was stronger and couldn't propose that 😉

matias-capeletto:
Just one thing I didn't like about `ignoredSet` is that it reads like it will be a noop. I also played with `silentSet` because of this (some idea as with lay, that is defined as: put (something) down gently or carefully).

antfu:
IDK, can't find a good pair for `peek` 🤣

- `secertlySet`
- `slientlySet`
- `slientSet`
- `change`
- `swap`

Issue #320, opened 2021-02-04
The binding ref is different in Vue2 and Vue3


likeswinds:
```
const target = templateRef('target') 
```
`target.value`  is dom in vue3  
`target.value`  is VueComponent  in vue2 
`target.value.$el`  is dom in vue2 

Can it be align with VUE3

```
  const elements = event.composedPath() 
// That was one of the bug.  onClickOutside  in  vue2 
 if (targetRef.value.$el === event.target || elements.includes(targetRef.value.$el!))
      return
```

antfu:
Can you share the template you are using? Depends on what you set `ref` at, the template could be a dom element nor a Vue component

likeswinds:
I made a mistake，The binding ref are the same In Vue2 and Vue3 ，At the same time ，You should use `target.value.$el` instead of `target.value` to get the DOM
``` vue
<template>
  <img alt="Vue logo" src="./assets/logo.png" />
  <HelloWorld ref="modal" msg="Welcome to Your Vue.js App" />
</template>

<script>
import { onMounted, defineComponent } from "vue";
import HelloWorld from "./components/HelloWorld.vue";
import { onClickOutside, templateRef } from "@vueuse/core";

export default defineComponent({
  name: "App",
  components: {
    HelloWorld,
  },
  setup() {
    const modal = templateRef("modal");
    onMounted(() => {
      console.log(modal.value);
      console.log(modal.value.$el);
    });
    onClickOutside(modal, (event) => {
      console.log(event);
      modal.value = false;
    });
  },
});
</script>
```

antfu:
If you understand currently, it's resolved for you right? Closing for now.

likeswinds:
Should I create a new issue instead？
`onClickOutside` is not work When binding ref in VueComponent .

likeswinds:
```
 const listener = (event: EventType) => {
    if (!targetRef.value)
      return

    const elements = event.composedPath()
    if (targetRef.value === event.target || elements.includes(targetRef.value!))
      return
   // add this
    if (targetRef.value.$el === event.target || elements.includes(targetRef.value.$el!))
      return
    handler(event)
  }
```

antfu:
@likeswinds valid point, you can create a PR directly. Thanks.

PR #322, opened 2021-02-05
Binding ref to VueComponent does not working

likeswinds:
#320

### DIFF
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -31,7 +31,7 @@ export function onClickOutside(
       return
 
     const elements = event.composedPath()
-    const el = targetRef.value.$el || targetRef.value
+    const el = targetRef.value?.$el || targetRef.value
     if (el === event.target || elements.includes(el))
       return
 
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -31,7 +31,8 @@ export function onClickOutside(
       return
 
     const elements = event.composedPath()
-    if (targetRef.value === event.target || elements.includes(targetRef.value!))
+    const el = targetRef.value.$el || targetRef.value
+    if (el === event.target || elements.includes(el))
       return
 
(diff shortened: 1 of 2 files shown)


packages/core/useMouseInElement/index.ts
```diff
@@ -39,7 +39,7 @@ export function useMouseInElement(
     stop = watch(
       [targetRef, x, y],
       () => {
-        const el = targetRef.value
+        const el = targetRef.value.$el || targetRef.value
```
antfu, line 42:
```suggestion
        const el = targetRef.value?.$el || targetRef.value
```

packages/core/onClickOutside/index.ts
```diff
@@ -31,7 +31,8 @@ export function onClickOutside(
       return
 
     const elements = event.composedPath()
-    if (targetRef.value === event.target || elements.includes(targetRef.value!))
+    const el = targetRef.value.$el || targetRef.value
```
antfu, line 34:
```suggestion
    const el = targetRef.value?.$el || targetRef.value
```

likeswinds:
image
I can't do it this way @antfu 
 ``` 
const el = (targetRef.value as any).$el || targetRef.value
```

**Comment by likeswinds** [comment]
A similar problem exists in all of these functions，such as `useElementVisibility`,`useElementBounding`,`useElementSize`,`useMutationObserver`,`useCssVar` and so on 
Or I should program it this way 
``` vue
const modal = templateRef('modal')
onMounted(() => {
  onClickOutside(modal.value.$el, event => {
    console.log(event)
  })
})
```
Here are some code snippets
```
// useElementVisibility , 32 (row)
const rect = element.value.getBoundingClientRect()  // element.value.$el.getBoundingClientRect() 
```
```
// useResizeObserver, 69(row)
observer!.observe(newValue, observerOptions)
```

antfu:
@likeswinds Thanks for the feedback, now it's supported since v4.1

PR #325, opened 2021-02-09
upgrade useWebSocket.

shuperry:
- enable send ping msg.
- enable reConnect.

PR #326, opened 2021-02-09
feat(useWebSocket)!: rework useWebSocket

### LINKED ISSUE #283: Websocket retry connection strategy
In case of problems with the Internet, a mechanism for reconnecting to the server is required

antfu:
BREAKING CHANGES:
- the `ws` option in the return of `useWebSocket` is not a ref instead of plain WebSocket instance
- `useWebSocket`'s `send` will now buffers data before the connection established instead of discarding sliently
- `useWebSocket`'s returned `state` is renamed to `status`

----

Continued from the work in #325 by @shuperry. Close #283.

> Note: This PR will be held until v4.2

### DIFF
--- a/indexes.json
+++ b/indexes.json
@@ -685,7 +685,7 @@
       "package": "core",
       "docs": "https://vueuse.js.org/core/useWebSocket/",
       "category": "Misc",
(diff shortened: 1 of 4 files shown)


shuperry:
Thanks, this is more like what i want, sorry that i wasn't familiar with existing api.

Issue #327, opened 2021-02-11
Function Proposal: useFetch


jacobclevenger:
I'm proposing a new function called `useFetch`. I think maybe the API could be similar to `useAxios` but for `fetch` since people may not want to add Axios to their project.

I'm thinking something like this

```ts
const { finished, response, error, data } = useFetch('http://example.com', { options })
```

I imagine it will return something like this
```ts
export interface FetchOptions {
  // Indicates if the fetch request has completed
  finished: boolean

  // The raw response from request
  response: Response | null

  // The error if there was one during the request
  error: any

  // Either the JSON or Text response of the request
  data: object | string
}
```

I've started implementing this in a personal project but thought it might fit well here.

Some of the features I'd like to include are the following:

- **Using an Object for headers**
  This would only make a slight adjustment to the current fetch options. Basically it would allow the user to pass in the normal options for headers as well as an object that would automatically be converted to a Headers object. Although perhaps you'd like us to avoid changing the default options, let me know what you think is best if you like this proposal.

```ts
export interface FetchConfig extends Omit<RequestInit, 'headers'> {
  headers: HeadersInit | ObjectHeaders
}
```

- **Returning either json or text in the data object**
  Typically when using fetch you will have to do something like `fetch().then(response => response.json()).then(json => ...)` but this function would read the `content-type` headers and automatically assign the `data` ref to what the response is, either text or json

Anyway let me know what you think of this feature. If you like it I'd be happy to make a PR. Thanks 😄

antfu:
Sounds cool, please go ahead. Would love to have a ref showing the status of the request and abort it as well. Thanks

https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort

dheimoz:
Hey guys. Thanks for the great work. I wonder if you have seen https://github.com/sindresorhus/ky

A wrapper or a plugin would be excellent and will save a few hours of your precious time.

jacobclevenger:
@antfu Sounds good, I'll add the status ref and look into the Abort controllers

PR #330, opened 2021-02-12
feat(useFetch): new function

jacobclevenger:
Here is the initial implementation of useFetch, as mentioned in #327.

**General Feature Overview**
- Abort Requests if the browser supports it
- Auto refetch requests if the url is a ref and the user has enabled autoRefetch
- Reactive status, fetching state, finished stated, error, and canAbort
- Ability to prevent the first initial fetch and manually call the fetch with the `execute` function

The signature for `useFetch` provides multiple overrides so it can be used in any of the following ways

```ts
useFetch(url: MaybeRef<string>)
useFetch(url: MaybeRef<string>, useFetchOptions: UseFetchOptions)
useFetch(url: MaybeRef<string>, options: RequestInit, useFetchOptions?: UseFetchOptions)
```

Let me know what you think and if you would like anything changed!


packages/core/useFetch/index.ts
```diff
@@ -0,0 +1,161 @@
...
+interface UseFetchReturn {
+  /**
+   * Indicates if the fetch request has finished
+   */
+  isFinished: Ref<boolean>
+
+  /**
+   * The status of the fetch request
+   */
+  status: Ref<number | null>
```
antfu, line 13: Do you mean the `statusCode` of the response here?

packages/core/useFetch/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  execute: Fn
+}
+
+interface UseFetchOptions {
+  /**
+   * Will automatically run fetch when `useFetch` is used
+   *
+   * @default true
+   */
+  autoFetch?: boolean
```
antfu, line 57:
```suggestion
  immediate?: boolean
```

packages/core/useFetch/index.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  /**
+   * Will automatically refetch when the URL is changed if the url is a ref
+   *
+   * @default false
+   */
+  autoRefetch?: MaybeRef<boolean>
```
antfu, line 64:
```suggestion
  refetch?: MaybeRef<boolean>
```

packages/core/useFetch/index.ts
```diff
@@ -0,0 +1,161 @@
...
+export function useFetch(url: MaybeRef<string>, useFetchOptions: UseFetchOptions): UseFetchReturn
+export function useFetch(url: MaybeRef<string>, options: RequestInit, useFetchOptions?: UseFetchOptions): UseFetchReturn
+
+export function useFetch(url: MaybeRef<string>, ...args: any[]): UseFetchReturn {
+  const supportsAbort = typeof AbortController === 'function'
+
+  let fetchOptions: RequestInit = {}
+  let options: UseFetchOptions = { autoFetch: true, autoRefetch: false }
+
+  if (args[0]) {
```
antfu, line 77: Let's check the overload with `args.length` instead of relying on their properties.

packages/core/useFetch/index.ts
```diff
@@ -0,0 +1,161 @@
...
+        const contentType = fetchResponse.headers.get('content-type')
+
+        if (contentType && contentType.includes('application/json'))
+          fetchResponse.json().then(json => data.value = json)
+        else
+          fetchResponse.text().then(text => data.value = text)
+
+        // see: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/
+        if (!fetchResponse.ok)
+          throw new Error(fetchResponse.statusText || 'Something went wrong!')
```
antfu, line 131:
```suggestion
          throw new Error(fetchResponse.statusText)
```

Would like to avoid text strings in VueUse.

jacobclevenger:
Everything you mentioned should be fixed now

packages/core/useFetch/index.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  const abort = () => {
+    if (supportsAbort && controller instanceof AbortController)
+      controller.abort()
+  }
+
+  const execute = () => {
+    isFetching.value = true
+    isFinished.value = false
+    error.value = null
```
antfu, line 110:
```suggestion
    error.value = null
    statusCode.value = null
```

packages/core/useFetch/index.ts
```diff
@@ -0,0 +1,161 @@
...
+interface UseFetchReturn {
+  /**
+   * Indicates if the fetch request has finished
+   */
+  isFinished: Ref<boolean>
+
+  /**
+   * The statusCode of the HTTP fetch response
+   */
+  status: Ref<number | null>
```
antfu, line 13:
```suggestion
  statusCode: Ref<number | null>
```

packages/core/useFetch/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  const data = ref<string | object | null>(null)
+
+  const canAbort = computed(() => {
+    return supportsAbort && isFetching.value
+  })
+
+  let controller: AbortController | null
+
+  const abort = () => {
+    if (supportsAbort && controller instanceof AbortController)
```
antfu, line 103:
```suggestion
    if (supportsAbort && controller)
```

packages/core/useFetch/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  const status = ref<number | null>(null)
+  const response = ref<Response | null>(null)
+  const error = ref<any>(null)
+  const data = ref<string | object | null>(null)
+
+  const canAbort = computed(() => {
+    return supportsAbort && isFetching.value
+  })
+
+  let controller: AbortController | null
```
antfu:
```suggestion
  let controller: AbortController | undefiend
```
So it matches with the initial value

packages/core/useFetch/index.ts
```diff
@@ -0,0 +1,161 @@
...
+      .then((fetchResponse) => {
+        response.value = fetchResponse
+        status.value = fetchResponse.status
+
+        const contentType = fetchResponse.headers.get('content-type')
+
+        if (contentType && contentType.includes('application/json'))
+          fetchResponse.json().then(json => data.value = json)
+        else
+          fetchResponse.text().then(text => data.value = text)
```
antfu, line 129: Could we make this usage simpler with some API?

packages/core/useFetch/index.ts
```diff
@@ -0,0 +1,161 @@
...
+        isFetching.value = false
+      })
+  }
+
+  if (isRef(url)) {
+    if (isRef(options.refetch))
+      watch([options.refetch, url], () => unref(options.refetch) && execute())
+    else if (options.refetch)
+      watch(url, () => execute())
+  }
```
antfu, line 147:
```suggestion
watch(
()=>{
unref(url)
unref(options.refetch)
}, 
() => unref(options.refetch) && execute()
)
```

Sorry for the formatting

packages/core/useFetch/index.ts
jacobclevenger: Oh right I should have renamed that, fixed now

packages/core/useFetch/index.ts
jacobclevenger: Oh yeah fixed that in my last commit

packages/core/useFetch/index.ts
jacobclevenger: No problem, I can fix it in a sec

packages/core/useFetch/index.ts
jacobclevenger: Hmm, is there something out there that would allow us to simplify it? I'm not aware of anything, but if you know I'll update it

packages/core/useFetch/index.ts
jacobclevenger: This doesn't actually seem to work. Can you watch an `unref`? I was under the impression it just gave you back the raw value if it was a ref

antfu: Did you test it? It's not watching an `unref` but the "accessing". `unref` is just a unified way to accessing here, the return value doesn't matter. Should work I think.

jacobclevenger: Ok it looks like it does work, but the watch needs to have `deep: true` set

packages/core/useFetch/index.ts
antfu: Not really. But maybe keep your implementation and make a `useFetchJson` wrapper with preconfigured header? Moreover, something like `usePost` and `usePostJson`?

Another approach might be similar to [`ky`](https://github.com/sindresorhus/ky)

```ts
useFetch(URL, {}).json()
```

WDYT?

packages/core/useFetch/index.ts
antfu: Em? I don't think it's needed... 🤔

packages/core/useFetch/index.ts
jacobclevenger: Hmm doing more research into the fetch API they actually have more than just `text()` and `json()`, https://developer.mozilla.org/en-US/docs/Web/API/Body#methods

If we were to make a wrapper for all of these we would have `useFetch`, `useFechJson`, `useFetchBlob`, `useFetchFormData`, and `useFetchArrayBuffer`. But I feel like most of these would not be used very often and it would create quite a few methods. Perhaps another option would be to have an option in the config to set how the data is parsed. Something like this perhaps?

```ts
useFetch(url, {
  // Not sure what a good name for this would be
  // It could be defaulted toauto which would try to automatically detect the type from the response headers
  readAs: 'auto' //could also be json, formData, blob, arrayBuffer, or text
})
```

I do like the idea of wrappers for `usePost` and `usePostJson`. If we do that would we also want ones for form-data and multipart form data?

packages/core/useFetch/index.ts
jacobclevenger: Weird, it only seems to work when I set `deep: true`. Not really sure why

jacobclevenger: I just created a repo to demonstrate the issue. It is a bit of a contrived example, but it should show the issue. It is in vite so you should be able to just clone it and run it

antfu:
@jacobclevenger got a few ideas of the api design, will give a try when I am back from vacation. Please wait for this a while, thanks.

jacobclevenger:
> @jacobclevenger got a few ideas of the api design, will give a try when I am back from vacation. Please wait for this a while, thanks.

Alright no problem, I look forward to it. Enjoy your vacation!

antfu:
Made some changes to the API (haven't tested it out, and maybe we should have some unit tests), wondering what you think? Thanks!

jacobclevenger (commented):
API design looks good, I'll test out functionality and see about some tests

packages/core/useFetch/index.ts
```diff
@@ -213,21 +211,15 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
...
-    }
-    return undefined
+    config.method = method
+    config.payload = payload
+    config.payloadType = payloadType || typeof payload === 'string' ? 'text' : 'json'
+    return shell as any
   }
 
   const setType = (type: DataType) => () => {
-    if (initialized) {
```
jacobclevenger: Since initialized was not called until the execute function was called I opted to remove this which fixed a bug when calling `.json()` or any of the other methods. Is there a particular reason to have this or is it okay to remove?

packages/core/useFetch/index.test.ts
```diff
@@ -0,0 +1,24 @@
...
+  const urls = {
+    get: 'https://httpbin.org/get',
+    post: 'https://httpbin.org/get',
+    error: 'https://httpbin.org/status/500',
+  }
+
+  test('basic get request', async() => {
+    const { statusCode } = useFetch(urls.get)
+
+    /**
```
jacobclevenger, line 13: Would like some input here on how to go about writing the tests for this type of problem described below

jacobclevenger, line 13: Would like some input on the best way to go about writing these tests in this situation

jacobclevenger: The comment starts on line 13

packages/core/useFetch/index.ts
antfu: Ah, it's `!initialized` 😅

jacobclevenger: Haha righto just fixed it

packages/core/useFetch/index.test.ts
antfu: We can make the `execute()` async. And they can also use `when`

```ts
await when(isFinished)
```

jacobclevenger: Sounds good, I'll make those changes and create some basic tests

packages/core/useFetch/index.test.ts
```diff
@@ -8,17 +8,10 @@ describe('useFetch', () => {
   }
 
   test('basic get request', async() => {
```
jacobclevenger, line 10: Looks like jest doesn't support fetch, so it seems like we need to mock it. There are a few libraries online that provide fetch mocks, like this one https://www.npmjs.com/package/jest-fetch-mock, but I wanted to get some input from you before I did anything

antfu: Sure, go for it!

jacobclevenger:
@antfu Added in some basic tests and fixed a few bugs with the useFetch function. Let me know what you think and if you'd like anymore tests 😄

antfu (approved):

PR #331, opened 2021-02-13
fix(useEventListener): return cleanup fn

colgin:
useEventListener return stop fn, which is not defined. see #332


antfu:
Thanks for spotting this, but the fix is a bit different, see cde4563f80c25b23aaefb3fcf75ed16336bfd0c0

PR #333, opened 2021-02-15
feat(useSwipe): new function

### LINKED ISSUE #324: New function proposal: useSwipe
Hello @antfu,

I'm back with another function proposal 😄. Few weeks ago I worked on a composable which handles swipe detection based on the `touchStart`, `touchMove` and `touchEnd` event.

```javascript
export enum SwipeDirection {
  UP = "UP",
  DOWN = "DOWN",
  RIGHT = "RIGHT",
  LEFT = "LEFT",
}

export function useSwipe(
  el: MaybeRef<HTMLElement | null>,
  threshold: number,
  onSwipeEnd: (direction: SwipeDirection, range: number) => void,
  onSwipe?: (e: TouchEvent, rangeX: number, rangeY: number) => void
)
```

I can open a PR if you think this will add value to your library 😄

lstoeferle:
Hello again,

this is the implementation of my `useSwipe` function proposal ✨ 

Kind regards,
Lukas

Closes #324


packages/core/useSwipe/index.ts
```diff
@@ -0,0 +1,119 @@
...
+ * @param target
+ * @param options
+ */
+export function useSwipe(
+  target: MaybeRef<HTMLElement | null | undefined>,
+  options?: {
+    threshold?: number
+    onSwipe?: (e: TouchEvent) => void
+    onSwipeEnd?: (e: TouchEvent, direction: SwipeDirection) => void
+  },
```
antfu, line 23:
```suggestion
  } = {},
```

packages/core/useSwipe/index.ts
```diff
@@ -0,0 +1,119 @@
...
+    else { e.preventDefault() }
+  }
+
+  const onTouchEnd = (e: TouchEvent) => {
+    isSwiping.value = false
+    if (direction.value)
+      onSwipeEnd?.(e, direction.value)
+  }
+
+  const addListeners = (el: HTMLElement | undefined | null) => {
```
antfu, line 76: You can use `useEventListener` which will auto unregister on unmount and target changes.

packages/core/useSwipe/index.ts
```diff
@@ -0,0 +1,119 @@
...
+    threshold?: number
+    onSwipe?: (e: TouchEvent) => void
+    onSwipeEnd?: (e: TouchEvent, direction: SwipeDirection) => void
+  },
+) {
+  const targetRef = ref(target)
+  const { threshold, onSwipe, onSwipeEnd } = options || {}
+
+  const coordsStart = ref([0, 0])
+  const coordsEnd = ref([0, 0])
```
antfu, line 29: Prefer `reactive({ x: 0, y: 0 })` here. We can also expose it.

packages/core/useSwipe/index.ts
```diff
@@ -0,0 +1,119 @@
...
+      }
+    },
+    { immediate: true },
+  )
+
+  return {
+    isSwiping,
+    direction,
+    lengthX: computed(() => Math.ceil(diffX.value)),
+    lengthY: computed(() => Math.ceil(diffY.value)),
```
antfu, line 117: Why do we need ceil? I think we can expose them as-is and for users to do the calculation.

packages/core/useSwipe/index.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      addListeners(el)
+
+      cleanup = () => {
+        removeListeners(el)
+        cleanup = noop
+      }
+    },
+    { immediate: true },
+  )
```
antfu, line 111:
```suggestion
```

This is no need with `useEventListener`

packages/core/useSwipe/index.ts
userquin: beware using passive event handlers at least on `touchstart`, `touchmove` and  `touchend`: on chromium based it not using passive event handler, swipe effect will freeze the window a few miliseconds

userquin: you can see an examle on: https://github.com/motss/app-datepicker/blob/main/src/tracker.ts

lstoeferle: I've now added `preventScrolling` to the composable options. The composable should also use `passive` now, as far the `preventScrolling` is `false`.

userquin: why not using a simplest way to detect `passive` event handler support? the function is redundant, can be inlined and using typescript definition, you dont need to defineProperty:
```typescript
let supportsPassive = false;
const fn = () => { /* empty listener */ };
const optionsBlock: AddEventListenerOptions = {
  get passive() {
    supportsPassive = true;
    return false;
  }
};
document.addEventListener('x', fn, optionsBlock);
document.removeEventListener('x', fn);
```

userquin: protect it when using SSR checking for document...

packages/core/useSwipe/index.ts
```diff
@@ -1,7 +1,8 @@
-import { computed, onBeforeMount, onBeforeUnmount, ref, unref, watch } from 'vue-demi'
 import { noop } from './../../shared/utils/is'
```
antfu, line 1:
```suggestion
import { noop } from '@vueuse/shared'
```

packages/core/useSwipe/index.ts
```diff
@@ -1,7 +1,8 @@
...
+import { computed, reactive, ref } from 'vue-demi'
 
 import { MaybeRef } from '..'
```
antfu, line 4:
```suggestion
import { MaybeRef } from '@vueuse/shared'
```

packages/core/useSwipe/index.ts
```diff
@@ -15,21 +16,22 @@ export enum SwipeDirection {
...
-  target: MaybeRef<HTMLElement | null | undefined>,
-  options?: {
+  target: MaybeRef<EventTarget>,
+  options: {
     threshold?: number
+    preventScrolling: boolean
     onSwipe?: (e: TouchEvent) => void
     onSwipeEnd?: (e: TouchEvent, direction: SwipeDirection) => void
-  },
+  } = { preventScrolling: false },
```
antfu:
```suggestion
  } = {},
```

it will be overridden when providing any of the options, should do it in the spreading below

packages/core/useSwipe/index.ts
```diff
@@ -15,21 +16,22 @@ export enum SwipeDirection {
...
 ) {
-  const targetRef = ref(target)
   const { threshold, onSwipe, onSwipeEnd } = options || {}
```
antfu, line 27:
```suggestion
  const { threshold, onSwipe, onSwipeEnd, preventScrolling = true } = options || {}
```

antfu: And why not being passive by default?

lstoeferle: passive by default would be `preventScrolling` = `false` ;)

antfu: how about just name this option to `passive` and default to true?

packages/core/useSwipe/index.ts
lstoeferle: Nice, thanks for sharing :)

antfu:
Made a few refactor, WDYT @lstoeferle?

userquin:
@antfu return also `stop` from `useEventListener` if somewho wants to use it on demand.

lstoeferle:
> Made a few refactor, WDYT @lstoeferle?

@antfu looks awesome, I was afraid touching this `useEventListener`, since it's used in several composables. Don't want to break things. Thanks for optimizing my code, this looks really good and readable now 👏 (nice to learn new coding styles and patterns). There is one minor thing I will change now, you put 50 as default `threshold`, but you added 0 in the interface declaration comments. I will change it to 50 also and try to add the suggestion from @userquin 👍

userquin:
One last question: `isSwiping`, `coordsStart` and `coordsEnd` shouldn't be readonly (they are the source of truth)?

userquin:
@antfu something pending here blocking merge?

antfu (approved):

userquin:
something is wrong:
imagen

userquin:
@antfu should destructuring with `...` before `useEventListener` => `const stops = [...useEventListener`

userquin:
@antfu `useEventListener` is wrong, `stop` doesn`t exists:

```typescript
export function useEventListener(...args: any[]) {
  let target: MaybeRef<EventTarget> | undefined
  let event: string
  let listener: any
  let options: any

  if (isString(args[0])) {
    [event, listener, options] = args
    target = defaultWindow
  }
  else {
    [target, event, listener, options] = args
  }

  if (!target)
    return

  let cleanup = noop

  watch(
    () => unref(target),
    (el) => {
      cleanup()
      if (!el)
        return

      el.addEventListener(event, listener, options)

      cleanup = () => {
        el.removeEventListener(event, listener, options)
        cleanup = noop
      }
    },
    { immediate: true },
  )

  tryOnUnmounted(stop)

  return stop
}
```

userquin:
created issue https://github.com/vueuse/vueuse/issues/335#issue-812610976

PR #339, opened 2021-02-22
feat(useWebcam): use getUserMedia to read web camera stream

YunYouJun:
## useWebcam

Use [MediaDevices.getUserMedia()](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia) to load camera stream for HTMLVideoElement.

`settings` can only be got after call `getUserMedia` in `mounted`. So `getUserMedia` is exported.


antfu:
I don't see how this helps with Vue Composition API from the demo, can you explain a bit? Otherwise, I think it's more like a general wrapper for Web APIs rather than a Vue composable.

YunYouJun:
Yeah. Initially I tried to use `onMounted` to automatically mount `getUserMedia` and `watch` `isFront` to toggle stream, but the exposed `setting` can only be obtained after `getUserMedia`. So i exported them.
Sorry, I didn't think of a better way to handle it. I will close it.

PR #341, opened 2021-02-23
fix(useAuth): fix imports to use only needed SDK & fixed typo in docs

jacobclevenger:
This fixes an issue with the `useAuth` function where it would import more from firebase than needed causing warnings in the console. It also fixes a small type in the docs for `useAuth`

The console error looked like this
image

### DIFF
--- a/packages/firebase/useAuth/index.ts
+++ b/packages/firebase/useAuth/index.ts
@@ -1,5 +1,6 @@
 import { computed, ComputedRef, ref, Ref } from 'vue-demi'
-import firebase from 'firebase'
+import firebase from 'firebase/app'
+import 'firebase/auth'
 
 export interface FirebaseAuthOptions {
   isAuthenticated: ComputedRef<boolean>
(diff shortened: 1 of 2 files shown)


packages/firebase/useAuth/index.ts
```diff
@@ -1,5 +1,6 @@
 import { computed, ComputedRef, ref, Ref } from 'vue-demi'
-import firebase from 'firebase'
+import firebase from 'firebase/app'
+import 'firebase/auth'
```
antfu, line 3: This introduced a side effect to this package which I would like to avoid. Is this possible to be done on user side?

jacobclevenger: Do you mean the 'import 'firebase/auth' line? If so then we can probably remove that and just keep `import firebase from 'firebase/app` line? If they are using the `useAuth` function then they will end up having to import `firebase/auth` anyway, so we should be able to safely remove it. Does that work for you?

antfu: Yep, thanks!

antfu (approved):

PR #342, opened 2021-02-23
feat(useAxios): axios instance support, close #273

### LINKED ISSUE #273: useAxios instances support
Any plans for adding a useAxios instances support?

jacobclevenger:
Adds support for Axios instances as mentioned in issue #273. 

Uses the following overrides so that it works with the existing API without breaking it.

```ts
export function useAxios<T = any>(url: string, config?: AxiosRequestConfig): UseAxiosReturn<T>
export function useAxios<T = any>(url: string, instance?: AxiosInstance): UseAxiosReturn<T>
export function useAxios<T = any>(url: string, config: AxiosRequestConfig, instance: AxiosInstance): UseAxiosReturn<T>
```

### DIFF
--- a/packages/integrations/useAxios/index.ts
+++ b/packages/integrations/useAxios/index.ts
@@ -1,5 +1,17 @@
 import { Ref, ref } from 'vue-demi'
-import axios, { AxiosError, AxiosRequestConfig, AxiosResponse, CancelTokenSource } from 'axios'
+import axios, { AxiosError, AxiosRequestConfig, AxiosResponse, CancelTokenSource, AxiosInstance } from 'axios'
+
+interface UseAxiosReturn<T> {
+  response: Ref<AxiosResponse<T> | undefined>
+  data: Ref<T | undefined>
+  finished: Ref<boolean>
+  canceled: Ref<boolean>
+  error: Ref<AxiosError<T> | undefined>
+}
+
+export function useAxios<T = any>(url: string, config?: AxiosRequestConfig): UseAxiosReturn<T>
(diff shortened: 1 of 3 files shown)


packages/integrations/useAxios/index.ts
```diff
@@ -8,10 +20,22 @@ import axios, { AxiosError, AxiosRequestConfig, AxiosResponse, CancelTokenSource
...
-export function useAxios<T = any>(
-  url: string,
-  config?: AxiosRequestConfig,
-) {
+export function useAxios<T = any>(url: string, ...args: any[]) {
+  let config: AxiosRequestConfig = {}
+  let instance: AxiosInstance = axios
+
+  if (args.length > 0) {
+    if ('request' in args[0])
```
antfu, line 28: This is a safe way to detect whether it's an axios instance?

jacobclevenger: I was going to use `instanceof` but it looks like you can't do that(https://github.com/axios/axios/issues/737) so I used the way that was suggested here. It doesn't seem to have any overlap with the `AxiosRequestConfig`, so it works from my testing. But I guess the only real way to enforce it is with the typescript types for parameters. If you have any ideas on another way to do it let me know.

antfu: Alright, go with it then. Can you add a comment above that line mentioning about the issue you linked? Thanks.

jacobclevenger: Sure thing!

antfu (approved):

PR #345, opened 2021-02-23
docs: include demos for Windi CSS, closes #344

### LINKED ISSUE #344: Tailwind classes are not available in demo files
It looks like the switch to WindiCSS broke some of the purgecss configuration. I noticed this looking at the docs for [`useTransition`](https://vueuse.org/core/useTransition/) that `.relative` does not have any effect. This appears to only be an issue in production, the dev environment still seems to work as expected.

**Steps to reproduce:**
Run `docs:build` command

**Expected result:**
`packages/.vitepress/dist/assets/style.css` should contain `.relative { position: relative }`

**Actual result:**
Compiled CSS is missing `.relative`

scottbedard:
I'm not an expert on WindiCSS or Vite, so if there is a better way to solve this feel free to close this PR 🙂 

Closes #344

### DIFF
--- a/packages/vite.config.ts
+++ b/packages/vite.config.ts
@@ -113,6 +113,7 @@ const config: UserConfig = {
     WindiCSS({
       scan: {
         dirs: ['.vitepress'],
+        include: ['core/**/*.vue'],
       },
       preflight: false,
     }),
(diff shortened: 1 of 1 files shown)


antfu:
Thanks. I have updated it a bit so it also includes other packages.

antfu (approved):

PR #353, opened 2021-02-28
feat(useFetch): new feature createFetch, close #352

### LINKED ISSUE #352: Add factory function for useFetch
Usecase:
```ts
//myPersonalProject/src/logics/api.ts
import { useFetchFactory } from '@vueuse/core'
const API_BASE_URL = process.env.API_BASE_URL!
//useFetchFactory returns the same properties as useFetch, just that it has the url prefix set to API_BASE_URL
export const useFetchWithBaseUrl = useFetchFactory(API_BASE_URL); 

```

```ts
//myPersonalProject/components/foo.vue
<script setup>
import { useFetchWithBaseUrl } from '@/logics/api.ts'
const { data } = useFetchWithBaseUrl('/users');
</script>
```

It'd be nice to have an equivalent of `axios.create({baseUrl: API_BASE_URL})` (I know useFetch uses fetch) for useFetch aswell.

Problems with this proposal:
This is basically ... reimplementing axios.create({baseUrl}), with the same dilemmas on the lib developer side like "should I add a / if the user does useFetchFactory('www.google.com') and then useFetchWithBaseUrl('abc') -> www.google.comabc" and so on

Alternatives:
Instead of adding a factory, add documentation on useFetch's side about how to create such a wrapper yourself. It's just tricky in TS since useFetch has 3 overloads

Obviously you can just do this instead:

```ts
//myPersonalProject/components/foo.vue
<script setup>
import { API_BASE_URL } from '@/utils/api.ts'
const { data } = useFetch(`${BASE_URL}/users`);
</script>
```

jacobclevenger:
Basic implementation of `createFetch` as mentioned in issue #352. Adds new test for the `createFetch` function.

I was also thinking it might be interesting to have some hooks on the `createFetch` object like `beforeRequest` and possibly `afterRequest`. These would return a promise which would allow one to do something like get an authorization token and attach it to the headers but the implementation of something like that might be a bit more difficult. So before doing anything like that I thought it would be better to discuss it.

Also I had to copy the function overloads to the `createFetch` function, if you have a better idea on how to do this let me know!

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -296,14 +296,6 @@ export function createFetch(config: CreateFetchOptions) {
   let options = config.options || {}
   let fetchOptions = config.fetchOptions || {}
 
-  /**
-   * The types here are a little complicated to get working properly
-   * Let me know if you have any ideas on how to simplify this.
-   */
-  function useFactoryFetch<T>(url: MaybeRef<string>): UseFetchReturn<T>
-  function useFactoryFetch<T>(url: MaybeRef<string>, useFetchOptions: UseFetchOptions): UseFetchReturn<T>
-  function useFactoryFetch<T>(url: MaybeRef<string>, options: RequestInit, useFetchOptions?: UseFetchOptions): UseFetchReturn<T>
-
   function useFactoryFetch(url: MaybeRef<string>, ...args: any[]) {
     const computedUrl = computed(() => joinPaths(unref(config.baseUrl), unref(url)))
 
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -89,6 +89,23 @@ export interface UseFetchOptions {
   refetch?: MaybeRef<boolean>
 }
 
+export interface CreateFetchOptions {
+  /**
+   * The base URL that will be prefixed to all urls
+   */
+  baseUrl: MaybeRef<string>
+
+  /**
+   * Default Options for the useFetch function
+   */
+  options?: UseFetchOptions
+
+  /**
+   * Options for the fetch request
+   */
+  fetchOptions?: RequestInit
+}
+
 export function useFetch<T>(url: MaybeRef<string>): UseFetchReturn<T>
 export function useFetch<T>(url: MaybeRef<string>, useFetchOptions: UseFetchOptions): UseFetchReturn<T>
 export function useFetch<T>(url: MaybeRef<string>, options: RequestInit, useFetchOptions?: UseFetchOptions): UseFetchReturn<T>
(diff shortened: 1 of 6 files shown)


packages/core/useFetch/index.ts
```diff
@@ -267,3 +284,51 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
...
+
+    if (args.length > 1) {
+      if ('immediate' in args[1] || 'refetch' in args[1])
+        options = { ...options, ...args[1] }
+    }
+
+    return useFetch(computedUrl, fetchOptions, options)
+  }
+
+  return useFactoryFetch
```
antfu, line 333:
```suggestion
  return useFactoryFetch as typeof useFetch
```

In this way the overload is no longer required I think

jacobclevenger: oh haha yeah, that's perfect! My bad 😅, thanks!

packages/core/useFetch/index.ts
```diff
@@ -267,3 +284,43 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
...
+export function createFetch(config: CreateFetchOptions) {
+  let options = config.options || {}
+  let fetchOptions = config.fetchOptions || {}
+
+  function useFactoryFetch(url: MaybeRef<string>, ...args: any[]) {
+    const computedUrl = computed(() => joinPaths(unref(config.baseUrl), unref(url)))
+
+    // Merge properties into a single object
+    if (args.length > 0) {
+      if ('immediate' in args[0] || 'refetch' in args[0]) { options = { ...options, ...args[0] } }
```
antfu, line 304: Not sure about this check, as we are likely to have more options in the future. Or maybe we can have a utils function that can be reuse as we are doing it four times now.

packages/core/useFetch/index.ts
```diff
@@ -267,3 +284,43 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
...
+
+  return `${start}${end}`
+}
+
+export function createFetch(config: CreateFetchOptions) {
+  let options = config.options || {}
+  let fetchOptions = config.fetchOptions || {}
+
+  function useFactoryFetch(url: MaybeRef<string>, ...args: any[]) {
+    const computedUrl = computed(() => joinPaths(unref(config.baseUrl), unref(url)))
```
antfu, line 300: This is great!

packages/core/useFetch/index.ts
jacobclevenger: Hmm yeah I was thinking about this as well. I'm not sure exactly what a clean solution might be for this. Off the top of my head perhaps like you said we could have a util function to handle this. Maybe something like

```ts
function hasProps(obj: object, props: string[]) {
  return Object.keys(obj).filter(key => props.includes(key)).length > 0
}

hasProps(args[0], ['immediate', 'refetch'])
```

In this case we would still need to update the parameters when we add more options, but it is a little cleaner than before. What do you think?

packages/core/useFetch/index.ts
```diff
@@ -89,6 +89,23 @@ export interface UseFetchOptions {
   refetch?: MaybeRef<boolean>
 }
 
+export interface CreateFetchOptions {
+  /**
+   * The base URL that will be prefixed to all urls
+   */
+  baseUrl: MaybeRef<string>
```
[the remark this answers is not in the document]
jacobclevenger: ah right good point

packages/core/useFetch/index.ts
antfu: I am thinking about this

```ts
// in shared utils
function containsProp(obj: object, ...props: string[]) {
  return props.some(k => k in obj)
}

// in this file
function isFetchOptions(obj: object): obj is UseFetchOptions {
  return containsProp(obj, 'immediate', 'refetch')
}

isFetchOptions(args[0])
```

We can also add a comment near `UseFetchOptions` saying remember to update that function

jacobclevenger: That looks good to me, I'll add it 😄

jacobclevenger: For now I put the `containsProp` function in `shared/utils/index.ts`. If you'd like it moved to a different file or want it in a new file just let me know

packages/core/useFetch/index.ts
```diff
@@ -1,5 +1,6 @@
 import { Ref, ref, unref, watch, computed, ComputedRef, shallowRef } from 'vue-demi'
 import { Fn, MaybeRef } from '@vueuse/shared'
+import { containsProp } from '@vueuse/shared/utils'
```
antfu, line 3: You can merge these two imports.

packages/core/useFetch/index.ts
```diff
@@ -267,3 +295,43 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
...
+  let options = config.options || {}
+  let fetchOptions = config.fetchOptions || {}
+
+  function useFactoryFetch(url: MaybeRef<string>, ...args: any[]) {
+    const computedUrl = computed(() => config.baseUrl ? joinPaths(unref(config.baseUrl), unref(url)) : unref(url))
+
+    // Merge properties into a single object
+    if (args.length > 0) {
+      if (isFetchOptions(args[0])) { options = { ...options, ...args[0] } }
+      else {
```
antfu, line 316:
```suggestion
    const computedUrl = computed(() => config.baseUrl 
      ? joinPaths(unref(config.baseUrl), unref(url)) 
      : unref(URL)
    )

    // Merge properties into a single object
    if (args.length > 0) {
      if (isFetchOptions(args[0])) { 
        options = { ...options, ...args[0] } 
      } else {
```

packages/core/useFetch/index.ts
jacobclevenger: Done

antfu (approved):

jacobclevenger:
Before you merge I'd still like to put on the table discussion about the 'beforeRequest' thing I mentioned earlier. Perhaps in a separate PR or something. But if you think it is outside the scope of this function I don't think that would be unreasonable either.

antfu:
@jacobclevenger sure thing, let's do it in another PR. Thanks!

PR #366, opened 2021-03-08
fix(useFirestore): sub-collection is wrongly detected as a document

### LINKED ISSUE #365: useFirestore does not work with sub-collections
Hi, I failed to use `useFirestore` with a sub-collection, for example:

`useFirestore(db.collection('docs').doc('foo').collection('bar'))`

But the collection is wrongly detected as a document by the `isDocumentReference` function:
https://github.com/vueuse/vueuse/blob/345de1aa9d14622df0bff62622416510191f2247/packages/firebase/useFirestore/index.ts#L29

dm4t2:
This PR fixes #365.

I modified the `isDocumentReference` which now checks the number of slash (`/`) occurences in the path:
* even -> collection
* odd -> document

### DIFF
--- a/packages/firebase/useFirestore/index.test.ts
+++ b/packages/firebase/useFirestore/index.test.ts
@@ -1,0 +1,25 @@
+import { useFirestore } from './index'
+import type firebase from 'firebase'
+
+describe('useFirestore', () => {
+  const expectArrayRef = (path: string, initialValue: any, result: boolean) => {
+    // @ts-ignore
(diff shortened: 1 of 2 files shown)


antfu:
Is this check reliable? Do you have any docs from firebase describe that, or can you explain a bit about how this works? Thanks

dm4t2:
Yes, this check is reliable because each odd node is a collection and each even node is a document:  https://firebase.google.com/docs/firestore/data-model?hl=en#hierarchical-data

antfu (approved):

PR #371, opened 2021-03-10
feat(useFetch): add beforeFetch option

jacobclevenger:
Hey @antfu this is the initial implementation of the `beforeFetch` feature I had in mind for useFetch. It isn't too complicated. This adds a new option to the `useFetch` options called `beforeFetch`. This new options accepts a promise that returns the url and request options. The basic usage would be as follows

```ts
useFetch('/api/me', {
  beforeFetch: async(url, options) => {
    options.headers = {
      ...options.headers,
      Authorization: `Bearer ${await getAuthToken()}`,
    }

    return {
      url,
      options,
    }
  },
})
```
Of course this option would also apply to the `createFetch` function, where it would be more likely to be used.

Let me know what you think about this API. If you like it I'll go ahead and write some tests and add docs, otherwise we can refactor 😄

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -92,7 +92,7 @@ export interface UseFetchOptions {
   /**
    * Will run immediately before the fetch request is dispatched
    */
-  beforeFetch?: (ctx: Context, abort: Fn) => Promise<Partial<Context>> | Partial<Context>
+  beforeFetch?: (ctx: Context) => Promise<Partial<Context>> | Partial<Context>
 }
 
 export interface CreateFetchOptions {
@@ -241,20 +241,20 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
       }
     }
 
-    let context = { url: unref(url), options: fetchOptions, cancel: abort }
+    const context = { url: unref(url), options: fetchOptions, cancel: abort }
 
     if (options.beforeFetch)
       Object.assign(context, await options.beforeFetch(context))
 
     return new Promise((resolve) => {
       fetch(
-        _url,
+        context.url,
         {
           ...defaultFetchOptions,
-          ..._fetchOptions,
+          ...context.options,
           headers: {
             ...defaultFetchOptions.headers,
-            ..._fetchOptions?.headers,
+            ...context.options?.headers,
           },
         },
       )
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -55,7 +55,7 @@ interface UseFetchReturnBase<T> {
 
 type DataType = 'text' | 'json' | 'blob' | 'arrayBuffer' | 'formData'
 type PayloadType = 'text' | 'json' | 'formData'
-type Context = { url: string; options: RequestInit }
+type Context = { url: string; options: RequestInit, cancel: () => void }
 
 interface UseFetchReturnMethodConfigured<T> extends UseFetchReturnBase<T> {
   // type
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -92,7 +92,7 @@ export interface UseFetchOptions {
   /**
    * Will run immediately before the fetch request is dispatched
    */
-  beforeFetch?: (ctx: Context, abort: Fn) => Promise<Context> | Context
+  beforeFetch?: (ctx: Context, abort: Fn) => Promise<Partial<Context>> | Partial<Context>
 }
 
 export interface CreateFetchOptions {
@@ -244,8 +244,12 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
     let _url = unref(url)
     let _fetchOptions = fetchOptions
 
-    if (options.beforeFetch)
-      ({ url: _url, options: _fetchOptions } = await options.beforeFetch({ url: _url, options: _fetchOptions }, abort))
+    if (options.beforeFetch) {
+      const { url: maybeUrl, options: maybeOptions } = await options.beforeFetch({ url: _url, options: _fetchOptions }, abort)
+
+      _url = maybeUrl || _url
+      _fetchOptions = maybeOptions || _fetchOptions
+    }
 
     return new Promise((resolve) => {
       fetch(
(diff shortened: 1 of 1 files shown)


packages/core/useFetch/index.ts
```diff
@@ -87,6 +87,8 @@ export interface UseFetchOptions {
    * @default false
    */
   refetch?: MaybeRef<boolean>
+
+  beforeFetch?: (url: string, options: RequestInit) => Promise<{ url: string; options: RequestInit }>
```
antfu, line 91: Think we can have non-promise returns as well. Do you have any plan to support canceling the fetch in this callback?

jacobclevenger: ah yeah, we should be able to do non-promise returns, I'll add that. As for canceling we can probably pass a third parameter to abort the request

antfu: I am thinking if we could unify the arguments and returns signature. Something like

```ts
beforeFetch?: (ctx: Context) => Promise<Context> | Context
```

jacobclevenger: Would you include the abort option in that context? Seems like it would make sense to have something like 

```ts
type Context = { url: string; options: RequestInit }
beforeFetch?: (ctx: Context, abort: Fn) => Promise<Context> | Context
```

antfu: How about 

```ts
beforeFetch?: (ctx: Context) => Promise<Partial<Context>> | Partial<Context>
```

I am not sure about either way tbh 😂

jacobclevenger: Hmm yeah partial should be fine I think

packages/core/useFetch/index.ts
```diff
@@ -235,15 +241,25 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
...
 
+    let _url = unref(url)
+    let _fetchOptions = fetchOptions
+
+    if (options.beforeFetch) {
+      const { url: maybeUrl, options: maybeOptions } = await options.beforeFetch({ url: _url, options: _fetchOptions }, abort)
+
+      _url = maybeUrl || _url
+      _fetchOptions = maybeOptions || _fetchOptions
+    }
```
antfu, line 252:
```suggestion
    let context = { url: unref(url), options: fetchOptions, cancel: abort }

    if (options.beforeFetch)
      Object.assign(context, await options.beforeFetch(context))
```

antfu: Would `abort` work before using fetch? So should we handle it with another approach (like not fire-up fetch at all)?

jacobclevenger: Oh haha yeah my bad 😂 

Perhaps we could also allow them to return a boolean? So they can return context, which would modify the request, `false`, which would not execute the request, and `true` which would execute the request without modifying the parameters. Does that sound okay to you?

```ts
beforeFetch?: (ctx: Context) => Promise<Context | boolean> | Context | boolean
```

userquin:
@jacobclevenger can you provide an `afterFetch` option for the response?

jacobclevenger:
> @jacobclevenger can you provide an `afterFetch` option for the response?

I think we could do that. Could you give an example use case so we can know what kind of usage you have in mind?

userquin:
do you have discord?

An example will to make response handlers to convert it to be http agnostic: fileupload max file size exceed => 413 response code then the hook will throw an app error

userquin:
On all my projects I have the concept of interceptor: then I configure them to the equivalent for `useFetch`: these interceptors can act `beforeFetch`, `afterFetch` and `onError` phases. All with pomises (similar to this but with an array), that is a chain: if you are familiar with backends, would be like a filter chain.

Then I can configure this chain for specific operations (app operations) abstracting the app to deal with http responses. I have advanced interceptor that can handle responses for CRUD operations: the server response for an insertion,  for example, contains a 200 but the response is a wrapper with validation info (was ok or not, if there are validation errors on the form, etc..)

userquin:
My discord user is `Joaquín#7667`

jacobclevenger:
Hmm so perhaps something like this where we provide the response as well as the response data? It might also be better to discuss this after we get the `beforeFetch` api finalized though so we can keep everything relatively consistent 

```ts
useFetch(url, {
  async afterFetch({ response, data }) {
    if (request.status === 413)
      throw new Error('Max File Size Exceeded')
  }
})
```

packages/core/useFetch/index.ts
antfu: My initial thought is to return boolean as well, but I am a little concerned about if the intention is clear enough. I was confused about [`return false` in event handlers](https://stackoverflow.com/questions/1357118/event-preventdefault-vs-return-false) for a long time, not sure if it means "cancel the event" or "cancel the handler" or just the "result".

So I guess it might be more clear and explicit to call `cancel()` or `abort()` in this case.

jacobclevenger: That's a good point. So we should go with your way then. As for what we call it I think we should go with `cancel()`(which you also had in your commit). Using `abort()` might cause confusion since the function already exports an `abort()` method and people may accidentally do something like

```ts
const { abort } = useFetch('', {
  beforeFetch({ options }) {
    abort()

    return {
      options,
    }
  },
})
```

packages/core/useFetch/index.ts
```diff
@@ -55,6 +55,7 @@ interface UseFetchReturnBase<T> {
 
 type DataType = 'text' | 'json' | 'blob' | 'arrayBuffer' | 'formData'
 type PayloadType = 'text' | 'json' | 'formData'
+type Context = { url: string; options: RequestInit }
```
antfu, line 58:
```suggestion
type Context = { url: string; options: RequestInit, cancel: () => void }
```

Ideally, we sure use an interface with some jsdocs

jacobclevenger: Do we want to keep it as `Context` or rename it to something more descriptive like `BeforeFetchContext`?

antfu: Sure, `BeforeFetchContext` sounds like a good idea 👍

antfu (approved):

mariusa:
Hi, filled an issue for `afterFetch` support: https://github.com/vueuse/vueuse/issues/456

Issue #364, opened 2021-03-06
useTransition v5 improvements


scottbedard:
The [`useTransition`](https://vueuse.org/core/useTransition/) function is a good building block, but I often find myself having to stack helpers on top of it. Since we're revisiting functions for 5.0, there are some changes I'd like to make. All of these changes are purely additive, and compatible with the current API.

### ~Controls https://github.com/vueuse/vueuse/pull/362, https://github.com/vueuse/vueuse/pull/397~

It would be great if we could expose some of the underlying utilities. The most obvious things that come to mind are the `pause` and `resume` functions that are being completely wasted right now! Here is what I'm thinking for the current implementation of the controls API.

```ts
const { output, pause, resume } = useTransition(source, { controls: true })
```

> **Withdrawn:** This feature ended up being more complex than anticipated, and I have decided not to implement it at this time. An alternative strategy for pause / resume functionality was described in https://github.com/vueuse/vueuse/pull/397

### Vectors #376

Currently `useTransition` only supports a single value. This is fine for simple use cases where the value is needed for text purposes, but is inadequate for non-trivial use cases like for game rendering. There is no good way to perform multiple transitions at once. For each value added, a separate requestAnimationFrame loop is needed, separate easing functions, separate `onFinished` logic, and so on...

There is no reason our function can't also support a vector as input, and produce a transitioned vector as output. This would give awesome flexibility when building graphs that require X/Y coordinates, working with libraries like Three.js, transitioning RGB values, or really anything where animations share a common duration. Here is the API I think that we should support.

```ts
// the number api will remain unchanged, under the hood this will be treated as [n]
const n = ref(0)

const output = useTransition(n, options) // <- ComputedRef<number>

// and the vector api will support any number of dimensions
const position = ref([x, y, z])

const output = useTransition(position, options) // <- ComputedRef<[number, number, number]>
```

### Enabling / disabling transitions #436

Our transitions right now simply watch the source number, and begin a requestAnimationFrame loop whenever that value changes. This makes it difficult to change that value _without_ transitioning, which there are plenty of situations that call for. It would be nice to also accepted a `Ref<boolean>` that allows us to turn the transition on and off.

```ts
const disabled = ref(true)

const output = useTransition(source, { disabled })
```

### Delayed transitions #386

[...]

PR #376, opened 2021-03-13
feat(useTransition): support for vectors

scottbedard:
This is the first of several PRs to implement features being designed in #364

There is a small breaking change here. The returned value is now a `ComputedRef` instead of a `Ref`, meaning anyone who manually set the value of a transitioned ref would no longer be able to do so. This was never part of the documented API though, so I'm not sure how much to worry about it. I think just making a mention in the release notes would be sufficient.

Vectors support any number of dimensions, and there is a quick track/sled example added to the demo.

### DIFF
--- a/packages/shared/utils/is.ts
+++ b/packages/shared/utils/is.ts
@@ -16,3 +16,8 @@ export const now = () => Date.now()
 export const timestamp = () => +Date.now()
 export const clamp = (n: number, min: number, max: number) => Math.min(max, Math.max(min, n))
 export const noop = () => {}
+export const rand = (min: number, max: number) => {
+  min = Math.ceil(min)
+  max = Math.floor(max)
+  return Math.floor(Math.random() * (max - min + 1)) + min
+}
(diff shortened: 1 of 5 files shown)


scottbedard:
Hmm, I see the deep watching test is not passing in 2.x. It's late for me, so I'll look in the morning and update the PR once it's resolved.

packages/shared/utils/is.ts
```diff
@@ -16,4 +16,8 @@ export const now = () => Date.now()
 export const timestamp = () => +Date.now()
 export const clamp = (n: number, min: number, max: number) => Math.min(max, Math.max(min, n))
 export const noop = () => {}
-export const rand = (min: number, max: number) => Math.floor(Math.random() * (max - min + 1)) + min
+export const rand = (min: number, max: number) => {
+  min = Math.ceil(min)
+  max = Math.floor(max)
+  return Math.floor(Math.random() * (max - min + 1)) + min
+}
```
scottbedard, line 23: Reverted, will generate random int between min (inclusive) and max (inclusive)

antfu:
Looks good, do you want to have this in 4.x or 5.0? Both works for me. If you are targeting 5.0, I have just created a new branch `next` for that and you can check the base of this PR. Thanks.

scottbedard:
@antfu Great! Since this does contain that breaking change, we should probably target 5.x. I'll also be submitting a few related PRs soon too, so it'll probably be easiest to just release all the changes together 🤷‍♂️

PR #382, opened 2021-03-15
refactor(useWebWorkerFn): Small doc and type improvements

Donskelle:
Thanks for that demo update, looks very impressive now.

Only some small types and docs fixes

### DIFF
--- a/packages/core/useWebWorkerFn/index.ts
+++ b/packages/core/useWebWorkerFn/index.ts
@@ -13,7 +13,15 @@ export type WebWorkerStatus =
   | 'TIMEOUT_EXPIRED'
 
 export interface WebWorkerOptions extends ConfigurableWindow {
+  /**
+   * Number of milliseconds before killing the worker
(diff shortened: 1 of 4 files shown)


packages/core/useWebWorkerFn/index.ts
```diff
@@ -21,17 +26,12 @@ export interface WebWorkerOptions extends ConfigurableWindow {
  */
 export const useWebWorkerFn = <T extends (...fnArgs: any[]) => any>(
   fn: T,
-  {
-    dependencies = [],
-    timeout,
-    window = defaultWindow,
-  }: WebWorkerOptions = {},
+  { dependencies = [], timeout, window = defaultWindow }: Partial<WebWorkerOptions> = {},
```
Donskelle, line 29: This is the only main change. Making optinons a Partial type.
It shouldn't be a breaking change

antfu (approved):

PR #383, opened 2021-03-16
feat(useMediaControls): new function

wheatjs:
Here is the initial implementation of `useMediaControls`. This function works for both `<audio>` and `<video>` elements. 

For the demo I ended up needing to create some components for the player preview, a scrubber and some menu stuff. I hope that's okay, if not I can remove them. They are in `useMediaControls/components`. I also needed to include the subtitles in a subfolder as well, `useMediaControls/data`.

As of now hear are most the features
- Playing/Pausing via ref
- Automatically injecting src from an array
- Automatically injecting closed-caption tracks from an array
- Volume controls
- Refs hooked up to events for nearly all properties that change
- Support for easily controlling the subtitles
- Picture in picture mode

I thinks that's most of it as of now. I also ended up using some other functions from `vueuse` in this, like `useIgnorableWatch` and `useEventListener` so I hope those are okay to use internally. 

Anyway, let me know what you think 😄


jacobclevenger:
Seems like the subtitles aren't showing up on the website demo, probably because of how I linked them in the demo.

antfu:
This is awesome @jacobclevenger, sorry to get it back late. Since you marked it as 5.0 are you trying to targeting 5.0 release? I am fine to introduce it in 4.x if you want as this is not a breaking change.

jacobclevenger:
@antfu No problem at all I know you've been busy. I only marked it as 5.0 as I figured the release for that was getting close, but if you want to release it in 4.x that is okay by me as well

antfu:
Alright, I guess it might still take a while to have 5.0 released. So I am going to have a minor release for this in 4.x. Love the demo!

packages/core/useMediaControls/index.ts
```diff
@@ -0,0 +1,481 @@
...
+      })
+    })
+  }
+
+  const stop = watch([target], () => {
+    usingElRef<HTMLMediaElement>(target, (el) => {
+      stop()
+
+      // Apply Options
+      watch(
```
antfu, line 305: Use nested watch you will need to capture the stopper for every one of them to clean up the side-effect. Which could be troublesome to handle correctly. I would suggest to use top-level `watch` and `watchEffect` for that.

For example, this could be rewritten as 

```ts
// Apply Options
watchEffect(() => {
  const el = unref(target)
  if (!el)
    return
  el.loop = unref(options.loop)!
  el.controls = unref(options.controls)!
  el.muted = unref(options.muted)!
  el.preload = unref(options.preload)!
  el.autoplay = unref(options.autoplay)!
  el.volume = unref(volume)!
  ;(el as HTMLVideoElement).playsInline = unref(options.playsinline)!
  ;(el as HTMLVideoElement).poster = unref(options.poster) || ''
})
```

And `useEventListener` accepts `MaybeRef<EventTarget>` which will do the cleanup automatically on target changes.

jacobclevenger: Oh good point, I'll make some changes. Thanks!

jacobclevenger: Alright most of it should be fixed now. For the tracks the event listeners actually come from the video text track property so I had to nest those listeners within a watch and then stop them manually. If you have an idea for a better way to do it let me know

antfu (approved):

PR #386, opened 2021-03-18
feat(useTransition): support for delayed transitions

scottbedard:
Just a quick helper option to make delayed and staggered transitions easier. Also, I take back my claim before that `onStarted` was unnecessary. It definitely serves a useful purpose now! 😅 

More info here https://github.com/vueuse/vueuse/issues/364

This also includes a tiny change to `useTimeoutFn`, in that now it supports a `MaybeRef<number>` as the interval.

### DIFF
--- a/packages/core/useTransition/index.ts
+++ b/packages/core/useTransition/index.ts
@@ -181,7 +181,7 @@ export function useTransition(
     onStarted()
   }
 
-  const timeout = useTimeoutFn(start, delay)
+  const timeout = useTimeoutFn(start, delay, false)
 
   watch(sourceVector, () => {
     if (unref(delay) <= 0) start()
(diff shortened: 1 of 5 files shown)


packages/core/useTransition/index.ts
```diff
@@ -172,6 +179,13 @@ export function useTransition(
 
     resume()
     onStarted()
+  }
+
+  const timeout = useTimeoutFn(start, delay)
```
antfu, line 184: Should we pass `immediate: false` here?

scottbedard: Good catch! Because of the way I wrote my tests, I didn't notice this.

Resolved 345d8c1bd377b48a586be19e4819b66d1aa53856

PR #391, opened 2021-03-24
feat(useAxios): new `loading` state in return object

lpj145:
Expressive way when requests is dispatch, loading is more idiomatic what finished, because every time you need to use them with states requests you need to translate to !finished.

### DIFF
--- a/packages/integrations/useAxios/demo.vue
+++ b/packages/integrations/useAxios/demo.vue
@@ -2,13 +2,13 @@
 import { stringify } from '@vueuse/docs-utils'
 import { useAxios } from '.'
 
-const { data, finished } = useAxios(
+const { data, loading } = useAxios(
   'https://jsonplaceholder.typicode.com/todos/1',
 )
 const text = stringify(data)
 </script>
 
 <template>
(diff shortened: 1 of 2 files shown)


jacobclevenger:
Hmm looks like this would be a breaking change. Not sure if it is a good idea to introduce a breaking change. Especially without deprecation warnings.

jacobclevenger:
Perhaps we could introduce this alongside the `finished` property instead? Or maybe wait for input from @antfu

lpj145:
> Perhaps we could introduce this alongside the `finished` property instead? Or maybe wait for input from @antfu

I don't, think about thousands people using five, six, seven... concurrency requests, how more `Ref` properties we have more we "joke" with users, maybe be a break change for 5x.

jacobclevenger:
Hmm just seems like a small benefit for a breaking change though. Let's see what Antfu says.

antfu:
Thinking we can have `loading: computed(() => !finished.value)` and since computed is lazy, it won't impact too much if users do use it.

antfu (approved):

lpj145:
Thank you, you are awesome.

PR #397, opened 2021-03-27
feat(useTransition): pause / resume controls

scottbedard:
This will be the largest PR in the #364 epic. Transitions can now be paused / resumed via the `controls` API.

```js
const { output, pause, resume } = useTransition(source, { controls: true })
```

This ended up being much more complicated than I anticipated, so let me break down what is happening here. If you need any help reviewing this PR, I'm happy to answer any questions or discuss the changes.

### 1. Clean up obsolete test file

Somehow we ended up with two copies of the `useTransition` tests ([see here](https://github.com/vueuse/vueuse/tree/c700fe9dfd5ae898708e0e40518a34020d219e93/packages/core/useTransition)), we must have had a bad merge at some point. I noticed that all other test files end in `.test.ts` instead of `.spec.ts`, so I've changed the files to match and deleted the obsolete ones.

89c603941d89f5b6a3f45ff1aba9672e6d6fd59a f45961b1a4e4e4f128984cf71ff93f19fa863d2e

### 2. Rewrite overloads for better controls typing

Given how much is going on in this function, I want to make sure to maintain proper type coverage. To do this I started from scratch, then migrated the functionality over once I had working types. I'm simplifying a bit here for readability, but these are the general shape of the types that we now have.

```ts
// numbers
useTransition(ref(1)) // ComputedRef<number>

useTransition(ref(1), { controls: true }) // { output: ComputedRef<number> }

// array vectors
useTransition(ref([1, 2])) // ComputedRef<number[]>

useTransition(ref([1, 2]), { controls: true }) // { output: ComputedRef<number[]> }

// tuple vectors
useTransition([ref(1), ref(2)]) // ComputedRef<[number, number]>

useTransition([ref(1), ref(2)], { controls: true }) // { output: ComputedRef<[number, number]> }
```

2e7f26aab8d0f71d6ba236e051787f233a1bb2c8

### 3. Create API for underlying `pause` and `resume` functions
I naively thought we could just expose `useRafFn`'s `pause` and `resume` functions, this was not the case. Additional logic had to be added to track how much time is remaining in a transition once it's paused. For example, if a transition lasts 1000ms, and it is paused after 750ms, the resumed transition must start from where it is currently paused and complete in 250ms.

1ea590c87b33879fced5df69a5743bb33cef2d8c

### 4. Handle edge cases

Additional logic had to be added to deal with situations like pausing a transition that is not running, starting a new transition while the previous one is paused, etc... While this ended up being a pain, it also allowed the controls API to include two additional refs, `isTransitioning` and `isPaused`. I'm open to renaming these if you think there are better options.

c4cae1b8e28a4217bccc4545a620901206c3132c

### 5. Avoid unnecessary work

I realized there was no need to create refs for data that we may not need. For example, if `controls` is not set to `true`, there is no reason to create/update refs for `pausedAt` and `isTransitioning`.

bcda2a1980d9d0a5162a74fb5dbf237eb1f49923


scottbedard:
Actually, I'm going to withdraw this PR. I'm noticing some perf issues when running many transitions at once. For now, if people need access to pause/resume functionality, it can be achieved by toggling an external ref, and setting its value to the state of the current transition when paused.

PR #406, opened 2021-03-31
fix(useBreakpoints): useBreakpoints smaller should be < rather than <=, closes #404

### LINKED ISSUE #404: useBreakpoints doesn't work correctly
The issue itself is quite straightforward, breakpoint service functions (`smaller`, `between`) shouldn't include last max-width pixel of media query into account. (e.g. 767px is "sm" but 768 isn't)

I've attached two screenshots where you can see the issue itself

useBreakpoints() vs tailwind:

### 767px

image

### 768px

image
(it's smaller and greater than "md" at the same time)

### My current workaround (hope you fix it ASAP, be me to it):

```
  install: (app: App): void => {
    const breakpoints = useBreakpoints(breakpointsTailwind)

    breakpoints.smaller = (k) => {
      return useMediaQuery(`(max-width: ${breakpointsTailwind[k] - 1}px)`)
    }

    app.config.globalProperties.$breakpoints = breakpoints
    app.provide('breakpoints', breakpoints)
  }
  ```
  
yet another thing, there is a typo in documentation. there is no `larger`, should be `greater` instead 😄 

image

P.S. thanks for the "vueuse" it's just sick!

wheatjs:


antfu:
Breakpoints can have units like `30rem`, see the updates

wheatjs:
> Breakpoints can have units like `30rem`, see the updates

Oh my bad 😅

antfu (approved):

PR #410, opened 2021-04-01
feat(filters): use number/ref with debounce & throttle filters

### LINKED ISSUE #409: The debounce or throttle time is not applied dynamically.
I want to use debounce or throttle time dynamically, but it doesn't apply

#### eg
```js
debouncedWatch(
  [() => props.mode, () => props.prev, () => props.current],
  () => { console.log('changed!') },
  { debounce: props.inputDelay }
)
```

The value does not change dynamically since the initial inputDelay value was applied

Is there a solution?

wheatjs:
Adds option to use `Ref<number>` for debounce and throttle filter along with functions that wrap those filters. Would resolve #409

### DIFF
--- a/packages/shared/utils/filters.ts
+++ b/packages/shared/utils/filters.ts
@@ -42,13 +42,15 @@ export function debounceFilter(ms: MaybeRef<number>) {
   let timer: ReturnType<typeof setTimeout> | undefined
 
   const filter: EventFilter = (invoke) => {
+    const duration = unref(ms)
+
     if (timer)
       clearTimeout(timer)
 
-    if (unref(ms) <= 0)
+    if (duration <= 0)
       return invoke()
 
-    timer = setTimeout(invoke, unref(ms))
+    timer = setTimeout(invoke, duration)
   }
 
   return filter
@@ -72,14 +74,17 @@ export function throttleFilter(ms: MaybeRef<number>, trailing = true) {
   }
 
   const filter: EventFilter = (invoke) => {
+    const duration = unref(ms)
     const elapsed = Date.now() - lastExec
 
     clear()
 
-    if (unref(ms) <= 0)
+    if (duration <= 0) {
+      lastExec = Date.now()
       return invoke()
+    }
 
-    if (elapsed > ms) {
+    if (elapsed > duration) {
       lastExec = Date.now()
       invoke()
     }
@@ -87,7 +92,7 @@ export function throttleFilter(ms: MaybeRef<number>, trailing = true) {
       timer = setTimeout(() => {
         clear()
         invoke()
-      }, unref(ms))
+      }, duration)
     }
   }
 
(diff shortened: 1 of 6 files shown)


packages/shared/utils/filters.ts
```diff
@@ -38,8 +38,8 @@ export const bypassFilter: EventFilter = (invoke) => {
  *
  * @param ms
  */
-export function debounceFilter(ms: number) {
-  if (ms <= 0)
+export function debounceFilter(ms: MaybeRef<number>) {
+  if (unref(ms) <= 0)
```
antfu, line 42: Then we should check this inside the filter

wheatjs: Good catch, will update

wheatjs:
@antfu Updated, let me know if that was an okay way to do it. It was a little weird since you can't just return `bypassFilter` in the filter function

packages/shared/utils/filters.ts
```diff
@@ -38,17 +38,17 @@ export const bypassFilter: EventFilter = (invoke) => {
...
-  if (ms <= 0)
-    return bypassFilter
-
+export function debounceFilter(ms: MaybeRef<number>) {
   let timer: ReturnType<typeof setTimeout> | undefined
 
-  const filter: EventFilter = (invoke) => {
+  const filter: EventFilter = (invoke, args) => {
+    if (unref(ms) <= 0)
+      return bypassFilter(invoke, args)
```
antfu, line 46:
```suggestion
      return invoke()
```

Just like this will do, but after clearTimeout.

wheatjs: 😞 Sorry I tried that originally then typescript got mad. Had to reload VSCode to fix the error :/

packages/shared/utils/filters.ts
```diff
@@ -79,6 +76,9 @@ export function throttleFilter(ms: number, trailing = true) {
 
     clear()
 
+    if (unref(ms) <= 0)
+      return invoke()
+
     if (elapsed > ms) {
```
antfu, line 82: Missed this? Maybe we can have an local variable to hold the value of `unref`

packages/shared/utils/filters.ts
```diff
@@ -38,17 +38,17 @@ export const bypassFilter: EventFilter = (invoke) => {
...
  */
-export function debounceFilter(ms: number) {
-  if (ms <= 0)
-    return bypassFilter
-
+export function debounceFilter(ms: MaybeRef<number>) {
   let timer: ReturnType<typeof setTimeout> | undefined
 
-  const filter: EventFilter = (invoke) => {
+  const filter: EventFilter = (invoke, args) => {
```
antfu, line 44:
```suggestion
  const filter: EventFilter = (invoke) => {
```

wheatjs:
Alright fixed all the issues I think and added some tests to make sure

packages/shared/utils/filters.ts
jefferycline1: []()

PR #416, opened 2021-04-04
feat(onKeyStroke): new functions

yyyang1996:
Add the onKeyStroke function to listen for keyboard events. Inspired from [react-use](https://streamich.github.io/react-use/?path=/story/sensors-usekey--docs).


antfu:
Preferred to name it `onKeyStroke`, and more over, we could have `onKeyDown` `onKeyUp` `onKeyPressed` alias.

And for the arguments, I think we can put `eventName`, `target`, and also `passive` into the options field. Meaning:

```ts
function onKeyStroke(key, handler, options) { }
```

To make this more useful, we could also have modifier support `^+P` or `Ctrl+Shift+P`, but what would be a bit out of scope, which we could have later.

yyyang1996:
> Preferred to name it `onKeyStroke`, and more over, we could have `onKeyDown` `onKeyUp` `onKeyPressed` alias.
> 
> And for the arguments, I think we can put `eventName`, `target`, and also `passive` into the options field. Meaning:
> 
> ```ts
> function onKeyStroke(key, handler, options) { }
> ```
> 
> To make this even more useful, we could also have modifier support `^+P` or `Ctrl+Shift+P`, but what would be a bit out of scope, which we could have later as `onHotKeys` or `onShortcuts`

I have changed it. 
For `onHotKeys`, I don't know  what the api looks like to make sense. 
Like `onHotKeys(key, handler, {modifiers: ['Ctrl', 'Shift']})`?

antfu:
> onHotKeys

Haha, just an idea, no pressure on that. I might have some ideas of the implementation where I will draft it later.

packages/core/onKeyDown/index.md
```diff
@@ -0,0 +1,24 @@
+---
+category: Sensors
+---
+
+# onKeyDown
```
antfu, line 5: Instead of creating folders for each of them, I actually think we can just put them all under `onKeyStroke` (multiple named exports).

Sorry, I didn't make that clear, I can handle that if you want. Thank you

yyyang1996: My first thought was this, but then I took them apart again.😂

antfu (approved):

PR #421, opened 2021-04-04
fix(useAxios): use shallowRef for resposne, data, and error

### LINKED ISSUE #415: useAxios: use shallowRef for response, data and error
Could [useAxios](https://github.com/vueuse/vueuse/blob/3e14ee39f94ceac4bc9684df456f5e41eb076cca/packages/integrations/useAxios/index.ts#L24) be modified to use [shallowRef](https://v3.vuejs.org/api/refs-api.html#shallowref) for [response](https://github.com/vueuse/vueuse/blob/3e14ee39f94ceac4bc9684df456f5e41eb076cca/packages/integrations/useAxios/index.ts#L45), [data](https://github.com/vueuse/vueuse/blob/3e14ee39f94ceac4bc9684df456f5e41eb076cca/packages/integrations/useAxios/index.ts#L46) and [error](https://github.com/vueuse/vueuse/blob/3e14ee39f94ceac4bc9684df456f5e41eb076cca/packages/integrations/useAxios/index.ts#L50)?

## Rationale

`shallowRef` would improve performance because `response`, `data` and `error` would not be made deeply reactive. If the content of `response.value`, `data.value` and `error.value` are not modified (which for me is the best practice), then making them reactive is completely redundant. The client code developers could explicitly decide, if they want to make the values of those refs deeply reactive using [reactive](https://v3.vuejs.org/api/basic-reactivity.html#reactive) or [ref](https://v3.vuejs.org/api/refs-api.html#ref).

wheatjs:
Fixes #415

### DIFF
--- a/packages/integrations/useAxios/index.ts
+++ b/packages/integrations/useAxios/index.ts
@@ -2,12 +2,12 @@ import { Ref, ref, shallowRef } from 'vue-demi'
 import axios, { AxiosError, AxiosRequestConfig, AxiosResponse, CancelTokenSource, AxiosInstance } from 'axios'
 
 export interface UseAxiosReturn<T> {
-  response: Ref<AxiosResponse<T> | undefined>
-  data: Ref<T | undefined>
+  response: Ref<AxiosResponse<T> | null>
+  data: Ref<T | null>
   finished: Ref<boolean>
   loading: Ref<boolean>
   canceled: Ref<boolean>
-  error: Ref<AxiosError<T> | undefined>
+  error: Ref<AxiosError<T> | null>
 }
 
 export function useAxios<T = any>(url: string, config?: AxiosRequestConfig): UseAxiosReturn<T>
@@ -42,12 +42,12 @@ export function useAxios<T = any>(url: string, ...args: any[]) {
       instance = args[1]
   }
 
-  const response = shallowRef<AxiosResponse<T> | null>(null)
-  const data = shallowRef<any>(undefined) as Ref<T | undefined>
+  const response = shallowRef<any>(null) as Ref<AxiosResponse<T> | null>
+  const data = shallowRef<any>(null) as Ref<T | null>
   const finished = ref(false)
   const loading = ref(true)
   const canceled = ref(false)
-  const error = shallowRef<AxiosError<T> | undefined>()
+  const error = shallowRef<any>(null) as Ref<AxiosError<T> | null>
 
   const cancelToken: CancelTokenSource = axios.CancelToken.source()
   const cancel = (message?: string) => {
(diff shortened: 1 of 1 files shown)


packages/integrations/useAxios/index.ts
```diff
@@ -42,12 +42,12 @@ export function useAxios<T = any>(url: string, ...args: any[]) {
       instance = args[1]
   }
 
-  const response = ref<any>(null) as Ref<AxiosResponse<T> | undefined>
-  const data = ref<any>(undefined) as Ref<T | undefined>
+  const response = shallowRef<any>(null) as Ref<AxiosResponse<T> | undefined>
```
antfu, line 45:
```suggestion
  const response = shallowRef<AxiosResponse<T> | null>(null)
```

Shouldn't we do this? Typed as undefined but passed as null feels a bit weird to me.

wheatjs: Ah yeah haha, not sure what was up with the types there to begin with 😅

packages/integrations/useAxios/index.ts
```diff
@@ -42,12 +42,12 @@ export function useAxios<T = any>(url: string, ...args: any[]) {
       instance = args[1]
   }
 
-  const response = ref<any>(null) as Ref<AxiosResponse<T> | undefined>
-  const data = ref<any>(undefined) as Ref<T | undefined>
+  const response = shallowRef<any>(null) as Ref<AxiosResponse<T> | null>
+  const data = shallowRef<any>(null) as Ref<T | null>
```
antfu, line 46: Since shalowRef does not unwrap, can we just have the type in the generic?

wheatjs: Hmm I'm not quite sure what you mean. I did what I thought you were talking about above but maybe that was a misunderstanding? Could we just do something like `const response = shallowRef<AxiosResponse<T>>()`

packages/integrations/useAxios/index.ts
```diff
@@ -42,12 +42,12 @@ export function useAxios<T = any>(url: string, ...args: any[]) {
       instance = args[1]
   }
 
-  const response = ref<any>(null) as Ref<AxiosResponse<T> | undefined>
-  const data = ref<any>(undefined) as Ref<T | undefined>
+  const response = shallowRef<AxiosResponse<T>>()
```
gkubisa, line 45: Should it be `const response = shallowRef<AxiosResponse<T> | undefined>()`?

wheatjs: image
It is since `shallowRef`'s signature is `export declare function shallowRef<T = any>(): Ref<T | undefined>;`

antfu: It's because you didn't pass the default value

```ts
const response = shallowRef<AxiosResponse<T> | null>(null)
```

wheatjs: I was going with the original typing that the functions had. Did you want them to be null instead of undefined?

antfu: Both work to me

antfu (approved):

PR #433, opened 2021-04-07
feat(useFocusTrap): new function

### LINKED ISSUE #420: useFocusTrap: new function
This is a proposal that would live under `@vueuse/integrations` as `useFocusTrap`. It would be a wrapper around https://github.com/focus-trap/focus-trap

My proposed api would look something like although I'm not 100% set on the names for the methods yet. Let me know if you think this would be a good addition and I'll open a PR

```ts
import { ref } from 'vue'
import { useFocusTrap } from '@vueuse/integrations'

const container = ref()
const { isTrapped, activate, deactivate  } = useFocusTrap(container, { options })
```

wheatjs:
Implementation of `useFocusTrap` function as discussed in #420.

Closes #420

### DIFF
--- a/packages/integrations/useFocusTrap/demo.vue
+++ b/packages/integrations/useFocusTrap/demo.vue
@@ -9,7 +9,7 @@ const { hasFocus, activate, deactivate } = useFocusTrap(target)
 <template>
   <div class="flex flex-col items-center">
     <button @click="activate()">
-      Trap Focus Within Form
+      {{ hasFocus ? 'Focus is trapped within form' : 'Trap focus within form' }}
     </button>
     <input
       type="text"
--- a/packages/integrations/useFocusTrap/demo.vue
+++ b/packages/integrations/useFocusTrap/demo.vue
@@ -9,7 +9,7 @@ const { hasFocus, activate, deactivate } = useFocusTrap(target)
 <template>
   <div class="flex flex-col items-center">
     <button @click="activate()">
-      Trap Focus
+      Trap Focus Within Form
     </button>
     <input
       type="text"
--- a/packages/integrations/useFocusTrap/demo.vue
+++ b/packages/integrations/useFocusTrap/demo.vue
@@ -1,0 +1,37 @@
+<script setup lang="ts">
+import { ref } from 'vue-demi'
+import { useFocusTrap } from '.'
+
+const target = ref()
+const { hasFocus, activate, deactivate } = useFocusTrap(target)
+</script>
+
+<template>
+  <div class="flex flex-col items-center">
+    <button @click="activate()">
+      Trap Focus
+    </button>
+    <input
+      type="text"
+      :placeholder="hasFocus ? 'You can\'t foucs me' : 'You can focus me'"
+    />
+
+    <div
+      ref="target"
+      class="shadow-lg bg-gray-600 bg-opacity-10 dark:(bg-gray-400 bg-opacity-10) rounded max-w-96 mx-auto p-8"
+    >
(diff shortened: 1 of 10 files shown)


packages/integrations/useFocusTrap/index.md
```diff
@@ -0,0 +1,119 @@
+---
+category: '@Integrations'
+---
+
+# useFocusTrap
+
+Reactive wrapper for [`useFocusTrap`](https://github.com/focus-trap/focus-trap)
```
antfu, line 7:
```suggestion
Reactive wrapper for [`focus-trap`](https://github.com/focus-trap/focus-trap)
```

packages/integrations/useFocusTrap/index.ts
```diff
@@ -0,0 +1,126 @@
...
+   *
+   * @link https://github.com/focus-trap/focus-trap#trapunpause
+   */
+  unpause: Fn
+}
+
+/**
+ * Reactive focus-trap
+ *
+ * @link https://vueuse.org/integrations/useFocusTrap/
```
antfu, line 57:
```suggestion
 * @link https://vueuse.org/useFocusTrap
```

We have HTTP alias set in Netlify

packages/integrations/useFocusTrap/index.ts
```diff
@@ -0,0 +1,126 @@
...
+      trap.unpause()
+      isPaused.value = false
+    }
+  }
+
+  const stop = watch(
+    () => unrefElement(target),
+    (el: HTMLElement) => {
+      if (el)
+        stop()
```
antfu, line 90: Should we support continue changing targets as other functions do?

packages/integrations/useFocusTrap/demo.vue
```diff
@@ -0,0 +1,37 @@
...
+</script>
+
+<template>
+  <div class="flex flex-col items-center">
+    <button @click="activate()">
+      Trap Focus
+    </button>
+    <input
+      type="text"
+      :placeholder="hasFocus ? 'You can\'t foucs me' : 'You can focus me'"
```
antfu, line 16: The demo looks interesting 👍

Maybe a line of "The focus is trapped in the form" somewhere could better showing the feature?

packages/integrations/useFocusTrap/index.ts
wheatjs: Hmm yeah that's not a bad idea. I'll see what I can do

packages/integrations/useFocusTrap/index.ts
```diff
@@ -113,7 +110,10 @@ export function useFocusTrap(target: MaybeElementRef, options: UseFocusTrapOptio
     }, { flush: 'post' })
 
   // Cleanup on unmount
-  tryOnUnmounted(deactivate)
+  tryOnUnmounted(() => {
+    stop()
```
antfu, line 114: No need for stopping a watch with this hook - it's called automatically.

wheatjs: Ah right, forgot 😅 

Fixed

packages/integrations/useFocusTrap/demo.vue
```diff
@@ -0,0 +1,37 @@
...
+
+const target = ref()
+const { hasFocus, activate, deactivate } = useFocusTrap(target)
+</script>
+
+<template>
+  <div class="flex flex-col items-center">
+    <button @click="activate()">
+      Trap Focus Within Form
+    </button>
```
antfu, line 13: We could also reflect the activated state on the button

PR #436, opened 2021-04-08
feat(useTransition): support for disabled transitions

scottbedard:
Transitions now support a `disabled` property, which can be useful when synchronously resetting a source value.

```ts
const disabled = ref(false)

useTransition(source, { disabled })
```

Some minor cleanup is also included in this PR. I removed an old test file that must have re-emerged after a bad merge, and renamed `.spec.ts` files to `.test.ts`

And with this, I'm considering #364 complete! Repeatable transitions are a fairly different type of action, and it'll be best to keep this function from becoming more complicated than it needs to be. I'll open an issue to brainstorm the API for a repeat transition function sometime in the future.


packages/core/useTransition/index.test.ts
```diff
@@ -1,146 +1,230 @@
...
+    disabled.value = true
+    source.value = 1
+
+    expect(transition.value).toBe(1)
     await promiseTimeout(150)
-    expect(onStarted.mock.calls.length).toBe(1)
-    expect(onFinished.mock.calls.length).toBe(1)
+    expect(onStarted).not.toHaveBeenCalled()
+    disabled.value = false
+    expect(transition.value).toBe(1)
```
scottbedard, line 228: Sorry for the messy diff here. I tried to keep this clean by deleting the old test file in it's own commit, but apparently that did not matter. The only real difference here is the addition of the disabled test.

antfu (approved):

PR #444, opened 2021-04-12
feat(autoResetRef): new function

### LINKED ISSUE #432: `autoResetRef`
This is a new ref function proposal.
Maybe it is better to be a composable function.

This is a value which will be reset to the default value after some time.

### Usage
```ts
const usePost = () => {
  const errorMessage = autoResetRef('', 1000)

  const post = (n: number) => {
    try {
      // something
    } catch {
      errorMessage.value = 'had error'
      // here the value will change to 'had error' but after 1000ms, it will change to ''
    }
    return { errorMessage, post }
  }
}
```

### Sample Implementation
```ts
import { customRef, Ref } from 'vue'

export const autoResetRef = <T>(defaultValue: T, afterMs = 10000): Ref<T> =>
  customRef<T>((track, trigger) => {
    let value: T = defaultValue
    let timeout = 0

    const resetAfter = () =>
      setTimeout(() => {
        value = defaultValue
        trigger()
      }, afterMs)

    return {
      get() {
        track()
        return value
      },
      set(newValue) {
        value = newValue
        trigger()

        clearTimeout(timeout)
        timeout = resetAfter()
      }
    }
  })
```

sapphi-red:
This PR is the implementation of #432.

closes #432

### DIFF
--- a/packages/core/autoResetRef/index.ts
+++ b/packages/core/autoResetRef/index.ts
@@ -1,0 +1,35 @@
+import { customRef, Ref } from 'vue-demi'
+
+/**
+ * Create a ref which will be reset to the default value after some time.
+ *
+ * @link https://vueuse.org/autoResetRef
+ * @param defaultValue The value which will be set.
+ * @param afterMs      A zero-or-greater delay in milliseconds.
+ */
+export function autoResetRef<T>(defaultValue: T, afterMs = 10000): Ref<T> {
+  return customRef<T>((track, trigger) => {
+    let value: T = defaultValue
+    let timer: any
+
+    const resetAfter = () =>
+      setTimeout(() => {
+        value = defaultValue
+        trigger()
+      }, afterMs)
+
+    return {
+      get() {
+        track()
+        return value
+      },
+      set(newValue) {
+        value = newValue
+        trigger()
+
+        clearTimeout(timer)
+        timer = resetAfter()
+      },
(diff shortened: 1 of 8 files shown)


packages/core/autoResetRef/index.ts
```diff
@@ -0,0 +1,35 @@
+import { customRef, Ref } from 'vue-demi'
+
+/**
+ * Create a ref which will be reset to the default value after some time.
+ *
+ * @link https://vueuse.org/autoResetRef
+ * @param defaultValue The value which will be set.
+ * @param afterMs      A zero-or-greater delay in milliseconds.
+ */
+export function autoResetRef<T>(defaultValue: T, afterMs = 10000): Ref<T> {
```
wheatjs, line 10: Maybe we could have `afterMs` be a ref so it can be changed?

wheatjs: or more specifically set the type to MaybeRef, and then just use `unref` to access it so users can pass in either just a number or a ref

sapphi-red: I have changed `afterMs` to `MaybeRef`!

wheatjs (approved):
Looks good to me, once Antfu has a chance to go over this once more it should be good to merge.

Nice work! 😄

antfu (approved):

PR #448, opened 2021-04-13
fix(useFirestore): optional Chaining on isDocumentReference

### LINKED ISSUE #447: useFirestore Error when using Query instead of Collection
This works perfectly

```js
const messagesCollection = db.collection('messages')

export const messages = useFirestore(messagesCollection)
```

But when I try to do this
```js
const messagesCollection = db.collection('messages')
const messagesQuery = messagesCollection.orderBy('createdAt', 'desc').limit(10)

export const messages = useFirestore(messagesQuery)
```

It fails with the error 
```
Uncaught TypeError: Cannot read property 'match' of undefined
    at isDocumentReference (index.esm.js:36)
    at useFirestore (index.esm.js:49)
    at useFirebase.js?t=1618288519311:23
```

rromanv:
Fixes #447

### DIFF
--- a/packages/firebase/useFirestore/index.ts
+++ b/packages/firebase/useFirestore/index.ts
@@ -27,7 +27,7 @@ function getData<T>(
 }
 
 function isDocumentReference<T>(docRef: any): docRef is firebase.firestore.DocumentReference<T> {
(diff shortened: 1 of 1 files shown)


lstoeferle (approved):
LGTM, thanks!

Issue #363, opened 2021-03-06
Preparing v5.0 (Draft)


antfu:
We are preparing our next major updates to VueUse. 

## Planned changes

- Discontinue APIs that currently marked as `@depreacted`
- Introducing `controls` options, #362
- Renderless components `@vueuse/components`, #486
- Revisit functions
- ...

We are expecting to have breaking changes in this update. If you have any idea of refactoring/redesign current functions, do hesitate to let us know, thanks!

antfu:
Nap, just need some time slot for getting back on this project, and also #486. Hopefully I could get it done in a week or two.

PR #454, opened 2021-04-15
Update `useFetch` to allow more calls

### LINKED ISSUE #442: `useFetch` cannot post with different data
Is it possible to call once
```js
const {isFetching, post, ...} = userFetch('someURL', {
    immediate: false
})
```
and after, when some data have changed call more times?

```js
// This is inside a callback
post({data: '...'}).execute()
```

I'm able to do so once, but not more times. This is really handy when you have a form, the user submits it, some error happens (probably validation), user changes the data and submits it again. 

I see the code responsible for this  is 
```js
function setMethod(method: string) {
    return (payload?: unknown, payloadType?: PayloadType) => {
      if (!initialized) {
        config.method = method
        config.payload = payload
        config.payloadType = payloadType || typeof payload === 'string' ? 'text' : 'json'
        return shell as any
      }
      return undefined
    }
  }
```
Is there any reason for doing so that I'm missing?
Same for `setType`?

isgj:
fixes #442 

Now there is only one interface `UseFetchReturn` that is returned, when you call `useFetch` the first time, set the method or update the type. Of course you cannot change the method or the type while it's making a request.

In addition this fixes (unless it was intended) the loading state when the request is cancelled in `beforeFetch` .

I didn't touch the tests nor the docs. before let me know what you think about this approach.

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -227,9 +227,13 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
       controller.abort()
   }
 
+  const loading = (isLoading: boolean) => {
+    isFetching.value = isLoading
+    isFinished.value = !isLoading
+  }
+
   const execute = async() => {
-    isFetching.value = true
-    isFinished.value = false
+    loading(true)
     error.value = null
     statusCode.value = null
     aborted.value = false
@@ -269,8 +273,10 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
     if (options.beforeFetch)
       Object.assign(context, await options.beforeFetch(context))
 
-    if (isCanceled || !fetch)
+    if (isCanceled || !fetch) {
+      loading(false)
       return Promise.resolve()
+    }
 
     return new Promise((resolve) => {
       fetch(
@@ -300,8 +306,7 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
           error.value = fetchError.message || fetchError.name
         })
         .finally(() => {
-          isFinished.value = true
-          isFetching.value = false
+          loading(false)
         })
     })
   }
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -57,21 +57,19 @@ interface UseFetchReturnBase<T> {
 type DataType = 'text' | 'json' | 'blob' | 'arrayBuffer' | 'formData'
 type PayloadType = 'text' | 'json' | 'formData'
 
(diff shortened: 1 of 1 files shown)


isgj:
> `useFetch` is designed to be used for one response type at once.

This is still true, you can't change the type nor the method (and data) while making a call.

More than different types, this is about different data as explained in the linked issue. I can update this PR so for the functions that set the type the returned interface is `base` + method functions (but it wont change much).

packages/core/useFetch/index.ts
```diff
@@ -361,9 +362,9 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
 
   function setType(type: DataType) {
     return () => {
-      if (!initialized) {
+      if (!isFetching.value) {
```
wheatjs, line 365: I think this should still be `!initalized`, right? Changing the payload seems fine to me, but we don't want to change the type.

isgj: In this PR `initialized` is not used. It was set to true the first time `execute` was called and stayed so.
Now in master now you can do

```js
const {post} = useFetch(..., {immediate: false})
// somewhere else
post(...).json().execute() // of course only once
```
If you'll allow more calls `json` will return `undefined` and you cannot chain `execute` and ... (you got the idea).

In the last commit I changed the types so you can do
```js
const {post} = useFetch(..., {immediate: false}).json()
// somewhere else
post(...).execute() // even more times
```

### To note
This is a breaking change as now the order you chain the method is strict `post -> json (method --> type)`, here instead is the inverse `json --> (post X times)`.
If the order shouldn't be strict then those functions should return the same interface, otherwise keeping track what was called and should be returned would be unnecessary

antfu:
I see. That's a good point. Since it's a breaking change, I'd more like to have it in the v5 release #363. Will handle it soon

schelmo:
👍
the next cool thing would be refetching when the the payload is changing!
like 
```js
const payload = ref({some: 'data'})
useFetch('/', {refetch: true}).post(payload)
payload.value.more = 'moar' // -> refetches
```
maybe even watch the whole config object? (so, changes to type, method, payload, payloadType will trigger a refetch)

wheatjs:
> 👍
> the next cool thing would be refetching when the the payload is changing!
> like
> 
> ```js
> const payload = ref({some: 'data'})
> useFetch('/', {refetch: true}).post(payload)
> payload.value.more = 'moar' // -> refetches
> ```
> 
> maybe even watch the whole config object? (so, changes to type, method, payload, payloadType will trigger a refetch)

Hmm probably would just want to watch the payload for refetching. I don't think we really want to be changing methods

isgj:
> Hmm probably would just want to watch the payload for refetching.

I agree. And resending some data with `POST` (or `PUT`, ...)  you're not refetching but changing something on the server.

PR #459, opened 2021-04-17
feat(and,or,not): new functions

antfu:
Logic reactive operations - `and` `or` `not`

Not sure if it's a good idea or not...


patak-js:
I think this looks good, another way to use the name `and` could be for a function that creates a ref that only triggers when both refs have changed. Imagine something like:
```js
watch( and( refA, refB ), ...
```
But I don't know how useful this is (maybe waiting for several fetches to resolve? or only doing an action if every field has been touched in a form?). The `or` is just passing an array, so instead of `and` probably better to offer
```js
watchAll( [refA, refB], ... // kind of equivalent to promise.all
```
So... all this to say that I think it is safe to use the names `and` `or` `not` here 👍🏼

antfu:
Oh, for me, I am implementing shortcuts navigations using `useMagicKeys` where I would seomtimes need to disable the shortcuts work users focus on `input` or something.

For now I need to have this:

```ts
const { space, right, left, up, down } = useMagicKeys()

whenever(() => space.value && !paused.value, next)
whenever(() => right.value && !paused.value, next)
whenever(() => left.value && !paused.value, prev)
whenever(() => up.value && !paused.value, prevSlide)
whenever(() => down.value && !paused.value, nextSlide)
```

But with these API, I can write them as


For now I need to have this:

```ts
const { space, right, left, up, down } = useMagicKeys()

const enabled = not(paused)

whenever(and(space, enabled), next)
whenever(and(right, enabled), next)
whenever(and(left, enabled), prev)
whenever(and(up, enabled), prevSlide)
whenever(and(down, enabled), nextSlide)
```

Just some preference to make less `.value` 😄. It's not that urgent, just sending the proposal here for discussions. 

-----

Re @patak-js: Not sure if it's common to use `and` as both changed? That's could be a bit unclear to behavior, like what if `a` changed twice while `b` does not change?

But inspiring from this, I do think something like `const ms = useLastChanged(refA)` could be a bit more useful and atomic on this? Like

```ts
const time = ref(Date.now())

cosnt msA = useLastChanged(a)
cosnt msB = useLastChanged(b)

watch([a, b], () => {
  if (msA.value > time.value && msB.value > time.value) {
    time.value = Date.now() // <--
    /* ... */
  } 
})
```

wheatjs:
Would something like this be possible?

```ts
whenever(and(cond1, not(cond2), or(cond3, cond4)), next)
```

antfu:
> Would something like this be possible?
> 
> ```ts
> whenever(and(cond1, not(cond2), or(cond3, cond4)), next)
> ```

Yes of course.

wheatjs:
Oh yeah sorry I guess that is obvious, only did a quick look at it on my phone. I like this feature though, think it makes everything pretty clean.

PR #461, opened 2021-04-17
feat(reactivePick): allow ref picks

### LINKED ISSUE #460: [feature] reactivePick: support computed keys
Currently `reactivePick` allows you to pick options from an object and have the resulting object be reactive. While this is already useful, the function could be made a lot more useful if it was possible to also have the picking action itself be reactive, i.e. the keys to pick supporting a `Ref`. In my use case I have a component wrapping three subcomponents, and I want to pick the subcomponents props from the wrapper components props since not all subcomponents use all props (I have to keep this design for backwards compatibility).

I propose the following change (and I'll try to create a PR later):

Add a new overload to `reactivePick`:
```typescript
reactivePick<T extends object, K extends keyof T>(obj: T, keys: MaybeRef<K[]>): Ref<{ [S in K]: UnwrapRef<T[S]> }>
```

This overload will react to changes in the `keys` parameter and modify the returned computed ref accordingly.

TimonLukas:
Closes #460.

This PR adds the ability to pass not only variadic keys, but also an array of keys or a ref array of keys. This is done while (I think) keeping BC. Tests work fine on 3, on 2 Jest completely hangs up got me, even with `--runInBand --detectOpenHandles --forceExit`. I hope these are the usual Jest shenanigans and not a real problem.

### DIFF
--- a/packages/shared/reactivePick/index.md
+++ b/packages/shared/reactivePick/index.md
@@ -19,6 +19,10 @@ const obj = reactive({
 })
 
 const picked = reactivePick(obj, 'x', 'elementX') // { x: number, elementX: number }
+// or
+const pickedArr = reactivePick(obj, ['x', 'elementX']) // { x: number, elementX: number }
+// or
+const pickedArrRef = reactivePick(obj, ref(['x', 'elementX'])) // { x: number, elementX: number }
 ```
 
 ### Scenarios
@@ -43,6 +47,8 @@ const props = defineProps({
 })
 
 const childProps = reactivePick(props, 'color', 'font')
+// or
+const childPropsArr = reactivePick(props, Object.keys(ChildComp.props))
 </script>
 
 <template>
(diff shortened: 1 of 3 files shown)


TimonLukas:
It appears the Jest problems are not local to my computer. I'll try to debug them, but help would be greatly appreciated - this seems really weird!

antfu:
Thanks for your contributions, but I am afraid if it's a good idea to introduce such complexity to these simple utils. I saw you found a solution for your problem https://github.com/vueuse/vueuse/issues/460#issuecomment-821888254, do you still feel these changes needed? (since that approach is much cleaner 👍)

Thanks!

TimonLukas:
I think it could still be useful, but you're right that the increased complexity probably isn't worth it :)

antfu:
Maybe we can further break it down into smaller functions to compose it. A good one might be `replaceReactive` there will delete extras keys when the target does not have them. So we can separate the concern and make the codebase more maintainable. WDTY? If you want, feel free to open a new PR. Thanks

antfu:
Thanks for contributions. Close for now.

PR #477, opened 2021-04-28
docs: replaced `@link` to `@see` in jsdoc

cawa-93:
`@link` is **inline** directive. Therefore, using it as
```js
/**
 * @link https://...
 */
```
 is actually a wrong syntax. As a result, some editors do not handle links correctly:

зображення
Note that the link is `https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver|MutationObserver`.

Also, in rendered mode `@link` are not parsed at all, 
зображення

Because the `@link` is used as a block directive, I replaced it with the `@see` directive, and changed the syntax for anchors in some places. Now:

зображення
зображення

Alternatively, you could use the link directive, but in an inline mode:
```js
/**
 * {@link https://...}
 */
```

### DIFF
--- a/packages/core/asyncComputed/index.md
+++ b/packages/core/asyncComputed/index.md
@@ -107,7 +107,7 @@ export declare type AsyncComputedOptions = {
 /**
  * Create an asynchronous computed dependency.
  *
- * @link https://vueuse.org/asyncComputed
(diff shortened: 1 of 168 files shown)


antfu:
Thanks a lot!

Issue #479, opened 2021-04-29
useTitle: Not react on title changes from another component


cawa-93:
I have two components. One downloads the data and changes the header based on it. The second - just displays this title.

```js
// Data fetcher Component
{
  setup() {
        const title = useTitle();
        watch(someData, () => title.value = someData.value.someValue);
  }
}
```

```vue
// CustomPageTitle Component
<template>
  <span>{{ title }}</span>
</template>

<script>
export default defineComponent({
  setup() {
    const title = useTitle();
    return {title};
  },
});
</script>
```

When the application works, the page title and `document.title` are constantly changing, but the `CustomPageTitle Component` does not respond to these changes.

PR #483, opened 2021-04-30
feat(useTitle): new option `observe` to watch title changes

cawa-93:
Should resolve https://github.com/vueuse/vueuse/issues/479


antfu:
Great, I think we can also provide an option allowing disable it (fine with default true)

antfu:
Made some changes, after rethink about this, I think observing title is rather a rare operations and properly expensive. Changed the default to disabled to match with previous behavior. Thanks

PR #486, opened 2021-05-03
feat: introduce `components` & `directives`

wheatjs:
Creating a draft PR for the components and directives so there can be open discussion about it as it is developed. 

There are currently a few components and one directive. We still need to setup the infrastructure for actually exporting the components like we do the other utils. Also wondering how we should handle the documentation for them. We could include them in the index.md file for each component?


cawa-93:
Haven't you thought about putting all the components in a separate package, let's say `@vueuse/components`?

wheatjs:
That was the original intention, but then we thought it would be easier to test and keep up to date if they were closer together.

antfu:
Yeah, they will be put under the same folder as each function, but published as `@vueuse/components`

antfu:
Could spread out, like

```html
<UseNow v-slot="{ now }" :interval="10">
  {{ now }}
</UseNow>
```

And when you want to do options way:

```html
<UseNow v-slot="{ now }" v-bind="{ interval: 10 }">
  {{ now }}
</UseNow>
```

wheatjs:
Sounds good to me

wheatjs:
Wondering how we should handle the docs for this. Should they be part of the same docs or maybe they  should be separated, by something like tabs maybe? I guess the main thing we need right now is an easy way to test/demo them. It would be easier to see if everything is working like expected if we had a way to demo the main functions.

wheatjs:
Ok so in total for our first release it looks like we will have 30 components including the following

- onClickOutside
- useActiveElement
- useBattery
- useBrowserLocation
- useDark
- useDeviceLight
- useDeviceMotion
- useDeviceOrientation
- useDevicePixelRatio
- useDevicesList
- useDocumentVisibility
- useElementBounding
- useElementSIize
- useElementVisibility
- useFullscreen
- useGeolocation
- useIdle
- useMouse
- useMouseInElement
- useMousePressed
- useNetwork
- useNow
- useOnline
- usePageLeave
- usePreferredColorScheme
- usePreferredDark
- usePreferredLanguages
- useTimeAgo
- useTimestamp
- useWindowSize

Each component includes minimal documentation showing basic usage.

antfu (approved):

PR #487, opened 2021-05-05
useOneWayTransition

scottbedard:
I was originally going to tackle this in #364, but decided to submit it as its own function instead of shoe-horning it into `useTransition`. Transitions that go between fixed values are almost always what I find myself needing, I think having this built-in would be handy.

```ts
import { useOneWayTransition, TransitionPresets } from '@vueuse/core'

const { transition, value } = useOneWayTransition({
  duration: 1000,
  transition: TransitionPresets.easeInOutCubic,
})

transition({ from, to })
```

Let me know what you think, if we decide to move forward with this I need to clean it up with docs and tests. I'm also not sure about the name. I looked at `useFixedTransition` as well, but that didn't seem quite right either.


scottbedard:
Nevermind, looking at this again with a fresh set of eyes and it seems a bit heavy-handed. The whole promise API was just to deal with under/overflow during interrupted transitions. But in my applications, it's easier just to guard against this myself which lets the one-way transition function be much lighter.

Issue #449, opened 2021-04-13
Add some features from useFetch also to useAxios


3nuc:
Currently useFetch has execute(), but useAxios does not. This means that with useFetch you can first create useFetch, set {immediate: false}, then execute() the call later, but with useAxios it fires immediately when you call useAxios.

Abort is also missing in useAxios.

Some properties don't make sense to add into axios such as statusText (since that's just in the axios response).

Another thing is that the returned property names are different in useAxios and useFetch, but changing that'd be a breaking change.

Exposed logic from both logics:
<details>
<summary>useAxios</summary>

```
  response: Ref<AxiosResponse<T> | undefined>
  data: Ref<T | undefined>
  finished: Ref<boolean>
  loading: Ref<boolean>
  canceled: Ref<boolean>
  error: Ref<AxiosError<T> | undefined>
```
</details>

<details>
<summary>useFetch</summary>

```
  isFinished: Ref<boolean>
  statusCode: Ref<number | null>
  response: Ref<Response | null>
  error: Ref<any>
  data: Ref<T | null>
  isFetching: Ref<boolean>
  canAbort: ComputedRef<boolean>
  aborted: Ref<boolean>
  abort: Fn
  execute: () => Promise<any>
```
</details>

3nuc:
Or perhaps it's better to just use https://vueuse.org/core/useAsyncState/ for requests in general?

lstoeferle:
Hey @3nuc,

I would love to see the `execute()` logic in `useAxios` too. AFAIK there is already an option to cancel request in `useAxios` (I can remember me adding it).

But I don't see any value in introducing breaking changes while refactoring the return type to match `useFetch`. IMO there are only a few rare cases, where somebody is using axios besides the fetch api and vice versa. Typically you only use one of both, which eliminates the risk of mixing the return types of `useAxios` and `useFetch`.

Feel free to open a PR 😉

Kind regards,
Lukas

antfu:
Would love to align those interfaces. We can make aliases and for APIs and mark some as depreacted. Then drop them in the next major.

wheatjs:
Which one are we looking to align against? Make `useAxios` align with the `useFetch` interface or `useFetch` align with the `useAxios` interface?

antfu:
Aligning with useFetch looks better to me

wheatjs:
Sounds good, I'll make a PR sometime this week!

raphaelsaunier:
[...]

PR #498, opened 2021-05-12
feat(useAxios): bring API into line with useFetch

wheatjs:
See #449 for more deatils.

Initial attempt to bring the `useAxios` API more inline with the `useFetch` API. This PR doesn't include the `execute()` option that was mentioned in #449, as we would probably need another config option to enable something like that. It would probably be best to avoid too many overrides as we already have 3. Let me know if you guys have a good idea on how you'd like to implement something like that.


wheatjs:
Oops no idea what happened here, sorry

PR #500, opened 2021-05-12
fix(useMediaControls): Doesn't rewrite default media properties

### LINKED ISSUE #492: useMediaControls: Overwrites the values set by the attributes
https://github.com/vueuse/vueuse/blob/ecd649621c4f7a6703f3cf2d897eb81a2f47a5db/packages/core/useMediaControls/index.ts#L313-L320

I'm not sure if this was intended, but it overwrites the default (browser) values and those that were set by the attributes.

Example:
```html
<video controls src="">
```
```js
useMediaControls(video, {})
```
Expected:
```js
video.controls // true // set by attribute
video.preload // "metadata" // default browser behaviour
```
Actual:
```js
video.controls // false
video.preload // "auto"
```

cawa-93:
Resolve #492

### DIFF
--- a/packages/core/useMediaControls/index.ts
+++ b/packages/core/useMediaControls/index.ts
@@ -303,29 +303,29 @@ export function useMediaControls(target: MaybeRef<HTMLMediaElement | null | unde
       return
 
     const loop = unref(options.loop)
-    if (loop) el.loop = loop
+    if (loop !== undefined) el.loop = loop
 
     const controls = unref(options.controls)
-    if (controls) el.controls = controls
+    if (controls !== undefined) el.controls = controls
 
     const muted = unref(options.muted)
-    if (muted) el.muted = muted
+    if (muted !== undefined) el.muted = muted
 
     const preload = unref(options.preload)
-    if (preload) el.preload = preload
+    if (preload !== undefined) el.preload = preload
 
     const autoplay = unref(options.autoplay)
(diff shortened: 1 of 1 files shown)


packages/core/useMediaControls/index.ts
```diff
@@ -310,14 +302,32 @@ export function useMediaControls(target: MaybeRef<HTMLMediaElement | null | unde
...
-    el.preload = unref(options.preload)!
-    el.autoplay = unref(options.autoplay)!
+    const loop = unref(options.loop)
+    if (loop) el.loop = loop
+
+    const controls = unref(options.controls)
+    if (controls) el.controls = controls
+
+    const muted = unref(options.muted)
+    if (muted) el.muted = muted
```
wheatjs, line 312: Hmm, maybe it would be better to check for the existence of the property rather than the actual value. For example if the user set something like this, then the video would still be muted. I think the settings in `useMediaControls` should override any attributes.

```ts
useMediaControls(target, {
  muted: false
})
```
and the video element had
```html
<video muted></video>
```
The expected outcome of above should be `muted: false` but it would be `muted: true`.

But in this case below
```ts
useMediaControls(target)
```
and the video element had
```html
<video muted></video>
```
The expected outcome should be `muted: true`

cawa-93: I Agree

wheatjs (approved):
Look good to me I think, thank!

PR #507, opened 2021-05-13
feat(useWebWoker): return worker

### LINKED ISSUE #381: useWebWorker: return also the worker
I use web workers with [Comlink](https://github.com/GoogleChromeLabs/comlink) and cannot use `useWebWorker` function, I need the web worker to proxy it.

wheatjs:
Closes #381

### DIFF
--- a/packages/core/useWebWorker/index.ts
+++ b/packages/core/useWebWorker/index.ts
@@ -1,6 +1,6 @@
 /* this implementation is original ported from https://github.com/logaretm/vue-use-web by Abdelrahman Awad */
 
-import { ref, Ref } from 'vue-demi'
+import { ref, Ref, shallowRef } from 'vue-demi'
 import { tryOnUnmounted } from '@vueuse/shared'
 import { ConfigurableWindow, defaultWindow } from '../_configurable'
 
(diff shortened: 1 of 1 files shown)


packages/core/useWebWorker/index.ts
```diff
@@ -55,5 +55,6 @@ export function useWebWorker(
     data,
     post,
     terminate,
+    worker: worker!,
```
antfu, line 58: The problem here is that the worker is a variable and might changes during the lifecyle. Better  to make it a shallowRef internally and expose the ref.

wheatjs: Ah yeah I started doing that at first but had some issues with the typing, I'll try to update it to use shallowRef again

wheatjs:
For the types I'm no sure how to keep it consistent with what it was before, so I created a dummy variable. Let me know if you have a better idea on how we should handle that

packages/core/useWebWorker/index.ts
```diff
@@ -22,38 +22,41 @@ export function useWebWorker(
   } = options
 
   const data: Ref<any> = ref(null)
-  let worker: Worker
+  let _worker: Worker
+  const worker = shallowRef<Worker>()
 
-  const post: typeof worker.postMessage = function post(val: any) {
-    if (!worker)
+  const post: typeof _worker.postMessage = function post(val: any) {
```
antfu, line 28:
```suggestion
  const post: typeof Worker['postMessage'] = function post(val: any) {
```

Would this work?

wheatjs: Doesn't seem to, gives the following error

```
Property 'postMessage' does not exist on type '{ new (stringUrl: string | URL, options?: WorkerOptions | undefined): Worker; prototype: Worker; }'
```

antfu (approved):

PR #508, opened 2021-05-13
feat(onClickOutside): default to just pointerDown 

### LINKED ISSUE #451: potential bug: onClickOutside
It looks like `onClickOutside` is being called multiple times when the element target is not removed after the `onClickOutside` event is fired. This doesn't show up on the VueUse demo page because the targets are removed, but if they aren't it will be fired twice.

Here is a reproduction on [Vue SFC Playground](https://sfc.vuejs.org/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCB7IHJlZiB9IGZyb20gJ3Z1ZSdcbmltcG9ydCB7IG9uQ2xpY2tPdXRzaWRlIH0gZnJvbSAnQHZ1ZXVzZS9jb3JlJ1xuXG5jb25zdCB0YXJnZXQgPSByZWYoKVxuY29uc3QgY2xpY2tlZCA9IHJlZigwKVxuXG5vbkNsaWNrT3V0c2lkZSh0YXJnZXQsICgpID0+IGNsaWNrZWQudmFsdWUgKz0gMSlcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxkaXYgc3R5bGU9XCJkaXNwbGF5OiBncmlkOyBwbGFjZS1jb250ZW50OiBjZW50ZXI7IHBsYWNlLWl0ZW1zOiBjZW50ZXI7IGhlaWdodDogMTAwJTsgZ2FwOiAxcmVtOyB1c2VyLXNlbGVjdDogbm9uZTtcIj5cbiAgICBDbGlja2VkIE91dHNpZGUge3sgY2xpY2tlZCB9fSB0aW1lKHMpXG4gICAgPGRpdiByZWY9XCJ0YXJnZXRcIiBzdHlsZT1cImJhY2tncm91bmQ6IHdoZWF0OyBwYWRkaW5nOiAycmVtOyBib3JkZXItcmFkaXVzOiAwLjI1cmVtO1wiPlxuICAgICAgQ2xpY2sgT3V0c2lkZSBvZiBtZVxuICAgIDwvZGl2PlxuICA8L2Rpdj5cbjwvdGVtcGxhdGU+XG4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG5cdFx0XCJAdnVldXNlL2NvcmVcIjogXCJodHRwczovL3VucGtnLmNvbS9AdnVldXNlL2NvcmUvZGlzdC9pbmRleC5lc20uanNcIixcbiAgICBcIkB2dWV1c2Uvc2hhcmVkXCI6IFwiaHR0cHM6Ly91bnBrZy5jb20vQHZ1ZXVzZS9zaGFyZWQvZGlzdC9pbmRleC5lc20uanNcIixcbiAgICBcInZ1ZS1kZW1pXCI6IFwiaHR0cHM6Ly91bnBrZy5jb20vdnVlLWRlbWkvbGliL2luZGV4LmVzbS5qc1wiXG4gIH1cbn0ifQ==)
 
Is this behavior intentional? Maybe this behavior was introduced in this issue https://github.com/vueuse/vueuse/issues/323? If this is intentional maybe we should add an option to only fire the event once.

wheatjs:
default to just pointerDown and add option to include more events. Closes #451

### DIFF
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -7,7 +7,7 @@ const events = ['mousedown', 'touchstart', 'pointerdown'] as const
 type EventType = WindowEventMap[(typeof events)[number]]
 
 export interface OnClickOutsideOptions extends ConfigurableWindow {
-  events: ('mousedown' | 'touchstart' | 'pointerdown')[]
+  events: string[]
 }
 
 /**
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -1,13 +1,11 @@
-import { Fn, tryOnUnmounted } from '@vueuse/shared'
+import { tryOnUnmounted } from '@vueuse/shared'
 import { MaybeElementRef, unrefElement } from '../unrefElement'
 import { useEventListener } from '../useEventListener'
 import { ConfigurableWindow, defaultWindow } from '../_configurable'
 
-const events = ['mousedown', 'touchstart', 'pointerdown'] as const
-type EventType = WindowEventMap[(typeof events)[number]]
-
-export interface OnClickOutsideOptions extends ConfigurableWindow {
-  events: string[]
+export type AllowedEvents = Pick<WindowEventMap, 'mousedown' | 'mouseup' | 'touchstart' | 'touchend' | 'pointerdown' | 'pointerup'>
+export interface OnClickOutsideOptions<E extends keyof AllowedEvents> extends ConfigurableWindow {
+  event: E
 }
 
 /**
@@ -18,11 +16,11 @@ export interface OnClickOutsideOptions extends ConfigurableWindow {
  * @param handler
  * @param options
  */
-export function onClickOutside(
+export function onClickOutside<E extends keyof AllowedEvents = 'pointerdown'>(
   target: MaybeElementRef,
-  handler: (evt: EventType) => void,
-  options: OnClickOutsideOptions = {
-    events: ['pointerdown'],
+  handler: (evt: AllowedEvents[E]) => void,
+  options: OnClickOutsideOptions<E> = {
+    event: 'pointerdown' as E,
   },
 ) {
   const { window = defaultWindow } = options
@@ -30,7 +28,7 @@ export function onClickOutside(
   if (!window)
     return
 
-  const listener = (event: EventType) => {
+  const listener = (event: AllowedEvents[E]) => {
     const el = unrefElement(target)
     if (!el)
       return
@@ -41,13 +39,7 @@ export function onClickOutside(
     handler(event)
   }
 
-  let disposables: Fn[] = options.events
-    .map(event => useEventListener(window, event, listener, { passive: true }))
-
-  const stop = () => {
-    disposables.forEach(stop => stop())
-    disposables = []
-  }
+  const stop = useEventListener(window, options.event, listener, { passive: true })
 
   tryOnUnmounted(stop)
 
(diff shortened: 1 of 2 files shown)


packages/core/onClickOutside/index.ts
```diff
@@ -6,6 +6,10 @@ import { ConfigurableWindow, defaultWindow } from '../_configurable'
 const events = ['mousedown', 'touchstart', 'pointerdown'] as const
 type EventType = WindowEventMap[(typeof events)[number]]
 
+export interface OnClickOutsideOptions extends ConfigurableWindow {
+  events: ('mousedown' | 'touchstart' | 'pointerdown')[]
```
antfu, line 10:
```suggestion
  events: string[]
```

packages/core/onClickOutside/index.ts
```diff
@@ -17,7 +21,9 @@ type EventType = WindowEventMap[(typeof events)[number]]
 export function onClickOutside(
   target: MaybeElementRef,
   handler: (evt: EventType) => void,
-  options: ConfigurableWindow = {},
+  options: OnClickOutsideOptions = {
+    events: ['pointerdown'],
+  },
 ) {
   const { window = defaultWindow } = options
```
antfu, line 28: Let's assign the default value here.

antfu (changes requested):

packages/core/onClickOutside/index.ts
wheatjs: Which part do you mean?

sibbng:
I my opinion we should only accept single string of whitelisted event names. Contrary to my earlier suggestion I don't think uses cases which needs to listen multiple pointer events are that common. 

Also to return proper type on event handler, end users have to mark their events array with `as const` which is a extra overhead in my opinion. As long as current status of this PR only accept array, it is also an overhead for anyone who needs to listen only one event.

wheatjs:
> I my opinion we should only accept single string of whitelisted event names. Contrary to my earlier suggestion I don't think uses cases which needs to listen multiple pointer events are that common.
> 
> Also to return proper type on event handler, end users have to mark their events array with `as const` which is a extra overhead in my opinion. As long as current status of this PR only accept array, it is also an overhead for anyone who needs to listen only one event.

So we should only listen to `pointerdown` then?

packages/core/onClickOutside/index.ts
antfu:
```ts
 const { window = defaultWindow, events = ['pointerdown'] } = options
```

sibbng:
I pushed the changes I suggested above. Also it gives proper event type to handler based on `event` option. I'm happy with this changes.

wheatjs:
This looks good to me

packages/core/onClickOutside/index.ts
```diff
@@ -1,10 +1,12 @@
-import { Fn, tryOnUnmounted } from '@vueuse/shared'
+import { tryOnUnmounted } from '@vueuse/shared'
 import { MaybeElementRef, unrefElement } from '../unrefElement'
 import { useEventListener } from '../useEventListener'
 import { ConfigurableWindow, defaultWindow } from '../_configurable'
 
-const events = ['mousedown', 'touchstart', 'pointerdown'] as const
-type EventType = WindowEventMap[(typeof events)[number]]
+export type AllowedEvents = Pick<WindowEventMap, 'mousedown' | 'mouseup' | 'touchstart' | 'touchend' | 'pointerdown' | 'pointerup'>
```
antfu, line 6:
```suggestion
export type OnClickOutsideAllowedEvents = Pick<WindowEventMap, 'mousedown' | 'mouseup' | 'touchstart' | 'touchend' | 'pointerdown' | 'pointerup'>
```

sibbng: I knew it 😂 What about `OnClickOutsideEvents`? We are using it a lot on code and this is little bit longer.

antfu: Sure.

sibbng (approved):

antfu (approved):

PR #510, opened 2021-05-13
fix: Change `onMediaError` to `onSourceError`

cawa-93:
@wheatjs 

https://github.com/vueuse/vueuse/blob/87a464dbfc2bd489eb412a022df4b4b425b0bd43/packages/core/useMediaControls/index.ts#L468

The listener of `error` events on the media element mostly does not make sense.
`Error` never will be triggered on `<video>` or `<audio>` with `<source>` tag inside. This will be triggered directly on the `<source>` tag whose resource could not be loaded. And this event will not pop up. Check console in [simple demo](https://codepen.io/cawa-93/pen/wvJMZdK?editors=1010).

### DIFF
--- a/packages/core/useMediaControls/index.ts
+++ b/packages/core/useMediaControls/index.ts
@@ -381,6 +381,15 @@ export function useMediaControls(target: MaybeRef<HTMLMediaElement | null | unde
     el.load()
   })
 
+  // Remove source error listeners
+  tryOnUnmounted(() => {
+    const el = unref(target)
+    if (!el)
+      return
+
+    el.querySelectorAll('source').forEach(e => e.removeEventListener('error', sourceErrorEvent.trigger))
+  })
+
   /**
    * Watch volume and change player volume when volume prop changes
    */
--- a/packages/core/useMediaControls/index.ts
+++ b/packages/core/useMediaControls/index.ts
@@ -236,7 +236,7 @@ export function useMediaControls(target: MaybeRef<HTMLMediaElement | null | unde
   const supportsPictureInPicture = document && 'pictureInPictureEnabled' in document
 
   // Events
-  const mediaErrorEvent = createEventHook<Event>()
+  const sourceErrorEvent = createEventHook<Event>()
 
   /**
    * Disables the specified track. If no track is specified then
@@ -369,6 +369,8 @@ export function useMediaControls(target: MaybeRef<HTMLMediaElement | null | unde
       source.setAttribute('src', src)
       source.setAttribute('type', type || '')
 
+      useEventListener(source, 'error', sourceErrorEvent.trigger)
+
       el.appendChild(source)
     })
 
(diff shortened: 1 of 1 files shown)


antfu:
This is a breaking change, better to add backward compatibility, and mark the previous one deprecated.

cawa-93:
If I'm not mistaken it hasn't been published yet. MediaErrors was added (https://github.com/vueuse/vueuse/commit/a6442832882ea7e032b0a390d85a5e5c5c763eb3) after 4.10 was released

antfu:
Oh yeah, good point, I was waiting for #507 :P

wheatjs:
Ah good catch, wasn't aware that it fired on the actual source. But maybe we should keep mediaError as well so we can also listen for errors on that?

wheatjs:
Well actually never mind that since you said an error will never trigger on the actual media tag itself

packages/core/useMediaControls/index.ts
```diff
@@ -369,6 +369,8 @@ export function useMediaControls(target: MaybeRef<HTMLMediaElement | null | unde
       source.setAttribute('src', src)
       source.setAttribute('type', type || '')
 
+      useEventListener(source, 'error', sourceErrorEvent.trigger)
```
wheatjs, line 372: I don't think this will be automatically cleaned up since it is nested inside of a watchEffect and effectScope isn't supported yet, so I think we will need to manually clean this up like we did starting on line 476. It might be cleaner if rename `listeners` to `disposables` and push each listener into the array of disposables instead of manually assigning them like I previously did.

cawa-93: @wheatjs Since the `sources` is never reactive, it doesn't seem to make sense to use `useEventListener`. Maybe should use the built-in `addEventListener`?

```js
source.addEventListener('error', sourceErrorEvent.trigger)
```

And if I'm not mistaken, the listener of events on the DOM node will be automatically deleted together with this DOM node.

https://github.com/vueuse/vueuse/blob/59a47a1e7f7929383cf5ba7fc76055caf36b356c/packages/core/useMediaControls/index.ts#L363

wheatjs: If they will be removed when the element is unmounted then `addEventListener` is fine

packages/core/useMediaControls/index.ts
```diff
@@ -360,7 +360,10 @@ export function useMediaControls(target: MaybeRef<HTMLMediaElement | null | unde
       sources = [src]
 
     // Clear the sources
-    el.querySelectorAll('source').forEach(e => e.remove())
+    el.querySelectorAll('source').forEach((e) => {
+      e.removeEventListener('error', sourceErrorEvent.trigger)
```
cawa-93, line 364: It looks like the event listener may not be deleted if a node references is stored somewhere. Therefore, I added this for prevention

wheatjs: This is only called when `options.src` is changed though, right? We also want these listeners to be cleaned up when the component is unmounted.

cawa-93: You are right. I forgot: 3

wheatjs (approved):
Looks good to me now, thanks!

PR #514, opened 2021-05-15
refactor(useMediaControls): Deprecate options that can simply be set as attributes

cawa-93:
If you change the volume or muted values by native controls (or in devtools) it does not effect reactive variables. [Here is a demonstration](https://sfc.vuejs.org/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCB7IHJlZiB9IGZyb20gJ3Z1ZSdcbmltcG9ydCB7IHVzZU1lZGlhQ29udHJvbHMgfSBmcm9tICdAdnVldXNlL2NvcmUnXG4gIFxuY29uc3Qgb3JpZ2luU291cmNlID0gJ2h0dHBzOi8vaW50ZXJhY3RpdmUtZXhhbXBsZXMubWRuLm1vemlsbGEubmV0L21lZGlhL2NjMC12aWRlb3MvZmxvd2VyLndlYm0nXG5jb25zdCBzcmMgPSByZWYob3JpZ2luU291cmNlKVxuXG5jb25zdCBtdXRlZCA9IHJlZihmYWxzZSlcbmNvbnN0IHZpZGVvRWxlbWVudCA9IHJlZigpXG5jb25zdCB7cGxheWluZywgdm9sdW1lfSA9IHVzZU1lZGlhQ29udHJvbHModmlkZW9FbGVtZW50LCB7XG4gIGF1dG9wbGF5OiBmYWxzZSwgLy8gPC0tXG4gIGNvbnRyb2xzOiB0cnVlLCBcbiAgcHJlbG9hZDogJ21ldGFkYXRhJyxcbiAgc3JjLFxuICBtdXRlZFxufSlcblxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPGRpdiBzdHlsZT1cImRpc3BsYXk6IGdyaWQ7IHBsYWNlLWNvbnRlbnQ6IGNlbnRlcjsgcGxhY2UtaXRlbXM6IGNlbnRlcjsgaGVpZ2h0OiAxMDAlOyBnYXA6IDFyZW07IHVzZXItc2VsZWN0OiBub25lO1wiPlxuICAgIDx2aWRlbyAgcmVmPVwidmlkZW9FbGVtZW50XCI+PC92aWRlbz5cbiAgICBcbiAgICB2b2x1bWU6IHt7dm9sdW1lfX1cbiAgICBtdXRlZDoge3ttdXRlZH19XG4gIDwvZGl2PlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcblx0XHRcIkB2dWV1c2UvY29yZVwiOiBcImh0dHBzOi8vdW5wa2cuY29tL0B2dWV1c2UvY29yZS9kaXN0L2luZGV4LmVzbS5qc1wiLFxuICAgIFwiQHZ1ZXVzZS9zaGFyZWRcIjogXCJodHRwczovL3VucGtnLmNvbS9AdnVldXNlL3NoYXJlZC9kaXN0L2luZGV4LmVzbS5qc1wiLFxuICAgIFwidnVlLWRlbWlcIjogXCJodHRwczovL3VucGtnLmNvbS92dWUtZGVtaS9saWIvaW5kZXguZXNtLmpzXCJcbiAgfVxufSJ9).

### DIFF
--- a/packages/core/useMediaControls/demo.vue
+++ b/packages/core/useMediaControls/demo.vue
@@ -10,9 +10,9 @@ import { useMediaControls } from '.'
 
 const video = ref<HTMLVideoElement>()
 const loop = ref(false)
+const poster = 'https://bitmovin.com/wp-content/uploads/2016/06/sintel-poster.jpg'
 
 const controls = useMediaControls(video, {
-  loop,
   src: {
     src: 'https://upload.wikimedia.org/wikipedia/commons/transcoded/f/f1/Sintel_movie_4K.webm/Sintel_movie_4K.webm.1080p.vp9.webm',
     type: 'video/webm',
@@ -32,7 +32,6 @@ const controls = useMediaControls(video, {
       srcLang: 'fr',
     },
   ],
-  poster: 'https://bitmovin.com/wp-content/uploads/2016/06/sintel-poster.jpg',
 })
 
 const {
@@ -66,7 +65,14 @@ const formatDuration = (seconds: number) => new Date(1000 * seconds).toISOString
     @keydown.left="currentTime -= 10"
   >
     <div class="relative bg-black">
-      <video ref="video" crossorigin="anonymous" class="w-full block" @click="playing = !playing" />
+      <video
+        ref="video"
+        crossorigin="anonymous"
+        class="w-full block"
+        :poster="poster"
+        :loop="loop"
+        @click="playing = !playing"
+      />
       <div v-if="waiting" class="absolute inset-0 grid place-items-center pointer-events-none bg-black bg-opacity-20">
         <Spinner />
       </div>
--- a/packages/core/useMediaControls/demo.vue
+++ b/packages/core/useMediaControls/demo.vue
@@ -12,7 +12,6 @@ const video = ref<HTMLVideoElement>()
 const loop = ref(false)
 
 const controls = useMediaControls(video, {
-  controls: true,
   loop,
   src: {
     src: 'https://upload.wikimedia.org/wikipedia/commons/transcoded/f/f1/Sintel_movie_4K.webm/Sintel_movie_4K.webm.1080p.vp9.webm',
--- a/packages/core/useMediaControls/demo.vue
+++ b/packages/core/useMediaControls/demo.vue
@@ -9,11 +9,10 @@ import { stringify } from '@vueuse/docs-utils'
 import { useMediaControls } from '.'
 
 const video = ref<HTMLVideoElement>()
-const muted = ref(false)
 const loop = ref(false)
 
 const controls = useMediaControls(video, {
-  muted,
+  controls: true,
   loop,
   src: {
     src: 'https://upload.wikimedia.org/wikipedia/commons/transcoded/f/f1/Sintel_movie_4K.webm/Sintel_movie_4K.webm.1080p.vp9.webm',
@@ -46,6 +45,7 @@ const {
   waiting,
   selectedTrack,
   volume,
+  muted,
   isPictureInPicture,
   supportsPictureInPicture,
   togglePictureInPicture,
--- a/packages/core/useMediaControls/index.md
+++ b/packages/core/useMediaControls/index.md
@@ -15,10 +15,8 @@ import { ref } from 'vue'
 import { useMediaControls } from '@vueuse/core'
 
 const video = ref()
-const loop = ref(false)
 const { playing, currentTime, duration } = useMediaControls(video, { 
   src: 'video.mp4',
-  loop,
 })
 </script>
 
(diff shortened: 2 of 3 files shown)


wheatjs (dismissed):

wheatjs:
Good catch, I actually noticed these issues recently but hadn't had the chance to go back and fix them. Thanks for picking up my slack on these 😄

wheatjs:
Maybe consider adding a check if `el` exists to get rid of the error

cawa-93:
I hastened with it. How could we update the `mute` that comes from `options`? My first think was:
```js
    if (options.muted !== undefined && isRef(options.muted))
      options.muted.value = (unref(target))!.muted
```

But TS prompts me that `options.muted.value` is readonly. 
```ts
UseMediaControlsOptions.muted: Ref<boolean> | ComputedRef<boolean>
```

wheatjs:
Hmm, I'll take a look tomorrow if you don't figure it out before then.

cawa-93:
What if return new `muted` value? In the same way as all other parameters return.

```js
{
  muted // <-- Reactive value
} = useMediaControls(video, {
  muted: MaybeRef<boolean>, // <-- Initial value
}

**Comment by wheatjs** [comment]
Hmm, maybe we should just remove the volume and muted objects from the initial parameters all together? Maybe things like `muted` and `volume` should just be returned from the object like you suggested. I feel like having the initial parameters and also returned might be over complicating it. I think what would be best is things that have a 2 way data flow, like volume, `currentTime`, and `muted` are returned from the function while things that have a one-way data flow like `poster` should be treated as configuration options.

**Comment by cawa-93** [comment]
Sounds good.

How about to deprecate one-way data that can be set by `<video>` attibutes and does't have any "reactive logic": `loop`, `controls`, `preload`, `autoplay`,  `poster`, `playsInline` and `autoPictureInPicture`.

**Don't**
```js
const poster = ref('[link to poster image]')
const {} = useMediaControls(target, {
  poster,
})
```

**Do**
```vue
<template>
  <video :poster="poster">
</template>

<script setup>
const poster = ref('[link to poster image]')

const {} = useMediaControls(target, {})
</script>
```
What do you thinks?

cawa-93:
@wheatjs Please review how I create deprecation message. If all goes well, I'll do the same for the rest of the options I talked about earlier.

packages/core/useMediaControls/demo.vue
```diff
@@ -9,11 +9,10 @@ import { stringify } from '@vueuse/docs-utils'
 import { useMediaControls } from '.'
 
 const video = ref<HTMLVideoElement>()
-const muted = ref(false)
 const loop = ref(false)
 
 const controls = useMediaControls(video, {
-  muted,
+  controls: true,
```
cawa-93, line 15: Ooops... This should not be here 😅

wheatjs:
> @wheatjs Please review how I create deprecation message. If all goes well, I'll do the same for the rest of the options I talked about earlier.

Sure thing, sorry for the delay, I'll review this sometime today 😄

antfu:
> How about to deprecate one-way data that can be set by <video> attibutes and does't have any "reactive logic": loop, controls, preload, autoplay, poster, playsInline and autoPictureInPicture.

Yeah, that sounds better to me to be as well. User could also use `v-bind` when they want to provide them altogether as an object.

wheatjs:
> @wheatjs Please review how I create deprecation message. If all goes well, I'll do the same for the rest of the options I talked about earlier.

The deprecation messages look good to me. Once you add them for the rest of the functions we can do another review and make sure everything works like expected

cawa-93:
I don't changed demo in last commit for checking backward compatibility

wheatjs (approved):
Nice job with the console warns as well, it looks good too me

antfu:
I was trying to avoid console logs in VueUse whenever possible. Since we don't have different builds for dev and prod, using consoles would also ships extra bits and hard-coded strings to production.

I would suggest to mark the API depredated on type level but keep the behavior the same, and then we can remove the code in major versions later

wheatjs:
> Nice job with the console warns as well, it looks good too me

Sorry I forgot about us trying to avoid console logs 😅

wheatjs:
This is only slightly relevant to this PR, but it seems like my links for the subtitles expired for the demo so if you want you can add them locally maybe? Otherwise we can do it in another PR

Here are the gists for the subtitles:

English https://gist.github.com/wheatjs/a85a65a82d87d7c098e1a0972ef1f726
French https://gist.github.com/wheatjs/38f32925d20c683bf77ba33ff737891b

PR #515, opened 2021-05-15
fix(toMatch): UntilBaseInstance#toMatch typedef

### LINKED ISSUE #482: `UntilBaseInstance#toMatch` assumes that `v` is always of the input type, but it's `unref`ed
I have following statement where `portals` is of type `ComputedRef<Portal[]>`

```ts
await until(portals).toMatch((v) => v.length > 0);
//                                   ~~~~~~~ expect to use v.value.length
```

`UntilBaseInstance#toMatch` now assumes that its `v` argument is also of type `ComputedRef<Portal[]>` but it's `Portal[]`

The current workaround is to use `unref`

```diff
- await until(portals).toMatch((v) => v.length > 0);
+ await until(portals).toMatch((v) => unref(v).length > 0);
```

The solution could be using `MaybeRef` or something similar 🤔 

---

- [x] I'm willing to provide a PR for that  
      This would be my first contribution and really helpful to jump into the code and find out how it's working 🙂 
      _I need to find a time-slot in the next few days, that's why I'm creating an issue instead of directly opening a PR to fix it 😄_

Shinigami92:
closes #482

### DIFF
--- a/packages/shared/until/index.md
+++ b/packages/shared/until/index.md
@@ -77,7 +77,7 @@ await until(ref).not.toBeTruthy()
 ```typescript
 export interface UntilToMatchOptions {
   /**
-   * Milseconds timeout for promise to resolve/reject if the when condition does not meet.
+   * Milliseconds timeout for promise to resolve/reject if the when condition does not meet.
    * 0 for never timed out
    *
    * @default 0
@@ -104,7 +104,7 @@ export interface UntilToMatchOptions {
 }
 export interface UntilBaseInstance<T> {
   toMatch(
-    condition: (v: T) => boolean,
+    condition: (v: UnwrapRef<T>) => boolean,
     options?: UntilToMatchOptions
   ): Promise<void>
   changed(options?: UntilToMatchOptions): Promise<void>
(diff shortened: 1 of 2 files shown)


Shinigami92:
@antfu @wheatjs I assume that the value for the condition is always unwrapped, but I'm not totally sure right now :thinking:  
I'm also not sure if it should be a `ShallowUnwrapRef` or not

antfu:
It's better to let the function do the unwrapping, so the T will be the plain value's type, see
https://github.com/vuejs/vue-next/blob/870f2a7ba35245fd8c008d2ff666ea130a7e4704/packages/runtime-core/src/apiWatch.ts#L107-L108

Shinigami92:
> It's better to let the function do the unwrapping, so the T will be the plain value's type, see
> [vuejs/vue-next@`870f2a7`/packages/runtime-core/src/apiWatch.ts#L107-L108](https://github.com/vuejs/vue-next/blob/870f2a7ba35245fd8c008d2ff666ea130a7e4704/packages/runtime-core/src/apiWatch.ts#L107-L108)

I'm not sure what you try to say :thinking: 
I currently didn't changed any runtime, but only a compile-time typedef bug

antfu:
See the update ;)

PR #517, opened 2021-05-16
feat(useVibrate): new function

### LINKED ISSUE #516: useVibrate
Reactive hook to provide physical feedback with device vibration hardware using the Vibration API.

bharathvaj-ganesan:
Closes #516


crutch12:
Don't you know: does it API supports gamepads? :)

bharathvaj-ganesan:
> Don't you know: does it API support gamepads? :)

Pardon me. I don't get this question.

packages/core/useVibrate/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+```js
+import { useVibrate } from '@vueuse/core'
+
+const vibrating = ref(false)
+
+useVibrate(vibrating)
+
+// vibrating.value = true => vibration will start
+// vibrating.value = false => vibration will end
```
cawa-93, line 19:
```suggestion
const startVibration = () => vibrating.value = true
const stopVibration = () => vibrating.value = false
const toggleVibration = () => vibrating.value = !vibrating.value
```

packages/core/useVibrate/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+const vibrating = ref(false)
+
+const { pause, resume } = useVibrate(vibrating, {
+  pattern: [300, 100, 200, 100, 1000, 300],
+  loop: true,
+})
+
+// resume() => vibration will start
+// pause() => vibration will end
```
cawa-93, line 36: For me is not clear what differentness between using `vibrating.value = true/false` and using `pause/resume` controls.

cawa-93 (commented):
You may also be interested in reading the [discussion](https://github.com/vueuse/vueuse/discussions/518#discussion-3370085)

packages/core/useVibrate/index.md
wheatjs: This is probably something we should discuss in #518 as well so we can make recommendations on when to use play/pause vs controlling a value like `vibrating`

antfu:
Close in favor of #1082

PR #519, opened 2021-05-18
fix(onClickOutside): duplicate code

JserWang:
`tryOnUnmounted(stop)` has been called inside `useEventListener` 
https://github.com/vueuse/vueuse/blob/6423bf88c3ec49bc6b6fe66412d44349872be5b8/packages/core/useEventListener/index.ts#L117-L124

### DIFF
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -1,4 +1,3 @@
-import { tryOnUnmounted } from '@vueuse/shared'
 import { MaybeElementRef, unrefElement } from '../unrefElement'
 import { useEventListener } from '../useEventListener'
 import { ConfigurableWindow, defaultWindow } from '../_configurable'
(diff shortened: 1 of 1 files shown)


sibbng (approved):

packages/core/onClickOutside/index.ts
```diff
@@ -39,7 +38,5 @@ export function onClickOutside<E extends keyof OnClickOutsideEvents = 'pointerdo
 
   const stop = useEventListener(window, event, listener, { passive: true })
```
sibbng, line 39:
```suggestion
  return useEventListener(window, event, listener, { passive: true })
```

packages/core/onClickOutside/index.ts
```diff
@@ -39,7 +38,5 @@ export function onClickOutside<E extends keyof OnClickOutsideEvents = 'pointerdo
...
 
-  tryOnUnmounted(stop)
-
   return stop
```
sibbng, line 41:
```suggestion
```

sibbng (changes requested):

sibbng (approved):

PR #520, opened 2021-05-18
refactor(useSwipe): switch to PointerEvents

### LINKED ISSUE #511: useSwipe: make it work also with mouse
Since `useSwipe` only works with `touch` events, there is no way to make it work using the mouse.

This issue can be solved using [this](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events#event_properties) as pointed on https://github.com/vueuse/vueuse/issues/451#issuecomment-819164542.

lstoeferle:
Hi there,

it was not that easy as I thought but I managed to switch `useSwipe` to [PointerEvents](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events). Unfortunately there are some breaking changes... and I needed to add a css workaround via [touch-action](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action) to keep the `Touch` functionality (else the events are canceled by the browser which wants to keep scrolling enabled). This means we could drop the `passive` option, since it's no longer possible to have it passive.

Closes #511

### DIFF
--- a/indexes.json
+++ b/indexes.json
@@ -772,7 +772,7 @@
       "package": "core",
       "docs": "https://vueuse.org/core/useSwipe/",
       "category": "Sensors",
-      "description": "reactive swipe detection based on [`TouchEvents`](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent)"
+      "description": "reactive swipe detection based on [`PointerEvents`](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events)"
     },
     {
       "name": "useTimeAgo",
--- a/packages/core/useSwipe/demo.vue
+++ b/packages/core/useSwipe/demo.vue
@@ -17,7 +17,7 @@ const reset = () => {
 const { direction, isSwiping, lengthX, lengthY } = useSwipe(
   target, {
     passive: false,
-    onSwipe(e: TouchEvent) {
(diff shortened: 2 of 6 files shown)


antfu:
I am not sure if it's worth to have this trade-off (and the style side-effect could be dangerous, at least we should document it). For myself, I actually relaying on the behavior of `useSwipe` only works for touch. So, I am thinking, maybe it's better to rename the previous one to `useTouchSwipe` and introduce the breaking change to `useSwipe` in v5.0. WDYT?

cawa-93:
@antfu Maybe it will be better if just keep `useSwipe` as is and add new `usePointerSwipe` function. We could add deprecation warning ti  `useSwipe`  and propose use `usePointerSwipe`. So we could release  `usePointerSwipe` now. It will not be a breaking change. And users will be able to gradually move on to use  `usePointerSwipe`  and to report any problems with which they are faced. This, as I'm see, will be a smoother and easy migration.

lstoeferle:
I'm fine with keeping `useSwipe` like it is. If you agree, I will create a new function `usePointerSwipe` like @cawa-93 suggested.

antfu:
Sure, good to me.

lstoeferle:
I will close this MR, since the refactoring should be part of a new function.

PR #531, opened 2021-05-24
feat(createEventHook): added interface

### LINKED ISSUE #529: request to add createEventHook result type
It would be nice if we have this return function createEventHook typed.

````
export type EventHook<T> = {
  on: (
    fn: (param: T) => void
  ) => {
    off: () => void
  }
  off: (fn: (param: T) => void) => void
  trigger: (param: T) => void
}
````
So i can type things like
````
selectEventHook: Ref<EventHook<Something>>
````

wheatjs:
Closes #529


wheatjs:
Added an `EventHook` interface that includes the on, off and trigger properties. I also included individual types for `EventHookOn`, `EventHookOff`, and `EventHookTrigger` as it may be useful to use those individually like we do in `useFetch`

PR #535, opened 2021-05-25
docs(guidlines): added guidelines

wheatjs:
Here is an initial draft for the guidelines. Let me know if you guys have any suggestions or corrections.

@antfu @cawa-93

### DIFF
--- a/packages/guidelines.md
+++ b/packages/guidelines.md
@@ -1,0 +1,105 @@
+---
+sidebar: auto
+---
+
+# Guidelines
+
+Writing good composables can be difficult, so we've created a small set of guidelines you can follow to improve your composables!
+
+## General
+- Avoid using console logs in VueUse
+- Use configurableWindow (etc.) when using global variables like window to be flexible when working with multi-windows, testing mocks, and SSR.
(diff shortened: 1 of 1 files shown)


cawa-93:
1. I would like see some examples (code or links to code) in general section, Because some points may not be clear. What I mean:
    > Use configurableWindow (etc.) when using global variables like window
  
    It's not clear what is `configurableWindow`. Link to some function with usage will be very helpful. 
    
2. How about add a bit points about "Where from import any functions"? What I mean: Today I just experimented. I Wrote `compute` and my IDE proposed me auto import from `vue`, `@vue/reactive-api` and `vue-demi`. And it happens often. And every time I worry about doing everything right. I would be calmer if the rules set what and where I can import.

cawa-93:
And, just interesting
@antfu 

> Use 'ref' over 'reactive' whenever possible

Why? 😅

antfu:
> Use 'ref' over 'reactive' whenever possible

https://www.youtube.com/watch?v=IMJjP6edHd0

wheatjs:
@cawa-93 good suggestions, I'll add some examples tonight

antfu:
Let's have it first, we could update them later.

PR #544, opened 2021-05-29
feat(typedef): add return typedefs (#543)

### LINKED ISSUE #543: Add return types for all composables
I personally make use of heavy TypeScript by using the eslint rule [typedef](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/typedef.md)

So we could add a return type and export these explicitly for every function.
We also don't need to type them safly but just create somewhat aliases.

We can do that like so

```ts
export function useMouse(options: MouseOptions = {}) { /*...*/ }

export type UseMouse = ReturnType<typeof useMouse>
```

Are you open for such an addition?

---

Currently I'm forced to define these myself: https://github.com/Shinigami92/vizscript/blob/c3d26fd3f921787dab0bf25cb9d024d170824225/src/views/Viz.vue#L55-L56

Shinigami92:
closes #543

### DIFF
--- a/packages/core/createGlobalState/index.md
+++ b/packages/core/createGlobalState/index.md
@@ -43,6 +43,9 @@ export default defineComponent({
 export declare function createGlobalState<T extends object>(
   stateFactory: () => T
 ): () => T
+export declare type CreateGlobalStateReturn = ReturnType<
+  typeof createGlobalState
+>
 ```
 
 ## Source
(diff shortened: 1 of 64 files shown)


Shinigami92:
@wheatjs For now I just looked into the `core`

I orientated myself at some already build in types that are mostly ending with `UseXyzReturn`

antfu:
Nice, thanks!

Issue #501, opened 2021-05-12
`useWebSocket` add `immediate` option


Shinigami92:
Add an ~~`autoConnect`~~ `immediate` option to `useWebSocket`s options that is default to `true`

_I'm willing to add this feature_

antfu:
Great, let's see the PR. I'd prefer to name it `immedidate` to align with other functions. You can make the default to `false` for now. And then we can change it to `true` in v5.0.

PR #545, opened 2021-05-29
refactor(useWebSocket)!: change immediate default for 5.0.0

Shinigami92:
https://github.com/vueuse/vueuse/issues/501#issuecomment-839726620

### DIFF
--- a/packages/core/useWebSocket/index.ts
+++ b/packages/core/useWebSocket/index.ts
@@ -59,7 +59,7 @@ export interface WebSocketOptions {
   /**
    * Automatically open a connection
    *
-   * @default false
+   * @default true
    */
   immediate?: boolean
 }
@@ -223,7 +223,7 @@ export function useWebSocket<Data = any>(
     heartbeatResume = resume
   }
 
-  if (options.immediate) _init()
+  if (options.immediate === false) _init()
 
   const open = () => {
     close()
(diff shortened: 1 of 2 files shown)


packages/core/useWebSocket/index.ts
```diff
@@ -223,7 +223,7 @@ export function useWebSocket<Data = any>(
     heartbeatResume = resume
   }
 
-  if (options.immediate) _init()
+  if (options.immediate === false) _init()
```
antfu, line 226: emm?

Shinigami92: Oops, seems it was very late yesterday 🤣
Will change it when I'm out of bed

packages/core/useWebSocket/index.ts
```diff
@@ -223,7 +223,7 @@ export function useWebSocket<Data = any>(
     heartbeatResume = resume
   }
 
-  if (options.immediate) _init()
+  if (options.immediate !== false) _init()
```
wheatjs, line 226: could just be `if (options.immediate) _init()` right? 😅

Shinigami92: No, cause the new default is that if it's `undefined` than it should also count as `true`
That's the whole point of this PR 😛

wheatjs: My bad 😅

antfu: Better to assign default value in L121 instead.

PR #549, opened 2021-05-30
refactor(useFetch): change internal logic

### LINKED ISSUE #526: Can't acess to useFetch response's body on error ?
On a HTTP 400 response sent by the server (missing required fields in a form), useFetch only pass the statusText to the Error object. [source code](https://github.com/vueuse/vueuse/blob/068144d86e3fabf933edb35f1e5c5ee1c3ce2adb/packages/core/useFetch/index.ts#L327)

I'd like to access to the json body response where I can access to the details of the error (statusText is irrelevant in my case). Response's body is already consumed (Stream locked) and `data` object has not been populated since the error is thrown before it gets the chance.

Is this some kind of non-fetch-standard issue and should I use useAxios instead ?

### LINKED ISSUE #539: useFetch and formData doesn't work
There are only one way I made send the FormData correctly

```
const formData = new FormData()
formData.append('id', id)
formData.append('file', blob)

const { data } = useFetch(path, { method: 'PUT', body: formData })
```

I do not understand what the [formData()](https://github.com/vueuse/vueuse/blob/5e15d7a9622d7caa516bbff8ad7c066f157a617b/packages/core/useFetch/index.ts#L76) method was created for. The _multipart/form-data_ never added but I suppose it [should](https://github.com/vueuse/vueuse/blob/5e15d7a9622d7caa516bbff8ad7c066f157a617b/packages/core/useFetch/index.ts#L292)

I've try without success
```
useFetch(url).formData().put(formData)
useFetch(url).put(formData).formData()
useFetch(url).put(formData, 'formData').formData()
```

Am I not understanding something or am I doing something wrong?

### LINKED ISSUE #542: useFetch: multipart form data header must be removed instead added
`useFecth` line 292 should be:
```typescript
if (config.payloadType === 'formData') {
  headers.remove('Content-Type')
} else {
  headers['Content-Type'] = 'text/plain'
}
```
The browser `fetch` will do the same check and will break the execution.

We need to check if the formData has any blob/file, in that case we just remove the header.

I think in that case it would be necessary to force the user to include the header, and so `useFetch` usage will be concise, but it is a `breaking change`, but if ppl do not create an  issue no one is using it (yet ;) ).

I suggest to force `multipart` usage to be explicity in the request (header) and so we can just remove the header once detected.

Here is a snippet from my own code building the `fetch` body:

```javascript
const buildBody = (body, headers) => {
  if (!body) {
    return body;
  }
  if (body instanceof FormData) {
    return body;
  }
  if (body instanceof Object) {
    const contentType = headers.get(CONTENT_TYPE);
    if (contentType === FORM_URL_ENCODED) {
      return encodeBodyObject(body, contentType);
    } else if (contentType === MULTIPART_FORM_DATA) {
      headers.delete(CONTENT_TYPE);
      const formData = new FormData();
      const jsonBody = {};
      Object.keys(body).forEach(key => {
        const value = body[key];
        if (value instanceof File) {
          formData.append(key, value, value.name);
        } else {
          jsonBody[key] = value;
        }
      });
      formData.append(
        "JSON_BODY",
        new Blob([JSON.stringify(jsonBody)], {
          type: APPLICATION_JSON
        })
      );
      return formData;
    } else {
      headers.set(CONTENT_TYPE, APPLICATION_JSON);
      return JSON.stringify(body);
    }
  }
  return body;
};

### LINKED ISSUE #553: `useFetch` Have access to the response body when the status is not ok
At the moment when the status is not `ok` an error is thrown while fetching. In this case `data` will be null and the error will be populated. It will be nice to have the body of the response set to `data`. Even if the status code is not in `200 ...(ok)` the response still has a body.

I think just throwing after setting the `data` will be ok

```js
// see: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/
if (!fetchResponse.ok)

if (options.afterFetch)      
    ({ data: responseData } = await options.afterFetch({ data: responseData, response: fetchResponse }))

data.value = responseData as any
```

userquin:
closes #542
closes #539
closes #553
closes #526

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -439,7 +439,7 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
 
   function setMethod(method: string) {
     return (payload?: unknown, payloadType?: PayloadType) => {
-      if (!initialized || (payloadType && config.payloadType === payloadType)) {
+      if (!initialized || !config.payloadType || (config.payloadType === payloadType)) {
         config.method = method
         config.payload = payload
         if (payloadType)
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -206,21 +206,19 @@ function toFormData(body: any): FormData {
       if (value instanceof File) {
         formData.append(key, value, value.name)
       }
+      else if (value instanceof Blob || typeof value === 'string') {
+        formData.append(key, value)
+      }
+      else if (value) {
+        formData.append(
+          key,
+          new Blob([JSON.stringify(value)], {
+            type: 'application/json',
+          }),
+        )
+      }
       else {
-        if (value instanceof Blob || typeof value === 'string') {
-          formData.append(key, value)
-        }
-        else if (value) {
-          formData.append(
-            key,
-            new Blob([JSON.stringify(value)], {
-              type: 'application/json',
-            }),
-          )
-        }
-        else {
-          formData.append(key, '')
-        }
+        formData.append(key, '')
       }
     })
   }
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -389,7 +389,7 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
       if (!initialized) {
         config.method = method
         config.payload = payload
-        config.payloadType = payloadType || typeof payload === 'string' ? 'text' : 'json'
+        config.payloadType = payloadType || (typeof payload === 'string' ? 'text' : 'json')
         return shell as any
       }
       return undefined
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -310,10 +310,9 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
       return Promise.resolve()
 
     return new Promise((resolve) => {
-      let useHeaders: Headers
+      const useHeaders = new Headers(defaultFetchOptions.headers)
       const contextHeaders = context.options?.headers ? new Headers(context.options.headers) : null
       if (contextHeaders) {
-        useHeaders = new Headers(defaultFetchOptions.headers)
         // just override existing or add context ones
         contextHeaders.forEach((hv, h) => {
           useHeaders.set(h, hv)
@@ -322,18 +321,17 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
         if (config.payloadType === 'formData')
           useHeaders.delete('Content-Type')
       }
-      else {
-        useHeaders = new Headers(defaultFetchOptions.headers)
-      }
+      // defaultFetchOptions headers was used in the above logic
       const { method, body } = defaultFetchOptions
-      const { headers, ...rest } = context?.options || {}
+      // headers was merged on above logic, and so will be ignored
+      const { headers, ...contextOptions } = context?.options || {}
       fetch(
         context.url,
         {
           method,
           body,
           headers: useHeaders,
-          ...rest,
+          ...contextOptions,
         },
       )
         .then(async(fetchResponse) => {
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -152,7 +152,7 @@ export interface UseFetchOptions {
 
   /**
    * Will run immediately after the fetch request is fetched.
-   * If provided, then it must extract the response and deal with the errors.
+   * If provided, then it must extract the payload from the response and deal with the errors.
    * @param response The response from `fetch`.
    */
   responseHandler?: (response: Response) => Promise<CustomResponseContext>
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -438,10 +438,9 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
   }
 
   function setMethod(method: string) {
-    if (!initialized)
-      config.method = method
-
     return (payload?: unknown, payloadType?: PayloadType) => {
+      if (!initialized)
+        config.method = method
       config.payload = payload
       if (payloadType)
         config.payloadType = payloadType
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -310,15 +310,26 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
       return Promise.resolve()
 
     return new Promise((resolve) => {
+      let headers: Headers
+      const contextHeaders = context.options?.headers ? new Headers(context.options.headers) : null
+      if (contextHeaders) {
+        headers = new Headers({
+          ...defaultFetchOptions.headers,
+          ...contextHeaders,
+        })
+        // since we are using Headers object, it will handle case insensiive
+        if (config.payloadType === 'formData')
+          headers.delete('Content-Type')
+      }
+      else {
+        headers = new Headers(defaultFetchOptions.headers)
+      }
       fetch(
         context.url,
         {
           ...defaultFetchOptions,
           ...context.options,
-          headers: {
-            ...defaultFetchOptions.headers,
-            ...context.options?.headers,
-          },
+          headers,
         },
       )
         .then(async(fetchResponse) => {
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -181,6 +181,20 @@ async function toURLSearchParams(body: any): Promise<URLSearchParams> {
   return new URLSearchParams()
 }
 
+function toFormData(body: any): FormData {
+  const formData = new FormData()
+  if (body) {
+    Object.keys(body).filter(key => Object.prototype.hasOwnProperty.call(body, key)).forEach((key) => {
+      const value = body[key]
+      if (value instanceof File)
+        formData.append(key, value, value.name)
+      else
+        formData.append(key, value)
+    })
+  }
+  return formData
+}
+
 export function createFetch(config: CreateFetchOptions = {}) {
   let options = config.options || {}
   let fetchOptions = config.fetchOptions || {}
@@ -315,6 +329,9 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
       }
       else {
         if (config.payloadType === 'formData' || payload instanceof FormData) {
+          // build the form data for the user
+          if (!(payload instanceof FormData))
+            payload = toFormData(payload)
           // we just add it here to remove later once merged with context.options.headers
           headers['Content-Type'] = 'multipart/form-data'
         }
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -109,6 +109,16 @@ export interface AfterFetchContext<T = any> {
 
 }
 
+export interface CustomResponseContext<T = any> {
+
+  data: T | null
+
+  error: boolean
+
+  errorMessage?: string | undefined
+
+}
+
 export interface UseFetchOptions {
   /**
    * Fetch function
@@ -139,6 +149,13 @@ export interface UseFetchOptions {
    * Runs after any 2xx response
    */
   afterFetch?: (ctx: AfterFetchContext) => Promise<Partial<AfterFetchContext>> | Partial<AfterFetchContext>
+
+  /**
+   * Will run immediately after the fetch request is fetched.
+   * If provided, then it must extract the response and deal with the errors.
+   * @param response The response from `fetch`.
+   */
+  responseHandler?: (response: Response) => Promise<CustomResponseContext>
 }
 
 export interface CreateFetchOptions {
@@ -378,16 +395,26 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
           response.value = fetchResponse
           statusCode.value = fetchResponse.status
 
-          let responseData = await fetchResponse[config.type]()
+          if (options.responseHandler) {
+            const {
+              data: useData,
+              error = false,
+              errorMessage = undefined,
+            } = await options.responseHandler(fetchResponse)
+            data.value = useData
+            if (error)
+              throw new Error(errorMessage)
+          }
+          else {
+            let responseData = await fetchResponse[config.type]()
+            if (!fetchResponse.ok)
+              throw new Error(fetchResponse.statusText)
 
-          // see: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/
-          if (!fetchResponse.ok)
-            throw new Error(fetchResponse.statusText)
+            if (options.afterFetch)
+              ({ data: responseData } = await options.afterFetch({ data: responseData, response: fetchResponse }))
 
-          if (options.afterFetch)
-            ({ data: responseData } = await options.afterFetch({ data: responseData, response: fetchResponse }))
-
-          data.value = responseData as any
+            data.value = responseData as any
+          }
           responseEvent.trigger(fetchResponse)
           resolve(fetchResponse)
         })
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -308,7 +308,9 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
     if (payload) {
       const headers = defaultFetchOptions.headers as Record<string, string>
       if (config.payloadType === 'json') {
-        payload = JSON.stringify(payload)
+        // avoiding double encoding
+        if (typeof payload !== 'string')
+          payload = JSON.stringify(payload)
         headers['Content-Type'] = 'application/json'
       }
       else {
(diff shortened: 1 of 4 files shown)


userquin:
Dont merge yet, I Will change this PR to correct content-type and we need also handle form data in the assignment if the payload type is not provided.

userquin:
the headers should be handled before the `fetch` call since the user can provide them from the context call.

userquin:
I need to do one more thing, encode the data when using `formEncoded`, so don't merge yet, I'm working on it rigth now.

userquin:
I need to add some tests for `application/x-www-form-urlencoded` and `multipart/form-data` payload types.

userquin:
@wheatjs it is confusing the logic inside `execute` method: since we can override the `body` from `context` options, and before calling `fetch` we are using `context.options` after `defaultOptions` when merging to create the `RequestInit` object, the logic handling the payload and the headers can be moved inside the `Promise` and we can also use `context.options.body` if provided instead the `config` one.

I think we can change the behavior and so we can use the `context.options` to reuse the execute method. I will wait your feedback before changing anything.

userquin:
@wheatjs I will change it so that you can compare with the original, looking at changes you will see what I mean.

userquin:
@wheatjs I cannot figure out how to include jest tests for `multipart/form-data` and for `application/x-www-form-urlencoded`, since we are mocking the request.

userquin:
@wheatjs the only way I have found is to include it on the demo (I will also add `application/x-www-form-urlencoded` in the page):

imagen

once executed (can be executed multiple time, with blob content changing between request)

imagen

userquin:
Also `application/x-www-form-urlencoded` included:

imagen

once executed (can be executed multiple times, with form content changing between request)

imagen

userquin:
@wheat

With this PR `useFetch` will work as expected, I'll try to fix the change of the content type once execute is called first time.

I don't know why don't allow to change the config object once initialized, I think it can be modified just not allowing to change it while fetch is in progress.

Why is `setMethod` returning `undefined` if initialized is `true`? Is this intentional or something that was included time ago and now it is not necessary?

wheatjs:
> Why is `setMethod` returning `undefined` if initialized is `true`? Is this intentional or something that was included time ago and now it is not necessary?

This is on purpose as we don't want to allow users to change the method after they've already set it. I think this is probably okay to merge though @antfu. There are a few other changes that need to be made to `useFetch` before v5.0 official release so I'd like   to get this finished so I can work on the others without to many merge conflicts. 

Also I'm not entirely sure if the demo is necessary, we could probably just get by with some tests

packages/core/useFetch/index.ts
```diff
@@ -386,13 +439,22 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
 
   function setMethod(method: string) {
     return (payload?: unknown, payloadType?: PayloadType) => {
-      if (!initialized) {
+      if (!initialized)
```
isgj, line 442: I don't know if this PR will be merged to master (if this has no breaking changes) or to `next` branch.
I prefer the check there https://github.com/vueuse/vueuse/blob/338b4a678f071a25eb76d77345e651e7ff5c4b1f/packages/core/useFetch/index.ts#L396. 

From my use case I have a component that first get some data, and if the user makes changes it `put` to the same url (simple REST), Using the `next` branch is really convenient creating only one `useFetch`
```js
const {isFetching, data, put, error} = useFetch('/resource').json().get()

// elsewhere
put({/*updated data*/}).execute()
```
the same `isFetching` or `error` are used.

Here if more `put`, `post`, `delete` are called it will use the first method registered and it will be confusing.

userquin: We are waiting for @antfu review, I think changing method will not be a problem (if well documented and including a breaking change like throwing an error while fetching...), your use case sounds reasonable, and including another `useFetch` is a pain, just see demo page in this PR. 

I'm awaiting for Jacob to see if I include here to provide the payload on error. I can confirm that it is necessary, since the user can use an error with payload: for example, I send a 413 when the size exceeds the server max size, and in general I always send a JSON payload from the server.

I suggest this changes (maybe we can be more concise, suach as including `responseError?: boolean | undefined`):

imagen

imagen

imagen

userquin: In fact I think there is a problem to directly consuming the response body: for example, downloading a file can result in 403 or 401 or 404, that is, unauthorized , permission denied or not found, but there is no blob to download.

The user must have the control via the callback, and its responsability will be to extract the response body after check if there is an error.

So a second change will be to delegate the responsability to the user when the `afterFetch` is provided.

userquin: Maybe this will be necessary to be moved to a dicussion.

userquin: One last thing, if the server is down or the client is offline, the response will be null or undefined so we need handle this case, maybe using `useOffline` before executing the request...

The browser will throw a generic error, I handle the situation using literals:
`failed to fetch` and `networkerror when attempting to fetch resource.`

packages/core/useFetch/index.ts
```diff
@@ -276,44 +303,70 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
...
+          headers['Content-Type'] = 'multipart/form-data'
+        }
+        else if (config.payloadType === 'text') {
+          headers['Content-Type'] = 'text/plain'
+        }
+        else {
+          headers['Content-Type'] = 'application/x-www-form-urlencoded'
+          // if the payload is not URLSearchParams nor a string, we need to encode it
+          // if it is a string it should be already encoded: avoiding double encoding
+          if (!(payload instanceof URLSearchParams) && typeof payload !== 'string')
```
wheatjs, line 340: Wondering if we shouldn't leave this up to the user to do. I'm not 100% if we want to be modifying the payload. We could always provide a separate function that the user could use to encode it instead. Also the fetch function is getting quite large and since this is something that probably won't be used a lot which makes me think even more that we can just provide an encode function instead

isgj: I think none of this code is needed, not even as a separate function. `fetch` works fine even with `application/x-www-form-urlencoded` and `multipart/form-data`.

This will set content type to `application/x-www-form-urlencoded;charset=UTF-8`
```js
const res = await fetch('server.com/blabla', {
    method: 'post',
    body: new URLSearchParams({ email: 'email@email.com', password: 'asdasd' })
});
```

This to `multipart/form-data; boundary...`
```js
const fd = new FormData();
fd.append('email', 'email@e.com');
fd.append('password', 'passw');
const res = await fetch('server.com/blabla', {
    method: 'post',
    body: fd
});
```
Strings will map to `text/plain;`.

The problem with `useFetch` is that it's setting the type to `text` for body other then typeof object. Why not set the content type only if no `payloadType` was provided and the payload is of type object. For the rest you just pass them as they are.
Also this part can be updated before making the call
```diff
else {
        defaultFetchOptions.body = config.payload;
-        headers['Content-Type'] = config.payloadType === 'formData'
-               ? 'multipart/form-data'
-               : 'text/plain';
 }
```
No code is better code.

userquin: We are not modifying the payload, we are encoding the payload, this is equivalent to `JSON.stringnify` when using json payload. Just check the devtools of the payload sent in demo page, see original data on the network tab, just include &%<> in any form field.
If the data must be encoded by the user, what can we do if just providing an object?

isgj: I think the API provided by `useFetch` should be simpler, if the payload is an object the content type will be set to `application/json` and the payload will be stringified, else payload will go to `fetch` as is also the content type if provided. This way these should work:
```js
// This will set the type to json, `fetch` does not accept an object as body so we suppose
// it's intended to be used as json
const {data} = useFetch('/api').post({data: 'some data'})
```
```js
// Content type is set to `application/x-www-form-urlencoded` and properly encoded
// `useFetch` should just pass the payload and it works
// The same will be true also with `multipart/form-data` and `new FormData()`
const {data} = useFetch('/api').post(new URLSearchParams({data: 'some data &%+ '}))
```

I would strongly recommend the use of `URLSearchParams` and `FormData` instead of custom encoding functions.

userquin: Then, why are you encoding when using `json` payload? The implementation should be consistent, `useFetch` should deal with all encodings or none of them.

In the other hand, using a `multipart/form-data` to submit a form? This will have server side effects, mappings should be added/modified to just handle a `multipart/form-data`.

I have changed the `application/x-www-form-urlencoded` to build a `URLSearchParams` from the body object and so we don't need to encoded it explicitly. The method added `toURLSearchParams` can be exported, right now is internal.

About the `payloadType`, in this PR I have already change how we must handle it. In fact this was the cause I make this PR: the payload handling should check both, the `payloadType` and the `payload` itself, since the user can provide incoherent configuration and payload.

isgj: > Then, why are you encoding when using `json` payload? The implementation should be consistent, useFetch should deal with all encodings or none of them

Because only `object` can't be in body, `string`, `URLSearchParams` and `FormData` can be and the content type is handled by default by `fetch`.

And yes I'd prefer `useFetch` to be a thin wrapper to `fetch` which adds only the reactivity part without touching the payload, modifying the header or __throwing an error when status is not ok__ (I really don't like this behaviour :shrug:). If someone wants these behaviours they can use:
```js
const useCustomFetch = createFetch({
    beforeFetch() {/*do what you think is ok, encode, stringify ...*/},
    afterFetch() {/*you' can throw here if you want*/}
})

Comment (userquin) [reply packages/core/useFetch/index.ts:326]: What do you mean with modifying the payload? `useFecth` is not modifiying the payload, it is encoding to what `fetch` expects.

Take a look at #542, the method building the body uses the `content-type` header provided by the user. Since `useFetch` has `payloadType` configuration option, it is a pain, since we must deal with `fetch` on `multipart/form-data` and `application/x-www-form-urlencoded` and the type of the payload (like original does with `json` payload). 

Just comment lines https://github.com/vueuse/vueuse/blob/37221e486ba718bd7b6c10949b8e3a2892ebc5f1/packages/core/useFetch/index.ts#L341 and https://github.com/vueuse/vueuse/blob/37221e486ba718bd7b6c10949b8e3a2892ebc5f1/packages/core/useFetch/index.ts#L342, start the app, go to demo page, and send the multipart, the files entry from the server response will be missing.

We need to remove the header for `multipart/form-data`, since we are breaking `fetch`: it will be not possible to parse the multipart on server (once set `content-type`, the parts that `fetch` add to the request are not correlated to this header: it will change between executions, and if configured, that is, not removed, then it will be fixed).

Comment (userquin) [reply packages/core/useFetch/index.ts:326]: Commeting out the lines result in:

imagen

and the response from server result in:

imagen

while removing the header will be:

imagen

with the response from server:

imagen

Comment (userquin) [reply packages/core/useFetch/index.ts:326]: The problem here is the boundary added to the `content-type` header by `fetch`, that cannot be configure on the parts...

Comment (userquin) [reply packages/core/useFetch/index.ts:326]: We also need to avoid double encoding on `json` payload:
```javascript
const x = JSON.stringify({a: 1, b: 2})
const y = JSON.stringify(x)
```
```console
"{\"a\":1,\"b\":2}"
"\"{\\\"a\\\":1,\\\"b\\\":2}\""
```

isgj: > What do you mean with modifying the payload? useFecth is not modifiying the payload, it is encoding to what fetch expects.

But why handle some cases and not let the user decide how the data is sent? `fetch` will handle most of the cases a part from the `object`. For consistency we ignore this fact and let the user stringify the body and set the content type. Or since json is a common type, `useFetch` does the encoding only in this particular case, the rest is up to the user.

> Since `useFetch` has `payloadType` configuration option, it is a pain

Yes, it has a problem and it is a pain. That's why I'm saying `useFetch` shouldn't add `content-type` or deal with that.

> The problem here is the boundary added to the `content-type` header by `fetch`

Exactly, don't deal with content-type, do not add and remove it. It will work like `post(new FormData())` for multipart, `post(new URLSearchParams())` for url encoded.  It's easy, it's up to the user

userquin: I think the user expects that passing an object, `useFetch` can deal with it via `payloadType`. How many lines you must write to use `URLSearchParams` or `FormData` while if using a simple object (see demo page), the user can forget about the encoding logic?.

In fact, the same approach can be done for `multipart/form-data`, that is, the user provides a raw object, and `useFetch` can deal with it: create a  `FormData` adding each key to the  `FormData`, and if a `File` entry then configure also the name.

As a said before, deal with all encodings or none of them: why `json` type and not `formEncoded` or `formData`? Because the simplicity of the conversion?

userquin: Of course, if the user provides the object `fetch` expects, `useFetch` will do nothing... 

Another story would be if we had to square the circle to make the encodings, but if it is 2 lines of code!!!.

userquin: I think the behavior I am trying to add to `useFetch` is good from a user usage perspective.

The number of lines that we save and the abstraction in the encoding that we will provide to the users will be huge.

Give `useFetch` an object and tell him which encoding you want to apply, which will take care of what it has to do to send it to the server. 

This is how I think `useFetch` should work.

userquin: I'll push changes to build the multipart from a raw object.

This PR will have the code with the functionality I think `useFetch` should have.

There is one thing left to do about error handling, @isgj can you see my comments on #553?

We can have later a second round...

userquin: I'll provide some docs here, once reviewed last changes I'll add to the docs...

userquin:
I think the PR is done, waiting for reviews, tmr I'm taking a vacation until next Monday / Tuesday ... but I'll do it with an eye here.

userquin:
One last thing I must include: some tests for encoding. I cannot figure out how to test `formEncoded` and `formData`  with `jest`, since we need to mock request and response and there is no real server behind.
See you next week...

wheatjs:
I feel like this PR is doing to much and it is getting hard to review. Perhaps you can refocus this PR on the original problem is was solving and we can address the other fetch issues on another branch

userquin:
> I feel like this PR is doing to much and it is getting hard to review. Perhaps you can refocus this PR on the original problem is was solving and we can address the other fetch issues on another branch

Can you create a branch and merge this pr on it? Maybe this will help to review it.

packages/core/useFetch/index.ts
```diff
@@ -168,6 +185,46 @@ function isFetchOptions(obj: object): obj is UseFetchOptions {
   return containsProp(obj, 'immediate', 'refetch', 'beforeFetch', 'afterFetch')
 }
 
+async function toURLSearchParams(body: any): Promise<URLSearchParams> {
```
benlesh, line 188: I'm not sure I understand the purpose of this function. The goal seems to be to deal with the fact that the user may have accidentally passed a search param that is an object, rather than a primitive value type.

1. I'm not sure why it's asynchronous. I don't see any async work being done.
2. It doesn't seem like a good idea to JSON encode _everything_, because then whatever is on the other end of the network will have to JSON decode _each value_. In other words, it will get a bunch of key/value pairs, and then be forced to `JSON.parse` (or the server language equivalent) for every single value in there, even if they're all strings.

I strongly think the onus of doing this work (and deciding if it's even necessary) should be on the developer making the fetch, and we should throw this logic out and let them handle it. They know their server (presumably). And this is covering a use case that generally doesn't exist in the 99% case. Most of the time, if someone is sending search params, they're value-type primitives, not objects. This is penalizing the 99% use case to accommodate a one-off scenario.

packages/core/useFetch/index.ts
```diff
@@ -276,60 +333,105 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
...
+
     const defaultFetchOptions: RequestInit = {
       method: config.method,
       headers: {},
     }
 
-    if (config.payload) {
+    let payload = context.options?.body || config.payload
+
+    if (payload) {
```
benlesh, line 352: Most of the work below here does not need to be done. `fetch` and it's related types, like `Request`, will automatically figure out the proper `Content-Type` header.

image

It's worth noting that this is even the case with XHR. [I recently wrote some code related to this for XHR, which handles this](
image) that is much more concise. Hopefully that helps.

I _strongly_ advise that we don't do things like check a `payloadType` configuration or the like, because it will lead to user confusion and odd behaviors.

benlesh (changes requested):
Sorry for the drive-by review. I saw this PR in here while I was submitting one of my own, and I've recently done a lot of work in this area.

This PR looks like it was a lot of very well thought out work and consideration, but I would advise against merging it in its current state for a few reasons:

1. It adds a lot more weight to what should otherwise be a pretty light-weight helper.
2. The handling of setting content-type is weighty and incorrect.
3. The area that deals with `URLSearchParams` will add additional JSON parsing burden to any server consuming what is (in the majority case) supposed to be simple value types. And it's a lot of unnecessary work.
4. Most of the ergonomics gains that could be made from this PR seem like they may result in unintentional pain and a lot of extra weight to cover what, in my opinion, would be edge cases that developers could easily handle with their own code.

Again, this is just my general review that was completely unsolicited, and I really respect the author for putting in the effort and sticking to the debate in this PR.

I think maybe a much leaner, smaller, incremental step, with obvious value, might be attainable by refactoring this PR.

packages/core/useFetch/index.ts
userquin: We need to avoid to include a breaking change.
It is a pain `payloadType` option, we must deal with it.
The content type must be addressed since can be wrong, for example, adding it when using multipart, you can see the pictures included in this pr.

From a user perspective, it Will be more useful to provide the payload as a raw object rather than using fetch ones,.

I'm waiting to be included in a separated branch.

wheatjs:
Going to close this out as we plan on introducing `useKy` instead for advanced fetch usage.

PR #562, opened 2021-06-07
fix: Reduced the contrast of a dark theme

cawa-93:
* Made the `--c-bg` less saturated.
* Increased the brightness for `--c-brand-dark` to meet the WCAG requirements

### DIFF
--- a/packages/.vitepress/theme/styles/vars.css
+++ b/packages/.vitepress/theme/styles/vars.css
@@ -67,7 +67,7 @@
 }
 
 html.dark {
-  --c-bg: #111;
+  --c-bg: #222;
   --c-bg-semi: rgba(17,17,17,0.8);
   --c-bg-secondary: #222;
   --c-text: #f5f7fa;
@@ -82,6 +82,8 @@ html.dark {
 
   --c-disabled-bg: #333;
   --c-disabled-fg: #aaa;
+  
+  --c-brand-dark: #44bb83
 }
 
 /** Fallback Styles */


antfu:
Emmm, I don't really like it tbh. To me, it's more like "Dim mode" than "Dark mode". Maybe instead of brighten background, we can darken the foreground (you can refer to GitHub's dark mode, which has similar background brightness to what we have). WDYT?

cawa-93:
GitHub "Default dark" also causes discomfort (especially on the amoled display). "Dark dimmed" is perceived better. Dark dimmed is perceived a little better.

cawa-93:
> I am not sure how you define "discomfort"

I look at the text. And after a few seconds the eyes begin to ache. The eyes get tired quickly. When you look away from the site, for some time you see a bright imprint of text on your retina. Over time, headaches and dizziness begin.

You cited GitHub as an example. At first, they only added a dark theme. In the forums began to show a message that too high contrast and GitHub added the Dimmed theme.

The idea of a dark theme for websites is still new and good practices have not yet been developed.

Look out for dev tools or code editors where the idea of a dark theme has been around for a long time. For example, a Edge devtools in the dark mode has a background `rgb(36,36,36)`. However, most of the text there is not white. The brightest white color what I founded is `rgb(212,212,212)`- which is rather gray. The default theme for VS Code also has a similar color scheme.

> save battery

OK. It's true. But how expedient is this saving? I mean, which is more priority for the website: accessible content or energy saving? In addition, there are studies that show that the difference in power consumption between a absolutely "black" pixel and a "dark gray" pixel is negligible.

antfu:
Alright, does https://windicss.org 's contrast work for you? That's the most bright dark mode I could accept. Otherwise, we could consider introduce a third dim more as GitHub and Twitter. But that we require much more work as we need to have `useDark` to support that (actually might be `useColorSchema`) and maybe windicss to support the dim selector, etc.

cawa-93:
> Alright, does https://windicss.org 's contrast works for you?

With a short session is not bad.

antfu:
Working on a fresh new theme, close for now.

PR #567, opened 2021-06-08
feat(useObservable): now supports optional error handling via `onError`.

### LINKED ISSUE #566: useObservable: should allow the consumer to pass an error handler
Look at [where we're currently subscribing](https://github.com/vueuse/vueuse/blob/303705deaa3388fcaff990fa1159690d3f8248fc/packages/rxjs/useObservable/index.ts#L7), there's no error handler being passed.

This means that any error that happens in the observable will be "unhandled", which is fine, however, it might be worth giving users the option to handle the error like so:  `useObservable(source$, { onError: err => { /* do something */ } })`. The observable subscription would, of course, still terminate, however, RxJS would no longer treat errors that propagated to the point of subscription as "unhandled errors". (provided that the user supplied an error handler). If no `onError` is supplied, then we should allow RxJS to do it's thing.

benlesh:
Adds a few feature to allow users to configure error handling if they so choose. Without this feature, any error from the source observable would be treated as "unhandled" by RxJS, and thrown in a new call stack.

Resolves #566

### DIFF
--- a/packages/rxjs/useObservable/index.md
+++ b/packages/rxjs/useObservable/index.md
@@ -39,12 +39,12 @@ const count = useObservable(
       if (n === 10) {
         throw new Error('oops')
       }
-      return n + n;
+      return n + n
     })
   ),
   {
     onError: err => {
-      console.log(err.message); // "oops"
+      console.log(err.message) // "oops"
     }
   }
 )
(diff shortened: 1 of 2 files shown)


antfu (approved):

antfu:
Thank you!

antfu (approved):

PR #572, opened 2021-06-10
fix(useFetch): parse body even if status is not ok

### LINKED ISSUE #526: Can't acess to useFetch response's body on error ?

### LINKED ISSUE #553: `useFetch` Have access to the response body when the status is not ok

isgj:
~~fixes #553~~
fixes #526

While not yet clear what path we should take in order to have access to the response body when the status code is not `ok` I propose this PR.

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -326,14 +326,15 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
 
           let responseData = await fetchResponse[config.type]()
 
-          // see: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/
-          if (!fetchResponse.ok)
-            throw new Error(fetchResponse.statusText)
-
           if (options.afterFetch)
             ({ data: responseData } = await options.afterFetch({ data: responseData, response: fetchResponse }))
 
           data.value = responseData as any
+
+          // see: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/
+          if (!fetchResponse.ok)
+            throw new Error(fetchResponse.statusText)
+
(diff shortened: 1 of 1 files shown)


NoBrainSkull:
Looks fair enough to me. What do you think @wheatjs ?

wheatjs:
I think this is good, maybe we can also pass the error to the responseData to the error as well?

isgj:
Do you mean

```js
if (!fetchResponse.ok) {
  const error = new Error(fetchResponse.statusText)
  error.data = responseData
  throw error
}
```
like suggested https://github.com/vueuse/vueuse/issues/526#issuecomment-853070245

antfu (approved):

PR #573, opened 2021-06-10
feat(useFetch): set content type only if it is json

### LINKED ISSUE #539: useFetch and formData doesn't work

### LINKED ISSUE #542: useFetch: multipart form data header must be removed instead added

isgj:
fixes #539 
fixes #542

In this PR only `json` is handled as content type.
This is an alternative to #549

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -65,14 +65,14 @@ interface UseFetchReturnBase<T> {
 }
 
 type DataType = 'text' | 'json' | 'blob' | 'arrayBuffer' | 'formData'
-type PayloadType = 'text' | 'json' | 'formData'
+type HttpMethod = 'get' | 'post' | 'put' | 'delete'
 
 interface UseFetchReturnTypeConfigured<T> extends UseFetchReturnBase<T> {
   // methods
   get(): UseFetchReturnBase<T>
-  post(payload?: unknown, type?: PayloadType): UseFetchReturnBase<T>
-  put(payload?: unknown, type?: PayloadType): UseFetchReturnBase<T>
-  delete(payload?: unknown, type?: PayloadType): UseFetchReturnBase<T>
+  post(payload?: unknown, type?: string): UseFetchReturnBase<T>
+  put(payload?: unknown, type?: string): UseFetchReturnBase<T>
+  delete(payload?: unknown, type?: string): UseFetchReturnBase<T>
 }
 
 export interface UseFetchReturn<T> extends UseFetchReturnTypeConfigured<T> {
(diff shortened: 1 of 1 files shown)


wheatjs:
Sorry for the delay on this PR, I'll review in a few hours so we can merge 🙂

wheatjs (approved):

PR #583, opened 2021-06-18
fix(usePointerSwipe): use `style.setProperty` instead of `setAttribute`

### LINKED ISSUE #548: usePointerSwipe: target inital styles removed on click
When I have initial styles on **target** element and click to it, my initial styles removed and changed to **touch-action: initial**
https://gfycat.com/nicesoftcoelacanth
https://github.com/ilkome/usepointerswipe/blob/master/src/App.vue

jeffrey-dot-li:
fixes #548


packages/core/usePointerSwipe/index.ts
```diff
@@ -99,7 +99,7 @@ export function usePointerSwipe(
     useEventListener(target, 'pointerdown', (e: PointerEvent) => {
       isPointerDown.value = true
       // Disable scroll on for TouchEvents
-      targetRef.value?.setAttribute('style', 'touch-action: none')
+      targetRef.value?.style.setProperty("touch-action", "none")
```
userquin, line 102: shouldn't be?:
`targetRef.value?.style?.setProperty("touch-action", "none")`

userquin: also use single quote
`targetRef.value?.style?.setProperty('touch-action', 'none')`

userquin: you forget also change on `pointerup` event... https://github.com/vueuse/vueuse/blob/main/packages/core/usePointerSwipe/index.ts#L130

jeffrey-dot-li: K thanks, was wondering as well if on line 51
https://github.com/vueuse/vueuse/blob/main/packages/core/usePointerSwipe/index.ts#L51
the Ref should be typed as HTMLElement instead of Element because `style` doesn't always exist on Element?

userquin: I think you should wait for author review...

PR #591, opened 2021-06-27
feat(useVForRefs): new function

### LINKED ISSUE #590: [feature] useVForRefs - Shorthand for v-for Array Refs
Hi, Sometimes we need to use refs on templates, and this can be encapsulated in a hook.

Possible implementation:

```typescript
const useVForRefs = () => {
  const refs = ref([])

  const setRef = (el) => {
    if (el) {
      refs.value.push(el)
    }
  }

  onBeforeUpdate(() => {
    refs.value.length = 0
  })

  return {
    refs,
    setRef,
  }
}
```

Example:

```vue
<template>
  <div v-for="item in list" :ref="listRef.setRef"></div>
</template>

<script setup>
const listRef = useVForRefs()
</script>
```

Ref:
https://v3.vuejs.org/guide/migration/array-refs.html#migration-strategy

Thanks for the useful and awesome library ❤️.

iendeavor:
ref: https://github.com/vuejs/vue-next/issues/4010

closes #590


antfu:
Close in favor of #645, thanks for the contribution!

PR #605, opened 2021-07-05
fix(usePointerSwipe): set capture on event target instead of root element

### LINKED ISSUE #584: UsePointerSwipe: `setPointerCapture` disables click events on chrome
When using usePointerSwipe, click events are disabled for that element and its children - including @click, button and `<a href=...` events when opened in chrome. Works normally for firefox and safari from my testing.
Result of https://github.com/vueuse/vueuse/blob/cfc1ed8525ae1fabdb21b248999702a059ea9212/packages/core/usePointerSwipe/index.ts#L104

and setPointerCapture. Line should be changed to `e.target?.setPointerCapture(e.pointerId)` instead, which I found from this reply on the google chrome repo. 
https://github.com/GoogleChromeLabs/pointer-tracker/issues/4#issuecomment-593929716

jeffrey-dot-li:
fixes #584

### DIFF
--- a/packages/core/usePointerSwipe/index.ts
+++ b/packages/core/usePointerSwipe/index.ts
@@ -101,8 +101,8 @@ export function usePointerSwipe(
       // Disable scroll on for TouchEvents
       targetRef.value?.style?.setProperty('touch-action', 'none')
       // Future pointer events will be retargeted to target until pointerup/cancel
-      targetRef.value?.setPointerCapture(e.pointerId)
-
+      const eventTarget = e.target as HTMLElement | undefined
+      eventTarget?.setPointerCapture(e.pointerId)
       const { clientX: x, clientY: y } = e
       updatePosStart(x, y)
       updatePosEnd(x, y)


packages/core/usePointerSwipe/index.ts
```diff
@@ -101,8 +101,8 @@ export function usePointerSwipe(
       // Disable scroll on for TouchEvents
       targetRef.value?.style?.setProperty('touch-action', 'none')
       // Future pointer events will be retargeted to target until pointerup/cancel
-      targetRef.value?.setPointerCapture(e.pointerId)
-
+      const eventTarget = e.target as HTMLElement | undefined
+      eventTarget?.setPointerCapture(e.pointerId)
```
lstoeferle, line 105:
```suggestion
        (e.target as HTMLElement | undefined)?.setPointerCapture(e.pointerId)
```

antfu: I don't actually mind to have two lines for type assertion. But I am not so sure of the logic here, does it make sense to capture on the event target instead @lstoeferle?

lstoeferle: In the end `e.target` is the same element like `targetRef.value`. Maybe there is some browser specific implementation that breaks the click events on chrome (black-box for me).

Issue #576, opened 2021-06-11
[object Object] after storing object after null in useLocalStorage


xorik:
```js
import { useLocalStorage } from '@vueuse/core'

const data = useLocalStorage('test', null)
data.value = {hello: 'world'}
```

[Demo](https://play.vueuse.org/#N4IgDghgxg1hDmBTAziAXAbVAOwgW0XRADcBXRAWgBNE8BLEAGhGQHtSAnKQtEU7MDHhMQNZFA50wAFzqtsRAGrkABABFadFQAoAFhAA2AMxV1sKgGIdE2KLoCUp5CogqaxRAdZgz8FaVkDOmkATxcDLwB3ZxD2FWlWFUjJaUQVAFVsOg8OZEMVZTSAGToAIw4ISRQVI1YOAtUAJhUAMhUAZhFOAyJdaWkwZDQAemH+QXgAOihWPGGyShp6YaDS4bMaAA9JlDxJgCtUAF9GHHweEAABBdJkRGHkfWsqETZObiJxoRExCSlZeREADKT0QVAaiHSd38gWCdBQky6HB6vD6AyGoy+Uxmc2u5Fu90elTBlwArJMAAyTACM62wWx2yD2hxAJzOBCIeMQBOGM2sr3YXAuWJ+KD+MjkCl4AGFWBFEFAAeZWCYUHdsLJ8oUVLK8GBWMg4fIMrDZCgkSiQGjBiMxgIhNNZsMuTy+YgyZSaXSGbsDscALrMIx0AzmtCgACCYDAkwW6FAqT1BggqSIr3F0iIdD1dWkKmA-juRVYUEMQISFSQKiONQ4sxUAHIXXdeXVEA2ADrYLszbDIPNUFOuAC8hcQxdLBnLdQQiG0DdS-YbjBU2FIEXsXcH0ggscMqlHwF0ni8aEbkTqBioDaOrKORyAA)

![Screen Shot 2021-06-11 at 08 04 09](https://user-images.githubusercontent.com/347044/121638652-9be9fc80-ca8b-11eb-9efd-a5f4498ace50.png)

If I replace `null` with `{}` it will work as expected

jeffrey-dot-li:
@mxmvshnvsk This issue is caused because null is typed as 'any', which is basically 'string', by default when sent to local storage. The only way I can think of to fix this is to store null as an object by default. In theory when parsing unknown values you can default to assuming it is an object with `JSON.parse`:
```ts
[...]

PR #606, opened 2021-07-05
fix(useStorage): read type with default null as null instead of string

### LINKED ISSUE #597: [bug] useStorage default value `null` get's written to localstorage and read as `"null"`
If you set `null` as the default value of `useStorage`, this value get's written to the localStorage on initial execution and is then read as the string `"null"`. This behaviour was most likely introduced with #528 and I suspect it isn't intended. 
All versions >=4.11.1 are affected.

Example:
```
import { useStorage } from "@vueuse/core";
const test = useStorage("key", null);
console.log("Key:", typeof test.value, test.value);

// First load (expected behaviour)
// > Key: object null
// Subsequent loads (broken)
// > Key: string "null"
```

jeffrey-dot-li:
fixes #597

### DIFF
--- a/packages/core/useStorage/index.test.ts
+++ b/packages/core/useStorage/index.test.ts
@@ -98,6 +98,26 @@ describe('useStorage', () => {
     })
   })
 
+  it('null value', () => {
+    localStorage.setItem(KEY, 'null')
+
+    useSetup(() => {
+      const ref = useStorage(KEY, null)
+
+      expect(ref.value).toBe(null)
+    })
+  })
+
+  it('null string', () => {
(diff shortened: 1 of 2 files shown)


adrianjost (approved):
I can confirm that this solves my issue and the code LGTM.

mxmvshnvsk:
@jeffrey-dot-li please, check this issue too #576, I think your PR can fix it.

jeffrey-dot-li:
> 
> 
> I can confirm that this solves my issue and the code LGTM.

Aight ill take a look today thanks

PR #612, opened 2021-07-06
fix(useFetch): reject for `execute` on error

### LINKED ISSUE #611: `useFetch` never resolves when response is not ok
`useFetch` returns `execute` function that can be called when `immediate = false`, pretty useful, especially when used when `post`, `put`... after some user interaction.
If we want to wait for call we can do 
```ts
await post({...}).execute()
```
but when `response.ok === false` the promise returned by `execute` is never resolved.

isgj:
fixes #611 

Since `useFetch` is a reactive function it should resolve always. The error is saved in the reactive variables.

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -308,7 +308,7 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
       return Promise.resolve()
     }
 
-    return new Promise((resolve) => {
+    return new Promise((resolve, reject) => {
       fetch(
         context.url,
         {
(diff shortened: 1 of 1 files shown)


wheatjs:
Wondering if you think we should use `reject` instead?

isgj:
Because it is part of a reactive function I think it shouldn't reject. Even though it's not `useFetch` that is throwing so I'm not really sure. (sorry if I'm raising more doubts :grin:)

wheatjs:
Hmm I guess my only concern would be if someone uses the `execute` function and expects to be able to catch the error. We also reject promises in other functions as well, see [useScriptTag](https://github.com/vueuse/vueuse/blob/499a6987128e3db6862f57bcc189095d6c501a67/packages/core/useScriptTag/index.ts)

Do you know of any issues that would be caused by rejecting instead of resolving?

isgj:
No I don't know any issue. You're right, it will be better if it's rejected.

PR #631, opened 2021-07-16
fix: add onFetchFinally. Fix headers issue #625

### LINKED ISSUE #625: [useFetch] Headers cannot be a `Headers` object
Works:

```ts
const {data} = useFetch("/page", {headers: {Accept: "application/json"}}).json();
```

Doesn't:

```ts
const headers = new Headers();
headers.append("Accept", "application/json");

const {data} = useFetch("/page", {headers}).json();
```

I think it's caused by the spreading here:

https://github.com/vueuse/vueuse/blob/b3d1e2c8c6715dd972392c85a2080dcfe52cab09/packages/core/useFetch/index.ts#L195-L198

And here:

https://github.com/vueuse/vueuse/blob/b3d1e2c8c6715dd972392c85a2080dcfe52cab09/packages/core/useFetch/index.ts#L316-L319

### LINKED ISSUE #628: [useFetch] Add `fetchError` and/or `fetchFinally` hooks
There's currently no way to execute code (for example, hiding a spinner) when there is an error.

`onFetchError` can only be used when calling `useFetch`, but as far as I know, you can't configure this before-hand using `createFetch`.

Would it be possible to add a `fetchError` and/or a `fetchFinally` hook, to enable this?

wheatjs:
Adds `onFetchFinally` which will run on both successful and error responses. 
Fix issue with headers as described in #625 

Closes #625 
Closes #628

### DIFF
--- a/packages/core/useFetch/index.test.ts
+++ b/packages/core/useFetch/index.test.ts
@@ -21,6 +21,20 @@ describe('useFetch', () => {
     expect(data.value).toBe('Hello World')
   })
 
+  test('should be able to use the Headers object', async() => {
+    fetchMock.mockResponse('Hello World', { status: 200 })
+
+    const myHeaders = new Headers()
+    myHeaders.append('Authorization', 'test')
+
+    const { statusCode, isFinished } = useFetch('https://example.com', { headers: myHeaders })
+
+    await until(isFinished).toBe(true)
+
+    expect(statusCode.value).toBe(200)
+    expect(fetchMock.mock.calls[0][1]!.headers).toMatchObject({ authorization: 'test' })
+  })
+
   test('should parse response as json', async() => {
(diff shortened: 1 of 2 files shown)


antfu (approved):

glen-84:
@wheatjs,

Thanks for doing this, but unfortunately it's not quite what I was looking for. It's only possible to use "hooks" (beforeFetch/afterFetch) when creating a custom instance with `createFetch`, for example:

```ts
const useMyFetch = createFetch({ 
  options: {
    beforeFetch() {
      showGlobalSpinner();
    },
    afterFetch() { // Doesn't run when there's an error. :(
      hideGlobalSpinner();
    },
  },
})
```

There's no way to set up events here, as they are only available when calling `useMyFetch`, as far as I know.

Would it be possible to provide one of these "hooks" (maybe "fetchError" or "fetchFinally") instead of or in addition to the events?

PR #642, opened 2021-07-22
fix(useStorage): default null values will not be stored in local storage

### LINKED ISSUE #597: [bug] useStorage default value `null` get's written to localstorage and read as `"null"`

jeffrey-dot-li:
`useStorage('key', null)` now no longer puts `{'key' : 'null'}` into local storage. All other useStorage behaviour is now working as expected: 
`state.value = null` now always deletes the item from local storage and sets `state.value` to be `null` as expected.
If the value of 'null' *is* stored within local storage, it is no longer treated any differently than any other string.
`useStorage('key', null)` now works as the default value for `useStorage` without any other strange behaviour.

fixes #597 properly


packages/core/useStorage/index.ts
```diff
@@ -11,23 +11,23 @@ export type Serializer<T> = {
...
   },
   any: {
-    read: (v: any) => (v != null && v !== 'null') ? v : null,
+    read: (v: any) => v,
     write: (v: any) => String(v),
   },
   string: {
-    read: (v: any) => v != null ? v : null,
+    read: (v: any) => v,
     write: (v: any) => String(v),
```
adamdehaven, line 31: @jeffrey-dot-li if the user passes `null` wouldn't this still convert it to a string of `"null"`? or am I missing something?

jeffrey-dot-li: Nah if the use passes `null` it doesn't call `serializer.write` in the first place so there is no string conversion - line 115:
```
        if (defaultValue !== null)
          storage.setItem(key, serializer.write(defaultValue))
```

antfu:
Can you target to the `next` branch instead. While I believe this is the correct way to do, but there might be some users already rely on this behavior, so we'd better to release it as a breaking change. Thanks!

adamdehaven:
> Can you target to the `next` branch instead. While I believe this is the correct way to do, but there might be some users already rely on this behavior, so we'd better to release it as a breaking change. Thanks!

@antfu any idea when the next branch will be released? I'm using this in a big MVP and would be a big help.

_Also, I wasn't able to test locally so I'm depending on someone else verifying this is the fix 😃_

antfu:
I release it quite frequently as beta, you can see the releases tab in the repo.

PR #643, opened 2021-07-23
docs: show default value example for useRouteQuery

marshallswain:
Also adds one line of whitespace to give the ts definitions some breathing room.  ;)


antfu:
Type definitions are generated - the change will not be applied.

Issue #422, opened 2021-04-05
Global Event Bus


brokolosov:
Any plans for adding a global event Bus? For example wrapper to [mitt](https://github.com/developit/mitt)

antfu:
Do you have any proposal API in mind?

brokolosov:
Just like the rest...

`
import { useBus } from '@vueuse/core'

export default defineComponent({
  setup() {
    const {emit, on, off} = useBus()

   emit('foo-event', {value: 'bar'})
  },
})
`

lstoeferle:
> Just like the rest...
> 
> ```es6
> import { useBus } from '@vueuse/core'
> 
> export default defineComponent({
>     setup() {
>     const {emit, on, off} = useBus()
> 
>     emit('foo-event', {value: 'bar'})
>   },
> })
> ```

I'm using [mitt](https://github.com/developit/mitt) for custom events in Vue3. Could be used as an integration function 😉

PR #647, opened 2021-07-26
feat(useEventBus): new function

TuiMao233:
#422 A simple Watcher | Publish and Subscribe mode hook.

....
The submitted information is wrong, don’t care

~~fix(computedInject): After unsubscribing, when observeItem.size is em…~~

fix(usePubsub): After unsubscribing, when observeItem.size is em…


antfu:
What's the different for this between an event emitter? I would think the name `usePubsub` could be a bit unclear by just reading it. Also think the api been `on` / `off` / `emit` could match more with the common practice in JavaScript.

TuiMao233:
It can be used in cross-component application scenarios where it is difficult to find the corresponding relationship.

TuiMao233:
As for why this is made, this is a bit of a long story. When I use useStorage, it is very convenient, but the only shortcoming is that when I use useStorage on multiple pages, one page change will not cause other pages to change.

So I implemented a useStorage that can monitor other page changes.

image

Then I thought, if the option of useStorage itself can have this field, it can be realized, then I think it will be very convenient.

But this implementation needs to use pubsub-js or other libraries, because I read a lot, @vue/use does not have this implementation.

I read the contribution document and it says `In the core package, try not to introduce 3rd-party`

So first make a lighter Watcher | Publish and Subscribe mode hook.

TuiMao233:
As for the name, I decided based on pubsub-js, sorry I can't think of a better name.

TuiMao233:
Well, how about the name of useEventBus, and then the exposed method is `on off emit`.

antfu:
Sounds cool, let's go for it!

wheatjs:
If targeting v6.x, it is probably preferable to use `onScopeDispose` instead of `onUnmounted`

TuiMao233:
@wheatjs 
Okay, but I think the latest version can also be uploaded, so I am submitting a merge in `next`.

antfu:
You don't actually need to create #649, I will sync it up once merged (so you don't need to do double effort for every change).

There is a few design thing I would like to discuss about.

1. instead of the listener "token", I would like to have the listener it self to do the unregister

```ts
const bus = useEventBus('foo')
const inc = () => counter.value++

bus.on(inc)

bus.off(inc) // <--
```

2. also, for the return of the `on` and `once`, we could return a simple `off` function (similar to stop from `watch`), also refer to: https://github.com/ai/nanoevents#remove-listener

```ts
const bus = useEventBus('foo')
const inc = () => counter.value++

const off = bus.on(inc)

off() // same as `bus.off(inc)`
```

3. we could improve the DX by binding the key with types, similar to vue's `InjectionKey` util.

```ts
const fooEventKey = Symbol() as UseEventBusKey<{ name: foo }>

const bus = useEventBus(fooEventKey)

bus.on((e) => {
	// `e` will be `{ name: foo }`
})
```

more details: https://antfu.me/notes#typed-provide-and-inject-in-vue

TuiMao233:
Wow, they all seem to be some good designs, and they are more convenient to use. I think I can try them.

I have understood the usage of InjectionKey, and it is really suitable to use in useEventBus.

But it may not be so fast, I have to make sure that the test passes, and I am a bit busy lately.

TuiMao233:
I think UseEventBusKey can be abbreviated as EventBusKey, what do you think.

TuiMao233:
Hmm..., I encountered a challenge when implementing it. The `once` method needs to encapsulate a layer of methods. However, if the `listener` can be passed in, the `once` cannot be found when it is off. I can't think of a solution to this method for the time being, and may have to abandon the `once` method.

TuiMao233:
Currently it is stored in the form of `Map<Key, Set<Function>>`

Delete the corresponding listener through set.has(Function)

But..., I encountered a challenge when implementing it. The `once` method needs to encapsulate a layer of methods. However, if the `listener` can be passed in, the `once` cannot be found when it is off. I can't think of a solution to this problem for the time being.

```js
const set = new Set()
const listener = () => {}
const onceListener = () => {
  // ...
  listener()
  off(onceListener)
  // ...
}
set.add(onceListener)
set.has(listener) // false
```

antfu:
> UseEventBusKey can be abbreviated as EventBusKey

Ok.

> `once` + `off`

To solve this, we could have another set of listeners for `once`. But that could make this function a bit heavy. I agree to drop the support atm as it's easy to do on the user land.

TuiMao233:
All right, now I also want to add a feature, that is, by default all `on` | `once` are asynchronous, so that you can specify whether a listener is processed synchronously.

Example 1
~~~js
const bus = useEventBus('foo')
// async
bus.on(() => console.log('-- 1 --'))
// sync
bus.on.sync(() => console.log('-- 2 --'))
bus.emit()
// '-- 2 --'
// '-- 1 --'
~~~

TuiMao233:
Yes, this will make this method look less concise, I am thinking about whether there is a better way.

TuiMao233:
Yes, if our goal is small size, we can consider another direction.

TuiMao233:
I added index.small.ts and it only retains the most basic functions, but it is very small.

antfu:
I think we can drop `once` for simplicity and smaller size. For ppl need more advanced features, they'd better to use other standalone emitter libraries instead.

TuiMao233:
Okay, this direction is also very good, so I will add test cases and documentation, and it will be almost done by then.

TuiMao233:
Okay, I think this should be the final version.

TuiMao233:
Already done, I think this should be the final version.

TuiMao233:
Already done, I think this should be the final version.

TuiMao233:
Already done, I think this should be the final version.

antfu:
@TuiMao233 I have made some changes to the PR, please see if they make sense to you.

antfu:
Moved to targeting `next` branch

TuiMao233:
Well, I didn’t check the mailbox in time and there is no problem

PR #664, opened 2021-08-11
feat(useStorage): watcher option.

TuiMao233:
Let multiple useStorage with the same key have data binding.

~~~js
// page-1.vue
const count = useStorage('count', 0)
// page-2.vue
const count = useStorage('count', 0, undefined, { watcher: true })


// page-1.vue
count.value++

// page-2.vue
count.value // 1
~~~

### DIFF
--- a/packages/core/useStorage/index.test.ts
+++ b/packages/core/useStorage/index.test.ts
@@ -270,4 +270,35 @@ describe('useStorage', () => {
 
     expect(localStorage.removeItem).toBeCalledWith(KEY)
   })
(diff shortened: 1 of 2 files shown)


antfu:
Thanks for the contribution. But I'd think it's worth to embedded another composable inside `useStorage`. Ideally, the composition should happen on the user side as possible. For what you want, you probably looking for https://github.com/vueuse/vueuse/blob/next/packages/shared/createSharedComposable/index.md

PR #669, opened 2021-08-13
feat(useVModel): new options `deep` to emit for deep changes

### LINKED ISSUE #613: `useVModel`: need attribute {deep: true} when using "watch" to trigger "emit"
If a props type is `object` or `array`, and use `useVModel` with `{passive: true}`, `watch` can't observe the internal changes of `object` and `array`, so we can add attribute `deep`  to the relevant `watch`

webfansplz:
fix #613

### DIFF
--- a/packages/core/useVModel/index.test.ts
+++ b/packages/core/useVModel/index.test.ts
@@ -79,6 +79,7 @@ describe('useVModel', () => {
     await nextTick()
 
     expect(emitMock.mock.calls[0][0]).toBe('update:data')
+    expect(emitMock).toHaveBeenCalledTimes(1)
     expect(JSON.stringify(emitMock.mock.calls[0][1])).toBe(JSON.stringify({ age: 20 }))
   })
 
@@ -97,6 +98,7 @@ describe('useVModel', () => {
     await nextTick()
 
     expect(emitMock.mock.calls[0][0]).toBe('update:data')
+    expect(emitMock).toHaveBeenCalledTimes(1)
     expect(JSON.stringify(emitMock.mock.calls[0][1])).toBe(JSON.stringify({ hobbys: ['coding', 'basketball'] }))
   })
 })
--- a/packages/core/useVModel/index.test.ts
+++ b/packages/core/useVModel/index.test.ts
@@ -57,7 +57,7 @@ describe('useVModel', () => {
       const data = useVModel(props, 'age', emitMock, { passive: true })
       data.value = 20
     })
-    expect(emitMock.mock.calls[0][0]).toBe(isVue2 ? 'input' : 'update:age')
+    expect(emitMock.mock.calls[0][0]).toBe('update:age')
     expect(emitMock.mock.calls[0][1]).toBe(20)
   })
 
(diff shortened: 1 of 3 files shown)


packages/core/useVModel/index.test.ts
```diff
@@ -48,4 +47,59 @@ describe('useVModel', () => {
...
+        hobbys: ['coding'],
+      },
+    }
+    useSetup(() => {
+      const data = useVModel(props, 'data', emitMock, { passive: true })
+      data.value.hobbys.push('basketball')
+    })
+
+    // This is used because the Watch Callback is Asynchronous
+    await new Promise(resolve => setTimeout(resolve, 0))
```
antfu, line 100:
```suggestion
    await nextTick()
```

packages/core/useVModel/index.ts
```diff
@@ -54,12 +54,18 @@ export function useVModel<P extends object, K extends keyof P, Name extends stri
...
+    const proxy = ref<P[K]>(defaultProxy)
+
+    watch(() => props[key!], v => proxy.value = v as UnwrapRef<P[K]>)
 
-    watch(() => props[key!], v => proxy.value = v as any)
     watch(proxy, (v) => {
       if (v !== props[key!])
         _emit(event, v)
+    }, {
+      deep: !!isObject,
```
antfu, line 68: I'd prefer to expose the deep option for user to specify.

webfansplz:
@antfu fix,thank you for your review~

packages/core/useVModel/index.ts
```diff
@@ -54,12 +61,16 @@ export function useVModel<P extends object, K extends keyof P, Name extends stri
   event = eventName || event || `update:${key}`
 
   if (passive) {
-    const proxy = ref<P[K]>(props[key!])
+    const defaultProxy = deep ? JSON.parse(JSON.stringify(props[key!])) : props[key!]
```
antfu, line 64: I am a worry about the clone here. As user might pass some object/instance that are not serializable (or not intended to be cloned) and introduce some performance cost. I don't think it's a proper solution here.

webfansplz: @antfu Thanks. i clone it is because  the `props`  and `proxy` data share the same mem address:
```ts
watch(proxy, (v) => {
  if (v !== props[key!]) // always equal
    _emit(event, v)
}, {
  deep,
})
```
maybe we can skip it  when user  setup the deep option ?  Or do you have some other proper solution ? 😊

```ts
watch(proxy, (v) => {
  if (v !== props[key!] || deep) // skip when user  setup the deep options
    _emit(event, v)
}, {
  deep,
})
```

antfu: Actually I am wondering what's the use case here? If you changed nested props, as they are the same reference for both child and parent, it should be no problem of passing the reactivity.

webfansplz: @antfu this purpose of  :
```ts
watch(proxy, (v) => {
  if (v !== props[key!]) // changed nested props,always equal. so can't trigger emit
    _emit(event, v)
}, {
  deep,
})
```

antfu: If you only want it to trigger the emit everytime, maybe this could be a solution

```ts
watch(proxy, (v) => {
  if (v !== props[key!] || deep) // skip when user  setup the deep options
    _emit(event, v)
}, {
  deep,
})
```

but I am just not really understand why you need this

webfansplz: @antfu  just want to fix #613 😂

antfu: ok, let's go with the approach above then. Thanks

webfansplz: @antfu Thanks~

antfu (approved):

PR #674, opened 2021-08-17
chore: use tryOnScopeDispose instead of tryOnUnmounted

webfansplz:
use tryOnScopeDispose instead of tryOnUnmounted


antfu:
This is left on purpose - it does not make sense to have `onMounted` to pair with `onScopeDispose`

PR #680, opened 2021-08-18
feat(useEventBus): new Function `once`

TuiMao233:
I am not very good at writing English documents. If I think this implementation is okay, I may have to trouble you to improve the documents.

### DIFF
--- a/packages/core/useEventBus/index.ts
+++ b/packages/core/useEventBus/index.ts
@@ -2,10 +2,7 @@ import { Fn } from '@vueuse/shared'
 import { getCurrentScope } from 'vue-demi'
 import { events } from './internal'
 
-export type EventBusListener<T = unknown> = {
-  (event: T): void
-  _?: EventBusListener<T>
-}
+export type EventBusListener<T = unknown> = (event: T) => void
 export type EventBusEvents<T> = EventBusListener<T>[]
 
 export interface EventBusKey<T> extends Symbol { }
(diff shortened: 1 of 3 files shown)


packages/core/useEventBus/index.ts
```diff
@@ -45,12 +54,22 @@ export function useEventBus<T = unknown>(key: EventBusIdentifer<T>): UseEventBus
...
   }
 
+  function once(listener: EventBusListener<T>) {
+    function _listener(...args: any[]) {
+      off(listener)
+      // @ts-ignore
+      listener(...args)
+    }
+    _listener._ = listener
+    return on(_listener)
```
antfu, line 64: Interesting solution! But honestly it feels a bit magic, how about:

```suggestion
    on(_listener)
    return () => off(listener)
```

And let's say we don't support to use `bus.off(...)` to remove listeners registered using `once` (which I believe is a very rare use case) but use the stop handler returned by `once`

TuiMao233: Yes, it is indeed a very interesting implementation, I saw when I checked the information.
There may indeed be this problem here, because when off, only the original method passed in will be compared.
This modification can avoid this problem.

packages/core/useEventBus/index.ts
```diff
@@ -15,6 +18,12 @@ export interface UseEventBusReturn<T> {
    * @returns a stop function to remove the current callback.
    */
   on: (listener: EventBusListener<T>) => Fn
+  /**
+   * Similar to on, but he finds it once
```
antfu, line 22:
```suggestion
   * Similar to `on`, but only fires once
```

antfu (approved):

PR #682, opened 2021-08-19
chore(useMediaControls): remove useless option

### LINKED ISSUE #678: useMediaControls always returns buffering = false
https://github.com/vueuse/vueuse/blob/main/packages/core/useMediaControls/index.ts#L158

The value is never set by anyone and always returns false.

webfansplz:
maybe the `buffering` option does not make sense .

fix #678


wheatjs (approved):

jefferycline1:
😁

PR #694, opened 2021-08-24
feat(useFetch): payload supports ref and follow changed

Jay214:
Set automatically refetch if the payload is a ref and changed.
This is a common scenario where we want to make the request again when the parameters change.


packages/core/useFetch/index.ts
```diff
@@ -129,7 +129,7 @@ export interface UseFetchOptions {
   immediate?: boolean
 
   /**
-   * Will automatically refetch when the URL is changed if the url is a ref
+   * Will automatically refetch when the URL is changed if the url is a ref or the payload is a ref
```
DarrylNoakes, line 132:
```suggestion
   * Will automatically refetch when:
   * - the URL is changed if the URL is a ref
   * - the payload is changed if the payload is a ref
```

packages/core/useFetch/index.ts
```diff
@@ -407,9 +412,10 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
         config.method = method
         config.payload = payload
         config.payloadType = payloadType
+        watchPayload()
         // Set the payload to json type only if it's not provided and a literal object is provided
         // The only case we can deduce the content type and `fetch` can't
-        if (!payloadType && payload && Object.getPrototypeOf(payload) === Object.prototype)
+        if (!isRef(payload) && !payloadType && payload && Object.getPrototypeOf(payload) === Object.prototype)
```
DarrylNoakes, line 418: What does this do?

Why not:

```suggestion
        if (!payloadType && unref(payload) && Object.getPrototypeOf(unref(payload)) === Object.prototype)
```

packages/core/useFetch/index.md
```diff
@@ -235,7 +235,8 @@ export interface UseFetchOptions {
    */
   immediate?: boolean
   /**
-   * Will automatically refetch when the URL is changed if the url is a ref
+   * Will automatically refetch when the URL is changed if the url is a ref 
+   * or the payload is a ref and changed
```
DarrylNoakes, line 239:
```suggestion
   * Will automatically refetch when:
   * - the URL is changed if the URL is a ref 
   * - the payload is changed if the payload is a ref
```

DarrylNoakes (changes requested):
Just some suggestions.
Take them as you please 😉

Jay214:
> Just some suggestions.
> Take them as you please 😉

thanks you suggestions. is useful

packages/core/useFetch/index.ts
Jay214: In fact, 'Object.getPrototypeOf(payload) === Object.prototype' will always return false if the payload is a ref

Issue #644, opened 2021-07-24
Checking for capslock on in useMagicKeys


Fanna1119:
I am trying to detect if capslock is on.
https://vueuse.org/core/useMagicKeys/

via mozilla docs it seems the event listeners do contain some magic to see if things like numlock and caps are true.
see:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState

enhancement perhaps?

PR #698, opened 2021-08-26
feat(useKeyModifier): new function

JMaylor:
#644 

Let me know what you think


packages/core/useModifierState/index.ts
```diff
@@ -0,0 +1,13 @@
+import { ref } from 'vue-demi'
+import { useEventListener } from '../useEventListener'
+
+export function useModifierState(modifier: string) {
```
antfu, line 4: I see the doc states only a few of possible options, can we better type them as literals?

antfu (changes requested):
Overall looks great, thanks!

I would think the name a bit too general when out-of-context. Thinking we could have `useKeyModifier` or `useKeyModifierState`.

JMaylor:
> Overall looks great, thanks!
> 
> I would think the name a bit too general when out-of-context. Thinking we could have `useKeyModifier` or `useKeyModifierState`. WDYT?

Makes sense - I originally used this name to align with the underlying `getModifierState` method - but I think it also makes sense to be as clear as possible so we can go with `useKeyModifierState`

JMaylor:
I gave this a bit of a rework. Looking at some other functions I tried to make the return value more consistent with other functions.

Before:
``` ts
// call function once per modifier and pass in the modifier name
const shift = useModifierState('Shift')
const capsLock = useModifierState('CapsLock')
```
After:
``` ts
// call function once and destructure what you need
const { shift, capsLock } = useKeyModifierState()
```


I also updated the demo page to be a little nicer. 
Appreciate any feedback on typing by the way - I literally picked it up TS to do this. Thanks!

antfu:
> rework

I actually prefer the previous one, as the current one will have too much perf overhead when you only want to use a few of them. If we want to go this style, we could actually merge this to `useMagicKeys` and use `getModifierState` based on the modifier lists (as `useMagicKeys` uses Proxy, the refs are generated on-demanded.

JMaylor:
Okay, thanks for the feedback.
I added a `Modifier` type literal based on the 12 non-deprecated key modifiers from the [docs](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState#browser_compatibility).

1 other thought: Mouse Events also have the `getModifierState` method, so we could add some of `mousedown`, `mouseup`, and `mousemove` to the event listeners used to update the state. Would adding these have much performance impact? If not, it would make the responsiveness a lot better.

antfu:
We could keep `mousedown` and `mouseup` for that by default and also expose them to be configurable. Refer to https://github.com/vueuse/vueuse/blob/68c7dc0987fe8d9b875b4c97ed27f47393564e97/packages/core/useIdle/index.ts

JMaylor:
Cool - hopefully that's implemented OK.

Other thoughts for enhancements could be:
1. Do we want user to be able to track multiple with one function call? i.e.
```ts
const [ capsLockState, shiftState ] = useKeyModifierState([ 'CapsLock' , 'Shift' ])
```
2. Option for setting initial state value? Currently initialised as null

packages/core/useKeyModifierState/index.ts
```diff
@@ -1,13 +1,25 @@
...
+}
+
+export function useKeyModifierState(modifier: Modifier, options: ModifierOptions = {}) {
+  const {
+    events = defaultEvents,
+  } = options
   const state = ref<null | boolean>(null)
-  listenerEvents.forEach((listenerEvent) => {
+  events.forEach((listenerEvent) => {
     useEventListener(document, listenerEvent, (evt: KeyboardEvent) => {
```
antfu, line 23: We'd better make `document` configurable as well. See https://vueuse.org/guidelines.html#configurable-globals

packages/core/useKeyModifierState/index.ts
```diff
@@ -1,24 +1,35 @@
...
 }
 
 export function useKeyModifierState(modifier: Modifier, options: ModifierOptions = {}) {
   const {
     events = defaultEvents,
+    document = defaultDocument,
+    initial = null,
   } = options
-  const state = ref<null | boolean>(null)
+  const state = ref(initial)
```
antfu, line 32: I suppose this will make the ref lose the type of `boolean`, no?

packages/core/useKeyModifierState/index.ts
```diff
@@ -1,24 +1,35 @@
...
   events.forEach((listenerEvent) => {
```
antfu, line 33: `if (docuement) { ...` for SSR compatible.

packages/core/useKeyModifierState/index.ts
JMaylor: [e799825](https://github.com/vueuse/vueuse/pull/698/commits/e79982510090845f78149310ce026eb29451c61a)

packages/core/useKeyModifierState/index.ts
JMaylor: [ebfaec5](https://github.com/vueuse/vueuse/pull/698/commits/ebfaec50d6c97f8abc26b2409e4675d088e9b5f3)

antfu:
I have renamed it to `useKeyModifier` for simplicity and improved the types a bit, thanks for your contributions!

antfu (approved):

PR #705, opened 2021-08-28
feat(useWebSocket): Added support for optional protocols

itpropro:
Added support for an optional array parameter called protocols. This second parameter to the WebSocket constructor contains one or more protocol strings to use e.g. subprotocols. If not provided, internally the parameter always defaults to an empty array (`[]`), so that was chosen as parameter initializer. More information can be found [here](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket).

### DIFF
--- a/packages/core/useWebSocket/index.ts
+++ b/packages/core/useWebSocket/index.ts
@@ -62,6 +62,13 @@ export interface WebSocketOptions {
    * @default true
    */
   immediate?: boolean
+
+  /**
+   * List of one or more sub-protocol strings
+   *
+   * @default []
+   */
+  protocols?: string[]
 }
 
 export interface WebSocketResult<T> {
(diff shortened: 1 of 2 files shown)


packages/core/useWebSocket/index.ts
```diff
@@ -113,10 +113,12 @@ function resolveNestedOptions<T>(options: T | true): T {
  *
  * @see https://vueuse.org/useWebSocket
  * @param url
+ * @param protocols
  */
 export function useWebSocket<Data = any>(
   url: string,
   options: WebSocketOptions = {},
+  protocols: string[] = [],
```
antfu, line 121: I would like to have this part of the options for future extensibility

itpropro: Thought about that. If you think that the WebSocket Class will be extended in the future, I add it to the options.

itpropro: Moved to WebSocketOptions

antfu (approved):

itpropro:
Thanks for accepting 😊

PR #708, opened 2021-09-01
feat(useResize): new function

sibbng:
Todos:

- [x] Options (disable edges, configurable window, max/min size)
- [x] Control (enable/disable resize functionality, let resizing part to events)
- [x] Events (emit `onResizeStart`, `onResizeMove`, `onResizeEnd`)
- [x] Edges have to be on foreground `Document.elementFromPoint()` and take border radius into account on corners
- [x] Refactor event listeners
- [x] Manual mode
- [x] Watch `width` and `height` refs to resize programmatically
- [ ] `position: absolute` support
- [ ] Clean code
- [ ] Component version
- [ ] Docs


packages/core/useResize/index.ts
```diff
@@ -0,0 +1,336 @@
...
+      && isEdgeActive('right')
+      && isOnForeground(right, currentY)
+    )
+      setCursorAndDirection('ew-resize', 'right')
+
+    else
+      setCursorAndDirection('', '', '')
+  }
+
+  function setSize(width: number, height: number) {
```
wheatjs, line 305: I believe we should leave this part up the user and just return the width and height like we do with `useDraggable`. Maybe we can also return a style string like `useDraggable`

packages/core/useResize/index.ts
```diff
@@ -0,0 +1,336 @@
...
+  const direction = ref('')
+
+  const pointer = reactive({ startX: 0, startY: 0, currentX: 0, currentY: 0 })
+
+  const widthRef = ref(0)
+  const heightRef = ref(0)
+
+  let cleanup: Fn[] = []
+
+  const start = () => {
```
wheatjs, line 73: I think we can get rid of the start function and cleanup array. We can move the useEventListener to the top level as well as everything else. This was actually the source of a bug I ran into. When I had two resizable elements on the page and removed one, then the other one would lose its event handlers as well for some reason. Not sure the source of this, but moving them to the top level fixed this.

packages/core/useResize/index.ts
```diff
@@ -0,0 +1,336 @@
...
+    const element = window!.document.elementFromPoint(x, y)
+    return isPathIncludesTarget.value || (element && target.value!.contains(element)) || (corner && getCorner(x, y))
+  }
+
+  const isEdgeActive = (edge: Edges) => {
+    return unref(edges).includes(edge)
+  }
+
+  function handlePointer({ currentX, currentY }: typeof pointer) {
+    if (isResizing.value || isOutside.value)
```
wheatjs, line 215: Should also check if `target.value` exists here since `useEventListener` will be on the top level

packages/core/useResize/index.ts
```diff
@@ -0,0 +1,336 @@
...
+
+  const stop = () => {
+    cleanup.forEach(fn => fn())
+    cleanup = []
+    isActive.value = false
+  }
+
+  watch(target, (value) => {
+    if (value) {
+      start()
```
wheatjs, line 94: Remove

packages/core/useResize/index.ts
```diff
@@ -0,0 +1,336 @@
...
+    }
+    else {
+      stop()
+    }
+  })
+
+  const setCursorAndDirection = (setCursor = '', setDirection = '', setTouchAction = 'none') => {
+    direction.value = setDirection
+    isOverEdge.value = !!setDirection
+    !disableCursor && window!.document.body.style.setProperty('cursor', setCursor)
```
wheatjs, line 109: Should be `!unref(disableCursor)`

packages/core/useResize/index.ts
wheatjs: We will also want to wrap the events in 
```
if (isClient) {
  ...
}
```

packages/core/useResize/index.ts
```diff
@@ -0,0 +1,336 @@
...
+    else
+      setCursorAndDirection('', '', '')
+  }
+
+  function setSize(width: number, height: number) {
+    target.value!.style.width = `${clamp(width, unref(minWidth) as number, unref(maxWidth) as number)}px`
+    target.value!.style.height = `${clamp(height, unref(minHeight) as number, unref(maxHeight) as number)}px`
+  }
+
+  function useElementSize() {
```
wheatjs, line 310: This can probably be moved to the top level as well so we don't have to manually clean it up

packages/core/useResize/index.ts
```diff
@@ -0,0 +1,336 @@
...
+    isActive,
+    isOverEdge,
+    isResizing,
+    onResizeStart: onResizeStart.on,
+    onResizeMove: onResizeMove.on,
+    onResizeEnd: onResizeEnd.on,
+  }
+}
+
+function clamp(value: number, min: number, max: number) {
```
wheatjs, line 334: This can be replaced with the existing clamp function here https://github.com/vueuse/vueuse/blob/499a6987128e3db6862f57bcc189095d6c501a67/packages/shared/utils/is.ts#L18

wheatjs (changes requested):
@sibbng Used this composable recently and have some suggestions based upon my experience using it

packages/core/useResize/index.ts
sibbng: This is definitely a bug. I will look into it.

packages/core/useResize/index.ts
sibbng: Exactly which line and why? 🤔

sibbng:
@wheatjs Thanks for your feedback. I pushed my latest "dirty updates". 

I use this function actively on my Tailwind playground. Soon I will update demo.vue with all intended use cases. Then we can make some cleanups and bug fixes.

wheatjs:
Sounds good!

packages/core/useResize/index.ts
```diff
@@ -198,16 +198,16 @@ export function useResize(element: MaybeElementRef, options: UseResizeOptions =
 
     const { left, top } = target.value!.getBoundingClientRect()
 
-    widthRef.value = clamp(newWidth, unref(minWidth as number), unref(maxWidth as number))
-    heightRef.value = clamp(newHeight, unref(minHeight as number), unref(maxHeight as number))
+    widthRef.value = clamp(newWidth, unref(Number(minWidth)), unref(Number(maxWidth)))
```
wheatjs, line 201: I believe you will want `Number` outside the `unref` like `Number(unref(minWidth))`. Otherwise anytime `minWidth` is a  ref it will return NaN

CodyBontecou:
Any update on this?

sibbng:
Hi, I've decided not to work on this function. The initial goal was to make an element resizable without the resize handlers in DOM. It worked, however, for such a utility the range of use cases is really big. That means it may not satisfy most people's needs in its current form and I don't have time to work on that. Even if we cover all the use cases, I don't think massive utilities are in the scope of VueUse. It's better for them to be a standalone library. So, I'm closing this PR now. 

If you are thinking to use this function in your projects, check out my implementation at UnoCSS Playground:
-  [Preview](https://uno.antfu.me/play/?html=DwEwlgbgBAFgtAMwK4BsVQC4FMAeG4DGWAdtgE5QIq5QDOW1B%2BxA9sVlAIZoBcGZnYrTAYwbOABYADFIB8AKChRQkKAFtOCpUpXRseOAFYc6BAHcAjDK7EwG7HABGLJMSJxuGG3c4OCL0jgwYgRgkQ5BHwcQJAFRcQtaLW0lVxYCWlpFbWAAenAIZJ0CqBYABwBmKUxcfBQAc0ozKuq1CyLtABUYDmDaDEEvcRAsDWIQKABBDBY1MAIoAGEAZWWoEnrgrAA6bJ18yA7ddQAmSmocKAArJH6wBABPQhJyGoMTk1LK6pgWCCwyABeABE5QAHFJgR0dJw9ikoGBCJwyM5iHAUCx6iw4JsMDAkI44Sl9Phgj0yCIidoYGQsAgQTAMBgyrQeLlcrj8Y5tv41Lk0hlaPzWILgVSlAMyPUsBgQQB9RwoQQAazF8KgsjymjheQKyV1h3kBsKRpKnEctBYKCQ2CgziZsyMUAp9UZcGq1AQ%2BGqJOepABXyqTRayWGo0EEwA7QihANSFBo8g0A8oARbjM5gAvc3UI0HQpAA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAFVcESgF84WKBBBwA5FxUS6dPUiQYAHpFjpsAQy4AbeJhz4iJYKQAUyVFDO9ABccADaVKhhehJcdDA6ALSBThh65IhwUk7QoXq0aHpwagC6VCWydAAW0DCxMHSh-qj6sfFJ1bX1enkwVjCJTqRwfZYD0Pa4pBhwNQBuGFDBowN99k5mqGqyIgqN4ZFyokoqPgCUssLyjCxsnDx8-OeXRwqS0n6HqHQS6xihAEYAHQAJheqAkaFweSqrDAjQA9Aj6CAgdUEZsWmoLuUkNikEA&options=N4IgTgpgzgDg9gOygSwG4RALgC5gK4QA0IA7sgCbYAWWAzAOwCsxVEyA5ldlvQAwBsAXyA) 
- [Playground.vue](https://github.com/unocss/unocss/blob/71a5f9843af80965604819b51f9935237eac1438/playground/src/components/Preview.vue) 
- [useResize.ts](https://github.com/unocss/unocss/blob/71a5f9843af80965604819b51f9935237eac1438/playground/src/useResize.ts)

thedamon:
Totally came to vueuse thinking this might be a thing they'd have as a beautiful complement to useDraggable. Looking at the code it is less straightforward than i might have thought! appreciate you looking into this.

PR #709, opened 2021-09-02
feat(tryOnBeforeUnmount): new function

xizher:
a hook like tryOnMounted;
in GIS System, sth. like the map restore or locate need to be used in before unmount when we want to destory a context


packages/shared/tryOnBeforeUnmount/index.md
```diff
@@ -0,0 +1,37 @@
+---
+category: Component
+---
+
+# tryOnUnmounted
```
DarrylNoakes, line 5: Was this supposed to be renamed too?

xizher: because tryOnBeforeUnmount use onBeforeUnmount, not onUnmounted

PR #723, opened 2021-09-07
feat(useAsyncState): Provide params to callback function

avfrancev:
It would be cool to be able to pass params to promise from `execute` function 

```javascript
const { state, isReady, execute } = useAsyncState(
  (args) => {
    let id = args?.id || 1
    return axios.get(`https://jsonplaceholder.typicode.com/todos/${id}`).then(t => t.data )
  },
  {},
  {
    delay: 2000,
    resetOnExecute: false,
  },
)
</script>

<template>
  <div>
    <note>Ready: {{ isReady.toString() }}</note>
    <pre lang="json" class="ml-2">{{ YAML.dump(state) }}</pre>
    <button @click="execute(2000, {id: 2})">
      Execute
    </button>
  </div>
</template>
```

### DIFF
--- a/packages/core/useAsyncState/index.ts
+++ b/packages/core/useAsyncState/index.ts
@@ -46,7 +46,7 @@ export interface AsyncStateOptions {
  * @param options
  */
 export function useAsyncState<T>(
-  promise: Promise<T> | (() => Promise<T>),
+  promise: Promise<T> | ((...args: any[]) => Promise<T>),
   initialState: T,
   options: AsyncStateOptions = {},
 ) {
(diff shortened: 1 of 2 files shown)


packages/core/useAsyncState/index.ts
```diff
@@ -61,18 +61,18 @@ export function useAsyncState<T>(
   const isReady = ref(false)
   const error = ref<unknown | undefined>(undefined)
 
-  async function execute(delay = 0) {
+  async function execute(delay = 0, ...args: any[]) {
     if (resetOnExecute)
       state.value = initialState
-
+
```
antfu, line 67:
```suggestion
```suggestion

```
```

Issue #736, opened 2021-09-10
feat: useScroll


webfansplz:
i have an idea , It provides some states for scroll event , such as :

```ts

const { x, y, scrolling, finished, arrivedStatus, ...} = useScroll(element , () => {...}, {...options} )

const { left, right, top, bottom } = arrivedStatus
```

antfu:
Sounds cool, PR welcome

webfansplz:
PR has merged

PR #739, opened 2021-09-11
feat(useScroll): new Function

webfansplz:
#736 

@antfu 

It's doesn't support the `window` scroll, because I think it's too much for it (Maybe it would be better to support it in `useWindowScroll` ) 

Do you think need to do compatibility for `window` ? 

If don't need, Maybe it's better to rename it `useElementScroll`？


antfu:
The function looks cool, thanks!

What's the problem of supporting window?

webfansplz:
The main concern is browser compatibility, But i added simple support for window,make sure it works in the standard browsers.  (I tested Chrome, Firefox, and Safari. They work well) 😊

```ts
const eventTarget = (e.target === document ? (e.target as Document).documentElement : e.target) as HTMLElement
```

packages/core/useScroll/index.ts
```diff
@@ -0,0 +1,121 @@
...
+   *
+   */
+  onStop?: () => void
+
+  /**
+   * Enabled direction for scroll.
+   *
+   * @default ['x','y']
+   */
+  enabledDirection?: Array<'x'|'y'>
```
antfu, line 33: When will this be used? Or I think we can just do it for both without the config, there should be much perf difference

packages/core/useScroll/index.ts
```diff
@@ -0,0 +1,121 @@
...
+ * Reactive scroll.
+ *
+ * @see https://vueuse.org/useScroll
+ * @param {MaybeElementRef} element
+ * @param {() => void} [handler=noop]
+ * @param {UseScrollOptions} [options={}]
+ */
+export function useScroll(
+  element: MaybeElementRef,
+  handler: () => void = noop,
```
antfu, line 53: Let's have this inside the option as `onScroll`

packages/core/useScroll/index.ts
```diff
@@ -0,0 +1,121 @@
...
+   */
+  throttle?: number
+
+  /**
+   * The check time when scrolling ends.
+   * This configuration will be setting to (throttle + checkStopTime) when the `throttle` is configured.
+   *
+   * @default 200
+   */
+  checkStopTime?: number
```
antfu, line 20:
```suggestion
  idle?: number
```

packages/core/useScroll/index.md
```diff
@@ -0,0 +1,27 @@
+---
+category: Sensors
+---
+
+# useScroll
+
+Reactive scroll
```
antfu, line 7:
```suggestion
Reactive scroll position and state
```

packages/core/useScroll/index.md
```diff
@@ -0,0 +1,27 @@
...
+</template>
+
+<script>
+  setup() {
+    import { useScroll } from '@vueuse/core'
+    const scrollRef = ref<HTMLElement | null>(null)
+    const { x, y, scrolling, finished, arrivedStatus } = useScroll(scrollRef)
+    return { x, y, scrolling, finished, arrivedStatus  }
+  } 
+</script>
```
antfu, line 24:
```suggestion
<script setup>
import { useScroll } from '@vueuse/core'

const scrollRef = ref<HTMLElement | null>(null)

const { x, y, scrolling, finished, arrivedStatus } = useScroll(scrollRef)
</script>
```

And let's move this before `<template>`

packages/core/useScroll/index.md
```diff
@@ -14,14 +14,11 @@ Reactive scroll
...
 
-<script>
-  setup() {
-    import { useScroll } from '@vueuse/core'
-    const scrollRef = ref<HTMLElement | null>(null)
-    const { x, y, scrolling, finished, arrivedStatus } = useScroll(scrollRef)
-    return { x, y, scrolling, finished, arrivedStatus  }
-  } 
+<script setup>
+	import { useScroll } from '@vueuse/core'
```
antfu, line 18: Let's use 2 spaces instead of tabs.

packages/core/useScroll/index.ts
webfansplz: By default, we do both. i think we should expose this config and let ppl setting it  to save perf overhead

antfu: I don't see there is any overhead on that. It's just variable assignment, and we don't have extra event listener.

webfansplz: Okay,Do you suggest that we remove this config and enable both by default ?

webfansplz: removed~

PR #740, opened 2021-09-12
feat(useRefHistory): new option `eventFilter`, new functions `useThrottledRefHistory` and `useDebouncedRefHistory`

harmyderoman:


packages/core/useRefHistory/index.ts
```diff
@@ -141,21 +141,23 @@ export interface UseRefHistoryReturn<Raw, Serialized> {
  * @see https://vueuse.org/useRefHistory
  * @param source
  * @param options
+ * @param eventFilter optional
  */
 export function useRefHistory<Raw, Serialized = Raw>(
   source: Ref<Raw>,
   options: UseRefHistoryOptions<Raw, Serialized> = {},
+  eventFilter: EventFilter = bypassFilter,
```
antfu, line 149: We should have the `eventFilter` inside the options.

packages/core/useThrottledRefHistory/index.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @see https://vueuse.org/useThrottledRefHistory
+ * @param source
+ * @param options
+ * @param ms
+ */
+export function useThrottledRefHistory<Raw, Serialized = Raw>(
+  source: Ref<Raw>,
+  options: UseRefHistoryOptions<Raw, Serialized> = {},
+  ms: MaybeRef<number>,
```
antfu, line 16: Same here, we could have the `ms` renamed to `throttle` and put inside of the options. Could have the type like: `options: Omit<UseRefHistoryOptions<Raw, Serialized>, 'filter'> & { throttle? MaybeRef<number> }`

packages/core/useThrottledRefHistory/demo.vue
```diff
@@ -0,0 +1,41 @@
...
+import { useCounter } from '@vueuse/shared'
+import dayjs from 'dayjs'
+import { Ref } from 'vue-demi'
+import { useThrottledRefHistory } from '.'
+
+const format = (ts: number) => dayjs(ts).format()
+const delay: Ref<number> = ref(1000)
+
+const { count, inc, dec } = useCounter()
+const { history, undo, redo, canUndo, canRedo } = useThrottledRefHistory(count, { capacity: 10 }, delay.value)
```
antfu, line 12:
```suggestion
const { history, undo, redo, canUndo, canRedo } = useThrottledRefHistory(count, { capacity: 10 }, delay)
```

antfu (changes requested):

packages/core/useRefHistory/index.ts
harmyderoman: updated

packages/core/useThrottledRefHistory/index.ts
harmyderoman: updated

packages/core/useRefHistory/index.ts
```diff
@@ -42,6 +42,10 @@ export interface UseRefHistoryOptions<Raw, Serialized = Raw> {
    * Deserialize data from the history
    */
   parse?: (v: Serialized) => Raw
+  /**
+   * Can get throttleFilter and debounceFilter
+   */
+  filter?: EventFilter
```
patak-js, line 48: This should be called `eventFilter`, no?

harmyderoman: Every other options is only one word, so eventFilter will contrast with them. But if you want, I can change it.

patak-js: `filter` is a good name looking only at `useRefHistory`, and if I recall correctly it was the name we used at one point but it was later replaced by `eventFilter` everywhere because it was more explicit. You can see the interface used in other composables (like `useStorage`, `useMouse`, ...) here https://github.com/vueuse/vueuse/blob/d6dd1ba2d5fab674771103f0955e631c6c8f436e/packages/shared/utils/filters.ts#L17
So it is better that we align the naming for the event filter with the rest of VueUse.

harmyderoman: Naming is the hardest part of programming😃

patak-js: Indeed 🙂. At least in this case the choice was already made by Anthony 😅

harmyderoman: Anthony also wrote me this declaration
`options: Omit<UseRefHistoryOptions<Raw, Serialized>, **'filter'**> & { throttle? MaybeRef<number> }`
so it confused me

harmyderoman: @patak-js updated

packages/functions.md
```diff
@@ -128,12 +128,14 @@
   - [`useAsyncState`](https://vueuse.org/core/useAsyncState/) — reactive async state
   - [`useCounter`](https://vueuse.org/shared/useCounter/) — basic counter with utility functions
   - [`useDebounce`](https://vueuse.org/shared/useDebounce/) — debounce execution of a ref value
+  - [`useDebouncedRefHistory`](https://vueuse.org/core/useDebouncedRefHistory/) — same as [useRefHistory](https://vueuse.org/core/useRefHistory/) but with debounse effect
```
patak-js, line 131:
```suggestion
  - [`useDebouncedRefHistory`](https://vueuse.org/core/useDebouncedRefHistory/) — same as [useRefHistory](https://vueuse.org/core/useRefHistory/) but with debounce effect
```

packages/core/useRefHistory/index.ts
antfu: Yeah, my bad. Would like to have `eventFilter` for it.

Issue #719, opened 2021-09-06
useSpeechRecognition: lang option can be reactive


jlucaso1:
To make the language option dynamic like in the [Chrome Demo](https://www.google.com/intl/pt/chrome/demos/speech.html).

PR #741, opened 2021-09-12
fix(useSpeechRecognition): reactive speech language

JMaylor:
Basic attempt at #719
I will update the demo another day so this uses something more appropriate for the non-English languages. For now you can use the demo to see that the language does change.


packages/core/useSpeechRecognition/index.ts
```diff
@@ -24,7 +24,7 @@ export interface SpeechRecognitionOptions extends ConfigurableWindow {
    *
    * @default 'en-US'
    */
-  lang?: string
+  lang?: string | Ref
```
wheatjs, line 27: Use `MaybeRef<string>` instead

packages/core/useSpeechRecognition/index.ts
```diff
@@ -69,12 +69,14 @@ export function useSpeechRecognition(options: SpeechRecognitionOptions = {}) {
 
     recognition.continuous = continuous
     recognition.interimResults = interimResults
-    recognition.lang = lang
+    recognition.lang = isRef(lang) ? lang.value : lang
```
wheatjs, line 72: Use `unref(lang)` instead

packages/core/useSpeechRecognition/index.ts
```diff
@@ -94,6 +96,7 @@ export function useSpeechRecognition(options: SpeechRecognitionOptions = {}) {
 
     recognition.onend = () => {
       isListening.value = false
+      if (isRef(lang) && recognition) recognition.lang = lang.value
```
wheatjs, line 99:
```js
if (recognition)
  recognition.lang = unref(lang)
```

wheatjs (changes requested):

packages/core/useSpeechRecognition/index.ts
JMaylor: For this - we only want to update `recognition.lang` if a ref is passed in - no point doing this if a plain string is used?

wheatjs: 👍

packages/core/useSpeechRecognition/index.md
```diff
@@ -34,3 +34,27 @@ The following shows the default values of the options, they will be directly pas
   continuous: true,
 }
 ```
+
+You can also bind the lang property to a ref object. Note that updating this ref value cannot take effect until the next time the recognition is stopped.
```
antfu, line 38: Have arguments accepts ref is a common practice in VueUse, I don't think we need to explicitly document it here. (but thanks!)

packages/core/useSpeechRecognition/index.ts
```diff
@@ -69,12 +69,14 @@ export function useSpeechRecognition(options: SpeechRecognitionOptions = {}) {
...
     recognition.continuous = continuous
     recognition.interimResults = interimResults
-    recognition.lang = lang
+    recognition.lang = unref(lang)
 
     recognition.onstart = () => {
       isFinal.value = false
     }
 
+    if (isRef(lang)) watch(lang, (lang) => { if (recognition && !isListening.value) recognition.lang = lang })
```
antfu, line 78: Can you format L78 and L99 a bit for better formatting?

PR #742, opened 2021-09-12
feat: added 'as' prop to renderable components

wheatjs:
Adds `as` prop to component that needs to render an element(`OnClickOutside` for example).


packages/core/onClickOutside/component.ts
```diff
@@ -1,8 +1,10 @@
 import { h, ref, defineComponent } from 'vue-demi'
 import { onClickOutside } from '@vueuse/core'
+import { RenderableComponent } from '@vueuse/core/_types'
```
antfu, line 3:
```suggestion
import { RenderableComponent } from '../types'
```

Please pull the latest branch, it's renamed to `types`

antfu (approved):

PR #744, opened 2021-09-14
feat(useDrauu): New function

wheatjs:
A slightly easier to use and more reactive function for drauu

### DIFF
--- a/packages/integrations/useDrauu/index.ts
+++ b/packages/integrations/useDrauu/index.ts
@@ -1,0 +1,135 @@
+import { ref, watch, reactive } from 'vue-demi'
+import { createDrauu, Drauu, Options, Brush } from 'drauu'
+import { MaybeElementRef, unrefElement, createEventHook } from '@vueuse/core'
+import { tryOnScopeDispose, Fn } from '@vueuse/shared'
+
+export type UseDrauuOptions = Omit<Options, 'el'>
+
+/**
+ * Reactive drauu
+ *
+ * @see https://vueuse.org/useDrauu
+ * @param target The target svg element
+ * @param options Drauu Options
+ */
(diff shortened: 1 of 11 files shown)


packages/integrations/useDrauu/index.ts
```diff
@@ -0,0 +1,122 @@
...
+  const onCanceledHook = createEventHook<void>()
+  const onComittedHook = createEventHook<void>()
+  const onStartHook = createEventHook<void>()
+  const onEndHook = createEventHook<void>()
+  const canUndo = ref(false)
+  const canRedo = ref(false)
+  const altPressed = ref(false)
+  const shiftPressed = ref(false)
+
+  const brush = {
```
antfu, line 30: I feel like `brush `itself might be better to be a ref (so you can swap the entire brush)

wheatjs: Hmm yeah I was having trouble figuring which way it should be. Because right now it is easy to hook up to a v-model for things like the brush size, but we aren't able to swap out brushes so I'm not sure what the best solution is.

wheatjs:
Sorry about the messy commits, still trying to sort things out. Seems like it includes stuff from my other PR 🥴

packages/integrations/useDrauu/index.ts
antfu: We have an extended `toRefs` utils that accept using passing an object ref. We could document that.

wheatjs: Sounds good, will update

wheatjs: Ok I updated the function to take a ref for the brush. I also created a new function that returns a reactive object with the brush properties so it is a bit easier to create brushes, let me know what you think of that. It worked out well in the demo and I was able to avoid using something like `toRefs`

PR #761, opened 2021-09-16
default use bubbling ...

zhangenming:
I don't know if there were other considerations


antfu:
I think we shouldn't implicitly include this default. Giving the web's default is not `true` so either we.

Issue #762, opened 2021-09-16
feat: watchAtMost


webfansplz:
i have an idea :
```ts

const watchAtMost = (source, fn, options = { max = 1 }) => {
  const { max } = options
  const count = ref(0)
  const stop = watch(source, (...args) => {
    fn(...args)
    count.value >= max && stop()
    count.value++
  }, options)
  return { count , stop }
}
```

harmyderoman:
why does the function not return `{ count, stop }`?

webfansplz:
> why does the function not return `{ count, stop }`?

good idea ～

harmyderoman:
Also, count in return must be readonly, so you can't change it outside the function.

webfansplz:
> Also, the `count` in return must be read-only, so you will not be able to change it outside the function.

Thanks for your suggest.

harmyderoman:
You are wellcome!

harmyderoman:
A good idea, but I think antfu will say that it is better to write a count filter so that it can then be applied to a whole family of functions.

webfansplz:
> A good idea, but I think antfu will say that it is better to write a count filter so that it can then be applied to a whole family of functions.

nice suggest ! I will do so if this idea is approved.

webfansplz:
PR has merged

PR #764, opened 2021-09-16
feat(watchAtMost): new function

webfansplz:
see #762


antfu:
Looks interesting, but can you share a bit context of why this is useful?

webfansplz:
My initial idea was watchOnce (I use it a lot in my projects，but with watch i need to manually call the stop function). i use watchWithCount instead of watchOnce that because i think it's more flexible.

patak-js:
Maybe `watchN(4, ...)` or `watchAtMost(4, ...)` to be closer to `watchOnce`. A `watchUntil(ms,...)` may also be interesting 🤔

patak-js:
Do you have a concrete example where you used this @webfansplz? I'm also curious about your real use cases

webfansplz:
```ts

const { stop } = watch(
                props.response!,
                () => {
                    nextTick(() => {
                        const floor = route.query.floor
                        onToggleActiveFloor(Number(floor))
                        stop()
                    })
                },
                { immediate: true },
            )

```
This is one of the scenes: I need to listen for data responses and do UI initialization with route parameters.

webfansplz:
> Maybe `watchN(4, ...)` or `watchAtMost(4, ...)` to be closer to `watchOnce`. A `watchUntil(ms,...)` may also be interesting 🤔

i agree with you. maybe `watchAtMost(4, ...)` is more appropriate ~

packages/shared/watchAtMost/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  options: WatchAtMostOptions<Immediate> = { max: 1 },
+): WatchAtMostReturn {
+  const {
+    max,
+    ...watchOptions
+  } = options
+  const { count, filter } = countFilter(max)
+  const stop = watch(
+    source,
+    createFilterWrapper(filter, (...args) => {
```
antfu, line 32: I don't think the filter is needed here (a bit overengineered from my point of view. Simply count and stop will be good enough.

antfu (changes requested):
I think it would be great to also include a shorthand `watchOnce` that uses this function under the hood

antfu:
I will made the changes directly

webfansplz:
@antfu thanks~

Issue #766, opened 2021-09-18
Feature: Random string generation utility


gkweb:
## Problem: 

Sometimes it's really handy to randomly generate ID's on elements especially when using semantic or aria related elements. Therefore a random string generation utility would be very helpful. 

## Solution: 

See PR I'll open soon. 

Thank you for your consideration.

antfu:
Sorry, but I don't think this is strongly bound to Vue's reactivity system, as the random string generators are often one-off functions, which reactivity does not really make sense. Can you provide more specific use cases?

gkweb:
> Sorry, but I don't think this is strongly bound to Vue's reactivity system, as the random string generators are often one-off functions, which reactivity does not really make sense. Can you provide more specific use cases?

@antfu

See here for a use case. `aria-labelledby` and `aria-controls` are using the random generator for ID's

https://stackblitz.com/edit/vue-hmmusw?file=src/components/Accordion.vue

antfu:
Why don't you just generate a string for that? I still didn't see the reactiveness use needed here.

```html
<template>
  <div id="accordionGroup" class="Accordion">
    <h3>
      <button :aria-expanded="isVisible ? 'true' : 'false'"
              class="Accordion-trigger"
              :aria-controls="contentId"
              :id="labelId" @click.prevent="isVisible = !isVisible">
        <span class="Accordion-title">
          <slot name="title"/>
        </span>
      </button>
    </h3>
    <div :id="contentId"
         role="region"
         :aria-labelledby="labelId"
         class="Accordion-panel" :hidden="isVisible ? null : 'true'">
      <div>
      <slot name="content" />
      </div>
    </div>
  </div>
</div>
</template>

<script setup>
const labelId = randomString()
const contentId = randomString()
</script>
```

antfu:
Sorry, I am gonna reject this. Thanks for the contributions! (you did great for the PR)

PR #767, opened 2021-09-18
Random string generation useRandomString #766

gkweb:
#766 

Adds random string generation capability in a composable function.

**Features:** 
- Ability to regenerate with returned function
- Ability to set prefix
- Ability to set character length


antfu:
As commented in #766

Issue #681, opened 2021-08-18
onClickOutside presents event management difficulties


Aaron-Pool:
I've recently that discovered that, since the onClickOutside listener does not allow you to target the basic `click` event, it creates issues with potentially interrupting the event.

Reproduction Example: https://rliqq.csb.app/

Imagine that the square is a content section for the user to interact with. I have a button to open the section, and then you can click off the section to collapse it. But, just in case this isn't obvious to the user I make the button a open/close toggle. So that the user can click off to close (if that is intuitive to them) _or_ the user can intentionally toggle the square closed.

Well, if you try to toggle it closed, both the clickOff and the click event fires and it starts to close and then reopens. If the onClickOutside usable used the `click` event, I could just handle this with some event phase/stop propagation settings. But since I can't use `click` they operate on a different even lifecycle and the two events don't interact.

Is there a reason the typing doesn't include `click` as a targetable event?

sibbng:
As a author of that function adding that event was in my mind for a while. I just missed it on my initial commit and didn't want to rush to add it before seeing real world use cases. Thanks for demo 👌.

Issue #323, opened 2021-02-05
onClickOutside add event pointerdown


cereschen:
When using a module like Threejs, clicking on the screen only triggers a pointerdown event

antfu:
Can't really help without a reproduction.

cereschen:
https://codesandbox.io/s/elastic-architecture-mp7n0?file=/src/components/HelloWorld.vue

here is  the reproduction

@antfu

PR #770, opened 2021-09-18
feat(onClickOutside): scrollbar and selection support

### LINKED ISSUE #641: onClickOutside do not trigger on scrollbar click
Using this lovely library for a modal system, where i want to close the modal if you click on the backdrop. However, targeting the inner content with: onClickOutside does so it closes if i click the vertical scrollbar, incase the content overflows.

Would it be possible to disable that?

Atm ive hack-fixed it with:

```
        const content = ref(null);

        onClickOutside(content, (event) => {
            if (event.clientX < document.documentElement.clientWidth - SCROLLBARWIDTH) {
                props.dialog.close();
            }
        });
```

sibbng:
Fix #641

### DIFF
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -48,9 +48,8 @@ export function onClickOutside<E extends keyof OnClickOutsideEvents = 'pointerup
   if (isUpEvent) {
     cleanup.push(
       useEventListener(window, 'pointerdown', () => (shouldListen.value = true), { passive: true }),
-      // `pointermove` event will be fired after `pointerup` event if pointer coordinates are changed during `pointerdown`
-      // We using this for 2 things: to prevent calling handler on scrolls and and selections (including inputs)
-      useEventListener(window, 'pointermove', e => (shouldListen.value = !e.movementX && !e.movementY), { passive: true }),
+      useEventListener(window, 'pointermove', () => (shouldListen.value = false), { passive: true }),
+      useEventListener(window, 'scroll', () => (shouldListen.value = false), { passive: true }),
     )
   }
 
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -1,9 +1,10 @@
+import { ref } from 'vue-demi'
 import { MaybeElementRef, unrefElement } from '../unrefElement'
 import { useEventListener } from '../useEventListener'
 import { ConfigurableWindow, defaultWindow } from '../_configurable'
 
 export type OnClickOutsideEvents = Pick<WindowEventMap, 'click' | 'mousedown' | 'mouseup' | 'touchstart' | 'touchend' | 'pointerdown' | 'pointerup'>
-export interface OnClickOutsideOptions<E extends keyof OnClickOutsideEvents> extends ConfigurableWindow {
+export interface OnClickOutsideOptions<E extends keyof OnClickOutsideEvents = 'pointerup'> extends ConfigurableWindow {
   event?: E
 }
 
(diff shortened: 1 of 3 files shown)


sibbng:
Currently there is a bug with click event even without these changes,

Set modal or dropdown event to `{ event: 'click' }`. It instantly runs handler function with `PointerEvent` 😕

sibbng:
> Currently there is a bug with click event even without these changes,

While debugging this bug I saw different behaviors between dev builds and prod build for same code. I don't know what was wrong there but it seems working as expected right now.

sibbng:
I realized click event is not firing already if we are on scrollbar (at least on Chrome & Firefox, idk what Safari does). 

So, click event wasn't needed to be included in `/(click|up|end)/` regex.

But at this time I realized another edge case. If selection starts and it ends outside, it is firing event handler. This causes bad user experience, especially if end user selecting the text inside an input. To prevent this I implemented `selectionstart` event, but it didn't work with input. Then I switched `pointermove` event, this is also eliminated the necessity of `scroll` event listener.

`pointermove` has special behavior, and I documented this code. I didn't test it in Safari. It would be nice if we confirm it works as expected in Safari.

packages/core/onClickOutside/index.ts
```diff
@@ -15,26 +16,44 @@ export interface OnClickOutsideOptions<E extends keyof OnClickOutsideEvents> ext
...
-  return useEventListener(window, event, listener, { passive: true })
+  const cleanup = [
+    useEventListener(window, event, listener, { passive: true }),
+  ]
+
+  if (isUpEvent) {
+    cleanup.push(
+      useEventListener(window, 'pointerdown', () => (shouldListen.value = true), { passive: true }),
+      useEventListener(window, 'pointermove', () => (shouldListen.value = false), { passive: true }),
+      useEventListener(window, 'scroll', () => (shouldListen.value = false), { passive: true }),
```
antfu, line 52: I don't really feel good about this solution. For solving the original case #641, could we just note the caveat and ask users to wrap a parent div and use scroll in the child if they absolutely need that? Giving it works pretty well in most of the cases, I don't think adding the three additional listener is worthwhile for this edge cases.

sibbng: Hi, let me tell you history of this function and what bring us to here.

At the beginning of this function we were firing handler for `mousedown` and `touchstart` events. Because I wanted to call handler asap to make clicking outside feels optimistic. Then we added `pointerdown` for https://github.com/vueuse/vueuse/issues/323 . Now we are listening 3 events and calling handler multiple times for same event. This issue mentioned in https://github.com/vueuse/vueuse/issues/451 . Then I suggested setting default event to `pointerdown` on [my comment](https://github.com/vueuse/vueuse/issues/451#issuecomment-821197541). Because I thought the issue on https://github.com/vueuse/vueuse/issues/323 was that `mousedown` and `touchstart` events not firing with canvas. I was wrong. The issue was comes from line 40 of reproduction where creates `new OrbitControls` which is prevents firing of our event listeners. I totally missed that line and suggested defaulting to `pointerdown`.

This PR intended to fix #641 by listening `pointerup` event default, but then I realized few other issues like `pointerup` fires when selection event ends outside target, and provide workaround for it.

The biggest problem of this functions was not defaulting to `click` event at the beginning. This raised some issues like https://github.com/vueuse/vueuse/issues/681 and click event have not issue with selecting inline text, selecting input text or scrollbar clicks.

Defaulting `click` event is breaking chance at the moment and needs some refactors. When clicking **`Show Modal`** button also our event handler runs and immediately close modal. There is two way to fix that:

- Adding `.stop` modifier to click event on user land `@click.stop="modal = true"`. This is not preferred way to do for me since we are planning to defaulting `click` event.
- Register `click` listener when `target.value` becomes truthy. This is preferred way.

After defaulting `click` event I don't think anyone who needs other event listeners (like https://github.com/vueuse/vueuse/issues/323) will look for scrollbar click or better selection support. If they need that stuff, they can call another `onClickOutside`.

I'm totally fine with skipping the changes in this PR for v6. All of these workaround can be achieved in current version by setting `{ event: 'click' }` on `onClickOutside` and adding `.stop` modifier to template. I will make another PR for next major version.

antfu: Thanks for the update. Looks like `click` is the way to go. I don't mind having a breaking change as I am also revising functions and working on the new docs recently (await for Vue's new docs).

If we go with `click`, I think we should probably not make it configurable in order to keep the function simple.

As for the clicking conflict, I guess we could solve that by adding `nextTick` somewhere, let's see the implementation first and I might be able to do some help.

Issue #697, opened 2021-08-26
Support new firebase SDK


wheatjs:
Since firebase has released a new modular/smaller SDK we should look into supporting it at some point. Just opening an issue so I don't forget. Will look into adding this sometime over the weekend if I get the time.

https://firebase.googleblog.com/2021/08/the-new-firebase-js-sdk-now-ga.html

FranciscoKloganB:
Any forecast on when this upgrade is going to be tackled?

antfu:
I personally don't use Firebase recently and have no bandwidth to do it. There is an existing work #775 by @peter50216 but closed due to conflicts. If anyone wants to get it supported, you can start fresh or based on @peter50216's work.

One notable thing is that we might want to also support v8 for a while longer, so as discussed in #775, we should support two sub-modules like `@vueuse/firebase/v8` and `@vueuse/firebase/v9` and redirect the default `@vueuse/firebase` to `v8` and move in the next major version. Thanks.

wheatjs:
So I was looking into making a PR for this earlier, but it seems kind of tricky. It seems like we will would need 2 separate npm packages to support this since v8 and v9 use different versions. Maybe there are ways to get around this, although I'm not sure exactly how to myself. 

I was thinking since our firebase functions are in their own package perhaps we could just upgrade to v9 and make a note in the docs for people using v8 to install a version that supports v8? I don't think there will be many changes coming to firebase v8, meaning our current composable should be pretty stable, so I feel like leaving it as an older package version wouldn't be a big issue, maybe I'm wrong?

peter50216:
> I personally don't use Firebase recently and have no bandwidth to do it. There is an existing work #775 by @peter50216 but closed due to conflicts. If anyone wants to get it supported, you can start fresh or based on @peter50216's work.
> 
> One notable thing is that we might want to also support v8 for a while longer, so as discussed in #775, we should support two sub-modules like `@vueuse/firebase/v8` and `@vueuse/firebase/v9` and redirect the default `@vueuse/firebase` to `v8` and move in the next major version. Thanks.

Wrong mention :P, @peterekjs for visibility.

antfu:
> I was thinking since our firebase functions are in their own package perhaps we could just upgrade to v9

That would also do, but since all packages in VueUse share the same version, doing this means we need to bump the major version for all packages. Could also work, but that might need to wait for a while.

peterekjs:
[...]

PR #775, opened 2021-09-20
feat(firebase): add support for firebase@9

peterekjs:
#697

Adds support for Firebase 9 SDK

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Collection of essential Vue Composition Utilities
 <a href="https://www.npmjs.com/package/@vueuse/core" target="__blank"><img src="https://img.shields.io/npm/v/@vueuse/core?color=a1b858&label=" alt="NPM version"></a>
 <a href="https://www.npmjs.com/package/@vueuse/core" target="__blank"><img alt="NPM Downloads" src="https://img.shields.io/npm/dm/@vueuse/core?color=50a36f&label="></a>
 <a href="https://vueuse.org" target="__blank"><img src="https://img.shields.io/static/v1?label=&message=docs%20%26%20demos&color=1e8a7a" alt="Docs & Demos"></a>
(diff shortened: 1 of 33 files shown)


antfu:
Thanks a lot for this! I'd expect that projects need time to take the migration and it would be better for us to support both versions for a while instead of cut the line and make breaking changes. Would you mind creating a new package like `@vueuse/firebase-9` instead so we release them together.

Or ideally, we could have sub modules `@vueuse/firebase/9` and `@vueuse/firebase/8` which the default export `@vueuse/firebase` could be an alias for either version (guess we stay `8` for now, and switch to `9` in the next major bumping for VueUse). This way might need some tweaks to the build scripts and might have some type conflicting, feel free to give it a try. Or if it's too complex, I am also fine with a separate package approach.

Thanks!

peterekjs:
Absolutely agree. The solution with 2 submodules (`@vueuse/firebase/9` and `@vueuse/firebase/8`) and using version `8` as default export for `@vueuse/firebase` before next major version seems to me as the best option here. I'll try playing with it a bit.

ub3rb3457:
when can we expect this functionality to be implemented?

peterekjs:
At the end I came up combining both approaches. To make sure, that `@vueuse/firebase` is compatible with `Firebase 8` (and it's imports) and to ensure, that library supporting `Firebase 9` will be installable via `npm` or `yarn`,  it is necessary to create a separate package, hence I added `@vueuse/firebase-9`. 

In order to mimic current `Firebase 9` module structure described by [upgrade guide](https://firebase.google.com/docs/web/modular-upgrade) I played a bit with bundling scripts to allow package nesting, so we have `modular` library accessible directly as `@vueuse/firebase-9` and `compat` library accessible as `@vueuse/firebase-9/compat`.

Could you, @antfu, please take a look into the commits?

PR #780, opened 2021-09-21
Fix typo in `useOnline` demo

MikealGo:


ahjam-taoufik:
this useonline not work

PR #788, opened 2021-09-23
Added webkitCompassHeading and webkitCompassAccuracy properties to th…

michael-observerly:
…e useDeviceOrientation() sensor.

Added webkitCompassHeading and webkitCompassAccuracy properties to the useDeviceOrientation() sensor. Added a global interface declaration for the DeviceOrientationEvent to override the default with the two new webkit properties. Includes amends to the documentation section for the useDeviceOrientation() sensor.


antfu:
Thanks for your contributions but Sorry I don't think we should support non-standard web APIs. I guess suggest you inlining your implementation in your codebase if you absolutely need that.

PR #789, opened 2021-09-23
feat(createUnrefFn): new function

shorwood:
Hello everyone,

While on a project, i had to dynamically call functions but the arguments I called them with were sometimes Refs. So my first thought was wrapping the function using the reactify utility; But then I remembered the wrapped function returns ComputedRef<T> and is evaluated on every changes of the arguments. And I want the function to only return T and only when I call it. I started looking for other utilities that might help and either did not found suitable one or d

This is where `unrefy` comes from. Below is it's documentation entry.

Issue: #790 

# Documentation

Convert a plain function into a function that unref it's aguments before every call.
The converted function accepts refs and raw values as its arguments
and returns the same value the unconverted function returns, with proper typing.

::: tip
Make sure you're using the right tool for the job. Using [`reactify`](/core/reactify/)
might be more pertinent in some cases where you want to evaluate the function on each changes of it's
arguments.
:::

## Usage

```ts
import axios from 'axios'
import { unrefy } from '.'

const url = ref('https://httpbin.org/post')
const data = ref({foo: 'bar'})
const post = (url, data) => axios.post(url, data)
const postUnrefied = unrefy(post)

post(url, data)           /* ❌ Will throw an error because the arguments are refs */
postUnrefied(url, data)   /* ✔️ Will Work because the arguments will be "unrefied" */
```

## Related Functions

- `reactify`


antfu:
Sorry it took a while. I was mainly concerned about the naming could be a bit not clear. I have changed it to `createUnrefFn` while you can still alias it to the name works for you while using it. Thanks.

PR #791, opened 2021-09-23
feat(loadify): new function

shorwood:
Hello everyone,

While making some services, i had a need for a function wrapper that bind a `Ref` to a `Promise`'s pending state. I first turned to `useAsyncState` but it ended up very verbosy :

```ts
const { isReady: isReadyFetchAll, execute: fetchAll, state } = useAsyncState(() => api.get(`/anything`).then(res => res.data))
const { isReady: isReadyFetch, execute: fetch, state } = useAsyncState(id => api.get(`/anything/${id}`).then(res => res.data))
const { isReady: isReadyPut, execute: put } = useAsyncState(id => api.put(`/anything/${id}`, state.value))
const { isReady: isReadyRemove, execute: remove } = useAsyncState(id => api.delete(`/anything/${id}`))
const isLoading = or(isReadyFetch, isReadyFetchAll, isReadyPut, isReadyRemove)
```

So, inspired by `useAsyncState`, I came up with `loadify`
This below is *almost* the same as above but using `loadify` instead of `useAsyncState`

```ts
const isLoading = ref(false)
const fetchAll =  loadify(isLoading, () => api.get(`/anything`).then(res => res.data))
const fetch =     loadify(isLoading, id => api.get(`/anything/${id}`).then(res => res.data))
const put =       loadify(isLoading, id => api.put(`/anything/${id}`, state.value))
const remove =    loadify(isLoading, id => api.delete(`/anything/${id}`))
```

Bear in mind `useAsyncState` and `loadify` are very similar but have key differences :
- `loadify` wrapper keeps the original function's return.
- `loadify` allows chaining of wrappers. Eg - `partial(loadify(isLoading, axios.post), "https://httpbin.org")`.
- `useAsyncState` initialises the `isReady` ref so you don't have to.
- `loadify` expects you have to provide the `isLoading` ref.
- `useAsyncState` store the result of the function in the `state` ref.
- **loadified** function returns the original result directly.
- calling multiple **loadified** functions bound to the same `isLoading` ref at the same time will result in undefined behaviors.

Issue: #792 

# Documentation

Convert a `Promise` or an async `Function` by binding it's to a ref.
Useful when you want to bind the loading state of multiple promises to a ref.

::: tip
Make sure you're using the right tool for the job. Using [`useAyncState`](/core/useAsyncState/)
might be more pertinent in some cases where you only want to bind the loading state of a single function 
or when you want to bind the loading states of multiple async functions individually.
:::

## Usage

```ts
import axios from 'axios'
import { loadify } from '@vueuse/core'

const api = axios.create({baseUrl: 'https://httpbin.org'})
const isLoading = ref(false)

const fetch =   loadify(isLoading, id => api.get(`/anything/${id}`).then(res => res.data))
const put =     loadify(isLoading, id => api.put(`/anything/${id}`, state.value))
const remove =  loadify(isLoading, id => api.delete(`/anything/${id}`))

const state = ref({})
state.value = fetch(42)
```

## Related Functions

- `useAsyncState`

### DIFF
--- a/indexes.json
+++ b/indexes.json
@@ -454,6 +454,13 @@
       "description": "combine computed and inject"
     },
     {
+      "name": "loadify",
+      "package": "core",
+      "docs": "https://vueuse.org/core/loadify/",
+      "category": "Utilities",
+      "description": "convert a `Promise` or an async `Function` by binding it's to a ref"
+    },
+    {
       "name": "onClickOutside",
       "package": "core",
       "component": true,
--- a/indexes.json
+++ b/indexes.json
@@ -454,13 +454,6 @@
       "description": "combine computed and inject"
     },
     {
-      "name": "loadify",
-      "package": "core",
-      "docs": "https://vueuse.org/core/loadify/",
-      "category": "Utilities",
-      "description": "convert a `Promise` or an async `Function` by binding it's to a ref"
-    },
-    {
       "name": "onClickOutside",
       "package": "core",
       "component": true,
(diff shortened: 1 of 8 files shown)


indexes.json
```diff
@@ -453,6 +453,13 @@
       "category": "Component",
       "description": "combine computed and inject"
     },
+    {
+      "name": "loadify",
+      "package": "core",
```
harmyderoman, line 458: `indexes.json` file now is in `.gitignore`. Delete this from your pr to pass tests. I had the same problem.

shorwood:
Went back to the documentation an I feel like this function is too close to [useAsyncState](https://vueuse.org/core/useasyncstate/). Closing.

shorwood:
Went back to the documentation an I feel like this function is too close to [useAsyncState](https://vueuse.org/core/useasyncstate/). Closing.

PR #804, opened 2021-09-27
feat(useConfirmDialog): new function

harmyderoman:
Creates event hooks to support modals and confirmation dialog chains.

### DIFF
--- a/packages/core/useConfirmDialog/index.md
+++ b/packages/core/useConfirmDialog/index.md
@@ -18,46 +18,26 @@ Functions can be used on the template, and hooks are a handy skeleton for the bu
 
 ### Using hooks
 
-Script:
-
-```ts
-import { defineComponent, ref } from 'vue'
-import { useConfirmDialog } from '@vueuse/core'
-
-export default defineComponent({
-  setup() {
-    const show = ref(false)
-    const {
-      reveal,
-      confirm,
-      cancel,
-      onReveal,
-      onConfirm,
-      onCancel,
-    } = useConfirmDialog(show)
-
-    onConfirm(() => {
-      console.log('Confirmed!')
-    })
-
-    return {
-      show,
-      reveal,
-      confirm,
-      cancel,
-    }
-  },
-})
-```
-
-Template:
-
 ```html
+<script setup>
+import { defineCirmDialog } from '@vueuse/core'
+
+const {
+  isRevealed,
+  reveal,
+  confirm,
+  cancel,
+  onReveal,
+  onConfirm,
+  onCancel,
+} = useConfirmDialog()
+</script>
+
 <template>
-  <button @click="reveal">Show Modal</button>
+  <button @click="reveal">Reveal Modal</button>
 
   <teleport to="body">
-    <div v-if="show" class="modal-bg">
+    <div v-if="isRevealed" class="modal-bg">
       <div class="modal">
         <h2>Confirm?</h2>
         <button @click="confirm">Yes</button>
@@ -68,54 +48,38 @@ Template:
 </template>
 ```
 
-### Another way of using this function. Promise
+### Promise
 
-There are actually two ways to use this feature. The first method uses hooks, this method was shown in the previous example, and the second uses the promise returned by the reveal function. If you prefer working with promises, you can use the second way.
-It is advisable not to mix them. Use either hooks or promises.
-
-Using promise:
-
-```ts
-import { defineComponent, ref } from 'vue'
-import { useConfirmDialog, onClickOutside } from '@vueuse/core'
-
-export default defineComponent({
-  setup() {
-    const show = ref(false)
-    const {
-      reveal,
-      confirm,
-      cancel,
-    } = useConfirmDialog(show)
-
-    const openDialog = async() => {
-      const { data, isCanceled } = await reveal()
-      if (!isCanceled.value) {
-        console.log(data)
-      }
-    }
-    // on click outside logic
-    const target = ref(null)
-    onClickOutside(target, () => cancel())
-
-    return {
-      show,
-      openDialog,
-      confirm,
-      target,
-    }
-  },
-})
-```
-
-Template:
+If you prefer working with promises:
 
 ```html
+<script setup>
+import { useConfirmDialog, onClickOutside } from '@vueuse/core'
+
+const {
+  isRevealed,
+  reveal,
+  confirm,
+  cancel,
+} = useConfirmDialog(show)
+
+const openDialog = async () => {
+  const { data, isCanceled } = await reveal()
+  if (!isCanceled.value) {
+    console.log(data)
+  }
+}
+
+// on click outside logic
+const target = ref(null)
+onClickOutside(target, () => cancel())
+</script>
+
 <template>
   <button @click="openDialog">Show Modal</button>
 
   <teleport to="body">
-    <div v-if="show" class="modal-layout">
+    <div v-if="isRevealed" class="modal-layout">
       <div ref="target" class="modal">
         <h2>Confirm?</h2>
         <button @click="confirm(true)">Yes</button>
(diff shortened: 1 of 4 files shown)


packages/core/useConfirmDialog/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  /*
+  * Opens the dialog
+  */
+  showDialog: Fn
+
+  /**
+   * Confirms and closes the dialog. Triggers a callback inside `onConfirm` hook.
+   * Can accept any data and to pass it to `onConfirm`.
+   */
+  confirm: (data?: any) => void
```
Shinigami92, line 14: Is it somehow possible to get this generic typed? (same for `cancel`)

harmyderoman: Let's see what I can do

packages/core/useConfirmDialog/index.ts
```diff
@@ -0,0 +1,67 @@
+import { Ref } from 'vue-demi'
+import { createEventHook, EventHook, Fn } from '@vueuse/core'
+
+export interface useConfirmDialogReturn<D, E> {
```
Shinigami92, line 4: We could also use default types like `<D = any, E = any>` or maybe `<D = unknown, E = unknown>`

Shinigami92: Oh nevermind... have now read line 43

harmyderoman: I have already used this on the function itself

harmyderoman:
@Shinigami92 I am thinking of creating another hook `onShowDialog` and remove the second param in `useConfirmDialog` for better aesthetics

harmyderoman:
> For what would the `onShowDialog` function be useful? Is it not the same as a `watch`er on the passed `show` ref?
I want it to replace the second param in the function. I think it will be useful to create reusable modals. But you are right, `watch`er solves the same problem

harmyderoman:
@Shinigami92 On the other side, with the `onShowDialog` hook it will be easier to do since I call the callback before invoking the dialog. This can also be done with a watcher, but more tricky.
I suggest not to make life more difficult for users.

packages/core/useConfirmDialog/index.ts
```diff
@@ -0,0 +1,74 @@
...
+   */
+  onCancel: (fn: (param: E) => void) => { off: () => void }
+}
+
+/**
+ * Hooks for creating confirm dialogs. Useful for modal windows, popups and logins.
+ *
+ * @see https://vueuse.org/core/useConfirmDialog/
+ * @param show ref boolean that handle a modal window
+ * @param onShowDialog a function to be called when the modal is creating with `showDialog()`
```
Shinigami92, line 45: I think you need to update the JSDoc

packages/core/useConfirmDialog/index.ts
```diff
@@ -0,0 +1,74 @@
...
+  confirm: (data?: D) => void
+
+  /**
+   * Cancels and closes the dialog. Triggers a callback inside `onCancel` hook.
+   * Can accept any data and to pass it to `onCancel`.
+   */
+  cancel: (data?: E) => void
+
+  /**
+   * Event Hook to be triggered right before dialog creating
```
Shinigami92, line 23:
```suggestion
   * Event Hook to be triggered right before dialog creating.
```

packages/core/useConfirmDialog/index.ts
```diff
@@ -0,0 +1,74 @@
+import { Ref } from 'vue-demi'
+import { createEventHook, EventHook, Fn } from '@vueuse/core'
+
+export interface useConfirmDialogReturn<T, D, E> {
+  /*
+  * Opens the dialog
```
Shinigami92, line 6:
```suggestion
  * Opens the dialog.
```

Shinigami92:
Ok, now I think it's on @antfu or @wheatjs side to approve it 😅

packages/core/useConfirmDialog/index.ts
```diff
@@ -0,0 +1,73 @@
+import { Ref } from 'vue-demi'
+import { createEventHook, EventHook, Fn } from '@vueuse/core'
+
+export interface useConfirmDialogReturn<T, D, E> {
+  /*
+  * Opens the dialog.
+  */
+  showDialog: Fn
```
antfu, line 8: I would like to have this named `reveal` or `prompt` and have it return a Promise that resolves on dialog closed. Something like

```ts
const { data, isCanceled } = await dialog.reveal()
```

antfu (commented):
Cool function!

packages/core/useConfirmDialog/index.ts
harmyderoman: Oh, I forgot to mention this in the interface, `showDialog` passes data from its attribute to the `onShowDialog` hook. I also thought about the promise, but in the end, I concluded that hooks completely replace the promise, and save us from the callback hell.

antfu: I would find it more useful to await the dialog and get the payload from it so I could continue the process. We could have them both.

harmyderoman: Ok, maybe it will be more useful for those who like to work with promises. Then I will add it, by the way, it was my first version of this function.

harmyderoman: I'll do it tomorrow, it's too many changes in all files.

Shinigami92: I'm with @antfu here 👍 this would be a nice addition

packages/core/useConfirmDialog/index.md
```diff
@@ -0,0 +1,238 @@
+---
+category: Utilities
+---
+
+# useConfirmDialog
+
+Creates event hooks to support modals and confirmation dialog chains. It returns 3 functions and 3 hooks.
```
Shinigami92, line 7:
```suggestion
Creates event hooks to support modals and confirmation dialog chains.
```

IMO this info is unnecessary / can be left out.
Assuming someone would add another function to this feature, they will miss this numbers to update

packages/core/useConfirmDialog/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+Functions can be used on the template, and hooks are a handy skeleton for the business logic of modals dialog or other actions that require user confirmation.
+
+## Functions and hooks
+
+- `reveal()` - triggers `onReveal` hook and sets `show.value` to `true`. Returns promise that resolves by `confirm()` or `cancel()`.
```
Shinigami92, line 13: sorry ^^' but just reading this `reveal()` now leads me to think that `prompt()` could be better 🤔 
But maybe we should ask some feedback from other parties 🤔 

_I'm even struggling with myself..._

packages/core/useConfirmDialog/index.md
```diff
@@ -0,0 +1,238 @@
...
+      confirm,
+      cancel,
+    } = useConfirmDialog(show)
+
+    openDialog( async() => {
+      const { data, isCanceled } = await reveal()
+      if(!isCanceled.value){
+        console.log(data)
+      }
+    })
```
Shinigami92, line 96: This is cool 😎, but where is the `openDialog` coming from?

packages/core/useConfirmDialog/index.md
Shinigami92: ... on the other side ... we are free to rename it on our own 🤷 So it should not block this PR to get merged

harmyderoman: I agree with you but `prompt` is a built-in function in js. So this can make a mess.

packages/core/useConfirmDialog/index.md
harmyderoman: Thanks!

packages/core/useConfirmDialog/index.md
harmyderoman: Maybe `evoke`?

Shinigami92: > ... is a built-in function in js

Oh 😲, nevermind, go with `reveal` then 👍

Shinigami92 (dismissed):

packages/core/useConfirmDialog/index.md
harmyderoman: `Confirm` is also a built-in function, but it doesn't make any problems.  Good names are always a problem. So `prompt` is next in line again.

packages/core/useConfirmDialog/index.ts
```diff
@@ -0,0 +1,85 @@
+import { Ref } from 'vue-demi'
+import { createEventHook, EventHook } from '@vueuse/core'
+
+export interface useConfirmDialogReturn<T, D, E> {
+  /*
+  * Opens the dialog. Create promise and return it. Triggers `onReveal` hook.
+  */
+  reveal: (data?: T) => Promise<{ data: D | E | undefined; isCanceled: boolean }>
```
Shinigami92, line 8: Maybe extract `{ data: D | E | undefined; isCanceled: boolean }` to an interface and reuse it

harmyderoman: Good point!

Shinigami92:
@harmyderoman The most common use case in our app is to 
- ask the customer for confirm loss of unsaved data when navigating away from the page without pressing save button
- ask confirmation of deleting something

Maybe we can use such an example 🤔

harmyderoman:
@wheatjs I have already tried this, and it does not work with the function, for such cases I made the UseFocusTrap component.  And it does a great job.

harmyderoman:
@antfu Oh, forget it. I saw that you made it optional. This is the best.

PR #812, opened 2021-10-01
feat(isDefined): new function

### LINKED ISSUE #803: feat: `defined` utility function (user-defined type guard)
I needed to check that a ref's `value` is defined (not `undefined` or `null`), so I wrote this [UDTG](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards):

```ts
function defined<T>(ref: Ref<T>): ref is Ref<Exclude<T, null | undefined>> {
    return ref.value !== null && ref.value !== undefined;
}
```

Usage:

```ts
const example = ref(Math.random() ? "example" : undefined); // Ref<string | undefined>

if (defined(example)) {
    example; // Ref<string>
}
```

I couldn't find something like this in vueuse, would it be useful to others?

(open to suggestions regarding the name)

glen-84:
Closes #803.


antfu:
Could you move it to shared? Thanks!

PR #813, opened 2021-10-02
feat(onKeyStroke): Added functionality to listen to multiple keys

heyitsarpit:
This PR adds the ability to listen to multiple key for onKeyStroke.
This simply adds a parameter type of `string[]` so the previous API is still valid and the PR is not a breaking change.

Example:
```js
import { onKeyStroke } from '@vueuse/core'

// Previous API
onKeyStroke('ArrowDown', (e) => {
  e.preventDefault()
})

// Now possible to pass an array of keys
onKeyStroke(['w', 'W', 'ArrowUp'], (e) => {
  e.preventDefault()
})
```

### DIFF
--- a/packages/core/onKeyStroke/demo.vue
+++ b/packages/core/onKeyStroke/demo.vue
@@ -5,22 +5,22 @@ import { onKeyStroke } from '.'
 const translateX = ref(0)
 const translateY = ref(0)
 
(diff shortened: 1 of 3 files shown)


maxrks:
How to listen to key with modifier key, i.e. `Ctrl + C`  or  `Ctrl + Esc`?

PR #820, opened 2021-10-04
feat(UseFocusTrap): component

harmyderoman:


packages/integrations/useFocusTrap/component.ts
```diff
@@ -0,0 +1,40 @@
+import { h, ref, defineComponent, watch, onUnmounted } from 'vue-demi'
+import { createFocusTrap } from 'focus-trap'
+import type { FocusTrap } from 'focus-trap'
+import { unrefElement } from '@vueuse/core'
+
+export interface RenderableComponent {
```
wheatjs, line 6: Use the existing interface in the types file instead

Here https://github.com/vueuse/vueuse/blob/main/packages/core/types.ts#L8-L15

harmyderoman: Thanks!

harmyderoman:
I didn't know that it blocks tree shaking, then of course it is better to use direct import.

Issue #825, opened 2021-10-07
Feat: Add useElementHover with mouseenter/mouseleave


ahsath:
Was surprised not to see this function/component, can't do it myself, dunno ts 🥲

ahsath:
Merged #830

PR #830, opened 2021-10-07
feat(useElementHover): new function

JMaylor:
This is a basic attempt at #825
As always let me know if any other features you think we need. I have written this below assuming it will be used in a 'Vue' way with refs.

If you think allowing user to pass a query selector `'#my-button'` (and optionally custom document object) would also be useful then I can look at adding this.

Usage:

```vue
<template>
  <button ref="myHoverableElement">
    {{ isHovered }}
  </button>
</template>

<script setup>
import { useElementHover } from '@vueuse/core'

const myHoverableElement = ref()
const isHovered = useElementHover(myHoverableElement)
</script>
```

Implementation:
```ts
import { MaybeRef } from '@vueuse/shared'

export function useElementHover(el: MaybeRef<EventTarget>) {
  const isHovered = ref(false)

  useEventListener(el, 'mouseenter', () => isHovered.value = true)
  useEventListener(el, 'mouseleave', () => isHovered.value = false)

  return isHovered
}

export type UseElementHoverReturn = ReturnType<typeof useElementHover>
```


ahsath:
hi, can we please expose the event object

Issue #838, opened 2021-10-10
useGeolocation() reactive shouldn't return { latitude, longitude } of {0, 0} by default


michael-observerly:
I'm thinking ... should we be returning 0,0 as the null default for latitude and longitude given that this is an actual location (roughly off the coast of Ghana)?

I feel that we should have latitude and longitude as maybe null if the API is unavailable? Or perhaps ... since a valid latitude is between -90 and +90, and a valid longitude is between -180 and +180, we could perhaps return something which defies this validation but is still type consistent? e.g.,

```ts
const coords: Ref<GeolocationPosition['coords']> = ref({
    accuracy: 0,
    latitude: Infinity,
    longitude: Infinity,
    altitude: null,
    altitudeAccuracy: null,
    heading: null,
    speed: null,
  })
```

That way, we know that the user / process / etc is actually at lat, long 0, 0 (rather than treating this as "invalid", which it is not...)

What are the core team's thoughts on this?

PR #839, opened 2021-10-11
fix(useGeolocation)!: latitude and longitude default to Infinity

michael-observerly:
useGeolocation API should return a valid number outside of the -90, + 90 / -180, +180 for coords.latitude and coords.longitude when the user isn't at coords { 0, 0 } and they haven't given permission to share their geoLocation.

This ensures we can make a determination between the user actually being at { 0, 0 } and not allowing us to access their geoLocation. 

If the value is Infinity, then we know "something" has gone wrong (i.e., we can confirm that the longitude or latitude is not valid)

### DIFF
--- a/packages/core/useGeolocation/index.ts
+++ b/packages/core/useGeolocation/index.ts
@@ -26,8 +26,8 @@ export function useGeolocation(options: GeolocationOptions = {}) {
   const error = ref<GeolocationPositionError | null>(null)
   const coords: Ref<GeolocationPosition['coords']> = ref({
     accuracy: 0,
-    latitude: 0,
-    longitude: 0,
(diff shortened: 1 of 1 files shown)


JMaylor:
#838

jkazimierczak:
For anyone who stumbles upon this, and is unsure how to make sure that coordinates exists without raising TS typecheck error - you can use [`isFinite`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite) to check that returned coordinates are indeed not an `Infinity`:
```ts
if (isFinite(coords.value.latitude) && isFinite(coords.value.longitude)) {
  // do something
}
```

PR #841, opened 2021-10-12
feat(useFetch): add onFetchError in UseFetchOptions

### LINKED ISSUE #628: [useFetch] Add `fetchError` and/or `fetchFinally` hooks

climba03003:
Fixes #628

Add `onFetchError` optoin in `UseFetchOptions` which allow a global function in `createFetch`.
In this function it is run before all event emitter and allow the modification of `data` and `error`.

The main reason to allow modification of `data` because some of the people may want to have a unified error handling function.
For example:
```js
const { isFinished, data } = await useFetch('http://example.com', {
  onFetchError(ctx) {
    // always null when it is network error
    if(ctx.data === null) {
      ctx.data = { }
    }
    ctx.data.isError = true

    return ctx
  }
})

await until(isFinished).toBe(true)
// in this case isError always has the value we need
if(data.value.isError) {
  alert('something went wrong')
}
```

### DIFF
--- a/packages/core/useFetch/index.md
+++ b/packages/core/useFetch/index.md
@@ -87,6 +87,21 @@ const { data } = useFetch(url, {
 })
 ```
 
+The `onFetchError` option can intercept the response data and error before it is updated.
+```ts
+const { data } = useFetch(url, {
+  onFetchError(ctx) {
+    // ctx.data can be null when 5xx response
+    if (ctx.data === null) 
+      ctx.data = { title: 'Hunter x Hunter' } // Modifies the resposne data
+
+    ctx.error = new Error('Custom Error') // Modifies the error
+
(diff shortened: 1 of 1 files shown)


glen-84:
I wonder if it'll be confusing having two `onFetchError` functions, the existing "event hook", and this new one. That's why I had suggested naming it something like `fetchError` (without the `on` prefix), but maybe it's not an issue?

Let's wait to hear from the reviewers.

Thanks!

PR #821, opened 2021-10-04
feat(useFuse): new function

DarrylNoakes:
Reactive wrapper for Fuse.js.

As discussed  in #806.

PR #857, opened 2021-10-19
feat(useFuse): improve the demo

darrylnoakes:
As I mentioned in [my comment in the original PR](https://github.com/vueuse/vueuse/issues/821#issuecomment-945017215).

Switched it from Grid to Flexbox and adjusted some related styling.
Fixed some incorrect WindiCSS classes. E.g. I had `m-top-0` instead of `mt-0` in at least one place.

I also made a minor change of removing unused parameters:
`data.value.sort((a, b) => Math.random() - 0.5)`
to
`data.value.sort(() => Math.random() - 0.5)`

### DIFF
--- a/packages/integrations/useFuse/demo.vue
+++ b/packages/integrations/useFuse/demo.vue
@@ -185,14 +185,15 @@ const { results } = useFuse(search, data, options)
       <h2 class="mt-0">
         Results
       </h2>
-      <ol v-if="results.length > 0" class="overflow-y-scroll space-y-2" start="0">
-        <li v-for="(result, index) in results" :key="index" class="mx-2 px-2 py-1 rounded-lg bg-gray-400/20">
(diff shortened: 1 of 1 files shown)


wheatjs:
No preference, it is fine the way it is as well.

DarrylNoakes:
Okay, I believe that's all then. I can't think of anything else as of yet.

wheatjs:
> @wheatjs I've reviewed it, and I think that is all. You can merge if/when you want to.
> 
> Edit:
> 
> I can change `Result Count: x` to `x Results`, if you want.
> 
> Sorry for randomly adding little bits and pieces like this after saying I'm done. Patience is not my strong point :).

It is fine the way it is. If you would like to change it consider that change approved as well. I'll leave the merging to Anthony. He is pretty busy so he'll get around to it when it can. Probably within the next week or two

PR #858, opened 2021-10-20
feat: change the time to change the status of copied

zifeiyu666:
I think we should chang the status of copied when the copy funciton is called, instead of change copied to false after some duration.


antfu:
This was designed to show some UI changes on copying - like you click the button it will turn to `Copied` for a few seconds and then change back.

Issue #615, opened 2021-07-08
`useDebounce` allow `maxWait`


isgj:
Do you consider adding a `maxWait` option to `debounce` family functions?
It should be greater than the debounce time.

antfu:
I am not sure if I understand what it does. Do you have any prior art that we could reference to?

isgj:
It's the same as `maxWait` of [lodash debounce](https://lodash.com/docs/4.17.15#debounce)

> The maximum time func is allowed to be delayed before it's invoked.

Ex: with `debounce` and `wait = 500ms`, if there is an interaction every `400ms` the callback will never be called, it will be debounced forever. Sometimes in order to improve UX you want the callback to be called if a `maxWait` has been passed.

`wait = 500` `mawWait = 1000`

|time|interaction|callback called|action
|--|--|--|--|
|0|yes|no|debounced after 500ms|
|400|yes|no|debounced after 500ms|
|800|yes|no|debounced after 200ms (800(`now`)+500(`wait`) > 1000(`maxWait`))|
|1000|no|yes|set last call to now (next debounce can't be before 1500,  next `maxWait` can't be before 2000ms )|

If `maxWait` < `wait` it becomes a `throttle`

PR #877, opened 2021-10-23
feat(useDebounce & useDebounceFn): accept maxMs parameter

JMaylor:
#615 

I added maxMs to the debounceFilter and consequentially useDebounce and useDebounceFn.

I updated the demos on useDebounce and useDebounceFn (maxMs = 5000ms) so you can see it working but can drop these.

### DIFF
--- a/packages/shared/utils/filters.ts
+++ b/packages/shared/utils/filters.ts
@@ -28,7 +28,7 @@ export interface DebounceFilterOptions {
    * The maximum time allowed to be delayed before it's invoked.
    * In milliseconds.
    */
-  maxWait: number | null
+  maxWait?: number
 }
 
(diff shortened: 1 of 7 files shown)


packages/shared/utils/filters.ts
```diff
@@ -23,6 +23,14 @@ export interface ConfigurableEventFilter {
   eventFilter?: EventFilter
 }
 
+export interface DebounceFilterOptions {
+  /**
+   * The maximum time allowed to be delayed before it's invoked.
+   * In milliseconds.
+   */
+  maxWait: number | null
```
antfu, line 31:
```suggestion
  maxWait?: number
```

PR #898, opened 2021-11-05
feat(useFps): reactive fps

okxiaoliang4:


packages/core/useFPS/index.md
```diff
@@ -0,0 +1,15 @@
+---
+category: Sensors
+---
+
+# useFPS
+
+reactive fps
```
antfu, line 7:
```suggestion
Reactive FPS
```

packages/core/useFPS/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  every?: number
+}) {
+  const fps = ref(0)
+
+  const every = options?.every ?? 10
+
+  let last = performance.now()
+  let ticks = 0
+  useRafFn(() => {
+    if (++ticks === every) {
```
antfu, line 14:
```suggestion
    if (++ticks >= every) {
```

antfu (dismissed):

DarrylNoakes:
Not sure a about anyone else, but I'm a little uncomfortable with the naming.
Almost all JS style guides recommend using camel case for naming functions, even when the name includes an abbreviation.

Would it be reasonable to rename it to `useFps()`?

The other VueUse functions with abbreviations are named like this, e.g. `useRafFn()`.

PR #904, opened 2021-11-05
feat(useVirtualList): adding support for reactive lists

### LINKED ISSUE #903: useVirtualList should support reactive refs
## Feature Request

I want to be able to have searchable virtual lists. The discussion below has a reproduction and desired API.

### Discussed in https://github.com/vueuse/vueuse/discussions/824

<div type='discussions-op-text'>

<sup>Originally posted by **JessicaSachs** October  6, 2021</sup>
Hello! The current implementation of `useVirtualList` works great for a static list, but I'm struggling with how to make it reactive. The current signature specifies an array instead of a reactive ref. Can I get an example of `useVirtualList` with a simple textfield filter?

Here's a reproduction of what I'd like to be able to do...

https://github.com/JessicaSachs/useVirtualList/blob/main/src/App.vue#L26-L44

```js
import { ref, computed } from 'vue'
import faker from 'faker'
import { useVirtualList } from '@vueuse/core'

const search = ref('')

const bigList = Array.from(Array(100).keys()).map(a => ({
  // name, avatar, etc
  ...faker.helpers.createCard(),
  id: faker.datatype.uuid()
}))

const filteredList = computed(() => {
  return bigList.filter(i => {
    return i.name.toLocaleLowerCase().indexOf(search.value.toLocaleLowerCase()) > -1
  })
})

// Ideally, I could pass something reactive into `useVirtualList`
const { list, wrapperProps, containerProps } = useVirtualList(filteredList.value, { itemHeight: 80 })
```

Screen Shot 2021-10-06 at 6 27 50 PM
</div>

JessicaSachs:
- Closes #903

https://user-images.githubusercontent.com/2801156/140575750-6928d1a8-da39-4890-bc3c-7d19d38f3e9b.mov

### Before merge
I'm struggling with the types for the list prop -- I think it should be a `MaybeRef`, but I'm not sure how to type it or properly "wrap" a non-reactive list.


packages/.vitepress/theme/styles/demo.css
```diff
@@ -99,29 +99,22 @@
     border-top: 0;
   }
 
-  input[type='text'] {
+  input {
```
JessicaSachs, line 102: Supports `input[type="search"]`

packages/core/useVirtualList/demo.vue
```diff
@@ -1,43 +1,64 @@
...
+
+const filteredItems = computed(() => {
+  return allItems.filter(i => i.size.startsWith(search.value.toLowerCase()))
+})
+
 const { list, containerProps, wrapperProps, scrollTo } = useVirtualList(
-  Array.from(Array(99999).keys()),
+  filteredItems,
   {
-    itemHeight: i => (i % 2 === 0 ? 42 + 8 : 84 + 8),
```
JessicaSachs, line 9: Previously this was index-based, this does not work with the reactive list -- the heights are calculated incorrectly.

packages/core/useVirtualList/index.ts
```diff
@@ -16,10 +17,17 @@ export interface UseVirtualListOptions {
   overscan?: number
 }
 
-export function useVirtualList <T = any>(list: T[], options: UseVirtualListOptions) {
+export type UseVirtualListItem<T> = {
+  data: T
+  index: number
+}
+
+export function useVirtualList <T = any>(list: MaybeRef<T[]>, options: UseVirtualListOptions) {
```
JessicaSachs, line 25: The types here and for `currentList.value` on line 85 are broken.

packages/core/useVirtualList/index.ts
```diff
@@ -16,10 +17,17 @@ export interface UseVirtualListOptions {
...
   const containerRef: Ref = ref<HTMLElement | null>()
   const size = useElementSize(containerRef)
-  const currentList: Ref = ref([])
+
+  const currentList: Ref<UseVirtualListItem<T>[]> = ref([])
+  const _list = ref(isRef(list) ? list.value : list)
```
JessicaSachs, line 30: I don't believe this line is correct and this is why there are type errors.

JessicaSachs (commented):
I left some notes on areas I need help with.

packages/core/useVirtualList/index.ts
JessicaSachs: I feel like I want to use the sync composable instead of creating a new ref like this.

antfu, line 30:
```suggestion
  const _list = ref(list)
```

Vue will check it's already a ref or not and reuse the existing one if possible

JessicaSachs: Nice, ok. Doing this locally. Do you have any suggestions for how to fix the types? I'm struggling a bit to understand why `_list.value` can't be assigned to `currentList.value`.

Screen Shot 2021-11-05 at 6 03 07 PM

PR #923, opened 2021-11-16
feat(useMemory): new function

okxiaoliang4:


DarrylNoakes:
Please note that `performance.memory` is non-standard and only supported in Chromium-based browsers. See [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Performance/memory).

Following the standard of this library, make the function return an object containing a `memory` property and an `isSupported` property, and only do anything if it is supported.

P.S. Antfu, if I am "interfering," please tell me :)

packages/core/useMemory/index.ts
```diff
@@ -13,10 +13,13 @@ type PerformanceMemory = Performance & {
...
   const memory = ref() as Ref<Memory>
+  const isSupported = performance && 'memory' in performance
 
-  useIntervalFn(() => {
-    memory.value = (performance as PerformanceMemory).memory
-  }, 1000, { immediate: true, immediateCallback: true })
+  if (isSupported) {
+    useIntervalFn(() => {
+      memory.value = (performance as PerformanceMemory).memory
+    }, 1000, { immediate: true, immediateCallback: true })
```
antfu, line 21: We should make this interval configurable

PR #927, opened 2021-11-17
fix(useFocus): resolve issues when used with v-if

### LINKED ISSUE #926: useFocus: Unexpected behavior when used with v-if
When using `useFocus` with a target that is conditionally rendered it will default to `true` if the element is hidden and you switch to another browser tab and then back.

Example:

```html
<template>
  <input v-if="show" ref="input" />
  <button @click="toggle">Toggle</button>
</template>
```
```typescript
const show = ref(false);
const toggle = () => { show.value = !show.value };
const input = ref();
const { focused } = useFocus({ target: input }); // Will be true after switching browser tabs while the element is hidden
```

I believe this is because the window is used as a fallback for the target when the ref isn't attached to an element.

VueUse version: 6.9.1

wtykirby:
Resolves #926

I noticed some issues when using `useFocus` with an element that was conditionally rendered. I refactored it to have the focus state be a computed based off of the existing `useActiveElement` function instead of trying to keep a ref in sync with value changes and events.

### DIFF
--- a/packages/core/useFocus/index.ts
+++ b/packages/core/useFocus/index.ts
@@ -1,8 +1,9 @@
 import { computed, Ref, watch } from 'vue-demi'
 import { MaybeElementRef, unrefElement } from '../unrefElement'
 import { useActiveElement } from '../useActiveElement'
+import { ConfigurableWindow } from '../_configurable'
 
-export interface FocusOptions {
+export interface FocusOptions extends ConfigurableWindow {
   /**
    * Initial value. If set true, then focus will be set on the target
    *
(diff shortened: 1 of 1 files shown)


packages/core/useFocus/index.ts
```diff
@@ -34,34 +33,23 @@ export interface FocusReturn {
...
+    },
+    set(value: boolean) {
+      if (!value && focused.value)
+        target.value?.blur()
+      if (value && !focused.value)
+        target.value?.focus()
+    },
+  })
+
+  watch(target, () => { focused.value = initialValue }, { immediate: true, flush: 'post' })
```
wtykirby, line 52: If the `intialValue` is set to true then this will result in the element being focused each time the target ref is updated. So for conditionally rendered elements with `v-if` it will cause them to be focused each time they render (not just the first time).

So for my example in #926 it will focus the input each time the show flag is toggled to true.

Not sure if this is desired or what the expected behavior should be.

PR #928, opened 2021-11-17
feat(useCounter): add options

damienroche:
Add as second parameter an options object with `minValue` and `maxValue` in order to create a limited counter.
can be useful for example to create a zoom 

```
const { count: zoom, inc, dec, reset } = useCounter(5, { minValue: 1, maxValue: 16 })
```

### DIFF
--- a/packages/shared/useCounter/index.ts
+++ b/packages/shared/useCounter/index.ts
@@ -1,16 +1,38 @@
 import { ref } from 'vue-demi'
 
+interface useCounterOptions {
+  minValue?: number
+  maxValue?: number
+}
+
 /**
  * Basic counter with utility functions.
  *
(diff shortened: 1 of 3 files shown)


packages/shared/useCounter/index.ts
```diff
@@ -1,16 +1,38 @@
 import { ref } from 'vue-demi'
 
+interface useCounterOptions {
+  minValue?: number
+  maxValue?: number
```
antfu, line 5:
```suggestion
  min?: number
  max?: number
```

damienroche: @antfu i push a new commit with `min` and `max` instead of `minValue` and `maxValue`

PR #938, opened 2021-11-19
fix: interface typos

### LINKED ISSUE #757: typo: Stopable should be named Stoppable
https://github.com/vueuse/vueuse/blob/d6dd1ba2d5fab674771103f0955e631c6c8f436e/packages/shared/utils/types.ts#L64

This triggers a spellcheck lint warning in my project 😆

Shinigami92:
fix #757


Shinigami92:
There is also `StorageSerializers`, but internet says that `serializers` is indeed okay. So it is a non registered word/plural-noun of the spellchecker 🤔

Issue #915, opened 2021-11-10
useMagicKeys weird behavior with meta key


damienroche:
```
const keys = useMagicKeys()
const ctrlK = keys['Meta+K']

watch(ctrlK, (v) => {
  if (v)
    console.log('Meta+K have been pressed')
})
```

With this code, I have a console log when I pressed Meta+K but after this first log, if I press only Meta key, console is fired. If I press another key then I press Meta key, I no longer have console.log (expected behavior)

```
Meta+K => console => Control+K have been pressed
Meta => console => Control+K have been pressed
Meta => console => Control+K have been pressed
Meta => console => Control+K have been pressed
K => void
Meta => void
Meta => console => Control+K have been pressed
```

francoism90:
@damienroche Could you try again with `{ deep: true }` as watch option?

damienroche:
@francoism90 I will try it.
also the bug is reproducible on the documentation here :
[https://vueuse.org/core/useMagicKeys/](https://vueuse.org/core/useMagicKeys/)

press a meta key and B key for example,
on release KeyB is still displayed on `Keys Pressed` section. You have to press the B key twice to hide it.

shayneo:
I also am seeing this

shayneo:
It looks like the order of releasing keys matters

Here is a quick example

```
const { command, a } = useMagicKeys()
```
Press and hold`cmd+a`
 - `command` is true
 - `a` is true

Release `command` then release `a`
 - `command` is false
 - `a` is false

Press and hold`cmd+a`, then release `a` then release `command`
 - `command` is false
 - `a` is **true**

CapitaineToinon:
I'm having the same issue with `useMagicKeys`, `useKeyModifier` and `onKeyStoke` too. It seems like the browser just never fires the on key up of a key after it's been combined with Meta?

m0dch3n:
I created my [own solution now](https://github.com/vueuse/vueuse/pull/946#issuecomment-1041811542)

antfu:
#1312

PR #946, opened 2021-11-22
fix(useMagicKeys): handle meta key

shayneo:
This is an attempt to clean up the behavior noted in this issue: https://github.com/vueuse/vueuse/issues/915

### DIFF
--- a/packages/core/useMagicKeys/demo.vue
+++ b/packages/core/useMagicKeys/demo.vue
@@ -3,7 +3,7 @@ import { computed } from 'vue-demi'
 import Key from './Key.vue'
 import { useMagicKeys } from '.'
 
-const { shift, v, u, e, s, v_u_e, u_s_e, current } = useMagicKeys()
+const { shift, v, u, e, s, v_u_e, u_s_e, current, command_bracketRight } = useMagicKeys()
 const keys = computed(() => Array.from(current))
 </script>
 
@@ -50,6 +50,10 @@ const keys = computed(() => Array.from(current))
         </Key>
         <Key :value="u_s_e">
           Use
+        </Key>
+
+        <Key :value="command_bracketRight">
+          command_bracketRight
         </Key>
       </div>
 
--- a/packages/core/useMagicKeys/demo.vue
+++ b/packages/core/useMagicKeys/demo.vue
@@ -3,7 +3,7 @@ import { computed } from 'vue-demi'
 import Key from './Key.vue'
 import { useMagicKeys } from '.'
 
-const { shift, v, u, e, s, v_u_e, u_s_e, current, command_bracketRight } = useMagicKeys()
+const { shift, v, u, e, s, v_u_e, u_s_e, current } = useMagicKeys()
 const keys = computed(() => Array.from(current))
 </script>
 
@@ -50,10 +50,6 @@ const keys = computed(() => Array.from(current))
         </Key>
         <Key :value="u_s_e">
           Use
-        </Key>
-
-        <Key :value="command_bracketRight">
-          command_bracketRight
         </Key>
       </div>
 
--- a/packages/core/useMagicKeys/index.ts
+++ b/packages/core/useMagicKeys/index.ts
@@ -82,25 +82,65 @@ export function useMagicKeys(options: UseMagicKeysOptions<boolean> = {}): any {
   const obj = { toJSON() { return {} }, current }
   const refs: Record<string, any> = useReactive ? reactive(obj) : obj
 
+  function safeSetRef(key: string, value: boolean) {
+    if (key in refs) {
+      if (refs[key] === value)
+        return
+
+      if (useReactive)
+        refs[key] = value
+
+      else
+        refs[key].value = value
+    }
+  }
+
+  function forceUpdateMetaDependents() {
+    const metaAliases = Object.entries(aliasMap).map(([key, value]) => {
+      if (value === 'meta')
+        return key
+
+      return ''
+    }).filter(k => k !== '')
+
+    const metaDeps = Object.keys(refs).map((k) => {
+      let key = ''
+      for (let idx = 0; idx < metaAliases.length; idx++) {
+        const alias = metaAliases[idx]
+        if (k.includes(alias))
+          key = k.replace(alias, '').replace(/[+_-]/g, '')
+      }
+      return key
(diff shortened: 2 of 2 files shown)


antfu:
I am not sure about the implementation, it feels a bit hacky and treats Meta too special. Do you know https://vueuse.org/core/useKeyModifier ? Maybe we could reuse this 🤔

shayneo:
> I am not sure about the implementation, it feels a bit hacky and treats Meta too special. Do you know https://vueuse.org/core/useKeyModifier ? Maybe we could reuse it 🤔

yeah it definitely felt a little hackish while writing since meta is treated as such a special case. I'll take a look at `useKeyModifier` and see if there's a better way forward.

rvmourik:
@shayneo Did you ever found a better solution? I am also running into the same problem as you mentioned.

shayneo:
> @shayneo Did you ever found a better solution? I am also running into the same problem as you mentioned.

Hey @rvmourik , like Anthony mentioned, you might be able to use the `useKeyModifier` function to track the meta key properly in combination with `useMagicKeys`. Though I haven't attempted this yet.

If you get a working implementation, we should post here and perhaps add a note to the docs offering guidance on this case, as it seems common enough that folks will reach for `useMagicKeys` to implement `cmd+{whatever key}` shortcuts

rvmourik:
> 

Hi @shayneo,

I tried it in combination with useKeyModifier, but the behaviour is the same in the end. The first time you try a combination with cmd + .. it works, but after releasing both keys, the event is only triggered by pressing the meta key and not the original combination.

Yeah, you find a lot of apps using the combination. But probably not with useMagicKeys :-) @antfu is there something that can be done to improve this implementation?

ManuelLeiner:
I helped myself with something like the following using `onKeyDown`:

```
  onKeyDown("f", (e) => {
    if (e.metaKey) {
      e.preventDefault();
      search.value.focus();
    }
  });
```

m0dch3n:
I'm also facing this issue...

@antfu Do you think the proposed pull request is something, which can soon be merged and published officially ? 

Since the request is pending since 4 month, I just wanted to know, if I should simply wait another few weeks, or if I should better create my own work around in the mean time?

m0dch3n:
After having tried this commit, I noticed, it is still not working correctly, so I created myself a solution:

```javascript
import { computed, reactive, ref, unref } from 'vue'
import {
  DefaultMagicKeysAliasMap,
  defaultWindow,
  MagicKeys,
  useEventListener,
  UseMagicKeysOptions,
} from '@vueuse/core'

/**
 * Reactive keys pressed state, with magical keys combination support.
 *
 * @see https://vueuse.org/useMagicKeys
 */
export function useMagicKeysCustomMetaHandling(
  options?: UseMagicKeysOptions<false>
): MagicKeys<false>
export function useMagicKeysCustomMetaHandling(
  options: UseMagicKeysOptions<true>
): MagicKeys<true>
export function useMagicKeysCustomMetaHandling(
  options: UseMagicKeysOptions<boolean> = {}
): any {
  const {
    reactive: useReactive = false,
    target = defaultWindow,
    aliasMap = DefaultMagicKeysAliasMap,
    passive = true,
    onEventFired = noop,
  } = options
  const current = reactive(new Set<string>())
  const toClean = new Set<string>()
  const obj = {
    toJSON() {
      return {}
    },
    current,
  }
  const refs: Record<string, any> = useReactive ? reactive(obj) : obj

  function updateRefs(e: KeyboardEvent, value: boolean) {
    const key = e.key?.toLowerCase()
    const code = e.code?.toLowerCase()
    const values = [code, key].filter(Boolean)

    // current set
    if (code) {
      if (value) {
        current.add(e.code)
        // track all the codes who were hit during together the metaKey
        if (e.metaKey) {
          toClean.add(e.code)
        }
      } else {
        current.delete(e.code)
        // clean all the tracked codes
        if (key === 'meta') {
          toClean.forEach((code) => current.delete(code))
          toClean.clear()
        }
      }
    }

    for (const key of values) {
      if (key in refs) {
        if (useReactive) refs[key] = value
        else refs[key].value = value
      }
    }
  }

  if (target) {
    useEventListener(
      target,
      'keydown',
      (e: KeyboardEvent) => {
        updateRefs(e, true)
        return onEventFired(e)
      },
      { passive }
    )
    useEventListener(
      target,
      'keyup',
      (e: KeyboardEvent) => {
        updateRefs(e, false)
        return onEventFired(e)
      },
      { passive }
    )
  }

  const proxy = new Proxy(refs, {
    get(target, prop, rec) {
      if (typeof prop !== 'string') return Reflect.get(target, prop, rec)

      prop = prop.toLowerCase()
      // alias
      if (prop in aliasMap) prop = aliasMap[prop]
      // create new tracking
      if (!(prop in refs)) {
        if (/[+_-]/.test(prop)) {
          const keys = prop.split(/[+_-]/g).map((i) => i.trim())
          refs[prop] = computed(() => keys.every((key) => unref(proxy[key])))
        } else {
          refs[prop] = ref(false)
        }
      }
      const r = Reflect.get(target, prop, rec)
      return useReactive ? unref(r) : r
    },
  })

  return proxy as any
}
```

This problem happens, because Mac OS is not firing the `keyup` event anymore, once the meta key is pressed. So basically a proper solution is NOT POSSIBLE.

What I basically do, do hack around:

I simply keep track in `const toClean = new Set<string>()` track of all the keys which were hit together during the meta keydown.

```javascript
    // current set
    if (code) {
      if (value) {
        current.add(e.code)
        // track all the codes who were hit during together the metaKey
        if (e.metaKey) {
          toClean.add(e.code)
        }
      } else {
        current.delete(e.code)
        // clean all the tracked codes
        if (key === 'meta') {
          toClean.forEach((code) => current.delete(code))
          toClean.clear()
        }
      }
    }
```

When the meta is then released, I clear all the keys from `current`

The only caveat: if the user is still holding the other key, it will delete it from `current`, and then a few ms later add it again, when key repeat of the holding fires the keydown event again.

In my app I don't need to solve this. However a possible solution would be:

```javascript
const toClean = new Map<string, number>()
...
toClean.add(e.code, new Date().getMilliseconds())
...
// on meta key release
// check if last hit is some time ago, then consider it's no longer pressed, and can safely be deleted 
```

antfu:
#1312

PR #950, opened 2021-11-23
feat(useEyeDropper): add initialValue option, improve types

okxiaoliang4:


packages/core/useEyeDropper/index.ts
```diff
@@ -1,14 +1,28 @@
 import { ref } from 'vue-demi'
 
-export function useEyeDropper() {
+type WindowEyeDropper = Window & typeof globalThis & {
```
antfu, line 3: Why don't we just type `EyeDropper` and also make it available in the return type?

packages/core/useEyeDropper/index.ts
```diff
@@ -1,14 +1,28 @@
...
+  }
+}
+
+/**
+ * Reactive [EyeDropper API](https://developer.mozilla.org/en-US/docs/Web/API/EyeDropper_API)
+ *
+ * @see https://vueuse.org/useEyeDropper
+ * @param initialValue string
+ */
+export function useEyeDropper(initialValue = '') {
```
antfu, line 18: Prefer an object option here for future extensiablity

packages/core/useEyeDropper/index.ts
okxiaoliang4: like this?
```typescript
export interface EyeDropper {
  new(): {
    open: () => Promise<{ sRGBHex: string }>
    [Symbol.toStringTag]: 'EyeDropper'
  }
}
```

antfu: Yes

packages/core/useEyeDropper/index.ts
okxiaoliang4: ok~

PR #953, opened 2021-11-23
feat(useClamp): new function

okxiaoliang4:


packages/core/useClamp/index.ts
```diff
@@ -0,0 +1,29 @@
...
+      return clamp(_value.value, unref(min), unref(max))
+    },
+    set(value: number) {
+      _value.value = clamp(value, unref(min), unref(max))
+    },
+  })
+  return result
+}
+
+function clamp(value: number, min: number, max: number) {
```
antfu, line 25: IIRC we already have a clamp function in `@vueuse/share` that you can reuse.

PR #958, opened 2021-11-24
feat(useLockScroll): new function

### LINKED ISSUE #957: feat: useLockScroll
I have an idea :
```html
<script setup lang="ts">
import { useLockScroll } from '@vueuse/core'

const el = ref<HTMLElement | null>(null)
const { isLocked, lock, unlock } = useLockScroll(el)
</script>

<template>
  <div ref="el"></div>
</template>
```

webfansplz:
close #957


packages/core/useLockScroll/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const unlock = () => {
+    const ele = (unref(element) as HTMLElement)
+    if (!ele || !lockState.value) return
+    isIosDevice ? touchMoveListener?.() : ele.style.overflow = initialOverflow
+    lockState.value = false
+  }
+
+  return {
+    lockState,
```
wheatjs, line 61: I think this would be better named as `isLocked`

webfansplz: fixed, thanks !

PR #972, opened 2021-11-25
fix(useAsyncState): optional argument of `execute`

lsdsjy:
The first parameter `delay` of execute function should be optional(which has a default value of 0).


besserwisser:
Hopefully this is not a stupid question... But how does Typescript know with this commit, that the first argument is a "delay" or an "args"?

lsdsjy:
I think the first argument of `execute` is always `delay`, not some arguments passed to `promise` function, according to the implementation.

besserwisser:
Thank you for clarifying. So I did ask a stupid question :D

Issue #975, opened 2021-11-26
useWebNotification()


michealroberts:
I feel it would be very beneficial to have a useNotification() composable for pushing /displaying web notifications. 

Does anyone know if work has already started on this?

michael-observerly:
PR her for reference #979 cc @antfu

PR #979, opened 2021-11-26
feat(useWebNotification): new function

michael-observerly:


michael-observerly:
Linked Issue: #975

michael-observerly:
@antfu Would you like an accompanying test to go with this new feature? (Hope that is not a stupid question? :D)

packages/core/useWebNotification/index.ts
```diff
@@ -0,0 +1,120 @@
+import { onMounted, Ref, ref, onUnmounted } from 'vue-demi'
+
+interface WebNotificationOptions {
+  /*
+   *
+   * The body string of the notification as specified in the constructor's options parameter.
```
antfu, line 6: Can you follow the TSDocs format from other composables and remove the extract spaces?

michael-observerly: Will do. Will push shortly.

michael-observerly:
@antfu Amends pushed. Thanks for the super quick review.

michael-observerly:
@antfu I've spotted a few typos, I'll fix these shortly.

michael-observerly:
@antfu Some major and minor changes have been pushed, will require further review. Thanks!

antfu:
I will see it, there is no need to ping me :)

packages/core/useWebNotification/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  // Is the web notifications API supported?:
+  const isSupported: boolean = 'Notification' in window
+
+  // Request permission to use web notifications:
+  const requestPermission = async() => {
+    if ('permission' in Notification && Notification.permission !== 'denied') await Notification.requestPermission()
+  }
+
+  // Show notification method:
+  const show = (): void => {
```
antfu, line 125:
```suggestion
  const show = (options?: WebNotificationOptions): void => {
```

I think it's better to allow passing an option. The top-level option could be the default for merging.

packages/core/useWebNotification/index.ts
```diff
@@ -0,0 +1,172 @@
...
+ * Reactive useWebNotification
+ *
+ * @see https://vueuse.org/useWebNotification
+ * @param title
+ * @param options of type WebNotificationOptions
+ * @param methods of type WebNotificationMethods
+ */
+export const useWebNotification = (
+  title: string,
+  options: WebNotificationOptions = {},
```
antfu, line 111: I feel we could merge the title and options into one argument.

packages/core/useWebNotification/index.ts
```diff
@@ -0,0 +1,172 @@
+import { onMounted, ref, onUnmounted } from 'vue-demi'
+import type { Ref } from 'vue-demi'
+
+import { useEventListener } from '../useEventListener'
+
+interface WebNotificationOptions {
```
antfu, line 6:
```suggestion
export interface UseWebNotificationOptions {
```

packages/core/useWebNotification/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  /**
+   * A handler for the show event. It is triggered when the notification is displayed.
+   *
+   * @default null
+   */
+  onShow: ((e: Event) => void) | null
+}
+
+// Default web notification methods:
+const defaultWebNotificationMethods = {
```
antfu, line 94: For hooks, we could use https://vueuse.org/shared/createeventhook/

antfu (changes requested):

packages/core/useWebNotification/index.ts
michael-observerly: Added amends to this PR.

packages/core/useWebNotification/index.ts
michael-observerly: Added amends to this PR.

packages/core/useWebNotification/index.ts
michael-observerly: @antfu Could you give an example of how you would like this implemented?

michael-observerly: @antfu I've added my interpretation into commit 1f01933. Thank you for the suggestions. :)

michael-observerly: * Includes usage documentation changes as well.

michael-observerly:
@antfu This one seems to have not come through onto vueuse.org documentation...

PR #985, opened 2021-11-29
feat(useSafeArea): new function

okxiaoliang4:
get safe area in js

```ts
import { useSafeArea } from '@vueuse/core'

const { top, right, bottom, left } = useSafeArea()
```


wheatjs:
Can you give a better description of what this does?

okxiaoliang4:
@wheatjs 
> Can you give a better description of what this does?

Reactive `env(safe-area-inset-*)`

image

packages/core/useSafeArea/index.ts
```diff
@@ -0,0 +1,56 @@
+import { ref } from 'vue-demi'
+import { isClient, useCssVar, useDebounceFn, useEventListener } from '@vueuse/core'
+
+const topVarName = '--vue-use-safe-area-top'
```
wheatjs, line 4: I think `vueuse` here should be a single word, so `--vueuse-safe-area-top` instead

okxiaoliang4: ok~

antfu:
How about name it `useScreenSafeArea` to be more clear?

okxiaoliang4:
ok~

PR #989, opened 2021-11-30
chore(set): set ref object value

### LINKED ISSUE #986: `set` with 3 parameters
Why is it that set with 3 parameters does not handle `ref` but with 2 it does?
For example
```js
const a = ref(0)
set(a, 1)
console.log(a.value) // 1
```
totally fine, but
```js
const a = ref({ foo: 0})

set(a, 'foo', 1) // Argument of type '"foo"' is not assignable to parameter of type 'keyof Ref<{ foo: number}>'
```
instead you have to do this
```js
const a = ref({ foo: 0})

set(get(a), 'foo', 1)

console.log(a.value.foo) // 1
```
For me this is unintuative

okxiaoliang4:
resolve #986


okxiaoliang4:
@antfu @wheatjs 

While running Vue2 jest tests, I found that adding a new property to the ref.value object would cause 2 watch callbacks to be triggered.

Don't quite understand why this is so.

I think maby [composition-api](https://github.com/vuejs/composition-api)  has bug? i well check it~

antfu:
I don't feel `set` should auto unref, it should be explicitly done by usage.

okxiaoliang4:
👌🏻

PR #997, opened 2021-12-02
feat(useFocusWithin): new function

wtykirby:
Added a new function that will track if a target element contains the focused element. This is similar to what can be achieved with the `:focus-within` CSS psuedo-class. This functionality probably could be rolled into the existing `useFocus` function with a new option, but I think the behavior would be somewhat confusing since that function also allows setting the value manually to focus/blur the element and that doesn't make much sense in this context.

A couple of examples in which this would be useful:
- Be able to track if any element in a form currently has focus
- Ensuring a dropdown menu remains visible as long as any of its descendants have focus


packages/core/useFocusWithin/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  focused: Ref<boolean>
+}
+
+/**
+ * Track if focus is contained within the target element
+ *
+ * @see https://vueuse.org/useFocusWithin
+ * @param options
+ */
+export function useFocusWithin(options: FocusWithinOptions): FocusWithinReturn {
```
wheatjs, line 26: To be consistent with our other API's the first argument should be just the target ref itself. If we have more options they can go in a second options object parameter.

wtykirby: Thanks for the clarification. I've gone ahead and made that change. Want me to open a separate PR to update `useFocus` so that it is also consistent? Will be a breaking change, but I would say it's worth it to keep them all in sync.

wheatjs: Oh yeah I guess that one slipped by. Feel free to make a PR and we can merge it in the next major version. Thanks!

wheatjs (approved):

PR #999, opened 2021-12-02
feat(useMap): new function

### LINKED ISSUE #998: feat: useMap
I have an idea:

```ts
const { size, keys, values, get, set, has, setAll, remove, reset } = useMap({ a: 1, b: 2 })

console.log(size.value) // 2
console.log(get('a')) // 1

setAll({ c: 3 }) 
console.log(has('c')) // true

reset()
console.log(get('b')) // 2
console.log(has('c')) // false

```

webfansplz:
close #998


antfu:
I don't think you need this, I suppose Vue handles Map and Set natively.

PR #1000, opened 2021-12-02
chore: typo

ItsRyanWu:
1. Updated the outdated api name from `defineCirmDialog` to `useConfirmDialog` in the import declaration
2. Added missing `ref` declaration otherwise people may get confused where is it from


antfu:
Congrats on getting the PR number #1000! 🎉 And thanks!

PR #1010, opened 2021-12-04
refactor(useFocus)!: made target an required argument instead of an option

wtykirby:
This PR is intended to make the `useFocus` API more consistent with the rest of the package and other functions related to managing focus like `useFocusTrap` and the potential new function `useFocusWithin`. I've also cleaned up the tests and docs a bit as well while working on this.

See discussion on PR for additional context #997 

**NOTE: THIS WILL BE A BREAKING CHANGE**


wheatjs:
Looks good to me, let's ship this in v8.0

wtykirby:
@wheatjs sounds good, thanks for the feedback!

packages/core/useFocus/index.ts
```diff
@@ -29,28 +24,27 @@ export interface FocusReturn {
...
  *
  * @see https://vueuse.org/useFocus
+ * @param target The target element for the focus and blur events.
  * @param options
  */
-export function useFocus(options: FocusOptions = {}): FocusReturn {
-  const {
-    initialValue = false,
-  } = options
+export function useFocus(target: MaybeElementRef, options: FocusOptions = {}): FocusReturn {
```
wheatjs, line 30: Actually just had a thought, maybe we should rename `FocusOptions` to `UseFocusOptions` to also be consistent with our other APIs. Since this is already a breaking change we might as well.

wtykirby: Ok, I renamed the return type to be `UseFocusReturn` as well.

wheatjs (approved):

packages/core/useFocus/index.md
```diff
@@ -4,15 +4,15 @@ category: Sensors
...
+Reactive utility to track or set the focus state of a DOM element. State changes to reflect whether the target element is the focused element. Setting reactive value from the outside will trigger `focus` and `blur` events for `true` and `false` values respectively.
 
 ## Basic Usage
 
 ```ts
 import { useFocus } from '@vueuse/core'
 
-const focusableElement = ref();
-const { focused } = useFocus({ target: focusableElement })
+const target = ref();
```
FRSgit, line 14: If we change this name then let's add an additional info somewhere in the docs that only *focusable* elements will work with this composable. Let's spare potential users headache of "why it's not working" 😄

teena-dot-dev: > If we change this name then let's add an additional info somewhere in the docs that only _focusable_ elements will work with this composable. Let's spare potential users headache of "why is it not working" 😄

I'm following this PR keenly as I have run into an issue where I keep getting this error:
```
ERROR in ./node_modules/@vueuse/integrations/useFocusTrap.mjs 3:0-45
Module not found: Error: Can't resolve 'focus-trap' in '/home/teena/projects/****/node_modules/@vueuse/integrations'
```

So If I understood you correctly, you can only use this on focusable elements?

I'm building a modal, and will like to trap the focus in the modal when it is opened.

wtykirby: @teena-dot-dev I believe the error you are seeing is actually the result of a missing dependency. The `useFocusTrap` composable is just a reactive wrapper around the [focus-trap](https://github.com/focus-trap/focus-trap) library so you likely just need to install that dependency in order for it to work correctly.

For `useFocusTrap`, the target element does not actually need to be focusable. See it's documentation for more information on how it decides which element to focus when the focus trap is enabled.

teena-dot-dev: > @teena-dot-dev I believe the error you are seeing is actually the result of a missing dependency. The `useFocusTrap` composable is just a reactive wrapper around the [focus-trap](https://github.com/focus-trap/focus-trap) library so you likely just need to install that dependency in order for it to work correctly.
> 
> For `useFocusTrap`, the target element does not actually need to be focusable. See its documentation for more information on how it decides which element to focus when the focus trap is enabled.

So I'd need to install `focus-trap`?
It's strange because I used it before without errors. I recently updated my dependencies and all of a sudden I'm faced with this.
Thanks though.
I'll try it now and if it does work, I'll open an issue with more details

teena-dot-dev: Works again, thanks!

FRSgit: It's a bit of a pity that nobody responded me on this one :(

PR #1012, opened 2021-12-04
feat(useSubject): new function

### LINKED ISSUE #1028: feat(useSubject): new function
RxJS add-on already has **useObservable** hook that produces one-way data flow from observable into component's ref.

It would be handy to provide in similar manner a hook that can establish two-way communication between component refs and RxJS Subjects.

I've already created #1012 that covers this issue. Do you think it is worth adding?

peterekjs:
Add simple way for two-way binding between Vue.js refs and RxJS subjects

- Connecting BehaviorSubject will immediately use it's current value
- Connecting other Subject types will set Ref's value to undefined
- Update Ref value will propagate to the subject

Closes #1028


peterekjs:
Closes #1028

PR #1019, opened 2021-12-08
feat(useSticky): new function

webfansplz:
```ts
<script setup lang="ts">

import { UseSticky as USticky } from './component'
import { useSticky } from '.'

const el = ref<HTMLElement | null>(null)

const { isFixed } = useSticky(el, { offsetTop: 100 })

</script>

<template>
  <u-sticky
    v-slot="{isFixed}"
    :offset-top="50"
  >
    <div>
      Sticky State (Component): {{ isFixed }}
    </div>
  </u-sticky>

  <div ref="el">
    Sticky State (Hook) : {{ isFixed }}
  </div>
</template>

```


wheatjs:
How does this differ from just using `position: sticky`?

antfu:
Thank you. But I guess the native CSS sticky property would work more efficiently. Close this for now.

PR #1031, opened 2021-12-10
feat:usePageClose

lxhyl:
# feat:usePageClose

Show hint when close or reload page,if something does't save.


antfu:
Thanks. But I think this is simple enough to direct use `useEventListener`.

lxhyl:
Thank you. I understand

PR #1036, opened 2021-12-11
chore(useFullscreen): remove useless code

webfansplz:


antfu:
I think it's for demo purpose

PR #1037, opened 2021-12-11
feat(useCountDown): new function

okxiaoliang4:
countdown controller

[reference](https://github.com/youzan/vant/blob/dev/packages/vant-use/src/useCountDown/index.ts)


wheatjs:
I wonder if we want to use an event hook instead of an option for `onFinish` and `onChange`. Tbh I'm not sure when it is better to use one over the other? @antfu any thoughts on when we should use an event hook vs an option?

okxiaoliang4:
i well add options, in progress

antfu:
I am closing both for now. If you still think this is valuable, feel free to bring it up again. Thanks

PR #1044, opened 2021-12-12
fix(useElementByPoint): fix ssr

okxiaoliang4:


wheatjs:
Is this actually broken? I believe `useRafFn` will only run on the client

okxiaoliang4:
Try to access it directly, there will be exceptions [useElementByPoint](https://vueuse.org/core/useelementbypoint/)

```typescript
  // useRafFn
  function loop() {
    if (!isActive.value)
      return
    fn() // will throw a execption
    if (window)
      window.requestAnimationFrame(loop)
  }
```

wheatjs:
Hmm, guess so

antfu:
We could fix it in useRaf instead

PR #1045, opened 2021-12-12
feat(useMemoize): new function

### LINKED ISSUE #1024: [feat] useCachedData to cache result of async functions depending on args
Hey fellows,

I was creating some caching helper for API requests and wondering that there is currently no composable to keep this simple (Maybe I missed something?). Do you think this could fit into the library? My current intention is to resolve the data only once no matter if the request has been finished or is currently pending.

Here is my current draft. I just want to show you my idea and if you think this is good then I can create a PR:
```typescript
export const useCachedData = <Result, Args extends unknown[]> (
  resolver: (...args: Args) => Promise<Result>,
  options?: {
    getKey?: (...args: Args) => any
  }
) => {
  const cachedData = reactive(new Map<any, Promise<Result>>())

  /**
   * Generate key from args
   */
  const generateKey = (...args: Args) => options?.getKey
    ? options.getKey(...args)
    // Default key: Serialize args
    : JSON.stringify(args)

  /**
   * Load data and save in cache
   */
  const loadData = async (...args: Args): Promise<Result> => {
    const promiseData = resolver(...args)
    cachedData.set(generateKey(...args), promiseData)
    return promiseData
  }

  /**
   * Get data from cache or load if not cached
   */
  const getData = async (...args: Args): Promise<Result> => {
    // Get data from cache
    const cached = cachedData.get(generateKey(...args))
    if (cached) return cached

    return loadData(...args)
  }

  /**
   * Clear cached data
   */
  const clearData = () => {
    cachedData.clear()
  }

  return {
    cachedData,
    get: getData,
    load: loadData,
    clear: clearData,
    generateKey
  }
}
```

Here is a usage example
```typescript
const users = useCachedData(
  async (userId: number): Promise<UserData> => axios.get(`users/${userId}`)
    .then(response => response.data)
)

const user1 = await users.get(1) // Requesting user 1
const user2 = await users.get(2) // Requesting user 2

// ...

const user1 = await users.get(1) // Retrieving user 1 from cache
const user2 = await users.get(2) // Retrieving user 2 from cache
```

This can also be used for example with a [pinia](https://pinia.esm.dev/) store:
```typescript
const useUserStore = defineStore('users', () => useCachedData(
  async (userId: number): Promise<UserData> => axios.get(`users/${userId}`)
    .then(response => response.data)
))
```

Some more ideas:
* `Map` should be replaceable with another class to be able to use for example [lru-cache](https://www.npmjs.com/package/lru-cache)
* A method to invalidate the cache of a specific key

freakzlike:
Cache results of functions depending on arguments and keep it reactive. It can be used for both asynchronous and synchronous functions.

```ts
import { useMemoize } from '@vueuse/core'

const users = useMemoize(
  async (userId: number): Promise<UserData> =>
    axios.get(`users/${userId}`).then(({ data }) => data)
)

const user1 = users.get(1) // Request users/1
const user2 = users.get(2) // Request users/2
// ...
const user1 = users.get(1) // Retrieve from cache

// ...
const user1 = users.load(1) // Request users/1

// ...
users.delete(1) // Delete cache from user 1
users.clear() // Clear full cache
```

Closes #1024


packages/core/useMemoize/index.ts
```diff
@@ -0,0 +1,83 @@
...
+export interface MemoizeCache<Key, Value> {
+  // Get value for key
+  get (key: Key): Value | undefined
+  // Set value for key
+  set (key: Key, value: Value): void
+  // Return flag if key exists
+  has (key: Key): boolean
+  // Delete value for key
+  delete (key: Key): void
+  // Clear cache
```
antfu, line 17: Could we use jsdoc to explain each API? (you can refer to other functions

antfu (commented):
Thanks for the detailed PR!

freakzlike:
I have changed the return value so it is possible to directly call the function without the need of `.get()`.

import { useMemoize } from '@vueuse/core'

const getUser = useMemoize(
  async (userId: number): Promise<UserData> =>
    axios.get(`users/${userId}`).then(({ data }) => data)
)

const user1 = getUser(1) // Request users/1
const user2 = getUser(2) // Request users/2
// ...
const user1 = getUser(1) // Retrieve from cache

// ...
const user1 = users.load(1) // Request users/1

// ...
users.delete(1) // Delete cache from user 1
users.clear() // Clear full cache

packages/core/useMemoize/index.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+### Customize cache mechanism
+By default, the results are cached within a `Map`. You can implement your own mechanism by passing `cache` as options with following structure:
+```ts
+export interface MemoizeCache<Key, Value> {
+  /**
+   * Get value for key
+   */
+  get (key: Key): Value | undefined
```
antfu, line 67:
```suggestion
  get (key: Key, force?: boolean): Value | undefined
```

How about adding a second option to bypass the cache?

freakzlike: The `MemoizeCache` is only the container which stores the result and it uses the same interface as the [`Map` class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).
When you want to bypass the cache you can call for example `getUser.load(1)`. This will always call the function and overwrite the previous result in cache

packages/core/useMemoize/index.ts
```diff
@@ -0,0 +1,118 @@
...
+ */
+export interface MemoizedFn <Result, Args extends unknown[]> {
+  /**
+   * Get result from cache or call memoized function
+   */
+  (...args: Args): Result
+  /**
+   * Call memoized function and update cache
+   */
+  load (...args: Args): Result
```
antfu, line 42: Do you think we should support async? Or explicit `asyncGet` and `asyncLoad`? I think for many cases, the things we want to cache are time or computation heavy tasks, many of them might be async.

freakzlike: It is already possible to pass an async function. In this case the promise gets cached and will be returned no matter if still pending or fulfilled. I did not at the `async` so it can also be used for non async functions.

packages/core/useMemoize/index.ts
```diff
@@ -0,0 +1,118 @@
...
+ * Reactive function result cache based on arguments
+ */
+export const useMemoize = <Result, Args extends unknown[]> (
+  resolver: (...args: Args) => Result,
+  options?: {
+    getKey?: (...args: Args) => CacheKey
+    cache?: MemoizeCache<CacheKey, Result>
+  },
+): MemoizedFn<Result, Args> => {
+  const cache = reactive(options?.cache || new Map<CacheKey, Result>())
```
antfu, line 71: Vue 2 does not support making `Map` reactive. We would need either to get a around of not using it, or make this function as Vue 3 only

freakzlike: I've added a fallback for Vue 2 to use a custom object instead of a `Map`. If you are happy with that, then I will squash the commits

PR #1046, opened 2021-12-12
feat(useNavigatorLanguage): new function

michael-observerly:
Added the useNavigatorLanguage() composable which listens to the "onlanguagechange" event and returns the current main user preference for language. 

Utilises useEventListener() from @vueuse/core for the event listening.


michael-observerly:
Closes #1047

PR #1048, opened 2021-12-12
feat(useSelections): new function

webfansplz:
This hook is used for Checkbox group, supports multiple selection, single selection, select-all, select-none and semi-selected.

```ts
import { useSelections } from '@vueuse/core'

const data = [1, 2, 3, 4, 5]
const { selected, noneSelected, allSelected, partiallySelected, isSelected, toggle, selectAll, unSelectAll } = useSelections(data, [1, 2, 3])
```


antfu:
I am not sure about this. It seems a bit too verbose to me. We might need a cleaner abstraction.

antfu:
I think this might be better to use a reactive Set to solve. Closing this for now. Thanks.

PR #1050, opened 2021-12-13
fix punctuation

kimulisiraj:


antfu:
Thanks. But is intended.

PR #1056, opened 2021-12-14
feat(useFetch): allow useFetch to be awaited

wheatjs:
Allows `useFetch` to be awaited. For example

```html
<script setup lang="ts">

const { data }  = await useFetch(url)

console.log(data.value) // outputs resolved data
</script>
```

### DIFF
--- a/packages/core/useFetch/index.test.ts
+++ b/packages/core/useFetch/index.test.ts
@@ -324,4 +324,20 @@ describe('useFetch', () => {
 
     expect(data.value).toStrictEqual(JSON.stringify({ message: 'Hello World' }))
   })
+
+  test('should await request', async() => {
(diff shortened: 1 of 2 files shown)


kissu:
So, thanks to that we will be able to affect something directly to a variable, like
```js
variable = data.some.nested.field
```
or will it look like this rather?
```js
variable = data.value.some.nested.field
```

wheatjs:
It will still be a ref, so you will have to use it like `data.value.some.nested.field`. But when you use it, you won't have to worry about using `watch` or `until` to check if it has data

kissu:
Alright, perfect thanks ! 🙏🏻

antfu:
Cool

starkana:
`createFetch()` does not return a `useFetch()` function that can be awaited, will this be updated anytime soon?

PR #1057, opened 2021-12-14
fix(useWebSocket): close WebSocket gently

lishangbu:
The code must be either 1000, or between 3000 and 4999. 1012 is neither, too.
Related #1018
Related #995

### DIFF
--- a/packages/core/useWebSocket/index.ts
+++ b/packages/core/useWebSocket/index.ts
@@ -157,8 +157,8 @@ export function useWebSocket<Data = any>(
 
   let bufferedData: (string | ArrayBuffer | Blob)[] = []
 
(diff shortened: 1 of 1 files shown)


lishangbu:
In version ```7.3.0```,it throws Error ```Failed to execute 'close' on 'WebSocket': The code must be either 1000, or between 3000 and 4999. 1012 is neither.``` when we attempt to close websocket connection or reconnect it. This code must be 1000(for normal closure) ,3000-3999(customized by library/framework),4000-4999(customized by application) according to [the IANA website](https://www.iana.org/assignments/websocket/websocket.xml#close-code-number)

antfu:
@Scrum WDYT?

Scrum:
> @Scrum WDYT?

Yes, it sounds quite logical. This is better than my suggestion. we are really closing the connection because it worked correctly.

Approved.

PR #1060, opened 2021-12-14
feat: ssr handlers

antfu:
This PR introduced a tiny layer for ssr context in `@vueuse/core`. The main point of it is to make a way to provide custom SSR handling logic for functions like `useDark` or `useStorage` that rely on the client state.

The internal layer should be handled by the framework integrations like `@vueuse/nuxt`, which allow some composable functions to work out-of-box across the client and server-side.

The gist of usage would be like:

```ts
// before VueUse usages
import { setSSRContext } from '@vueuse/core'

if (process.server) {
  setSSRContext('updateHTMLAttrs', (selector: string, attribute: string, value: string) => {
    // inject tags to SSR HTML
  }) 
}
```

Combine with swapping the storage to cookies, it will allow the app to render the correct color mode preventing color mode flashing on the client-side.

### Currently exposed APIs

> APIs are experimental and not finalized

https://github.com/vueuse/vueuse/blob/feat/ssr/packages/ssr-context/index.ts

### TODO

- [x] Provide the SSR context in `@vueuse/nuxt` for Nuxt (under a flag)
- [ ] Consider more functions
  - [x] `useColorMode`
  - [x] `useStorage`
  - [x] `useStorageAsync`
  - [ ] `useWindowsSize`
  - [ ] `useMediaQuery`
  - [ ] `useNavigatorLanguage`
  - [ ] `useBreakpoints`


packages/nuxt/ssr-plugin.mjs
```diff
@@ -0,0 +1,48 @@
...
+      removeItem: (key) => {
+        if (!cookieMap.get(key))
+          cookieMap.set(key, useCookie(key))
+        cookieMap.get(key).value = undefined
+      },
+    }
+  })
+
+  if (process.server) {
+    setSSRContext('updateHTMLAttrs', (selector, attr, value) => {
```
pi0, line 28: This is pinning ssr context to a global shared variable. context should available in non-async executation and unset afterwards.

Issue #1026, opened 2021-12-09
Can we use `titleTemplate` like property (nuxt) with useTitle() method?


anburocky3:
can we use `titleTemplate` like feature in `useTitle()` ?
I'm saying like it: `Dynamic title: AppName`

michael-observerly:
Unfortunately, not at this moment. Looking through the source code it might be possible to introduce a titleTemplate property, but it's a little tricky because currently title is a ref ... so it can be set directly, with `title.value = "My Awesome Website"`. 

It might be more beneficial to create a new `useTitleTemplate()` composable that wraps the `useTitle()` composable.

cc. @antfu Question: RE: above comment, is this something you would be happy to accept as a PR? I can always take that workload if you're happy for me to ...

michael-observerly:
@anburocky3 I've made a PR for this here: https://github.com/vueuse/vueuse/pull/1062 I'm not core team, so it's subject to acceptance from Anthony et al.. Maybe have a play locally with it and let me know if that is what you envisioned.

I've followed the guidance from here: https://vue-meta.nuxtjs.org/guide/metainfo.html

michael-observerly:
@anburocky3 Feedback from core team was that it could be included inside useTitle without breaking changes. PR here for reference: #1087

michael-observerly:
@anburocky3 Looks like my PR addressing your use case has now been merged into core, expect it in the 7.5.0 release 👍

anburocky3:
Great, Looking forward. Thanks for your contribution.

michael-observerly:
@anburocky3 Usage will be added to useTitle().

anburocky3:
@michael-observerly @antfu 

Can we have an example of that?

Do i need to use 
```ts
const title = useTitle('New Title', { titleTemplate: '%s | My Awesome Website' })
```

Always? Is that so? what is the use of `titleTemplate`

**I currently did like these:**
1. Created nux3 project and created `layouts/default.vue` file and placed the above snippet there.
2. On my pages, I just used `useTitle('Homepage')`, but only `Homepage` is appearing, `titleTemplate`'s content is not appearing.

May be i might have done wrong. But i don't see any additional information on the docs.

PR #1062, opened 2021-12-14
feat(useTitleTemplate): new function

michael-observerly:
Added the useTitleTemplate() composable which wraps the useTitle() composable to follow the https://vue-meta.nuxtjs.org/guide/metainfo.html style of specifying the title.


michael-observerly:
Closes #1026

antfu:
How about adding a template options in `useTitle` directly?

michael-observerly:
> How about adding a template options in `useTitle` directly?

Yeh I think that might be a better alternative ... I'll see if I can implement it in a non-breaking way. Happy for you to close this PR.

PR #1075, opened 2021-12-18
docs: asynchronous functions

wheatjs:
Added documentation for creating asynchronous composables


packages/guidelines.md
```diff
@@ -157,6 +158,50 @@ export function useShare(
 }
 ```
 
+## Asynchronous Composables
+
+When a composable is asynchronous, like `useFetch`, it is a good idea to return a PromiseLike object from the composable
+so the user is able to await the function. This is especially useful in the case of Vue's `<Suspense>` api.
+
+- Use a `ref` to determine when the function should resolve e.g. `isFinished`
+- Store the return state in a varaible as it must be returned twice, once in the return and once in the promise.
```
userquin, line 167: typo on `varaible`:

Store the return state in a `variable` as it must be returned twice, once in the return and once in the promise.

PR #1076, opened 2021-12-19
feat(usePageVisibility): new function

michael-observerly:
Added the usePageVisibility() composable which listens to when a document becomes visible or hidden (as well as keeping a track of the page state, if needed).

Includes associated usage guidelines and a base test.

Includes updates the packages/functions.md for usePageVisibility.

### DIFF
--- a/packages/core/usePageVisibility/demo.vue
+++ b/packages/core/usePageVisibility/demo.vue
@@ -1,0 +1,9 @@
+<script setup lang="ts">
+import { usePageVisibility } from '.'
+
+const { visibilityState } = usePageVisibility()
+</script>
+
+<template>
+  <div>Page Visibility: <b>{{ visibilityState }}</b></div>
+</template>
--- a/packages/core/usePageVisibility/index.md
+++ b/packages/core/usePageVisibility/index.md
@@ -9,15 +9,24 @@ Reactive page visibility state.
 ## Usage
 
 ```ts
+import { defineComponent, ref } from 'vue'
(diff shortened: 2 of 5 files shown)


darrylnoakes:
Should the following issue with the `visibilitychange` event in Safari be mentioned in the docs for this?
From the [Can I Use table](https://caniuse.com/mdn-api_document_visibilitychange_event):

>  Doesn't fire the `visibilitychange` event when navigating away from a document, so also include code to check for the `pagehide` event (which does fire for that case in all current browsers). See WebKit bugs [116769](https://webkit.org/b/116769), [151234](https://webkit.org/b/151234), [151610](https://webkit.org/b/151610), and [194897](https://webkit.org/b/194897).

It could also listen for the `pagehide` event, and then check `hidden`?

wheatjs:
I believe this function already exists at [useDocumentVisibility](https://github.com/vueuse/vueuse/blob/f65707876e1d93211c44414c2a30dc90b1178d68/packages/core/useDocumentVisibility/index.ts). Going to close unless you think this is a mistake and we can reopen.

michael-observerly:
> I believe this function already exists at [useDocumentVisibility](https://github.com/vueuse/vueuse/blob/f65707876e1d93211c44414c2a30dc90b1178d68/packages/core/useDocumentVisibility/index.ts). Going to close unless you think this is a mistake and we can reopen.

Ahhh, yes, a lot of overlap for sure. Happy to close.

Issue #1014, opened 2021-12-06
useDraggable add a boundary


btea:
Is it possible to add an attribute to control the dragged element not to exceed the boundary of the parent element?

wheatjs:
You can combine multiple composable to get this working. Take a look at this stackblitz project to see https://stackblitz.com/edit/vitejs-vite-kvx39r?file=src/App.vue (Just use `npm run dev` in the terminal to start the project).

```html
<script setup lang="ts">
import { ref, computed } from 'vue';
import { useDraggable, useElementBounding, clamp } from '@vueuse/core';

const container = ref();
const draggable = ref();

const { left, right, top, bottom } = useElementBounding(container);
const { width, height } = useElementBounding(draggable);
const { x, y } = useDraggable(draggable);

const restrictedX = computed(() =>
  clamp(left.value, x.value, right.value - width.value)
);
const restrictedY = computed(() =>
  clamp(top.value, y.value, bottom.value - height.value)
);
</script>

<template>
  <div
    :style="{
      height: '100vh',
      background: 'teal',
      display: 'grid',
      placeItems: 'center',
    }"
  >
    <div
      :style="{
        background: 'white',
        width: '90%',
        height: '500px',
        position: 'relative',
      }"
      ref="container"
    >
      <div
        ref="draggable"
        :style="{
          userSelect: 'none',
          position: 'fixed',
          width: '100px',
          height: '100px',
          background: 'red',
          top: `${restrictedY}px`,
          left: `${restrictedX}px`,
        }"
      >
        I am draggable
      </div>
    </div>
  </div>
</template>

```

btea:
OK, I'll try. Thank you

laterdays:
@wheatjs   fixed does not fit the scene, Whether absolute can be supported

PR #1081, opened 2021-12-21
feat(add,subtract,multiply,divide): new functions

wheatjs:
Reactive operators for addition, subtraction, multiplication, and division.

Sometimes when using composable you will need to do basic operations such as addition. This can make simple things a bit more verbose, for example this https://github.com/vueuse/vueuse/issues/1014#issuecomment-986530188 could be simplified to

```html
<script setup lang="ts">
...

const restrictedX = useClamp(left.value, x.value, subtract(right.value, width.value))
const restrictedY = useClamp(top.value, y.value, subtract(bottom.value, height.value))
</script>
```
Not having to create a computed for simple arithmetic is pretty convenient I think, but I'm not really sure if this is a great idea, so I'd like some feedback.

cc @antfu @patak-dev


antfu:
The naming sounds a bit too general, how about `numericAdd`, `numericDivide` etc?

wheatjs:
Hmm are you thinking it might conflict with other functions is users codebases or that the naming might confuse people?

patak-dev:
Interesting. Maybe under an object like `reactive.add()`?

At one point I thought that we wouldn't use ref sugar in a library like VueUse, but I think that it may actually make the code a lot more maintainable in some of the composables. Maybe we should push for that?

```js
import { useDraggable, useElementBounding, clamp } from '@vueuse/core';

const container = $ref();
const draggable = $ref();

const { left, right, top, bottom } = $(useElementBounding(container));
const { width, height } = $(useElementBounding(draggable));
const { x, y } = $(useDraggable(draggable));

const restrictedX = $computed(() => clamp(left, x, right - width));
const restrictedY = $computed(() => clamp(top, y, bottom - height));
```

wheatjs:
I do think that the syntax sugar can help, but there are still situations where it might be not be useful. For instance if we have a function that take a ref number as a parameter, we would have to have a computed to use that. Like `useClamp` for example, I think it might not be unusual to have one of the parameters be something the user will need to calculate. To me having something like `add`, etc, would be much more "composable" then using a computed with a generic clamp function. Maybe I'm incorrect in this thinking though. 

To me though something like 
```ts
const x = useClamp(...)
``` 
is much more clear in an initial scanning of the code that `x` is a clamped value than something like 
```ts
const x = computed(() => clamp(...))
```

michael-observerly:
> The naming sounds a bit too general, how about `numericAdd`, `numericDivide` etc?

I feel like add, subtract, multiply and divide as a set are self-explanatory ... do we apply these mathematical operations to any other type other than number?

antfu:
Btw, you might want to try this
https://github.com/vueuse/vue-chemistry

antfu:
On a second thought, I think it's better to have them live in vue-chemistry.

PR #1085, opened 2021-12-22
feat(useBroadcastChannel): new function

michael-observerly:
Added the useBroadcastChannel() composable.

Includes associated demo (requires the user to have at least two tabs open at once) - adds a broadcast alert message across tabs.

Includes is defined test and usage examples.


Anoesj:
Quick question @michealroberts and @antfu. Theoretically, if multiple messages are sent really fast after each other by another subscriber of the BroadcastChannel, the `data` ref gets updated a few times in a row. Wouldn't that mean that there's a possibility of missing messages in a `data` watcher, as the Vue tick system will only remember the last message sent in that tick?

If so (I haven't tried, but I was wondering), I could update the docs to use `flush: 'sync'` and maybe make sure this composable provides a data watcher that uses `flush: 'sync'` itself, as that would definitely get rid of the problem :smile:

PR #1086, opened 2021-12-22
feat(useMouseInElement): added element distance to center and edge calcs

michael-observerly:
Added non-breaking change to detect distance between { x, y } and an element's centre point, as well as detecting the closest distance between { x, y } and the div's edge / bounding client*. 

Distance to boundary is returned as 0 when inside the rectangle.

This calculation is rotation agnostic*.

This was something that recently came up as a production requirement for a project I am working on, and would be really useful if this was in core.

I was also in two minds whether to create a new "wrapper" around useMouseInElement as this is better described as useMouseRelativeToElement ...

Thoughts and suggestions are warmly welcomed. Enjoy! :)

### DIFF
--- a/packages/core/useMouseInElement/index.ts
+++ b/packages/core/useMouseInElement/index.ts
@@ -1,4 +1,4 @@
-import { ref, watch } from 'vue-demi'
+import { computed, ref, watch } from 'vue-demi'
 import type { MaybeElementRef } from '../unrefElement'
 import { unrefElement } from '../unrefElement'
 import type { MouseOptions } from '../useMouse'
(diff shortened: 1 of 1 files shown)


antfu:
Since computed are lazy, it should not impact ppl who don't use it. The main concern is it increases the sizes for everyone using this function, and also introduces extra complexity. I think it would be great if you can provide more context about why this could be generally used, or otherwise I feel it's might be better to live in your project code base.

michael-observerly:
> Since computed are lazy, it should not impact ppl who don't use it on perf. The main concern is it increases the sizes for everyone using this function, and also introduces extra complexity. I think it would be great if you can provide more context about why this could be generally used, or otherwise I feel it's might be better to live in your project code base.

That's understandable, I think it would be beneficial to create a new composable then that wraps useMouseInElement. I will PR that today. 

My justification is simple, you can essentially detect the distance from the element for the coordinate of the current user's mouse ... so you could have some interactive effects based off of this calculation ... it acts as a kind of "predictive hover" ... user is moving towards an element ... useful for say, animating a CTA or similar when the user is close (but not hovering).

michael-observerly:
Closing, I've opened up a new PR for a wrapper function: #1094

PR #1091, opened 2021-12-23
feat: converts the rest of the vitest tests

JessicaSachs:
This was a great exercise in using Vitest! I think this is the last of the issues.


dheimoz:
great work @JessicaSachs. This is like an advance course of Testing in General. It might worth be part of Vitest documentation as a reference.
Thanks for taking the time to do this.

PR #1092, opened 2021-12-23
refactor: migrate to vitest

antfu:
Thanks a lot to @JessicaSachs 🙌


JessicaSachs:
So I was getting an issue before I upgraded to 0.0.108 where those tests would fail. (the "tests magically work now" commit). 

I think the upgrade to 0.0.108 made `vitest` fast enough that the race condition isn't apparent on my machine, but it still fails in CI (slower CPU, etc). All of the tests that are broken mock or other-wise reset window/DOM state. e.g. `document.location`, `window.localStorage`, `window.navigator`, `document.body`

To reproduce the failure, you must run all tests together.  Most of the time I spent was trying to debug why those tests were failing. I couldn't figure it out. I think, environmentally, this environment isn't as pure/sandboxed as jest's jsdom resets.

PR #1094, opened 2021-12-23
feat(useMouseDistanceToElement): new function

michealroberts:
This composable returns the scalar distance in pixels to the nearest edge point of an element relative to some mouse position { x, y }.

This useful when there is a need to detect use intentions - and could be utilised for lazy loading routes or resources by predictively detecting where a user is "going" prior to hovering (e.g., something like as a user moves close to a collapsed sidebar, as the user moves towards it it reveals).

This could also be used in the context of animations ... detecting intentions of a user, animating CTAs as a user moves "close" to the element etc.


michael-observerly:
@webfansplz Sorry to notify directly, but I see you are quite active at the moment in this repo ... it would be nice to get some feedback about this ... I feel like this would be really useful for detecting the pixel distance between the current mouse position to any element * **boundary** * on the page.

michael-observerly:
cc. @okxiaoliang4

okxiaoliang4:
Thank your contribute. But i can't review your pr,I'm out few days

michael-observerly:
> Thank your contribute. But i can't review your pr,I'm out few days sorry.

No worries, thank you for the prompt response. :)

webfansplz:
Thanks for the PR.

I'm not so sure about it, it seems just provide two computed state based on `useMouseInElement`.

If ppl has similar needed, perhaps the combination of states provided based on `useMouseInElement` would be more flexible ?

michael-observerly:
@webfansplz Totally understand, the original PR was #1086 which did exactly that. 

But with discussions with Anthony, it was decided for performance reasons* that we should maybe place this in an extended function to be used if needed ... or in a private library. However, I feel that this would be useful to add ... perhaps maybe it should just return the calculation needed for getting the distance to the boundary.

*(for me, on my Mac M1 I see minimal performance degradation, but maybe for other devices ... it's obviously heavy).

michael-observerly:
@webfansplz Also, I totally understand the combination aspect, but the maths is not entirely trivial or easy to calculate ... I have two degrees (Physics and Astrophysics) and it wasn't trivial to sketch out the problem ... and reduce to what I have added. (It involves understand the space around the box, relative to the box, and disseminating various sections around the box). 😝

okxiaoliang4:
Maybe we can add`useDistance(pointA: MaybeRef<Point>, pointB: MaybeRef<Point>)`in this pr

michael-observerly:
@okxiaoliang4 @webfansplz Yeh that could work, but the caveat is that it is hard to detect say Point A, the closest point to an element's boundary ... imagine a line from the center of the element, to PointB. PointA is on that line, but it's not a trivial calculation to know where PointA is :)

michael-observerly:
I'll provide a quick diagram ... * one moment

okxiaoliang4:
I mean maybe can add a other function `useDistance` in this pr :)

michael-observerly:
@okxiaoliang4 Ahhh ok, yeh I agree with that. I will refactor a few bits for that and get it pushed. I'll maybe wait for @antfu 's final decision on this before.

okxiaoliang4:
We may release a new package `@vueuse/math`. https://github.com/vueuse/vueuse/pull/1224#issuecomment-1034560523

michael-observerly:
Frame 2

michael-observerly:
> We may release a new package `@vueuse/math`. [#1224 (comment)](https://github.com/vueuse/vueuse/pull/1224#issuecomment-1034560523)

Yeh, maybe release the calculation in this package, and utilise it within this PR?

okxiaoliang4:
Wait for antf opinion :)

michael-observerly:
Yeh :) for sure, thank you for all of the discussions. I know I'm fighting hard for it but I think it's a rather elegant solution :)

michael-observerly:
@antfu Any thoughts on this?

michealroberts:
@antfu @okxiaoliang4 Closing this now. Disappointing.

PR #1101, opened 2021-12-26
feat(useForceUpdate): new function

okxiaoliang4:
see https://github.com/vuejs/vue-next/issues/3562


antfu:
Does this work on Vue 2?

antfu:
And yes I agree with the discussion in the thread. We should not recommend this in general. Giving it's fairly easy to implement in the user land, I am closing it. Thanks.

PR #1107, opened 2021-12-28
feat:(useLoading):new fucntion

macheteHot:
use case
```js
const [isLoading, useThisFn] = useLoadingFn(async() => {
  // await  some thing
})
```


wheatjs:
Seems like a duplicate of https://vueuse.org/core/useasyncstate/

macheteHot:
ohhh my bad.🤣

macheteHot:
Is not same like useasyncstate
when you call new function  will automatically toggle boolean `isLoading` usually use to  waiting status like waiting button , waiting some fetch

wheatjs:
Wouldn't it be functionally the same as

```ts
const { state, isReady, execute } = useAsyncState((args) => sleep(5000),
  {},
  { immediate: false },
)
```
You can then use `execute` as the instead of `useThisFn`

macheteHot:
Is not same  
use loading first time toggle `isLoading` value 
wd66o-idw03

wheatjs:
Seems to me that isLoading is just the inverted state of isReady. Is there any other difference besides that?

macheteHot:
When the function finishes executing  `isReady` is aways be true unless re-run 
if the button bind `isReady` to loading prop  will block the run

isReady => true => false 

isLoading => false => true => false

wheatjs:
Ah well if that is the case, then I suggest we add an `isLoading` property to `useAsyncState` instead

macheteHot:
> Ah well if that is the case, then I suggest we add an `isLoading` property to `useAsyncState` instead

This is a brilliant idea

PR #1117, opened 2021-12-31
chore(useMediaControls): add some functions that can control media better

SnowingFox:
You can control media better by using it
```js
const { controlMedia } = useMediaControls()
const { play, pause, stop } = controlMedia()
play() // Enable media to play
pause() // Enable media to pause
stop() // Equivalent to pause()
```

### DIFF
--- a/packages/core/useMediaControls/index.md
+++ b/packages/core/useMediaControls/index.md
@@ -15,7 +15,7 @@ import { onMounted, ref } from 'vue'
 import { useMediaControls } from '@vueuse/core'
 
 const video = ref()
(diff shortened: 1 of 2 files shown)


wheatjs:
Hmm I'm not really sure we need this. We are already able to control the play state via the `isPlaying` property https://github.com/vueuse/vueuse/blob/main/packages/core/useMediaControls/index.ts#L374-L384

SnowingFox:
> Hmm I'm not really sure we need this. We are already able to control the play state via the `isPlaying` property https://github.com/vueuse/vueuse/blob/main/packages/core/useMediaControls/index.ts#L374-L384

OK, thanks

PR #1119, opened 2021-12-31
fix(useAsyncState): improve types

macheteHot:
fix state error type


packages/core/useAsyncState/index.ts
```diff
@@ -61,20 +61,33 @@ export interface AsyncStateOptions {
  * @param initialState    The initial state, used until the first evaluation finishes
  * @param options
  */
+
+export function useAsyncState<T>(
+  promise: Promise<T> | ((...args: any[]) => Promise<T>),
+  initialState: T,
+  options: Omit<AsyncStateOptions, 'shallow'> & { shallow: false }): UseAsyncStateReturn<T, false>
```
antfu, line 68: Pass the true/false as a generic of `AsyncStateOptions`

macheteHot: Thanks a lot for your advice

JonasWijne:
@antfu and @macheteHot  can you check if this is right, it looks like  ShallowRef should be shallowRef,

PR #1123, opened 2022-01-04
feat(useDateFormat): new function 

webfansplz:
Related to  https://github.com/vueuse/vueuse/pull/1035#issuecomment-993013769


packages/shared/useDateFormat/index.md
```diff
@@ -0,0 +1,47 @@
...
+| `SSS`  | 000-999          | The millisecond, 3-digits             |
+
+## Usage
+
+```html
+<script setup lang="ts">
+import { ref, computed } from 'vue-demi'
+import { useDateFormat } from '@vueuse/core'
+
+const now = ref(Date.now())
```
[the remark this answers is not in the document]
antfu: Also let's example how to pass the format string

packages/shared/useDateFormat/demo.vue
```diff
@@ -0,0 +1,15 @@
+<script setup lang="ts">
+import { computed, ref } from 'vue-demi'
+import { useDateFormat } from '.'
+
+const now = ref(Date.now())
+const formatted = computed(() => useDateFormat(now))
```
antfu, line 6: Would be great to have a input box allowing user to play with the formatter

packages/shared/useDateFormat/index.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Get the formatted date according to the string of tokens passed in.
+ *
+ * @see https://vueuse.org/useDateFormat
+ * @param date
+ * @param formatStr
+ */
+
+export function useDateFormat(date: MaybeRef<UDate>, formatStr = 'HH:mm:ss') {
+  const parsedDate = parseDate(unref(date))
+  return formatDate(parsedDate, formatStr)
```
antfu, line 70: It should return a computed ref, instead of asking user to do that

okxiaoliang4:
Maybe formatter function move to `@vueuse/shared`?

webfansplz:
> Maybe formatter function move to `@vueuse/shared/utils`?

I'm not sure if this abstraction is necessary, because `useDateFormat` basically covers usage scenarios.

Maybe we can just marked it as `TODO`  And then migrate as needed.

packages/shared/useDateFormat/index.ts
```diff
@@ -0,0 +1,72 @@
+import type { MaybeRef } from '@vueuse/shared'
+import { computed, unref } from 'vue-demi'
+
+export type UDate = Date | number | string | undefined
```
antfu, line 4:
```suggestion
export type DateLike = Date | number | string | undefined
```

packages/shared/useDateFormat/index.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const REGEX_PARSE = /* #__PURE__ */ /^(\d{4})[-/]?(\d{1,2})?[-/]?(\d{0,2})[Tt\s]*(\d{1,2})?:?(\d{1,2})?:?(\d{1,2})?[.:]?(\d+)?$/
+const REGEX_FORMAT = /* #__PURE__ */ /\[([^\]]+)]|Y{1,4}|M{1,4}|D{1,2}|d{1,4}|H{1,2}|h{1,2}|a|A|m{1,2}|s{1,2}|Z{1,2}|SSS/g
+
+export const padStart = (string: number, length: number, pad: string) => {
```
antfu, line 9: Can't we just use padStart from the string prototype? If not, we'd better to move this to utils

packages/shared/useDateFormat/index.ts
```diff
@@ -0,0 +1,72 @@
...
+    m: String(minutes),
+    mm: padStart(minutes, 2, '0'),
+    s: String(seconds),
+    ss: padStart(seconds, 2, '0'),
+    SSS: padStart(milliseconds, 3, '0'),
+  }
+  return formatStr.replace(REGEX_FORMAT, (match, $1) => $1 || matches[match])
+}
+
+export const parseDate = (date: UDate) => {
```
antfu, line 43:
```suggestion
export const normalizeDate = (date: DateLike) => {
```

packages/shared/useDateFormat/index.ts
webfansplz: > Can't we just use padStart from the string prototype? If not, we'd better to move this to utils

I'm not so sure about that, because it may have poor compatibility, do we need to consider it ?

okxiaoliang4: > > Can't we just use padStart from the string prototype? If not, we'd better to move this to utils
> 
> I'm not so sure about that, because it may have poor compatibility, do we need to consider it ?

ES2017 include `String.prototype.padStart`
https://github.com/vueuse/vueuse/blob/main/tsconfig.json#L3

PR #1125, opened 2022-01-04
feat(useHash): new function

michealroberts:
Added a reactive useHash function which can reactive two-way bind the window.location.hash to a hash Vue ref().

### DIFF
--- a/packages/core/useHash/demo.vue
+++ b/packages/core/useHash/demo.vue
@@ -1,0 +1,9 @@
+<script setup lang="ts">
+import { useHash } from '.'
+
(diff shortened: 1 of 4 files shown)


wheatjs:
Seems like this is already covered by https://vueuse.org/core/useBrowserLocation/?

michael-observerly:
@wheatjs Yeh fair, I was looking at specifics around hash internal routing ... and making it easy to implement (we implement it all the time in landing pages).

I'm happy to close this issue if you feel it is superfluous to useBrowserLocation() composable ...

PR #1137, opened 2022-01-12
fix(useWebSocket): close WebSocket

Scrum:
The `beforeunload` event is called with the passing of the event parameters instead of code.

Related #1057 
Related #1018
Related #995


antfu:
Oh, then we should do `'beforeunload', ()=>close()` instead

Scrum:
> Oh, then we should do `'beforeunload', ()=>close()` instead

This is what I originally wanted, but this does not guarantee against new or unintended calls.

antfu:
TypeScript should make sure the input types, otherwise it would be a misusage (and you can't capture all the possiblities at runtime)

antfu (approved):

PR #1141, opened 2022-01-17
feat(useBreakpoints): return readonly reactive object

hunterliu1003:
I'm not sure if this is a good practice to return a reactive object. 
But I wanna directly use `breakpoints.md` in template.

```vue
<script setup lang="ts">
import { breakpointsTailwind, useBreakpoints } from '.'

const breakpoints = useBreakpoints(breakpointsTailwind)
</script>

<template>
  <div class="font-mono">
    <div> sm: <BooleanDisplay :value="breakpoints.sm" /></div>
    <div> md: <BooleanDisplay :value="breakpoints.md" /></div>
    <div> lg: <BooleanDisplay :value="breakpoints.lg" /></div>
    <div> xl: <BooleanDisplay :value="breakpoints.xl" /></div>
  </div>
</template>
```

### DIFF
--- a/packages/core/useBreakpoints/index.ts
+++ b/packages/core/useBreakpoints/index.ts
@@ -1,4 +1,5 @@
 import type { Ref } from 'vue-demi'
+import { reactive, readonly } from 'vue-demi'
 import { increaseWithUnit } from '@vueuse/shared'
 import { useMediaQuery } from '../useMediaQuery'
 import type { ConfigurableWindow } from '../_configurable'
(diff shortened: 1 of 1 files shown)


wheatjs:
In general we prefer to return a ref instead of a reactive in VueUse. I think this change would introduce a breaking change and would go against our general practices. As an alternative what you could do is wrap `useBreakpoints` in a `reactive` when using it like
```ts
const breakpoints = reactive(useBreakpoints())
```

mesqueeb:
@wheatjs 
> In general we prefer to return a ref instead of a reactive in VueUse.

I wonder why? Would love to learn something new : )

wheatjs:
> @wheatjs
> 
> > In general we prefer to return a ref instead of a reactive in VueUse.
> 
> I wonder why? Would love to learn something new : )

This article might be helpful https://antfu.me/posts/composable-vue-vueday-2021#ref-vs-reactive

PR #1142, opened 2022-01-17
docs: new theme

antfu:


packages/functions.md
dmaixner93, line 1: I'm new to using VueUse, so i'm not super familiar with the source code. But I believe this is causing the 'Functions' link and related dropdown items to lead to a blank page. (screenshot and url below).

Screenshot 2023-07-26 at 1 39 15 PM

[Functions URL](https://vueuse.org/functions.html)

dmaixner93: Update: It seems to be working now since typing the first comment.

PR #1154, opened 2022-01-19
feat(useStyleTag): new function

### LINKED ISSUE #619: useStylesheet
This is a function I ended up writing for the Vueuse playground, but thought maybe it we might want to have here.

The general idea is to provide a reactive [CSSStyleSheet](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet)

The current API I have looks like something like
```ts
const { rules } = useStyleSheet('stylesheet-name')

rules.value = {
  '.my-special-class': 'background-color: red; display: block;',
}
```

Perhaps this api could be improved a bit during a PR if we want this. Of course this is also a pretty niche use-case especially in a vue application so I thought maybe I should make an issue before creating a PR

sibbng:
Closes #619


wheatjs:
Looks good to me

wheatjs (dismissed):

packages/core/useStyleTag/index.md
```diff
@@ -4,7 +4,7 @@ category: Browser
 
 # useStyleTag
 
-Inject reactive `&lt;style&gt;` element in head.
+Inject reactive `<style>` element in head.
```
sibbng, line 7: @antfu This characters were causing this issue on function search:

image

antfu: could fix on the function list side. In `<code>` `&lt;` won't come back to `<`

PR #1156, opened 2022-01-19
docs: update reference

ilkome:
same text as here https://vueuse.org/core/useStorageAsync/


wheatjs:
Tbh I don't think this is a typo. The docs is just telling users to refer to the other documentation for more info. Prefer would be telling them to use the other function over this one

PR #1169, opened 2022-01-23
feat(useCached): new function

Kingwl:


antfu:
How about naming it `useCached`?

antfu (approved):

Issue #1163, opened 2022-01-22
Uncaught ReferenceError: globalThis is not defined  in QQ browser


fangpeidong:
`const _global = globalThis || this`
`const globalKey = '__vueuse_ssr_handlers__'`
`// @ts-expect-error inject global`
`_global[globalKey] = _global[globalKey] || {}`
`// @ts-expect-error inject global`
`const handlers: Partial<SSRHandlersMap> = _global[globalKey]`

PR #1133, opened 2022-01-07
fix(ssr-handlers): compatible with QQ browser

btea:
`globalThis` is not defined in qq browser

PR #1175, opened 2022-01-24
fix(ssr-handlers): fix globalThis is not defined

okxiaoliang4:
fix globalThis is not defined

link issues
#1133 #1163 #1187

### DIFF
--- a/packages/core/ssr-handlers.ts
+++ b/packages/core/ssr-handlers.ts
@@ -1,4 +1,5 @@
 import type { Awaitable } from '@vueuse/shared'
+import globalThis from 'globalthis'
 
 export interface StorageLikeAsync {
   getItem(key: string): Awaitable<string | null>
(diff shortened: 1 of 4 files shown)


okxiaoliang4:
#1133 #1163

packages/core/ssr-handlers.ts
```diff
@@ -21,12 +22,11 @@ export interface SSRHandlersMap {
   updateHTMLAttrs: (selector: string, attribute: string, value: string) => void
 }
 
-const _global = globalThis || this
 const globalKey = '__vueuse_ssr_handlers__'
```
antfu, line 25: I am not a fan of extra deps, use

```ts
const _global = typeof globalThis === 'undefined' ? this : globalThis

antfu (changes requested):

packages/core/ssr-handlers.ts
okxiaoliang4: but the build output `this` will transform to undefined

PR #1186, opened 2022-01-25
feat(whenever): improve typing of callback params

lsdsjy:
Parameters of callback should be type-narrowed to be truthy


antfu:
I would feel it's more like a foot gun as the check is in the runtime and sometimes you might want to return false to skip this round.

lsdsjy:
Sorry but I don't really get it.. Do you mean sometimes it's possible to call `cb` with a falsy `v`?

Issue #1187, opened 2022-01-25
Uncaught ReferenceError: globalThis is not defined


guixr:
"element-plus": "^1.3.0-beta.8",
"@vueuse/core": "^7.5.4",

UA: "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3405.0 Safari/537.36"

<img width="425" alt="企业微信截图_16430998517493" src="https://user-images.githubusercontent.com/25836014/150941390-8b46ac9f-83cd-4a4e-92b5-1ed8a96aafdb.png">

guixr:
<img width="591" alt="企业微信截图_16431001215617" src="https://user-images.githubusercontent.com/25836014/150942068-47ab8262-3bd7-4f4e-8300-1b1b746aca29.png">

guixr:
```
if (!globalThis) {
  globalThis = window;
}
```

PR #1190, opened 2022-01-26
fix(ssr-handlers): fix `_global` is undefined

okxiaoliang4:
In https://github.com/vueuse/vueuse/commit/3c8cdded6e949ccfbbdc2539f6d0d26e6c954b61 the build output is
```typescript
const _global = typeof globalThis === "undefined" ? undefined : globalThis;
```
`this` will transform to undefined.
This will cause an exception to occur if there is no `globalThis`
`Uncaught TypeError: Cannot read property '__vueuse_ssr_handlers__' of undefined`

### DIFF
--- a/packages/core/ssr-handlers.ts
+++ b/packages/core/ssr-handlers.ts
@@ -21,7 +21,14 @@ export interface SSRHandlersMap {
   updateHTMLAttrs: (selector: string, attribute: string, value: string) => void
 }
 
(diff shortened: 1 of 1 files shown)


btea (commented):
Maybe it can be written like this
```js
const _global = globalThis || windows || self || {}
```

okxiaoliang4:
fixes #1222 #1163 #1187 #1133

PR #1192, opened 2022-01-26
feat(useEagerComputed): allow specify watchOptions

Kingwl:


packages/shared/eagerComputed/index.ts
```diff
@@ -1,14 +1,18 @@
...
+  const { flush = 'sync' } = options ?? {}
 
-export function eagerComputed<T>(fn: () => T): Readonly<Ref<T>> {
   const result = shallowRef()
 
-  watchSyncEffect(() => {
+  watchEffect(() => {
     result.value = fn()
+  }, {
+    flush,
```
antfu, line 15: Don't we just pass all the options?

Kingwl: Done

antfu (approved):

Kingwl:
Well, I guess it's not really can avoid the effect. @lsdsjy

antfu (approved):

PR #1198, opened 2022-01-27
chore: decrease types castings in useScriptTag

dblazhkun:


antfu:
I don't mind either way, or how would this change help?

dblazhkun:
> 
Just an improvement in readability, in fact. :)
When we use `as`, it looks like it have no sense to check that value is exist on the next line. But if we pass `HTMLScriptElement ` as a type argument of `querySelector`, then we really need to check if the result is not `null`.

PR #1200, opened 2022-01-28
docs(useAsyncQueue): update index.md

donotloveshampo:
'result' attribute 'value' does not exist


wheatjs:
Interesting, in general I think we prefer to use `ref` over `reactive` wonder why we didn't here. Anyway I'll merge since the docs are wrong

wheatjs (approved):

PR #1201, opened 2022-01-28
feat(useLazyload): new function

### LINKED ISSUE #833: feat: useLazyload
I have an idea :

provide `useLazyload` hook based on `useIntersectionObserver`

unknownzjc:
close #833


wheatjs:
Hmm, I'm not really sure about this. Currently the API seems too specific. I think maybe we should either narrow down the function name to something like `useImageLazyLoad` or generalize the function to have some sort of callback instead.

In the case of us narrowing it down to something like `useImageLazyLoad` I think it would be better to have something like this instead instead of using html attributes

```html
<script setup>
const el = ref()
const { isLoading, src } = useImageLazyLoad(el, { src: '...' })
</script>

<template>
  <img ref="el" :src="src" />
</template>
```

This would allow users to have loading states for their images and would be more compatible with a component approach.

```html
<UseImageLazyLoad src="..." v-slot="{ src, isLoading }">
  <span v-if="isLoading">Loading Image...</span>
  <img :src="src" v-else />
</UseImageLazyLoad>

```

 If we do end up including this function, then in the documentation I suggest we include info about native image lazy loading https://web.dev/browser-level-image-lazy-loading/

SnowingFox:
I think the directive is the best appropriate instead of this hook, because I can configurate that in entry file, like
```html
<img v-lazy="http://xxx.com/xxx.jpg />"
```

unknownzjc:
This API looks good, indeed, the current implementation is a bit special, very closely related to HTML Attributes. However, I have some suggestions for your API, maybe we can use the image in the loading state when loading, and replace it with the real image when the loading is complete.

```
// when loading src use placeholder src, loading complete use truly src
<UseImageLazyLoad src="..." v-slot="{ src, isLoading }">
  <img :src="src" data-placeholder="src">
</UseImageLazyLoad>
```

wheatjs:
With the component usage it would be possible to use another image as a placeholder. It could be implemented in a manner of ways such as

```html
<UseImageLazyLoad src="..." v-slot="{ src, isLoading }">
  <img :src="isLoading ? placeholderSrc : src" />
</UseImageLazyLoad>
```

unknownzjc:
OK, let me refactor the code

lumdzeehol:
the refactor seems done , could this pr be merged? that would be helpful.

Issue #1165, opened 2022-01-22
feat(onClickOutside): request whitelisting element


chu121su12:
I am integrating `onClickOutside` and `flatpickr`, and due to `flatpickr` by default attaching itself outside the `onClickOutside` container (below `body`), clicking on it will trigger the close.

okxiaoliang4:
I will start work for this.

okxiaoliang4:
Released https://github.com/vueuse/vueuse/releases/tag/v7.6.0

PR #1205, opened 2022-01-29
feat(onClickOutside): add `ignore` option

okxiaoliang4:
#1165


packages/core/onClickOutside/index.ts
```diff
@@ -5,6 +5,10 @@ import { useEventListener } from '../useEventListener'
 import type { ConfigurableWindow } from '../_configurable'
 import { defaultWindow } from '../_configurable'
 
+export type OnClickOutsideOptions = {
```
antfu, line 8: use interface and extend

antfu (dismissed):

PR #1217, opened 2022-02-05
feat(useScrollValues): new function

chu121su12:
I'm trying to implement https://github.com/WickyNilliams/headroom.js/ . But using boolean state is less than ideal. ~Instead this PR adds `rates` for x & y scroll rates and `spaces` for the amount of pixel left in each directions.~

- Add new functions for clarity versus the existing `useScroll`
- Use `distances` instead of `arrivedState` show the amount of pixel to all 4 directions
- Use `rates` (please recommend more suitable name if any) for horizontal and vertical scroll rates (in px/tick)


wheatjs:
I feel like the naming is a little confusing

chu121su12:
Any name suggestion?

chu121su12:
@wheatjs ping for separate implementation update.

PR #1218, opened 2022-02-07
feat(useFileDialog): new function

wobsoriano:
Adds an open file dialog hook without struggling with file input.


packages/core/useFileDialog/index.ts
```diff
@@ -0,0 +1,47 @@
+import { readonly, ref } from 'vue-demi'
+
+export interface Options {
```
antfu, line 3:
```suggestion
export interface UseFileDialogOptions {
```

packages/core/useFileDialog/index.ts
```diff
@@ -0,0 +1,47 @@
...
+      const result = event.target as HTMLInputElement
+      files.value = result.files
+    }
+
+    input.click()
+  }
+
+  return {
+    files: readonly(files),
+    openFileDialog,
```
antfu, line 43:
```suggestion
    open,
```

We don't duplicate the context, and let user to do the rename when neccessary

packages/core/useFileDialog/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    input.onchange = (event: Event) => {
+      const result = event.target as HTMLInputElement
+      files.value = result.files
+    }
+
+    input.click()
+  }
+
+  return {
+    files: readonly(files),
```
antfu, line 42: Since it's readonly, we might need an extra function `reset`

wobsoriano: I agree, a `reset` function would be needed

packages/core/useFileDialog/index.ts
```diff
@@ -0,0 +1,52 @@
+import { readonly, ref } from 'vue-demi'
+
+export interface UseFileDialogOptions {
```
okxiaoliang4, line 3: I suggest adding a new option `capture`, to select the file input source.

packages/core/useFileDialog/index.ts
```diff
@@ -0,0 +1,52 @@
...
+export function useFileDialog(options?: Partial<UseFileDialogOptions>) {
+  const files = ref<FileList | null>(null)
+
+  const open = (localOptions?: Partial<UseFileDialogOptions>) => {
+    const _options = {
+      ...DEFAULT_OPTIONS,
+      ...options,
+      ...localOptions,
+    }
+    const input = document.createElement('input')
```
okxiaoliang4, line 28: Is it possible to move to the `useFileDialog` scope? I'm worried that using the `open` function multiple times will create too many `input`.

wobsoriano: Moved outside `useFileDialog` now. Thanks!

packages/core/useFileDialog/index.ts
wobsoriano: Can elaborate?

packages/core/useFileDialog/index.ts
```diff
@@ -0,0 +1,53 @@
+import { readonly, ref } from 'vue-demi'
+
+export interface UseFileDialogOptions {
+  multiple?: boolean
+  accept?: string
+}
```
okxiaoliang4, line 6:
```suggestion
export interface UseFileDialogOptions {
  multiple?: boolean
  accept?: string
  capture?: string
}
```

new options `capture`

okxiaoliang4, line 6:
```suggestion
export interface UseFileDialogOptions {
  multiple?: boolean
  accept?: string
  /**
   * Select the input source for the capture file.
   * @see [HTMLInputElement Capture](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture)
   */
  capture?: string
}
```

packages/core/useFileDialog/index.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const open = (localOptions?: Partial<UseFileDialogOptions>) => {
+    const _options = {
+      ...DEFAULT_OPTIONS,
+      ...options,
+      ...localOptions,
+    }
+    input.multiple = _options.multiple!
+    input.accept = _options.accept!
+
```
okxiaoliang4, line 38:
```suggestion
    input.multiple = _options.multiple!
    input.accept = _options.accept!
    input.capture = _options.capture!

```
When `capture` value is `user` or `environment` or other you can select the device to limit the input source

[HTMLInputElement Capture](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture)

packages/core/useFileDialog/index.ts
wobsoriano: Hey sorry I didn't see this last commit suggestion. Anyway, I copy pasted it.

packages/core/useFileDialog/index.ts
```diff
@@ -3,6 +3,11 @@ import { readonly, ref } from 'vue-demi'
 export interface UseFileDialogOptions {
   multiple?: boolean
   accept?: string
+  /**
+   * Select the input source for the capture file.
+   * @see [HTMLInputElement Capture](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture)
+   */
+  capture?: 'user' | 'environment'
```
okxiaoliang4, line 10:
```suggestion
  capture?: string
```

PR #1223, opened 2022-02-08
feat(useNumberRange): add function

Kingwl:


antfu:
The name `useSlidingWindow` feels like I can provide an array to slides. Maybe better name it `useNumberRange`, WDYT?

antfu:
Converting to draft now. Feel free to switch back when you back on it.

PR #1224, opened 2022-02-08
feat(useProjection): add function

Kingwl:


wheatjs:
Interesting, a few things I'd like to mention. In VueUse we prefer to return an object rather than an array. Secondly I've noticed you've created a couple PR's (#1223) for math related functions. I'm wondering if we would rather have a `@vueuse/math` library as some of the names might confuse users thinking it is something else. @antfu any input on a new package for these types of functions?

Kingwl:
I'm Ok with that. Would you mind tell me what should I do?

antfu:
Let's have a @vueuse/math then. Anyone interested in it feel free to kickstart it and send a PR

Kingwl:
I'm not a big fan about `@vueuse/math`:

1. Currently, other packages all have third party dependencies (electron, firebase, etc). But the `Math` does not.
2. If we have `/math`, why don't we have something like `/events`, `/convertion`, `/encoding`, or something like that ? It's seems a bit overkill.

How about just a new category called `Math`?

wheatjs:
We have `@vueuse/shared` and `@vueuse/components` which have the same dependencies as `@vueuse/core`. The reason I think it is a good idea is that both the PR's you have opened `useSlidingWindow` and `useProjection` have names that people might confuse as other functions. If they are under a math package it will be more clear and we will be less likely to run into similar confusion or possible conflicts in the future

Kingwl:
Have create a new package `/math`.

Kingwl:
ping 👀

wheatjs (approved):

PR #1063, opened 2021-12-14
feat(useScroll): wheel direction lock

okxiaoliang4:
see #1059

PR #1231, opened 2022-02-10
feat(useWheelScrollDirection): new function

### LINKED ISSUE #1059: horizontal scroll
I want to scroll the content of an element horizontally by mouse scrolling.

`useScroll` does not seem to be able to achieve this effect. Can you add a new option to achieve a similar effect? Or create a new function?

okxiaoliang4:
move from this: #1063

close #1059


okxiaoliang4 (commented):
self review

packages/core/useWheelScrollDirection/index.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Restrict wheel scroll direction.
+ *
+ * @param element
+ * @param options - UseWheelScrollDirectionOptions
+ * @see https://vueuse.org/useWheelScrollDirection
+ */
+export function useWheelScrollDirection(element: MaybeRef<HTMLElement | SVGElement | Window | Document | null | undefined>, options: UseWheelScrollDirectionOptions = {}) {
```
btea, line 17: `MaybeRef<HTMLElement | SVGElement | Window | Document | null | undefined>` can be extracted to `unrefElement`.

okxiaoliang4: Thanks.

packages/core/useWheelScrollDirection/index.ts
```diff
@@ -1,6 +1,8 @@
-import { unref } from 'vue-demi'
+import { computed, unref } from 'vue-demi'
 import type { MaybeRef } from '@vueuse/shared'
+import type { MaybeElementRef } from '@vueuse/core'
 import { useEventListener } from '@vueuse/core'
+import { unrefElement } from '../unrefElement'
```
btea, line 6: Maybe it can be written like this `import { unrefElement, type MaybeElementRef } from '../unrefElement'`

okxiaoliang4: Yes, but I think it would be better to separate.

btea: ok.

antfu:
Thank you. But this approach feels a bit hacky and I feel the usage of this can also be nitch. I am sorry but I think it's better for us to not have it for now. I think it fit better with in userland codebase. Thanks.

PR #1236, opened 2022-02-10
feat(useCssVar): support prop be ref

btea:


okxiaoliang4 (dismissed):

packages/core/useCssVar/index.ts
```diff
@@ -33,7 +35,7 @@ export function useCssVar(
     variable,
     (val) => {
       if (elRef.value?.style)
-        elRef.value.style.setProperty(prop, val)
+        elRef.value.style.setProperty(propRef.value, val)
```
antfu, line 38:
```suggestion
        elRef.value.style.setProperty(unref(prop), val)
```

And we don't need the extra computed then.

antfu (changes requested):

packages/core/useCssVar/index.ts
btea: Done.

antfu (approved):

PR #1244, opened 2022-02-12
fix(useFps,useMemory): ssr support

webfansplz:


packages/core/useFps/index.ts
```diff
@@ -12,6 +13,8 @@ export interface UseFpsOptions {
 
 export function useFps(options?: UseFpsOptions): Ref<number> {
   const fps = ref(0)
+  if (!isClient)
+    return fps
   const every = options?.every ?? 10
 
   let last = performance.now()
```
antfu, line 20: Can we just check if `performance` exists?

webfansplz: image

It'll throw ReferenceError , because it should be `Window.performance`

antfu: Yeah, I mean, if a performance object exists in Node, we shouldn't skip it. We need to do the check `if (typeof performance === 'undefined')` instead of `isClient`

webfansplz: Fixed

PR #1256, opened 2022-02-13
fix(useToggle): should return value while param type is ref

### LINKED ISSUE #1255: Typing Error in useToggle
https://github.com/vueuse/vueuse/blob/322cdcb8cd5469340675b4c57c17fdb7805081ac/packages/shared/useToggle/index.ts#L10-L31

the returned `toggle` function is declared to be `=> boolean`, but it's actually `=> void`.

webfansplz:
Close #1255


haoyun:
The else part is also missing a return.

webfansplz:
> The else part is also missing a return.

Thanks, fixed

okxiaoliang4 (approved):

PR #1258, opened 2022-02-13
feat(build): allow only pnpm

AllenYu0118:


antfu:
Thanks but I don't think this is required. We already documented it in https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md

AllenYu0118:
👌

PR #1262, opened 2022-02-14
feat(useScroll): removeEventListener support

webfansplz:
In my case:

We need to stop listening for the scroll event after hitting some condition .


antfu:
Won't `effectScope` covers the need?

webfansplz:
> effectScope

Maybe can’t, such as :

```ts
const { y, removeEventListener } = useScroll(el, {
  onScroll() {
    const conditions = y.value > 1000
    if (conditions)
      removeEventListener()
  },
})
```

PR #1267, opened 2022-02-14
docs: show breaking changes on changelog

sibbng:


antfu:
What's the reason for this? Refactors shouldn't change the logic either be listed in the changelog

antfu:
We could match for `!` instead of `refactor`

sibbng:
> We could match for `!` instead of `refactor`

I updated PR according to that.

PR #1269, opened 2022-02-15
feat(useContextMenu): new function

MelihAltintas:


packages/core/useContextMenu/component.ts
```diff
@@ -0,0 +1,46 @@
...
+    const style = computed(() => `position:fixed; top: ${position.value!.y}px; left:${position.value!.x}px;`)
+
+    const contextMenu = ref<HTMLElement>()
+
+    onClickOutside(contextMenu, () => {
+      contextMenu.value!.style.display = 'none'
+    })
+
+    return () => {
+      return h(
```
okxiaoliang4, line 18: Maybe `teleport` to body?

okxiaoliang4: Can we just give a wrap element? style is up to the developer

MelihAltintas: I wanted the user to be able to create a context menu without using any HTML. So I don't want to use teleport. You are right about styling. Maybe it should be options like that:
```
export interface ContextMenuOptions{
  backgroundColor:string
  hoverColor:string
}
```

What do you think about this way?

wheatjs:
Tbh I feel like this might be opening us up to quite a few problems we wouldn't want to deal with in this library. Maybe I'm wrong, but a context menu would be pretty complicated. We shouldn't be shipping any styles and I don't think we would be able to deal with the accessibility concerns right now. This might be the type of feature that is better left to a UI library like https://headlessui.dev/

MelihAltintas:
Okey :) thanks for your response. Maybe this can be done easily with electron using the electron menu in the electron section without still thinking styling. You can close this pr thanks.

michael-observerly:
It might be worth also adding that what happens to the context menu for literal edge cases ... when x, y are close to screen minimum and maximums (both vertically and horizontally). It might be worth exploring how popper.js achieves this...

packages/core/useContextMenu/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  x: Number
+  y: Number
+}
+
+export function useContextMenu(element: MaybeElementRef,
+  items: MenuItem[]) {
+  const target = computed(() => unrefElement(element))
+  const position = ref<MenuPosition>({ x: 0, y: 0 })
+
+  const menu = computed(() => createVNode(ContextMenu, { items, position }))
```
antfu, line 24: Thanks but no. I don't think we should create vnodes in compsoables.

antfu (changes requested):

packages/core/useContextMenu/index.ts
MelihAltintas: Thanks a lot @antfu

PR #1270, opened 2022-02-15
fix(useFetch): make `error` shallow

okxiaoliang4:
`shallowRef` types fix

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -1,4 +1,4 @@
-import type { ComputedRef, Ref, ShallowRef } from 'vue-demi'
+import type { ComputedRef, Ref } from 'vue-demi'
 import type { EventHookOn, Fn, MaybeRef, Stoppable } from '@vueuse/shared'
 import { containsProp, createEventHook, until, useTimeoutFn } from '@vueuse/shared'
 import { computed, isRef, ref, shallowRef, unref, watch } from 'vue-demi'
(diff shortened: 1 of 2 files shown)


packages/core/useFetch/index.ts
```diff
@@ -18,17 +18,17 @@ export interface UseFetchReturn<T> {
...
    * Any fetch errors that may have occurred
    */
-  error: Ref<any>
+  error: ShallowRef<any>
 
   /**
    * The fetch response body, may either be JSON or text
    */
-  data: Ref<T | null>
+  data: ShallowRef<T | null>
```
antfu, line 31: Not sure. I think `Ref<T>` is already shallow.

okxiaoliang4: image

in `useFetch` `data` and `response` is using `shallowRef` but the types is `Ref`.

https://github.com/vueuse/vueuse/blob/37a0205ed9668be06f65b51fad1ce36906084c43/packages/core/useFetch/index.ts#L300-L302

in `useAxios` `error` is using `shallowRef`. maybe in `useFetch` `error` should be shallow too.

https://github.com/vueuse/vueuse/blob/f464846a09906433c8dca876ee87c18630d0f33c/packages/integrations/useAxios/index.ts#L91-L96

antfu: Yeah, it makes sense to have error as `shallowRef`, but there is no need to change the types.

Issue #1266, opened 2022-02-14
`afterFetch` runs even after any non-2xx response, in useFetch.


haoyun:
According to the doc,

https://github.com/vueuse/vueuse/blob/b8a26678de12f0720f05e04082659df433682ccd/packages/core/useFetch/index.ts#L171-L175

`afterFetch` should run only on 2xx. However in the code,

https://github.com/vueuse/vueuse/blob/b8a26678de12f0720f05e04082659df433682ccd/packages/core/useFetch/index.ts#L380-L388

the `statusCode` is not checked. Therefore, even with 404, 500 that has error messages in the response body, `afterFecth()` will be executed.

PR #1271, opened 2022-02-15
fix(useFetch): fix run `afterFetch` condition

okxiaoliang4:
close: #1266


okxiaoliang4 (commented):
verified

webfansplz:
Maybe can add a test case for it ?

okxiaoliang4:
> Maybe can add a test case for it ?

Yes,later

PR #1272, opened 2022-02-15
feat(useFullscreen): autoExit support

### LINKED ISSUE #827: Should `useFullscreen` exit in `tryOnBeforeUnmount`?
I think it would be appropriate to add automatic exit of fullscreen mode to useFullscreen if the component that called it was unmount
```js
tryOnBeforeUnmount(exit)
```

Are there any thoughts on this?

webfansplz:
Close #827

### DIFF
--- a/packages/core/useFullscreen/index.ts
+++ b/packages/core/useFullscreen/index.ts
@@ -1,6 +1,7 @@
 /* this implementation is original ported from https://github.com/logaretm/vue-use-web by Abdelrahman Awad */
 
 import { ref } from 'vue-demi'
+import { tryOnScopeDispose } from '@vueuse/shared'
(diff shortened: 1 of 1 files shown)


packages/core/useFullscreen/index.ts
```diff
@@ -62,6 +63,15 @@ const functionsMap: FunctionMap[] = [
   ],
 ] as any
 
+export interface UseFullscreenOptions extends ConfigurableDocument {
+  /**
+   * Automatically exit
```
wheatjs, line 68: Maybe a more descriptive doc like "Automatically exit fullscreen when component is unmounted"

wheatjs (approved):

PR #1281, opened 2022-02-16
feat(utils): add getType

btea:


packages/shared/utils/is.ts
```diff
@@ -9,10 +9,9 @@ export const isBoolean = (val: any): val is boolean => typeof val === 'boolean'
 export const isFunction = <T extends Function> (val: any): val is T => typeof val === 'function'
 export const isNumber = (val: any): val is number => typeof val === 'number'
 export const isString = (val: unknown): val is string => typeof val === 'string'
-export const isObject = (val: any): val is object =>
-  toString.call(val) === '[object Object]'
-export const isWindow = (val: any): val is Window =>
-  typeof window !== 'undefined' && toString.call(val) === '[object Window]'
+export const getType = (val: unknown) => toString.call(val)
```
webfansplz, line 12: Maybe easier to use :
```ts
export const getType = (val: unknown) => toString.call(val).slice(8, -1)
```

btea: IMO, just return the initial string, and let the user handle it himself.

okxiaoliang4: > IMO, just return the initial string, and let the user handle it himself.

I agree.

antfu:
Can you provide more context why this util is needed?

btea:
> Can you provide more context why this util is needed?

I think it is convenient to judge some special data types, such as `File`, `Blob`...

PR #1292, opened 2022-02-18
feat(reactiveOmit): add reactiveOmit

### LINKED ISSUE #1291: Feature request reactiveOmit
It complements with reactivePick.

It will be more convenient when pick most fields from a reactive object. 

I can do it.

```ts
const props = defineProps({
  value: {
    default: 'value',
  },
  color: {
    type: String,
  },
  font: {
    type: String,
  }
})

const childProps = reactivePick(props, 'color', 'font')
// same as
const childProps = reactiveOmit(props, 'value')
```

tolking:
close #1291


packages/shared/reactiveOmit/index.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @see https://vueuse.js.org/reactiveOmit
+ */
+export function reactiveOmit<T extends object, K extends keyof T>(
+  obj: T,
+  ...keys: K[]
+): { [S in Exclude<keyof T, K>]: UnwrapRef<T[S]>; } {
+  const pickKeys = Object.keys(obj).filter(k => !keys.includes(k as any)) as Array<Exclude<keyof T, K>>
+
+  return reactivePick(obj, ...pickKeys)
```
antfu, line 15: I guess this is not ideal as it won't cover new keys added to that object

tolking: You're right, I didn't take into account this situation, let me think about it.

antfu (approved):

Issue #1268, opened 2022-02-14
useColorMode not reactive when using multiple instances


toniengelhardt:
When I use `useColorMode()` in different places, different refs are returned, which means reactivity between them doesn't work, e.g.

```html
// ColorMode toggle component

<template>
  Some color mode toggle
</tempalte>

<script setup lang="ts">
const color = useColorMode()
// manipulate color with toggle...
</script>
```

```html
// Some other component

<template>
  <div :style="{ 'background-color': bgColor }">
    Some HTML
  </div>
</tempalte>

<script setup lang="ts">
const color = useColorMode()
const bgColor = computed(() => color === 'light' ? '#fff' : '#000')
</script>
```

Now, when I use the toggle to change the color, the second component will not react...

okxiaoliang4:
`useColorMode` return a `Ref`,`computed` should use `color.value`
![image](https://user-images.githubusercontent.com/22525904/153969328-eea36ed2-e18d-4e84-a2b0-d895aef64bf6.png)

okxiaoliang4:
Because the `store` event is not triggered when the value is change, the `ref` value of the other ranges is not updated.

okxiaoliang4:
You can reuse one `ref` variable to resolve this.

```typescript
// colorMode.ts
import { useColorMode } from '@vueuse/core';
export const color = useColorMode();
```

toniengelhardt:
It throws an error, see updated example: https://stackblitz.com/edit/nuxt3-vueuse-color-mode-reactivity-bug

okxiaoliang4:
Sorry,I'm out you can try it in your project, I can't run stackBlitz in mobile😅

toniengelhardt:
Amazing @webfansplz, it works. I opened a PR to add this example to the docs: https://github.com/vueuse/vueuse/pull/1293.

okxiaoliang4:
> > It throws an error, see updated example: https://stackblitz.com/edit/nuxt3-vueuse-color-mode-reactivity-bug
> 
> Try it , https://stackblitz.com/edit/nuxt3-vueuse-color-mode-reactivity-bug-gfrbbx?file=composables%2FuseTheme.ts

Hi I'm back. I think it's a bit complicated. see this
https://stackblitz.com/edit/vitejs-vite-dfuy95?file=src%2Fuse%2FuseGlobalColorMode.ts&terminal=dev

toniengelhardt:
[...]

PR #1293, opened 2022-02-19
docs(useColorMode): example for reactivity (#1268)

toniengelhardt:
Adds an example on how to use a composable to make multiple instances of useColorMode reactive (credits @okxiaoliang4 and @webfansplz).

### DIFF
--- a/packages/core/useColorMode/index.md
+++ b/packages/core/useColorMode/index.md
@@ -54,3 +54,69 @@ const mode = useColorMode({
 - `useDark`
 - `usePreferredDark`
 - `useStorage`
+
+## Advanced Usage
+
+When accessing `useColorMode()` in multiple components, the different instances of mode will not be reactive between each other.
+
+Example:
+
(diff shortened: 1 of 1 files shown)


okxiaoliang4:
I think this is not necessary. we can use [createGlobalState](https://vueuse.org/shared/createGlobalState/) to create global state like this: 

https://stackblitz.com/edit/vitejs-vite-dfuy95?file=src%2Fuse%2FuseGlobalColorMode.ts&terminal=dev

webfansplz:
> I think this is not necessary. we can use [createGlobalState](https://vueuse.org/shared/createGlobalState/) to create global state like this:
> 
> https://stackblitz.com/edit/vitejs-vite-dfuy95?file=src%2Fuse%2FuseGlobalColorMode.ts&terminal=dev

I agree .

okxiaoliang4:
Thanks for your PR.

PR #1297, opened 2022-02-20
feat(useContainerQuery): new function

michealroberts:
A composable implementation of CSS container queries with JS. CSS container queries are tantalisingly powerful in what they will achieve. However, the specification is at an early stage and has not yet been adopted or implemented in any mainstream browser,  although they can be experimented with by utilising experiment flags in Chrome Canary.

**_This composable implements them in JS._**


michael-observerly:
[useQueryContainer Loom](https://www.loom.com/share/da1df44ebebf4e5688724a41918c9fe3)

michael-observerly:
https://user-images.githubusercontent.com/84131395/154844521-37475291-c9d5-4f06-84df-8c8ab5eb25f6.mp4

packages/core/useContainerQuery/index.ts
```diff
@@ -0,0 +1,92 @@
...
+  const activeBreakpoint = computed(() => {
+    for (const [key, { min = 0, max }] of Object.entries(breakpoints))
+      if (width.value >= min && (max === undefined || width.value <= max)) return key
+  })
+
+  return {
+    activeBreakpoint,
+    width,
+  }
+}
```
okxiaoliang4, line 92: Sorry. I accidentally pushed directly to your repository, I think this looks better. what do you think?

packages/core/useContainerQuery/index.test.ts
```diff
@@ -0,0 +1,7 @@
+import { useContainerQuery } from '../useContainerQuery'
+
+describe('useContainerQuery', () => {
```
okxiaoliang4, line 3: should need more test

packages/core/useContainerQuery/index.ts
```diff
@@ -0,0 +1,92 @@
...
+import type { MaybeElementRef } from '@vueuse/core'
+import { useResizeObserver } from '../useResizeObserver'
+
+export interface Breakpoint {
+  /**
+   *
+   * minimum value of breakpoint (in pixels)
+   *
+   */
+  min?: number
```
okxiaoliang4, line 11: default value: `0`

packages/core/useContainerQuery/index.md
```diff
@@ -0,0 +1,67 @@
...
+```ts
+import { defineComponent, ref } from 'vue'
+import { useContainerQuery } from '@vueuse/core'
+
+// setup a ref to an element on your template:
+const someTemplateRef = ref(null)
+
+const breakpoints = {
+  sm: {
+    min: 320,
```
okxiaoliang4, line 39: remove. `min` default value: `0`

packages/core/useContainerQuery/index.ts
michael-observerly: Hi @okxiaoliang4 I believe the computed prop for activeBreakpoint is going to be slightly less performant this way because we don't want to re-computed on every change of width ... just when it changes boundary.

packages/core/useContainerQuery/index.test.ts
michael-observerly: I think as well because of the refactoring, Italy slightly harder to test the breakpoint matching functionality...

packages/core/useContainerQuery/index.ts
okxiaoliang4: https://github.com/vueuse/vueuse/pull/1297/commits/caff3bd3c0236685ab743999b44650a38fd43f9b#diff-d745801b80f40092b894f70c524ddaa9b8e7338eb3bc921d5c7a80f5f4565b49R120-R124

but you are watching `width`, also re-computed every time `width` changes

```typescript
watch(width, (newWidth: number, prevWidth: number) => {
  if (newWidth === prevWidth) return

  handleElementResize()
})
```

michael-observerly: Ah no actually with the watcher removed, it's slightly better! It seems less memory intensive. Nice! :)

okxiaoliang4: and you can try this. only when boundary changes will the `ref` update be triggered

```typescript
watch(activeBreakpoint, value => console.log(value))
```

image

okxiaoliang4: and use computed to change breakpoints to `MaybeRef<QueryBreakpoints>`. can also trigger recalculation when `options.breakPoints` changes

michael-observerly: Yeh agree, it's more performant and should re-computed on breakpoint change.

wheatjs:
I'd suggest we hold off on publishing this until a major browser adds stable support for useContainerQuery

michael-observerly:
> I'd suggest we hold off on publishing this until a major browser adds stable support for useContainerQuery

Hmmm, but this is why I am adding it ... so we can use Vue to navigate around the lack of support. (This is a JS implementation of container queries, and doesn't involve CSS).

wheatjs:
Ah I see then. Maybe once they are released we can update this function to use sometime similar to ` window.matchMedia` if it is available for container queries. But yeah if that's the case then we could probably release it before they are widely supported

michael-observerly:
> Ah I see then. Maybe once they are released we can update this function to use sometime similar to ` window.matchMedia` if it is available for container queries. But yeah if that's the case then we could probably release it before they are widely supported

Yeh, there will be a planned redundancy for this package, but in time we can work with the correct CSS apis 🥇 :)

okxiaoliang4 (approved):

PR #1308, opened 2022-02-23
chore(useClamp): test case

okxiaoliang4:
### Description

test case

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/packages/core/useClamp/index.ts
+++ b/packages/core/useClamp/index.ts
@@ -15,7 +15,7 @@ export function useClamp(value: MaybeRef<number>, min: MaybeRef<number>, max: Ma
   const _value = ref(value)
   return computed<number>({
     get() {
-      return clamp(_value.value, unref(min), unref(max))
+      return _value.value = clamp(_value.value, unref(min), unref(max))
     },
     set(value) {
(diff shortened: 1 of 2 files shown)


packages/core/useClamp/index.ts
```diff
@@ -15,7 +15,7 @@ export function useClamp(value: MaybeRef<number>, min: MaybeRef<number>, max: Ma
   const _value = ref(value)
   return computed<number>({
     get() {
-      return clamp(_value.value, unref(min), unref(max))
+      return _value.value = clamp(_value.value, unref(min), unref(max))
```
antfu, line 18: Why would this be neccessary?

okxiaoliang4: I think the `value` should not change when `max` or `min` change.

```typescript
const value = ref(100)
const min = ref(0)
const max = ref(100)

const v = useClamp(value, min, max)

v.value = 100 // _value.value = 100, v.value === 100

max.value = 90 // _value.value === 100, v.value === 90
max.value = 91 // _value.value === 100, v.value === 91
```

PR #1318, opened 2022-02-27
feat(useScroll): add wheel listener

Liumingxun:
add:

- add wheel listener

update:

- doc & demo

### Description

I want to detect the mouse wheel change when the page has no scrollbar, so I enhanced this function. 

### Additional context

Now it can be used to detect the mouse wheel for elements without scrollbar, it may be used for turning pages, like PowerPoint

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other


Liumingxun:
Okay, I get it. I will resubmit a PR

PR #1319, opened 2022-02-27
chore(docs): dynamic functions count

webfansplz:
### Description

image

image

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/.vitepress/theme/components/Home.vue
+++ b/packages/.vitepress/theme/components/Home.vue
@@ -1,6 +1,6 @@
 <template>
   <main class="home" aria-labelledby="main-title">
-    <HomeHero />
+    <HomeHero :count="props.functionsCount" />
     <slot name="hero" />
     <HomeFeatures />
     <slot name="features" />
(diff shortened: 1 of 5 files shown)


antfu:
Not sure, how do you deal with GitHub's readme?

webfansplz:
> Not sure, how do you deal with GitHub's readme?

I don't really understand ，Do you mean here :
https://github.com/vueuse/vueuse/blob/53d198375359e049ae646b441bddd967d50b1b12/scripts/utils.ts#L278

antfu:
The badge is not only used on the website but also the README page on GitHub https://github.com/vueuse/vueuse , which I don't think it's possible to dynamic without a server

webfansplz:
> The badge is not only used on the website but also the README page on GitHub https://github.com/vueuse/vueuse , which I don't think it's possible to dynamic without a server

I know，we has already deal with it via https://github.com/vueuse/vueuse/blob/53d198375359e049ae646b441bddd967d50b1b12/scripts/utils.ts#L278

Maybe there is something wrong with my description, the solution is just to read the number of functions when we run the update command.

antfu:
I made the badge generated on the build time and serve with our docs: https://vueuse.org/badge-function-count.svg this way we could make both of them dynamic and make PRs cleaner. Thanks for the initiator

PR #1322, opened 2022-02-27
feat(useWheel): new Function for wheel listener

Liumingxun:
add:
- add `useWheel` function for wheel listener
- doc & demo

### Description

I want to detect the mouse wheel change when the page has no scrollbar, so I added this function. 
link #1318 

### Additional context

Now it can be used to detect the mouse wheel for elements without scrollbar, it may be used for turning pages, like PowerPoint

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other


packages/core/useWheel/index.ts
```diff
@@ -0,0 +1,99 @@
...
+    throttle = 0,
+    idle = 200,
+    onStop = noop,
+    eventListenerOptions = {
+      capture: false,
+      passive: true,
+    },
+  } = options
+
+  const isWheeling = ref(false)
```
wheatjs, line 56: I think we should either rename this to `isScrolling` or change `onScroll` to `onWheel`. Although I think "onWheeling" sounds a little funny haha

PR #1324, opened 2022-02-27
docs(useToggle): fix import location

### LINKED ISSUE #1323: There is no 'useToggle' in '@vueuse/core'
### Describe the bug

According to the [documentation](https://vueuse.org/shared/usetoggle/#usage) you can import `useToggle` from `@vueuse/core`, but this does not exist, it is located under `@vueuse/shared`.

image

But in all the demos it imports correctly, like under [useDark](https://github.com/vueuse/vueuse/blob/main/packages/core/useDark/demo.vue#L2).

In addition, it is also incorrectly referenced in some other parts of the documentation, such as under [useDark](https://vueuse.org/core/useDark/#basic-usage).

I think that is also meant here in the [issue](https://github.com/vueuse/vueuse/issues/1112).

### Reproduction

[Repo](https://github.com/vueuse/vueuse/tree/main/packages/core)

### System Info

```Shell
/
```

### Used Package Manager

npm

### Validations

MelihAltintas:
closes #1323

### DIFF
--- a/packages/core/useDark/index.md
+++ b/packages/core/useDark/index.md
@@ -9,7 +9,8 @@ Reactive dark mode with auto data persistence.
 ## Basic Usage
 
 ```js
-import { useDark, useToggle } from '@vueuse/core'
+import { useDark } from '@vueuse/core'
+import { useToggle } from '@vueuse/shared'
 
 const isDark = useDark()
 const toggleDark = useToggle(isDark)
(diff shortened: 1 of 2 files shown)


antfu:
Thanks but exports in `@vueuse/shared` are forward to code. Users don't need to care about `@vueuse/shared` normally.

PR #1333, opened 2022-02-28
feat(useUserAgent): new function

michael-observerly:
### Description

A zero-dependency reactive for detecting the userAgent of the device (in browser only).

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
When would you expect the user agent to be changed during the runtime? If not, I don't think it belongs to VueUse as it might not related to Vue's reactvitiy

michael-observerly:
> When would you expect the user agent to be changed during the runtime? If not, I don't think it belongs to VueUse as it might not related to Vue's reactvitiy

Yeh, so this is where I was totally in two minds with this PR. The UserAgent string won't change, and it doesn't need to be reactive.

antfu:
Thanks for the PR. But since it's not Vue specific, I think it's better to have it as a separate lib than VueUse.

PR #1336, opened 2022-03-01
fix(useFetch): uppercase HTTP methods in useFetch

Glandos:
### Description
By https://www.rfc-editor.org/rfc/rfc7231#section-4.1:
> The method token is case-sensitive because it might be
   used as a gateway to object-based systems with case-sensitive method
   names.

By http://www.iana.org/assignments/http-methods/http-methods.xhtml all existing method are uppercase.

I currently have a REST API that doesn't support `patch` method but works with `PATCH`.

### Additional context

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


okxiaoliang4 (changes requested):
default `method` should replace to `SET`.

https://github.com/Glandos/vueuse/blob/1864298eb09e0490d9df08b9067e4efd7e35514e/packages/core/useFetch/index.ts#L268

okxiaoliang4:
> Good catch. I also changed to GET instead of SET because I think it's way you meant 🤔

Aha

PR #1343, opened 2022-03-03
fix(useAxios): args param need limit

WuLianN:
### Description

when use the following demo, can't run, options's value is intance, not { immediate: true }
```ts
import axios from 'axios'
import { useAxios } from '@vueuse/integrations/useAxios'

const instance = axios.create({
  baseURL: '/api'
})

const { data, isFinished } = useAxios('/posts', { method: 'POST' }, instance)
```
the reason is
```js
if (args.length >= 2) {
      options = args[args.length - 1];
 }
```
### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/integrations/useAxios/index.test.ts
+++ b/packages/integrations/useAxios/index.test.ts
@@ -1,0 +1,133 @@
+import { watch } from 'vue-demi'
+import type { AxiosRequestConfig } from 'axios'
+import axios from 'axios'
+import { useAxios } from '.'
+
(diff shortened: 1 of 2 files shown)


okxiaoliang4 (commented):
Hi, Can you provide test cases?

hengshanMWC:
> 你好，可以提供测试用例吗？

Test cases have been supplemented

PR #1353, opened 2022-03-04
feat(integrations): add useTitleCase

hengshanMWC:
### Description

wrapper for  [`title-case`](github.com/blakeembrey/change-case) to useTitleCase

> Transform a string into title case following English rules.

### Additional context

catalogue: packages/integrations/useTitleCase

---

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
How about a more general util like `useStringCase` with an addtional argument for the case style?

hengshanMWC:
> How about a more general util like useStringCase with an addtional argument for the case style?

wrapper for [stringcase]('https://www.npmjs.com/package/stringcase')？Or make it `packages/core/useStringCase `. It seems that both of them need to be done again. We can only close this module

antfu:
I mean, since you are warpping `change-case`, that have a lot more cases. We could have the wrapper configurable like:

```ts
useStringCase(strRef, 'camelCase')
useStringCase(strRef, 'pascalCase')
// ...
```

wheatjs:
Maybe something like this. I think it is a good idea to have the package name in the function name for integrations.

```ts
const titleCase = useChangeCase(value, 'titleCase')
```

antfu:
Yeah, I agree with calling it `useChangeCase`

hengshanMWC:
ok, I'll try

hengshanMWC:
[New PR](https://github.com/vueuse/vueuse/pull/1371)
Then I'll close this PR

PR #1354, opened 2022-03-04
feat(useInfiniteScroll): new options

innocenzi:
### Description

This PR adds two new options to `useInfiniteScroll`:
- `direction`, to chose in which direction to use the infinite scroll
- `keepScrollPosition`, to automatically compute the scroll position so it stays in place

The current behavior stays the same without explicitly defining these options.

### Use case

I was building a chat with infinite scroll, which I want to behave kind of like Discord when you scroll up. For this, I had to revert the direction to `top` instead of the default `bottom`, and I needed to keep the scroll position the same after updating the message list.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useInfiniteScroll/index.ts
+++ b/packages/core/useInfiniteScroll/index.ts
@@ -1,5 +1,5 @@
 import type { UnwrapNestedRefs } from 'vue-demi'
-import { reactive, watch } from 'vue-demi'
+import { reactive, watch, unref, nextTick } from 'vue-demi'
 import type { MaybeRef } from '@vueuse/shared'
 import type { UseScrollOptions } from '../useScroll'
 import { useScroll } from '../useScroll'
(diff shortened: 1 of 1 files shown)


packages/core/useInfiniteScroll/index.ts
```diff
@@ -11,6 +11,20 @@ export interface UseInfiniteScrollOptions extends UseScrollOptions {
...
+   * @default 'bottom'
+   */
+  direction?: 'top' | 'bottom'
+
+  /**
+   * Whether to keep the current scroll position when loading more items.
+   *
+   * @default false
+   */
+  keepScrollPosition?: boolean
```
wheatjs, line 27: I think `preserveScrollPosition` might be a better name

innocenzi: You're right 👍

wheatjs (approved):

vincentbernat:
Did you consider adding `direction: both` as an option?

PR #1357, opened 2022-03-05
feat(useElementHover): directive support

webfansplz:
### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useElementHover/directive.test.ts
+++ b/packages/core/useElementHover/directive.test.ts
@@ -1,0 +1,42 @@
+import { defineComponent } from 'vue-demi'
+import type { VueWrapper } from '@vue/test-utils'
+import { mount } from '@vue/test-utils'
+
+import { vHover } from './directive'
+
+const App = defineComponent({
+  props: {
+    onHover: {
+      type: Function,
+      required: true,
+    },
+  },
+
+  template: `<template>
+  <div v-hover="onHover">Hover me</div>
+  </template>
+  `,
+})
+
+describe('vOnLongPress', () => {
+  let onHover = vi.fn()
+  let wrapper: VueWrapper<any>
+
+  beforeEach(() => {
+    onHover = vi.fn()
+    wrapper = mount(App, {
+      props: {
+        onHover,
+      },
+      global: {
+        directives: {
+          hover: vHover,
+        },
+      },
+    })
+  })
+
+  it('should be defined', () => {
+    expect(wrapper).toBeDefined()
+  })
+})
--- a/packages/core/useElementHover/directive.test.ts
+++ b/packages/core/useElementHover/directive.test.ts
@@ -18,7 +18,7 @@ const App = defineComponent({
   `,
 })
 
-describe('vHover', () => {
+describe('vElementHover', () => {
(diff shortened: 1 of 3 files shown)


packages/core/useElementHover/directive.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+  },
+
+  template: `<template>
+  <div v-hover="onHover">Hover me</div>
+  </template>
+  `,
+})
+
+describe('vOnLongPress', () => {
```
okxiaoliang4, line 21: typo

webfansplz: Thanks !

packages/core/useElementHover/index.md
```diff
@@ -22,3 +24,25 @@ const myHoverableElement = ref()
...
+import { vHover } from '@vueuse/components'
+
+const isHovered = ref(false)
+function onHover(state: boolean) {
+  isHovered.value = state
+}
+</script>
+
+<template>
+  <button v-hover="onHover">
```
antfu, line 43: I think we should match with the function name and call it vElementHover

packages/core/useElementHover/directive.ts
```diff
@@ -0,0 +1,15 @@
...
+
+type BindingValueFunction = (state: boolean) => void
+
+export const vHover: FunctionDirective<
+HTMLElement,
+BindingValueFunction
+> = (el, binding) => {
+  if (typeof binding.value === 'function') {
+    const isHovered = useElementHover(el)
+    watch(isHovered, v => binding.value(v))
```
okxiaoliang4, line 13: I think replace `watchEffect` has better performance.
```suggestion
    watchEffect(() => binding.value(unref(isHovered)))
```

packages/core/useElementHover/index.md
webfansplz: Fixed

packages/core/useElementHover/directive.ts
webfansplz: The watchEffect will triggered once immediately which will give The user unexpected behavior . (triggered `onHover` callback)

antfu:
Ouch, please reopen it against main 😅

packages/core/useElementHover/directive.ts
okxiaoliang4: but it looks like when `isHovered` change then trigger the callback and pass `isHovered`.

okxiaoliang4: maybe you want like this?
```typescript
if(isHovered) binding.value(unref(isHovered))
```

webfansplz: I think we also need triggered the callback  while the state is `false`

flybbjcs70:
it seems useEventListener ```tryOnScopeDispose(stop)``` cannot work in  directive  mount hook

flybbjcs70:
```js
binding.instance.scope.run(()=>{
})
```
This may be better?

PR #1361, opened 2022-03-05
feat(useMagicKeys): directive support

webfansplz:
### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
I am not sure about this, this feel that we lose the "magic" of dynamically get the key ref, I guess `onKeyStroke` would better fit the directive usage. WDYT?

webfansplz:
> I am not sure about this, this feel that we lose the "magic" of dynamically get the key ref, I guess `onKeyStroke` would better fit the directive usage. WDYT?

Ok, I agree.

PR #1373, opened 2022-03-07
feat(useCanvasContext): new function

### LINKED ISSUE #1346: proposal: useCanvas()

On a recent project I have utilised a composed way for working with the canvas, as well as detecting various changing properties of the canvas object. I've found that composing this was a neat way to abstract a lot away from a specific component, as well as focusing on the core canvas ideas.

### Suggested solution

I'm thinking a proposed solution would look something like:

```ts
const { canvasRef, canvas, context } = useCanvas({ context: '2D', alpha: true })
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

michael-observerly:
### Description

So this PR solves the usual setup of a canvas object, whilst simultaneously making it reactive to it's background element (utilising functionality from the useElementBoundin() composable).

Closes #1346 

### Additional context

If this PR is acceptable, I will also extend to another 3D rendering context composable (webgl), as well as PR'ing an integration for three.js if there is an appetite for it.

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other


packages/core/useCanvas2D/demo.vue
```diff
@@ -0,0 +1,52 @@
+<script setup lang="ts">
+import { computed, reactive, ref } from 'vue-demi'
+import YAML from 'js-yaml'
```
okxiaoliang4, line 3:
```suggestion
import { stringify } from '@vueuse/docs-utils'
```

packages/core/useCanvas2D/demo.vue
```diff
@@ -0,0 +1,52 @@
...
+  width,
+  height,
+} = useCanvas2D(canvas, bound, { alpha: false })
+
+const properties = computed(() => YAML.dump(
+  reactive({
+    width,
+    height,
+  }),
+))
```
okxiaoliang4, line 24:
```suggestion
const properties = stringify(reactive({
  width,
  height,
}))
```

packages/core/useCanvas2D/index.ts
```diff
@@ -0,0 +1,103 @@
...
+   */
+  desynchronized?: boolean
+}
+
+/**
+ * Reactive 2D Canvas API
+ *
+ * @param element (HTML Canavs Element) template ref
+ * @param attributes
+ * @param options
```
okxiaoliang4, line 45:
```suggestion
 * @param bound
 * @param attributes
```

packages/core/useCanvas2D/index.ts
```diff
@@ -0,0 +1,103 @@
...
+ * @see https://vueuse.org/useCanvas
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API
+ */
+export function useCanvas2D(
+  el: Ref<HTMLCanvasElement | null>,
+  bound: Ref<HTMLElement | null>,
```
okxiaoliang4, line 51: why we need this?

packages/core/useCanvas2D/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  watch(rectWidth, (widthValue) => {
+    if (!canvas.value) return
+    canvas.value.width = widthValue
+  })
+
+  watch(rectHeight, (heightValue) => {
+    if (!canvas.value) return
+    canvas.value.height = heightValue
+  })
```
okxiaoliang4, line 93: I think it might be unnecessary, we can use it like this.
```typescript
const canvas = ref<HTMLCanvasElement>()
const { ctx } = useCanvas2D(canvas)

const { width, height } = useElementBounding(canvas)
```

packages/core/useCanvas2D/index.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Reactive 2D Canvas API
+ *
+ * @param el (HTML Canavs Element) template ref
+ * @param attributes 2D Canvas Attributes
+ * @see https://vueuse.org/useCanvas
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext
+ */
+export function useCanvas2D(
+  el: Ref<HTMLCanvasElement | null>,
```
okxiaoliang4, line 48:
```suggestion
  el: MaybeRef<HTMLCanvasElement | null>,
```

packages/core/useCanvas2D/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (context && isCanvasRenderingContext2D(context))
+      return context
+
+    return undefined
+  }
+
+  // Watching the canvas to ensure we can get the canvas' context:
+  watch(canvas, (canvasValue) => {
+    if (!canvasValue) return
+    ctx.value = getContext()
```
okxiaoliang4, line 77:
```suggest
  attributes?: MaybeRef<UseCanvasAttributes>,
```

michael-observerly:
Hey @okxiaoliang4 I've now refactored the code with your suggestions and I have also added dual support for '2d' and 'webgl' contexts as these are the most widely used.

okxiaoliang4:
@michael-observerly Hi, I have initiated a pr in your warehouse to see if this can solve your needs. It has better type hints and less code.
https://github.com/michael-observerly/vueuse/pull/2

PR #1377, opened 2022-03-08
feat(useElementVisibility): directive support

AllenYu0118:
### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/useElementVisibility/index.ts
```diff
@@ -17,25 +17,26 @@ export interface VisibilityScrollTargetOptions extends ConfigurableWindow {
  * @param options
  */
 export function useElementVisibility(
-  element: Ref<Element|null|undefined>,
+  element: MaybeRef<Element|null|undefined>,
   { window = defaultWindow, scrollTarget }: VisibilityScrollTargetOptions = {},
 ) {
+  const _element = ref(element)
+  const _scrollTarget = unref(scrollTarget)
```
webfansplz, line 24: I think we don't need it, just use `unref` API to handle it.

packages/core/useElementVisibility/directive.ts
```diff
@@ -0,0 +1,24 @@
...
+
+type BindingValueArray = [BindingValueFunction, VisibilityScrollTargetOptions]
+
+export const vElementVisibility: FunctionDirective<
+HTMLElement,
+BindingValueFunction | BindingValueArray
+> = (el, binding) => {
+  if (typeof binding.value === 'function') {
+    const handler = binding.value
+    const isVisibility = useElementVisibility(el)
```
webfansplz, line 16: `isVisibility` => `isVisible`

packages/core/useElementVisibility/index.md
```diff
@@ -40,3 +40,32 @@ export default {
...
+```html
+<script setup lang="ts">
+import { ref } from 'vue'
+import { vElementVisibility } from '@vueuse/components'
+
+const target = ref(null)
+const isVisible = ref(false)
+
+function onElementVisibility(state) {
+  isVisible.value = stat
```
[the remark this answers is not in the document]
AllenYu0118: OK

packages/core/useElementVisibility/directive.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        directives: {
+          ElementVisibility: vElementVisibility,
+        },
+      },
+    })
+  })
+
+  it('should be defined', () => {
+    expect(wrapper).toBeDefined()
+  })
```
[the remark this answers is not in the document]
AllenYu0118: Ok, I'll add that

packages/core/useElementVisibility/directive.ts
```diff
@@ -0,0 +1,24 @@
...
+type BindingValueArray = [BindingValueFunction, VisibilityScrollTargetOptions]
+
+export const vElementVisibility: FunctionDirective<
+HTMLElement,
+BindingValueFunction | BindingValueArray
+> = (el, binding) => {
+  if (typeof binding.value === 'function') {
+    const handler = binding.value
+    const isVisible = useElementVisibility(el)
+    watch(isVisible, v => handler(v))
```
webfansplz, line 17: Maybe we should triggered it immediately which that ensure the initial state is correctly.

packages/core/useElementVisibility/index.ts
```diff
@@ -27,25 +27,26 @@ export function useElementVisibility(
...
-      const rect = element.value.getBoundingClientRect()
-
-      elementIsVisible.value = (
-        rect.top <= (window.innerHeight || document.documentElement.clientHeight)
-          && rect.left <= (window.innerWidth || document.documentElement.clientWidth)
-          && rect.bottom >= 0
-          && rect.right >= 0
-      )
+      const rect = unref(element)?.getBoundingClientRect()
+      if (rect) {
```
webfansplz, line 35: Why we need this conditions,It‘s seem to always return true.

AllenYu0118: Yes, don't need

packages/core/useElementVisibility/directive.ts
AllenYu0118: You're right!

webfansplz (approved):

Issue #1374, opened 2022-03-08
useAxios：Do you need a way not to pass the url？


hengshanMWC:
### Clear and concise description of the problem

At present, 'url' is required to be transmitted, and then set inertia through '{immediate: false}'. I saw that some people need a way not to transmit `url`( [issues](https://github.com/vueuse/vueuse/issues/1347), but the implementation method I think of is different from him.

### Suggested solution

Judge that the first parameter is not a string. The default is {immediate: false}, and then `execute` (executeurl: String = URL, config: axiosrequestconfig = {}). 
```js
const {execute } = useAxios()
// or
const { execute } = useAxios(config)
// or
const { execute } = useAxios(instance)
// or 
const { execute } = useAxios(config, instance)
execute(url)
```

The `execute` function `url` here is optional, and `url2` will replace the `url1`.
```ts
const { execute } = useAxios(url1, {}, {immediate: false})
execute(url2)
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Issue #1347, opened 2022-03-03
Allow useAxios without an url


aalsamoht:
### Clear and concise description of the problem

I would like to be able to call `useAxios` without an url, it would be great if the following method signatures could be supported:

```ts
useAxios() // implicit immediate=false or pass the option { immediate: false }
useAxios(AxiosInstance) // also immediate=false
useAxios(AxiosRequestConfig)
useAxios(AxiosRequestConfig, AxiosInstance)
```

With `useAxios(AxiosInstance)` it would be easier to extend `useAxios` with a shared config:

```ts
export const axiosWithToken = axios.create({
  baseURL: config.baseURL,
  timeout: config.timeout,
});

axiosWithToken.interceptors.request.use(async (axiosConfig) => {
  // ....

  axiosConfig.headers.Authorization = `Bearer ${response.accessToken}`;

  return axiosConfig;
});

// Import and use this version of useAxios for protected APIs
export function useTokenAxios(config?: AxiosRequestConfig) {
  return useAxios(config, axiosWithToken);
}
```

I would like to keep all my endpoint configuration in the same place:
```ts
export const bookEndpoints = {
  getAll(): AxiosRequestConfig {
    return { url: 'books' };
  },
  getPaged(pageSize: number, pagenumber: number): AxiosRequestConfig {
    return { url: 'books', params: { pageSize, pagenumber } };
  },
  getSingle(id: string): AxiosRequestConfig {
    return { url: `books/${encodeURIComponent(id)}`, transformResponse: transformData };
  },
  create(objData: unknown): AxiosRequestConfig {
    return { url: 'books', method: 'POST', data: objData };
  },
  update(id: string, objData: unknown): AxiosRequestConfig {
    return { url: `books/${encodeURIComponent(id)}`, method: 'PUT', data: objData };
  },
  delete(id: string): AxiosRequestConfig {
    return { url: `books/${encodeURIComponent(id)}`, method: 'DELETE' };
  },
};
```

API usage:
```ts
// create, update, delete or something else that is not going to be executed in setup
const { execute } = useAxios() // implicit immediate=false or pass the option { immediate: false }

// execute the request some time in the future when the user has updated the data
execute(bookEndpoints.update('aBzw', myUpdatedBook))

// others executed immediately
useAxios(bookEndpoints.getAll())
// or
useAxios(bookEndpoints.getAll(), axiosWithToken)
// or 
useTokenAxios(bookEndpoints.getAll())
```

### Suggested solution

[...]

PR #1388, opened 2022-03-10
feat(useAxios): allow not passing url in constructor

hengshanMWC:
### Description
#1347 A scheme of `url` delay delivery is needed

### Additional context
#1374 This more flexible solution is provided

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


hengshanMWC:
Waiting for your minesweeping game

packages/integrations/useAxios/index.ts
```diff
@@ -112,9 +142,18 @@ export function useAxios<T = any>(url: string, ...args: any[]): UseAxiosReturn<T
     isLoading.value = loading
     isFinished.value = !loading
   }
-  const execute = (config: AxiosRequestConfig = {}) => {
+  const execute: OverallUseAxiosReturn<T>['execute'] = (executeUrl: string | AxiosRequestConfig | undefined = url, config: AxiosRequestConfig = {}) => {
+    let _url = ''
```
wvffle, line 146: Shouldn't this be `let _url = url`?

Now when someone would do following, it'd request the same url that user has visited instead of api which may break existing stuff. 
```js
const { execute } = useAxios('/api/...', {}, { immediate: false })
execute()
```

hengshanMWC: Failed to consider this business scenario

PR #1406, opened 2022-03-13
feat(onClickOutside): customizable phase

### LINKED ISSUE #1405: Can't toggle by using '@click' with onClickOutside
### Describe the bug

onClickOutside doesn't let me toggle the menu item on click. It opens it, onClickOutside works fine but can't close the menu by clicking the button again.
There is a workaround but maybe this should be corrected?: https://github.com/vueuse/vueuse/issues/1245

### Reproduction

https://stackblitz.com/edit/vitejs-vite-7v9ofd/?file=src%2Fmain.js

### System Info

```Shell
System:
    OS: Windows 10 10.0.19044
    CPU: (16) x64 Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
    Memory: 19.21 GB / 31.77 GB
  Binaries:
    Node: 16.13.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.4.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (99.0.1150.39)
```

### Used Package Manager

pnpm

### Validations

sibbng:
Fix #1405

### Description

If we were using bubbling phase, issues like #1405 could have been fixed by appending `.stop` modifier. Switching to bubbling is breaking change and have different behaviour than capture phase in terms of execution order.

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -10,6 +10,11 @@ export interface OnClickOutsideOptions extends ConfigurableWindow {
    * List of elements that should not trigger the event.
    */
   ignore?: MaybeElementRef[]
+  /**
+   * Use capturing phase for internal event listener.
+   * @default true
+   */
+  capture?: boolean
 }
 
 /**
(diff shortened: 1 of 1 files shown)


sibbng:
I realized that using bubbling phase, will take priority from the click handlers defined in templates. This is a breaking change. 

I found  bubbling behaviour more extensible and predictable than capturing phase. I will mark this PR as Draft. We might consider it for the next major version.

PR #1414, opened 2022-03-14
fix(useElementSize): make directive only registered once

webfansplz:
### Description

Already tested.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/useElementSize/directive.ts
```diff
@@ -11,13 +11,22 @@ type BindingValueFunction = (size: ElementSize) => void
...
 
-export const vElementSize: FunctionDirective<
+export const vElementSizeHandler = (): FunctionDirective<
 HTMLElement,
 BindingValueFunction | BindingValueArray
-> = (el, binding) => {
-  const handler = typeof binding.value === 'function' ? binding.value : binding.value?.[0]
-  const options = (typeof binding.value === 'function' ? [] : binding.value.slice(1)) as RemoveFirstFromTuple<BindingValueArray>
+> => {
+  let mountedEle: HTMLElement | null = null
```
antfu, line 18: What if you apply the directive to multiple elements? I guess we need to maintain a weakset, and remove it when the element unmounted

webfansplz: Good catch , but the functional directive callback only triggered when element mounted and updated , seems hard to remove it.

webfansplz: Maybe we could migrate it to object directives and provide a generic hook to be compatible with Vue 2, what do you think ?

antfu: Yeah I think object style would be better

PR #1417, opened 2022-03-14
fix(useIntersectionObserver/useScroll directive): convert useIntersectionObserver/scroll directive to objectDirective

Ericteen:
### Description

Current intersectionObserverDirective implementation is a functionDirective type, during the lifecycle of any element who uses this directive, it will be triggered multiple times, which cases excessive performance loss. So, I converted it to the objectDirective type and bounded to the `mounted` directive hook, in this way, the directive will be initialed only once and make it more concise compared to the #1411 which imported some excessive code.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


Ericteen:
I modified the `useScroll` directive the same way, which will limit the performance loss.

webfansplz:
Duplicate of #1415 
The mounted hook only work fine in Vue3，but we should support Vue2 & 3.

Ericteen:
> Duplicate of #1411 #1415 Thanks for the PR.but the mounted hook only work fine in Vue3,we should support Vue2 & 3.

btw, how can you use the directive function in vue 2? You have to manually register this directive function to vue2 hooks, bind, inserted or something else, then what's the difference? manually register step included

Issue #1396, opened 2022-03-11
Proposal: useAdBlockerDetector


adirkandel:
### Clear and concise description of the problem

As a developer using VueUse i want to let website owners know if their users use ad blocker to intercept ads and ads data.

With this functionality the website owners could decide what they want to do with user that uses an ad blocker, maybe to hide their content for instance.

Ads are mechanism for getting paid for a free content website owners provide, with ad blockers the content creators don't get to be paid.

I intent to submit a PR for this proposal.

### Suggested solution

In module useAdBlockerDetector I will provide a functionality that detects ad blockers.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #1426, opened 2022-03-15
Creating useAdBlockerDetector for detecting ad blockers on the browser

adirkandel:
### Description
Creating useAdBlockerDetector for detecting ad blockers on the browser

### Additional context
Resolves vueuse/vueuse#1396

---

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/useAdBlockerDetector/index.ts
```diff
@@ -0,0 +1,23 @@
...
+const hasAdBlocker = ref(false)
+/**
+ * Check if the user has an active ad blocker.
+ *
+ * @see https://vueuse.org/core/useAdBlockerDetector
+ */
+export async function useAdBlockerDetector(): Promise<Ref<boolean>> {
+  try {
+    await defaultWindow?.fetch(HONEY_POT_URL)
+  }
```
okxiaoliang4, line 18:
```suggestion
interface UseAdBlockerDetectorOptions extends ConfigurableWindow {
}

/**
 * Check if the user has an active ad blocker.
 *
 * @see https://vueuse.org/core/useAdBlockerDetector
 */
export async function useAdBlockerDetector(options: UseAdBlockerDetectorOptions = {}) {
  try {
    const {
      window = defaultWindow,
    } = options
    await window?.fetch(HONEY_POT_URL)
  }
```

packages/core/useAdBlockerDetector/index.ts
```diff
@@ -0,0 +1,23 @@
+import type { Ref } from 'vue-demi'
+import { ref } from 'vue-demi'
+import { defaultWindow } from '../_configurable'
```
okxiaoliang4, line 3:
```suggestion
import type { ConfigurableWindow } from '../_configurable'
import { defaultWindow } from '../_configurable'
```

packages/core/useAdBlockerDetector/index.ts
```diff
@@ -0,0 +1,23 @@
...
+
+// Parameters taken from easylist general advert blocking filters
+// https://easylist.to/easylist/easylist.txt
+const HONEY_POT_URL = '/&ad_code=ad&ad_height=ad&ad_ids=ad&ad_network_ad&ad_slot=ad&ad_sub=ad&ad_system=ad'
+
+const hasAdBlocker = ref(false)
```
wheatjs, line 9: should be placed inside of the function

adirkandel: I'm not sure about that because there is a possibility to check for ad blocker more than once, in that case i don't think we should define `hasAdBlocker` again, instead just update it's value every time we call `useAdBlockerDetector`

packages/core/useAdBlockerDetector/index.ts
adirkandel: @okxiaoliang4 why is it better to set the window as a dependency of the function?

antfu:
Thanks, but since this does not need Vue's reactivity I think it should be a separate library instead of in VueUse.

wheatjs:
> Thanks, but since this does not need Vue's reactivity I think it should be a separate library instead of in VueUse.

I agree with this sentiment. Adblockers are pretty smart and would probably be too much to maintain in a project like this. I'd suggest using something like https://github.com/sitexw/BlockAdBlock instead. Going to close this PR, but thanks for the contribution. All contributions are appreciated event if they don't end up getting merged!

Issue #1440, opened 2022-03-21
useKeyModifier: Uncaught TypeError: s.getModifierState is not a function


Shinigami92:
### Describe the bug

```node
Uncaught TypeError: s.getModifierState is not a function
    at HTMLDocument.<anonymous> (vendor.28d1767a.js:formatted:13655:25)
(anonymous) @ vendor.28d1767a.js:formatted:13655
```

<img width="497" alt="image" src="https://user-images.githubusercontent.com/7195563/159268355-793436c7-a6f5-433f-aab4-0623c42b3b3b.png">

<img width="571" alt="image" src="https://user-images.githubusercontent.com/7195563/159268201-b36d3cef-7638-4ec9-8602-3bbec1e895c8.png">

---

I assume we need to check that `s` (in code it's `evt`) is "correct" 🤔 

https://github.com/vueuse/vueuse/blob/eb248317227d8b809d9cfc5cbcd42fa730990932/packages/core/useKeyModifier/index.ts#L40

So we may need something like this(?):
```ts
if (typeof evt.getModifierState === 'function') state.value = evt.getModifierState(modifier) 
```

### Reproduction

Need to setup later

### System Info

```Shell
System:
    OS: macOS 12.3
    CPU: (10) arm64 Apple M1 Max
    Memory: 341.14 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    npm: 8.3.1 - ~/.nvm/versions/node/v16.14.0/bin/npm
  Browsers:
    Chrome: 99.0.4844.74
    Safari: 15.4
  npmPackages:
    @vueuse/core: ~8.1.2 => 8.1.2 
    vue: ~3.2.31 => 3.2.31
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #1441, opened 2022-03-22
fix(useKeyModifier): `getModifierState` is not a function

okxiaoliang4:
### Description

fixes: #1440

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useKeyModifier/index.ts
+++ b/packages/core/useKeyModifier/index.ts
@@ -37,7 +37,8 @@ export function useKeyModifier<Initial extends boolean | null>(modifier: KeyModi
   if (document) {
     events.forEach((listenerEvent) => {
       useEventListener(document, listenerEvent, (evt: KeyboardEvent) => {
-        state.value = evt.getModifierState(modifier)
+        if (typeof evt.getModifierState === 'function')
+          state.value = evt.getModifierState(modifier)
       })
     })
   }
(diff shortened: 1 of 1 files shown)


antfu:
Not sure. I guess we could have an `isSupported` flag, instead of silently failing in production.

antfu:
Oh I see.

okxiaoliang4:
> Oh I see. I think instead we shouldn't listen to 'mousedown' and 'mouseup' since they have no effects anyway

For example, if you press ctrl when the browser is out of focus, the `mousedown` event can update the state when you click on the page

packages/core/useKeyModifier/index.ts
```diff
@@ -37,7 +37,8 @@ export function useKeyModifier<Initial extends boolean | null>(modifier: KeyModi
   if (document) {
     events.forEach((listenerEvent) => {
       useEventListener(document, listenerEvent, (evt: KeyboardEvent) => {
```
antfu, line 39: KeyboardEvent | MouseEvent

okxiaoliang4: how about this? `Event & { getModifierState: (modifier: KeyModifier) => boolean }`

antfu: It's a internal type, that won't affect users so I don't think it need to be that general

okxiaoliang4: ok

PR #1446, opened 2022-03-23
fix(useStorage): sync too strict

cimchd:
fix(useStorage): sync too strict

- add timeout to reset sync after 150ms

### Description

After the latest changes to useStorage I had a problem, that the synchronisation between two browser tabs did not work as before. The first change was always ignored because the reset of the sync var was only triggered on the storage event which did not happen there. 
The fix adds a timeout that ensures that the sync var will always set back to false after some time independend of a storage event. 

### Additional context

It is discusable what is the best time in ms. Feel free to set it to a different value.

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useStorage/index.ts
+++ b/packages/core/useStorage/index.ts
@@ -90,6 +90,11 @@ export interface StorageOptions<T> extends ConfigurableEventFilter, Configurable
    * @default false
    */
   shallow?: boolean
+
+  /**
+   * Ensure sync usage. Prevents endless update loops.
+   */
+  ensureSync?: boolean
 }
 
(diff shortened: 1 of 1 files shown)


antfu (changes requested):
We'd better have an option for that.

cimchd:
> We'd better have an option for that.

I can change that. What should be the default behaviour? With the syncReset or without? From my point of view it's a breaking change if we set the default to "without syncReset" because the live synchronisation between browser tabs was working prior to v8.x.

antfu:
I mean the magic number `150` should be configurable.

cimchd:
> I mean the magic number `150` should be configurable.

OK :-) 
I will change that.

PR #1461, opened 2022-03-26
fix(useMediaQuery): media query before mount

### LINKED ISSUE #1302: useMediaQuery wrong initial value
https://user-images.githubusercontent.com/5492960/154996247-24119077-cb34-437b-8ae9-2b12c00710e4.mov

As you can see when the page is first loaded, it returns false, then it changes to true, is there any way to get the "valid" value on first instance?

I'm using SSR but it shouldn't affect AFAIK.

ryanmoyo:
### Description

Fix #1302 

### Additional context

I add a new function called `tryOnBeforeMount` like `tryOnMounted` to slove this issue, is this allowed? Or is this pr valid?

---

- [x] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other


ryanmoyo:
@antfu Please approve running workflows, thx.

PR #1463, opened 2022-03-27
docs(useBroswerLocation): recommand `useRoute`

### LINKED ISSUE #1445: useWindowLocation does not update on vue router changes, or anchor clicks
### Describe the bug

The returned value of `useWindowLocation` doesn't update when using `createWebHistory` mode for the router, and when they navigate via an `<a>`.

This makes since based off the source, as `popstate` is triggered when using the browsers navigation, and then `hashchange` is only triggered (I assume) when using `createHashHistory` mode.

Locally I have made a copy of `useWindowLocation` and added the following:
```ts
let lastPath = location.pathname;
useEventListener(
	window,
	"click",
	() => {
		if (location.pathname != lastPath) {
			lastPath = location.pathname;
			state.value = buildState("routechange");
		}
	},
	{ passive: true },
);
```

It works, however, I'm not sure if this is the best solution or the solution that the library would want.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-xzhoad/?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: Windows 10 10.0.19043
    Memory: 17.15 GB / 31.94 GB
  Binaries:
    Node: 16.11.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.1.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (99.0.1150.46)
  npmPackages:
    @vueuse/components: ^7.3.0 => 7.4.0
    @vueuse/core: ^8.1.1 => 8.1.1
    @vueuse/head: ^0.7.4 => 0.7.4
    vue: ^3.2.29 => 3.2.29
```
```


### Used Package Manager

pnpm

### Validations

### PR DESCRIPTION
**Pull request by ryanmoyo**
### Description

Fix #1445, it seems like there is no good solution, Router's internal processing logic makes this method fail and should give some hint.

### Additional context

---

- [x] Bug fix(enhance)
- [ ] New Feature
- [x] Documentation update
- [ ] Other

### CODE REVIEW

File: packages/core/useBrowserLocation/index.ts [hunk 1]
```diff
@@ -27,6 +27,8 @@ export interface BrowserLocationState {
  * @param options
  */
 export function useBrowserLocation({ window = defaultWindow }: ConfigurableWindow = {}) {
+  if (getCurrentInstance()?.appContext.config.globalProperties.$router) throw new Error('Do not use `useBrowserLocation` with Vue Router installed, use `useRoute` provided by Vue Router instead.')
```
antfu, line 30: In VueUse we are trying to avoid throw runtime errors. Also, since we can't imagine what's the usage that ppl needs, we shouldn't babysit here. I think a line in the doc is enough.

Issue #1464, opened 2022-03-28
proposal: useSort


okxiaoliang4:
### Clear and concise description of the problem

Composable reactive sort array.

### Suggested solution

```typescript
// useSort/index.ts
import { computed } from '@vue/reactivity'
import type { MaybeRef } from '@vueuse/shared'
import { unref } from 'vue-demi'

export type UseSortCompareFn<T> = (a: T, b: T) => number

export type UseSortFn<T> = (arr: T[], compareFn: UseSortCompareFn<T>) => T[]

const defaultCompare: UseSortCompareFn<number> = (a, b) => a - b

export interface UseSortOptions<T> {
  compareFn?: UseSortCompareFn<T>
}

export function useSort<T>(arr: MaybeRef<T[]>, sortFn: UseSortFn<T>, options: UseSortOptions<T> = {}) {
  const { compareFn = defaultCompare } = options
  return computed(() => sortFn(unref(arr), compareFn as UseSortCompareFn<T>))
}
```

```typescript
// useQuickSort/index.ts
import type { MaybeRef } from '@vueuse/shared'
import type { UseSortCompareFn, UseSortOptions } from '../useSort'
import { useSort } from '../useSort'

export function quickSort<T>(arr: T[], compareFn: UseSortCompareFn<T>): T[] {
  // sorted
  if (arr.length <= 1) return arr

  const pivot = arr[0]
  const centerArr: T[] = [pivot]
  const left: T[] = []
  const right: T[] = []

  for (let i = 1; i < arr.length; i++) {
    const element = arr[i]
    const compare = compareFn(element, pivot)
    if (compare === 0)
      centerArr.push(element)
    else if (compare < 0)
      left.push(element)
    else
      right.push(element)
  }

  return quickSort(left, compareFn).concat(centerArr, quickSort(right, compareFn))
}

export function useQuickSort<T>(arr: MaybeRef<T[]>, options: UseSortOptions<T> = {}) {
  return useSort(arr, quickSort, options)
}
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

mdbetancourt:
[...]

PR #1465, opened 2022-03-28
feat(useSort, useQuickSort): new function

okxiaoliang4:
### Description

Composable reactive sort array.

If this PR passes the merge, I will start adding other common sorting algorithms.

### Additional context

closes: #1464

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


okxiaoliang4 (commented):
self reviewed

packages/core/useSort/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * reactive sort array
+ *
+ * @see https://vueuse.org/useSort
+ * @param source source array
+ * @param sortFn sort function
+ * @param options
+ */
+export function useSort<T>(source: MaybeRef<T[]>, sortFn: UseSortFn<T>, options: UseSortOptions<T> = {}) {
```
antfu, line 31:
```suggestion
export function useSorted<T>(source: MaybeRef<T[]>, sortFn: UseSortFn<T>, options: UseSortOptions<T> = {}) {
```

`sort` sounds like mutating, while we are trying to be immutable. `sorted` sounds like a better fit https://github.com/tc39/proposal-change-array-by-copy

antfu: I would also like to see it matches with JS's sort more with an optional comparator `(a, b) => number`

packages/core/useQuickSort/index.ts
```diff
@@ -0,0 +1,27 @@
+import type { UseSortCompareFn } from '../useSort'
+import { useSortWrapFn } from '../useSort'
+
+export function quickSort<T>(source: T[], compareFn: UseSortCompareFn<T>): T[] {
```
antfu, line 4: I am a bit hesitate about this. Could we force on the general `useSorted` first and then think about the possibly way to support more alrigthom?

antfu: Also, I feel JS's native sort should be quite fast already (https://stackoverflow.com/a/37245185), I have never written a sorting algorithm in JS actually. Is there any benefit to implementing a quick sort on our own?

okxiaoliang4: > I am a bit hesitated about this. Could we force on the general `useSorted` first and then think about the possible way to support more algorithms?

ok~.

> Also, I feel JS's native sort should be quite fast already (https://stackoverflow.com/a/37245185), I have never written a sorting algorithm in JS actually. Is there any benefit to implementing a quick sort on our own?

Not sure. Every browser type has it's own javascript engine implementation.
Just wanted to provide a `useSorted` implementation here.

packages/core/useSort/index.ts
okxiaoliang4: > I would also like to see it matches with JS's sort more with an optional comparator `(a, b) => number`

Already have
https://github.com/vueuse/vueuse/pull/1465/files#diff-57d9b1ba0ff31d653b57c370d5eaf1e56421d821a8aa79aba512a3da23a2747dR13

okxiaoliang4, line 31: @antfu Perhaps it would be better if I added sortFn to the options, providing a default sort method.

```typescript
export interface UseSortedOptions<T> {
  /**
   * sort algorithm
   */
  sortFn?: UseSortedFn<T>
  /**
   * compare function
   */
  compareFn?: UseSortedCompareFn<T>
  /**
   * change the value of the source array
   * @default false
   */
  dirty?: boolean
}

const defaultSortFn = <T>(source: T[], compareFn: UseSortedCompareFn<T>): T[] => source.sort(compareFn)

export function useSorted<T>(source: MaybeRef<T[]>, options: UseSortedOptions<T> = {}) {
  const { sortFn = defaultSortFn, compareFn = defaultCompare, dirty = false } = options
  // ...
}
```

antfu: Looks good to me

PR #1468, opened 2022-03-30
feat(eventRef): add function

aa900031:
### Description

Converts event listener to reactive Ref

- Unrestricted `addEventListener`: Can receive any event register(ex: wrap from [mitt](https://github.com/developit/mitt), [Swiper.js](https://swiperjs.com/) events)
- Reactive getter: Just use like `computed`
- Reactive register: Observe change in register, and call register again
- Lazy init

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
TBH I feel this is a bit too opinionated. Since we have `useEventListener` already, maybe we could have `useEventListenerRef` that expose the event object as ref

```ts
const event = useEventListenerRef(window, 'scroll')
```

```ts
const scrollTop = computed({
  get() { return event.value?.pageY ?? window.pageYOffset ?? window.scrollY },
  set(val) { window.scrollTo(0, val) }
})
```

to be more composable?

aa900031:
@antfu For example: It can composition with `socket io`, can easy get a reactive message

```vue
<script setup>
import { ref } from 'vue';
import { io } from "socket.io-client";
import { eventRef } from '@vueuse/core';

const socket = io('ws://localhost')

const news = eventRef((handler) => {
  socket.on('news', handler)
  return () => socket.off('news', handler)
}, (event) => event)
</script>
```

Issue #1450, opened 2022-03-23
Function proposal: `useTypeAhead()`


0x009922:
### Clear and concise description of the problem

In native UIs as well as in native `<select>` element there is a very useful a11y feature - "type ahead". It is easy to understand how it works from this demo:

https://user-images.githubusercontent.com/43530070/159703763-af3bc9e2-1852-4bfd-aa13-8539e82a7e0f.mov

In WAI-ARIA it is [described](https://www.w3.org/TR/wai-aria-practices-1.1/#listbox_kbd_interaction) as following:

> Type-ahead is recommended for all listboxes, especially those with more than seven options:
> - Type a character: focus moves to the next item with a name that starts with the typed character.
> - Type multiple characters in rapid succession: focus moves to the next item with a name that starts with the string of characters typed.

So, it would be great to have an easy-to-use composable that brings that so much useful feature! This feature is not so easy to implement, and it can greatly increase quality of an application and UX.

### Suggested solution

I've found an implementation of it in Material Components (in MDCList component): [typeahead.ts
](https://github.com/material-components/material-components-web/blob/f31a833fae6f132318068f30a24a12c6c0cc192f/packages/mdc-list/typeahead.ts). I think it can be ported.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

czh740132583:
That's a cool idea

michealroberts:
I’m going to tackle this tomorrow unless anybody else is working on it? :)

OrbisK:
a pr was closed with this comment: https://github.com/vueuse/vueuse/pull/1476#issuecomment-1116307947

closing this issue for now

PR #1476, opened 2022-04-03
feat(useTypeAHead) Simulate the implementation of the 'TypeAHead' feature in the native UI and native < Select > elements.

baiwusanyu-c:
## useTypeAHead
close: #1450 
Simulate the implementation of the 'TypeAHead' feature in the native UI and native < Select > elements. If you want to write a component with the 'TypeAHead' feature yourself, you can use 'useTypeAHead' to obtain the focus index after keyboard interaction

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


baiwusanyu-c:
wow, how often do you guys watch pr? I hope to get some replies or suggestions.😢

baiwusanyu-c:
I have modified the unit test. Could you please execute the workflow again？@antfu

antfu:
I think this one is a bit too complex and too specific that obeys the rules of VueUse to have general and composable functions. Thanks for the PR but sorry I think we probably won't take it as-is. Maybe you can refactor and see if we could have some general logic extracted from this to compose what you want to achieve.

PR #1502, opened 2022-04-18
fix(useRafFn): call `cancelAnimationFrame` on pause

r1ader:
### Description

This commit is out of my notice about the green button `Transition` in [useTransition demo](https://vueuse.org/core/usetransition/) 

If you use the performance panel of chrome dev tool to record , and then click button `Transition` quickily for several times.

And then you will find there are many `Animation Frame Fired` in just one frame.

This commit try to add `cancelAnimationFrame` to avoid the unnecessary invoke of the repeated raf callbacks

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useRafFn/index.ts
+++ b/packages/core/useRafFn/index.ts
@@ -46,7 +46,7 @@ export function useRafFn(fn: Fn, options: RafFnOptions = {}): Pausable {
 
   function pause() {
     isActive.value = false
-    if(rafId.value && window){
+    if(rafId.value && window) {
       window.cancelAnimationFrame(rafId.value)
       rafId.value = null
     }
--- a/packages/core/useRafFn/index.ts
+++ b/packages/core/useRafFn/index.ts
@@ -27,13 +27,14 @@ export function useRafFn(fn: Fn, options: RafFnOptions = {}): Pausable {
   } = options
 
   const isActive = ref(false)
+  const rafId = ref(null)
 
   function loop() {
     if (!isActive.value || !window)
       return
 
     fn()
-    window.requestAnimationFrame(loop)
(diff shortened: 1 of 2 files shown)


packages/core/useRafFn/index.ts
```diff
@@ -45,6 +46,10 @@ export function useRafFn(fn: Fn, options: RafFnOptions = {}): Pausable {
 
   function pause() {
     isActive.value = false
+    if(rafId.value && window){
```
okxiaoliang4, line 49:
```suggestion
    if(rafId.value && window) {
```

okxiaoliang4 (changes requested):

okxiaoliang4 (dismissed):

packages/core/useRafFn/index.ts
```diff
@@ -27,13 +28,14 @@ export function useRafFn(fn: Fn, options: RafFnOptions = {}): Pausable {
   } = options
 
   const isActive = ref(false)
+  const rafId = ref(null) as Ref<null | number>
```
antfu, line 31:
```suggestion
  const rafId: number | undefined
```

This does not need to be a ref

packages/core/useRafFn/index.ts
```diff
@@ -47,9 +46,9 @@ export function useRafFn(fn: Fn, options: RafFnOptions = {}): Pausable {
 
   function pause() {
     isActive.value = false
-    if (rafId.value && window) {
-      window.cancelAnimationFrame(rafId.value)
-      rafId.value = null
+    if (typeof rafId === 'number' && window) {
```
antfu, line 49:
```suggestion
    if (rafId != null && window) {
```

PR #1507, opened 2022-04-19
fix(refAutoReset): clear timeout on scope dispose

wvffle:
### Description

The timeout is not cleared on scope dispose in refAutoReset. This PR fixes that.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/shared/refAutoReset/index.ts
+++ b/packages/shared/refAutoReset/index.ts
@@ -1,6 +1,7 @@
 import type { Ref } from 'vue-demi'
 import { customRef, unref } from 'vue-demi'
 import type { MaybeRef } from '@vueuse/shared'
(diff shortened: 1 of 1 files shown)


packages/shared/refAutoReset/index.ts
```diff
@@ -20,6 +21,11 @@ export function refAutoReset<T>(defaultValue: T, afterMs: MaybeRef<number> = 100
         trigger()
       }, unref(afterMs))
 
+    tryOnScopeDispose(() => {
+      if (timer)
+        clearTimeout(timer)
```
antfu, line 26:
```suggestion
       clearTimeout(timer)
```

PR #1522, opened 2022-04-23
fix(onClickOutside): fallback to `pointerup` event if `click` event not propagate

### LINKED ISSUE #1520: onClickOutside not working on Safari iOS
### Describe the bug

On safari on iOS the feature doesn’t work at all. You can see it in the demo itself.

The video below shows how on safari tapping outside on the demo doesn’t hide the modal or tooltip. It only works if you tap on an element that triggers a pointer event.

https://user-images.githubusercontent.com/20112149/164793336-2d5eda46-7cd0-45ff-b30b-65ec960f31d1.MOV

The second video shows how on chrome does work as expected.

https://user-images.githubusercontent.com/20112149/164793899-5f2b180e-d69f-4081-ab77-9e78e4c39aad.MOV

### Reproduction

https://vueuse.org/core/onclickoutside/

### System Info

```Shell
-
```

### Used Package Manager

npm

### Validations

sibbng:
Fix #1520

### Description

[This article](https://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html) from 2014 clearly explains what is going on there. This PR simply workaround this issue by falling back to `pointerup` event. 

`click` event is fired after all release events (`mouseup`, `touchend`, `pointerup`). So, if click event is propagated normally within 50ms, timeout will be cleared. Initially, I was thinking to use `touchend` event. But the type of `touchend` event is not compatible with PointerEvent.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -37,7 +37,11 @@ export function onClickOutside(
 
   const shouldListen = ref(true)
 
+  let fallback: number
+
   const listener = (event: PointerEvent) => {
+    window.clearTimeout(fallback)
+
     const el = unrefElement(target)
     const composedPath = event.composedPath()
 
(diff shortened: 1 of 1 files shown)


antfu (approved):

hminghe:
event.composedPath() expires after setTimeout

Issue #1517, opened 2022-04-22
Proposal: `useToggle` new options


okxiaoliang4:
### Clear and concise description of the problem

Sometimes we need to toggle `0` or `1`. `useToggle` does not do this very well.

### Suggested solution

```ts
export interface UseToggleOptions<Truly, Falsely> {
  trulyValue?: MaybeRef<Truly>
  falselyValue?: MaybeRef<Falsely>
}

export function useToggle<Truly, Falsely, T = Truly | Falsely>(initialValue: Ref<T>, options?: UseToggleOptions<Truly, Falsely>): (value?: T) => T
export function useToggle<Truly = true, Falsely = false, T = Truly | Falsely>(initialValue?: T, options?: UseToggleOptions<Truly, Falsely>): [Ref<T>, (value?: T) => T]

/**
 * A boolean ref with a toggler
 *
 * @see https://vueuse.org/useToggle
 * @param [initialValue=false]
 */
export function useToggle(initialValue: MaybeRef<unknown> = false, options: UseToggleOptions<true, false> = {}) {
  const {
    trulyValue = true,
    falselyValue = false,
  } = options

  const valueIsRef = isRef(initialValue)
  const innerValue = ref(initialValue) as Ref<boolean>

  function toggle(value?: boolean) {
    // has arguments
    if (arguments.length) {
      innerValue.value = value!
      return innerValue.value
    }
    else {
      innerValue.value = innerValue.value === unref(trulyValue) ? unref(falselyValue) : unref(trulyValue)
      return innerValue.value
    }
  }

  if (valueIsRef)
    return toggle
  else
    return [innerValue, toggle]
}
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
Maybe use this https://vueuse.org/core/useCycleList ?

okxiaoliang4:
> Maybe use this https://vueuse.org/core/useCycleList ?

I think `useToggle` is more appropriate.

we can reduce judgment conditions like this
[...]

PR #1523, opened 2022-04-23
feat(useToggle): new options

okxiaoliang4:
### Description

support toggle `trulyValue` and `falselyValue`

### Additional context

#1517

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/shared/useToggle/index.ts
```diff
@@ -1,33 +1,44 @@
...
+      return innerValue.value
     }
-
-    return [boolean, toggle] as const
   }
+
+  if (valueIsRef)
+    return toggle
+  else
+    return [innerValue, toggle]
```
antfu, line 43:
```suggestion
    return [innerValue, toggle] as const
```

PR #1526, opened 2022-04-24
feat(useScreenOrientation): new function

michael-observerly:
### Description

This is a reactive composable that can relay the `OrientationLockType`, e.g., "primary-portrait" etc reactively, listening to the `orientationchange` event.

It is modelled around the [Screen Orientation API MDN](https://developer.mozilla.org/en-US/docs/Web/API/Screen_Orientation_API).

I'm not 100% convinced we need locking and unlocking, as this is almost a unused feature for most workflows, so please let me know if this is required.

Please view [useScreenOrientation Loom Video Demonstration](https://www.loom.com/share/39ac4c911b3e4b44b4d81a85ede79bd1) for demonstration.

- [ ] Bug fix
- [X] New Feature
- [X] Documentation update
- [ ] Other


packages/core/useScreenOrientation/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  const isSupported = !!(window && 'screen' in window && 'orientation' in window.screen)
+
+  const screenOrientation = isSupported ? window.screen.orientation : {} as ScreenOrientation
+
+  const orientation = ref<OrientationType | undefined>(screenOrientation.type)
+  const angle = ref(screenOrientation.angle || 0)
+
+  if (isSupported) {
+    useEventListener(screenOrientation, 'change', () => {
```
okxiaoliang4, line 25: I guess it would be better to register the listeners this way than on the window.But I tried it and found that the window event fired first, I will modify it back

okxiaoliang4 (changes requested):

okxiaoliang4 (dismissed):

packages/core/useScreenOrientation/index.ts
michael-observerly: Yeh there are deprecation changes that you needed to be aware of when putting this together.

michael-observerly:
@okxiaoliang4 I'm not sure I like the use of the Promise reject in the lockOrientation() call ...

okxiaoliang4:
> @okxiaoliang4 I'm not sure I like the use of the Promise reject in the lockOrientation() call ...

We could return `Promise.resolve()`. But I'm not sure which one is better.

michael-observerly:
@okxiaoliang4 Thank you for the suggestions, just updated the documentation for all correct orientation types. It will need your review again! Thanks :)

michael-observerly:
> > @okxiaoliang4 I'm not sure I like the use of the Promise reject in the lockOrientation() call ...
> 
> We could return `Promise.resolve()`. But I'm not sure which one is better.

I think the Promise.reject makes sense actually, it looks like `oreintation.lock()` returns a `Promise`, so yeh, I think you're right to return a rejection here.

michael-observerly:
@okxiaoliang4 Also, could you set me to the assignee please as I was the one that opened the PR...

packages/core/useScreenOrientation/index.md
```diff
@@ -0,0 +1,47 @@
...
+const {
+  isSupported,
+  orientation,
+  angle,
+  lockOrientation,
+  unlockOrientation,
+} = useScreenOrientation()
+```
+
+To lock the orientation, you can pass an [OrientationLockType](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/type) to the lockOrientation function:
```
okxiaoliang4, line 23:
```suggestion
To lock the orientation, you can pass an [OrientationLockType](https://github.com/microsoft/TypeScript/blob/06fb30725d4a70708b28f2a4613543ce8c4cd645/lib/lib.dom.d.ts#L18805) to the lockOrientation function:
```

packages/core/useScreenOrientation/index.md
```diff
@@ -0,0 +1,47 @@
...
+lockOrientation('primary-portrait')
+```
+
+and then unlock again, with the following:
+
+```ts
+unlockOrientation()
+```
+
+Accepted orientation types are one of `"landscape-primary"`, `"landscape-secondary"`, `"portrait-primary"`, `"portrait-secondary"`, `"any"`, `"landscape"`, `"natural"` and `"portrait"`.
```
okxiaoliang4, line 45:
```suggestion
Accepted orientation types are one of [OrientationLockType](https://github.com/microsoft/TypeScript/blob/06fb30725d4a70708b28f2a4613543ce8c4cd645/lib/lib.dom.d.ts#L18805) to the lockOrientation function:
```

okxiaoliang4 (commented):
Maybe we can link to the type definition [lib.dom.d.ts](https://github.com/microsoft/TypeScript/blob/06fb30725d4a70708b28f2a4613543ce8c4cd645/lib/lib.dom.d.ts#L18805)

okxiaoliang4:
> Maybe we can link to the type definition [lib.dom.d.ts](https://github.com/microsoft/TypeScript/blob/06fb30725d4a70708b28f2a4613543ce8c4cd645/lib/lib.dom.d.ts#L18805)

doesn't seem like a good idea.

okxiaoliang4 (approved):

okxiaoliang4:
> Also, could you set me to the assignee please as I was the one that opened the PR

sure, I only want follow this pr

michael-observerly:
> > Also, could you set me to the assignee please as I was the one that opened the PR
> 
> sure, I only want follow this pr

Awesome 🙏 thank you. Just so I have the blame on this one if there are issues, so new Issues can be assigned to me. :)

PR #1539, opened 2022-04-28
feat(useSliceUpload): new function

Asaki-M:
### Description

Add a new function.
This function supports an upload feature base on a big file slice upload.

When the file slice, it needs to use an MD5 mark. So I use a 'spark-md5' library.

### Additional context

---

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other


wheatjs:
Not too sure about this one. I think we want to avoid adding third-party dependencies to the core library if we can. Also there is not much reactivity in this function, typically we would return some objects, in this case you we would want to return something like `uploadProgress: Ref<number>` and other things like that. Last comment is I think functions like these can become pretty complicated in their implantations and might create issues for users.

Thinking this might be better off as its own package maybe?

Asaki-M:
> Not too sure about this one. I think we want to avoid adding third-party dependencies to the core library if we can. Also there is not much reactivity in this function, typically we would return some objects, in this case you we would want to return something like `uploadProgress: Ref<number>` and other things like that. Last comment is I think functions like these can become pretty complicated in their implantations and might create issues for users.
> 
> Thinking this might be better off as its own package maybe?
Hmmm, I see the core library uses the 'md5' dependencies, maybe using it can replace the 'spark-md5' dependencies. And provide some responsive properties. for example, upload progress.
I think this function becomes pretty complicated in their implantation as you said. So I don't know if this PR is correct.

wheatjs:
Going to close this as I feel this is better left up to being its own library, thanks for the PR though

PR #1545, opened 2022-05-01
feat(useFetch): chain callbacks from createFetch and useFetch

### LINKED ISSUE #1310: chain beforeFetch/afterFetch/onFetchError from createFetch and useFetch

Currently, a handler defined in `createFetch` can be overwritten in `useFetch`.

For example, we can define a `createFetch` with a global `beforeFetch` option to set a bearer token, and in the `useFetch` call, add a `beforeFetch` option to set some other needed headers.

### Suggested solution

useFetchFactory should be able to chain beforeFetch/afterFetch/onFetchError instead of overwriting them.

### Alternative

_No response_

### Additional context

Manually chain for now:
```typescript
const injectToken = function ({ options }: BeforeFetchContext) {
  options.headers = {
    ...options.headers,
    Authorization: `Bearer ${Vue.$auth.token()}`
  }

  return { options }
}

const useApiFetch = createFetch({
  baseUrl: '/api/v1',
  options: {
    beforeFetch: injectToken
  }
})

// fetchOptions comes from an external caller
useApiFetch(path, {
    immediate: false,
    initialData: defaultValue,
    // Complicated stuff to be able to workaround beforeFetch overwrite
    ...fetchOptions,
    ...(fetchOptions?.beforeFetch === undefined ? {} : {
      async beforeFetch (ctx) {
        ctx = {...ctx, ...injectToken(ctx)}
        ctx = {...ctx, ...(await fetchOptions.beforeFetch?.(ctx))}
        return ctx
      }
    })
  })
```

### Validations

tobyzerner:
### Description

Chain beforeFetch/afterFetch/onFetchError callbacks from `createFetch` and `useFetch`, rather than `useFetch` overwriting them.

### Additional context

closes #1310

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


wheatjs (dismissed):

climba03003:
It is actually a `breaking change`. 
And `overwrite` of the functions is valuable function.

It should allow the user to choose which behavior they desire.

antfu:
@climba03003 That makes sense, do you want to introduce the option to toggle the behavior via another PR to `next` branch? Thanks.

antfu (approved):

climba03003:
I will send a PR when I have time.

PR #1549, opened 2022-05-02
feat(useUniqueId): new function

xyhxx:
### Description
add new function
generate globally unique string id

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
Any reason it needs to be Vue specific? If not I would recommend directly use `nanoid` instead.

xyhxx:
i guess you are right

PR #1550, opened 2022-05-02
fix(onLongPress): make directive work normal

### LINKED ISSUE #1543: v-on-long-press: multiple call on upon subsequent long-press
### Describe the bug

When using directive `v-on-long-press`, the callback gets called multiple times on subsequent long-press. However, It is working as expected when using as `onLongPress()`

```vue
<script setup lang="ts">
import { ref } from 'vue';
import { onLongPress } from '@vueuse/core';
import { vOnLongPress } from '@vueuse/components';

const count = ref(0);
const longPressBtn = ref<HTMLElement | null>(null);
onLongPress(longPressBtn, (e: PointerEvent) => {
  count.value++;
});
</script>

<template>
  <h1>Hello VueUse!</h1>
  <a href="https://vueuse.org/" target="_blank">Documentation</a>

  <div>
    {{ count }}
  </div>

  <div>
    <!-- !error: multiple call on v-on-long-press upon subsequent long-press -->
    <button
      v-on-long-press="
        () => {
          count++;
        }
      "
    >
      v-on-long-press
    </button>
  </div>

  <div>
    <!-- work as expected -->
    <button ref="longPressBtn">button</button>
  </div>
</template>

<style></style>

```

### Reproduction

https://stackblitz.com/edit/vitejs-vite-cdcr11?file=src/App.vue

### System Info

```Shell
System:
    OS: Windows 10 10.0.19044
    CPU: (4) x64 Intel(R) Core(TM) i5-7260U CPU @ 2.20GHz
    Memory: 2.54 GB / 15.89 GB
  Binaries:
    Node: 14.17.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.15 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.5 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 100.0.4896.127
    Edge: Spartan (44.19041.1266.0), Chromium (101.0.1210.32)
```

### Used Package Manager

pnpm

### Validations

webfansplz:
### Description

Closes #1543

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/onLongPress/directive.test.ts
```diff
@@ -47,13 +46,6 @@ describe('vOnLongPress', () => {
     it('should be defined', () => {
       expect(wrapper).toBeDefined()
     })
-
-    it('should trigger longpress after 500ms', async() => {
```
webfansplz, line 51: It was removed because `@vue/test-utils` only support `Vue3`, which will never trigger vue2's directive `inserted` hook. 

So maybe we just need to test it manually to make sure it works .

antfu: Could wrap it with `isVue3` from `vue-demi`

webfansplz: Ok, done.

PR #1557, opened 2022-05-04
fix(useVModel): returning undefined on false values

wheatjs:
### Description

In the latest release we introduced a default value to v-model. But if the value is false it will return undefined instead which is not desired behavior 
image

### Additional context

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useVModel/index.ts
+++ b/packages/core/useVModel/index.ts
@@ -86,7 +86,7 @@ export function useVModel<P extends object, K extends keyof P, Name extends stri
   else {
     return computed<P[K]>({
       get() {
-        return typeof props[key!] !== 'undefined' ? props[key!] : defaultValue!
+        return props[key!] ?? defaultValue!
       },
       set(value) {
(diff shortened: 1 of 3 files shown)


packages/core/useVModel/index.ts
```diff
@@ -86,7 +86,7 @@ export function useVModel<P extends object, K extends keyof P, Name extends stri
   else {
     return computed<P[K]>({
       get() {
-        return props[key!] || defaultValue!
+        return typeof props[key!] !== 'undefined' ? props[key!] : defaultValue!
```
antfu, line 89: Can't we just use `??`?

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: 5.4
+lockfileVersion: 5.3
```
antfu, line 1: You might want to upgrade your local pnpm

PR #1562, opened 2022-05-04
fix(useScrollLock): stop listener properly

okxiaoliang4:
### Description

stop event on scope dispose.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useScrollLock/index.ts
+++ b/packages/core/useScrollLock/index.ts
@@ -1,6 +1,7 @@
 import { computed, ref, unref, watch } from 'vue-demi'
 import type { Fn, MaybeRef } from '@vueuse/shared'
-import { isClient } from '@vueuse/shared'
+import { isClient, tryOnScopeDispose } from '@vueuse/shared'
+
 import { useEventListener } from '../useEventListener'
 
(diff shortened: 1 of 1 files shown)


packages/core/useScrollLock/index.ts
```diff
@@ -63,11 +64,13 @@ export function useScrollLock(
     const ele = (unref(element) as HTMLElement)
     if (!ele || !isLocked.value)
       return
-    isIOS && touchMoveListener?.()
+    isIOS && stopTouchMoveListener?.()
     ele.style.overflow = initialOverflow
     isLocked.value = false
   }
 
+  tryOnScopeDispose(unlock)
```
okxiaoliang4, line 72: I'm not sure about this.
Maybe we just need to recycle events?

```ts
tryOnScopeDispose(() => {
  stopTouchMoveListener?.()
})
```

antfu: useEventListener already disposes automatically and we don't need this.

okxiaoliang4: Unfortunately, the listener is started within the `lock`, should be in the global scope

PR #1568, opened 2022-05-05
chore(useVModel): unified value

okxiaoliang4:
### Description

In #1567, I don't think null should use the default value.

Update the passive option case.

### Additional context
#1567 
#1565 

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/packages/core/useVModel/index.ts
+++ b/packages/core/useVModel/index.ts
@@ -70,8 +70,10 @@ export function useVModel<P extends object, K extends keyof P, Name extends stri
 
   event = eventName || event || `update:${key}`
 
+  const getValue = () => isDef(props[key!]) ? props[key!] : defaultValue
+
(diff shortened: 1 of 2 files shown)


packages/core/useVModel/index.ts
```diff
@@ -70,8 +70,10 @@ export function useVModel<P extends object, K extends keyof P, Name extends stri
 
   event = eventName || event || `update:${key}`
 
+  const getValue = () => isDef(props[key!]) ? props[key!] : defaultValue
```
okxiaoliang4, line 73: Or we can determine if the property exists in the object and thus determine if the default value is obtained

```ts
  const getValue = () => Object.hasOwnProperty.call(props, key!) ? props[key!] : defaultValue
```

PR #1571, opened 2022-05-06
fix(useFetch): return PromiseLike after setting method

tobyzerner:
### Description

This fixes an oversight where if you call any of the method-setting methods (`post`, `patch`, etc.), the return value is not `PromiseLike` as the typings would suggest.

```ts
useFetch('http://example.com').post({ foo: 'bar' }).then(() => {}) // undefined is not a function
```

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -481,7 +481,13 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
         if (!payloadType && unref(payload) && Object.getPrototypeOf(unref(payload)) === Object.prototype)
           config.payloadType = 'json'
 
-        return shell as any
+        return {
+          ...shell,
+          then(onFulfilled: any, onRejected: any) {
+            return waitUntilFinished()
+              .then(onFulfilled, onRejected)
+          },
+        } as any
       }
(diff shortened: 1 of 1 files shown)


packages/core/useFetch/index.ts
```diff
@@ -481,7 +481,13 @@ export function useFetch<T>(url: MaybeRef<string>, ...args: any[]): UseFetchRetu
...
           config.payloadType = 'json'
 
-        return shell as any
+        return {
+          ...shell,
+          then(onFulfilled: any, onRejected: any) {
+            return waitUntilFinished()
+              .then(onFulfilled, onRejected)
+          },
+        } as any
```
okxiaoliang4, line 490: maybe like this
```suggestion
        } as UseFetchReturn<T> & PromiseLike<UseFetchReturn<T>>
```

tobyzerner: Yeah fair suggestion, I just copied the implementation of `setType` below :) https://github.com/vueuse/vueuse/pull/1571/files#diff-0db62ed388d474bc80cd69e2ff1593cd6c9719b8794d8f88459a1704e8da25b5L508

PR #1580, opened 2022-05-08
fix(useFocus): `focused` should be `false` when `element` is undefined

### LINKED ISSUE #1560: useFocus demo abnormal when refresh
### Describe the bug

when u click the left sidebar menu , the router path is : https://vueuse.org/core/useFocus/  
**useFocus** the F is uppercase
and the focus state  of all three button is correct

BUT

when u refresh the page, the url path is :   https://vueuse.org/core/usefocus/

**usefocus** the F is lowercase

then the three buttons are in abnormal focus state

### Reproduction

https://vueuse.org/functions.html

### System Info

```Shell
System:
    OS: Windows 10 10.0.17134
    CPU: (8) x64 Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
    Memory: 4.99 GB / 15.78 GB
  Binaries:
    Node: 16.13.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.1.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (42.17134.2208.0)
    Internet Explorer: 11.0.17134.1
  npmPackages:
    @vueuse/core: ^8.4.1 => 8.4.1
    vue: ^3.2.31 => 3.2.31
```

### Used Package Manager

yarn

### Validations

webfansplz:
### Description

Closes #1560 

Caused by : both of the element to be `undefined` when the `Vitepress` generate HTML page.

image

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useFocus/index.ts
+++ b/packages/core/useFocus/index.ts
@@ -1,5 +1,6 @@
 import type { Ref } from 'vue-demi'
 import { computed, watch } from 'vue-demi'
+import { isDef } from '@vueuse/shared'
 import type { MaybeElementRef } from '../unrefElement'
 import { unrefElement } from '../unrefElement'
 import { useActiveElement } from '../useActiveElement'
@@ -36,9 +37,7 @@ export function useFocus(target: MaybeElementRef, options: UseFocusOptions = {})
   const targetElement = computed(() => unrefElement(target))
   const focused = computed({
     get() {
-      if (activeElement.value === undefined && targetElement.value === undefined)
-        return false
-      return activeElement.value === targetElement.value
+      return isDef(activeElement.value) && isDef(targetElement.value) && activeElement.value === targetElement.value
     },
     set(value: boolean) {
       if (!value && focused.value)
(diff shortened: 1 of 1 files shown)


packages/core/useFocus/index.ts
```diff
@@ -36,6 +36,8 @@ export function useFocus(target: MaybeElementRef, options: UseFocusOptions = {})
   const targetElement = computed(() => unrefElement(target))
   const focused = computed({
     get() {
+      if (activeElement.value === undefined && targetElement.value === undefined)
+        return false
       return activeElement.value === targetElement.value
```
okxiaoliang4, line 41:
```suggestion
      return isDef(activeElement.value) && isDef(targetElement.value) && activeElement.value === targetElement.value
```

okxiaoliang4 (approved):

PR #1583, opened 2022-05-09
fix(useUrlSearchParams): should return initialValue

odex21:
### Description
Now the empty url will overwrite the initial value, making the initial value ineffective.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useUrlSearchParams/index.ts
+++ b/packages/core/useUrlSearchParams/index.ts
@@ -44,7 +44,7 @@ export function useUrlSearchParams<T extends Record<string, any> = UrlParams>(
   if (!window)
     return reactive(initialValue) as T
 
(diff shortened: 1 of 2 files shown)


packages/core/useUrlSearchParams/index.ts
```diff
@@ -44,7 +44,7 @@ export function useUrlSearchParams<T extends Record<string, any> = UrlParams>(
   if (!window)
     return reactive(initialValue) as T
 
-  const state: Record<string, any> = reactive(initialValue)
+  const state: Record<string, any> = reactive({})
```
impeiran, line 47: It won't trigger the watcher callback by initializing the reactive state with `{}` in vue2

Issue #1301, opened 2022-02-21
bug(useEyeDropper): "open" button is disabled when loading the page directly


darrylnoakes:
When loading the `useEyeDropper` page directly, the "open" button momentarily shows the "not supported" message, and then the demo finishes loading (the output text appears above the button) and the button text switches to "Open Eye Dropper", but the button remains disabled.

Navigating to the page from another page on the site is fine, with no loading period at all.

ryanmoyo:
I tried some methods to slove this, the only way is put it in onMounted, but it will be wired that it first showed `not sppport ...` and then became normal and available.

darrylnoakes:
It seems to be something to do with `isSupported` not being a ref. The button for some reason starts as disabled with the disabled text, and then the text changes but not the disabled state.

[...]

Issue #949, opened 2021-11-23
useEyeDropper demo bug


okxiaoliang4:
Visit [eyeDropper](https://vueuse.org/core/useeyedropper/) directly or after refreshing the page isSupported is true but the button is still disabled.
This problem does not occur when running locally.

PR #1598, opened 2022-05-13
fix(useEyeDropper): improve SSR compatibility (#949)(#1301)

zhowiny:
### Description
Visit [eyeDropper](https://vueuse.org/core/useeyedropper/) directly or after refreshing the page isSupported is true but the button is still disabled.

### Additional context

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


okxiaoliang4:
I'm not sure about this. Because `isSupported` of other functions is also not reactive

zhowiny:
> I'm not sure about this. Because `isSupported` of other functions is also not reactive

I found that other functions have similar problems, such as useClipboard. If the above method is determined to be effective, I can fix all places

antfu:
Close in favor of #1800

PR #1610, opened 2022-05-17
feat(useDropZone): new function

Julien-Martin:
### Description

Create a zone where you can drop files.

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


wheatjs:
API feels a little weird. Typically in VueUse we pass in an element reference rather than returning handlers. I feel like we should be consistent here

Julien-Martin:
> API feels a little weird. Typically in VueUse we pass in an element reference rather than returning handlers. I feel like we should be consistent here

You right, i just changed it, and that make more sense like you say and you can track which element (ref) is using the dropZone :+1:

wheatjs (dismissed):

wheatjs:
Looks good to me

packages/core/useDropZone/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import { isClient } from '@vueuse/shared'
+import { useEventListener } from '../useEventListener'
+
+export interface UseDropZoneReturn {
+  isOverDropZone: Ref<boolean>
+}
+
+export function useDropZone(target: MaybeRef<HTMLElement | null>, onDrop: (files: File[] | null) => void): UseDropZoneReturn {
+  const isOverDropZone = ref(false)
+  const counter = ref(0)
```
wheatjs, line 13: Quick question, any reason we are using a counter here? If there is we can probably just use a normal variable and not a ref

Julien-Martin: Actually we used a counter to reset `isOverDropZone` because `dragleave` event seems to not have the same behaviour when you move horizontally and vertically :thinking: 

I just change to a normal variable :white_check_mark:

wheatjs: Interesting, anyway it seems good to me. Good work 🙂

wheatjs (approved):

packages/core/useDropZone/index.ts
Julien-Martin: Thank you :pray:

packages/core/useDropZone/index.md
```diff
@@ -0,0 +1,29 @@
...
+}
+
+const { isOverDropZone } = useDropZone(onDrop)
+
+<template>
+  <div ref="dropZoneRef">
+    Drop files here
+  </div>
+</template>
+</script>
```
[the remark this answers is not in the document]
Julien-Martin: Ups, thank for the fix :wink:

PR #1619, opened 2022-05-20
feat(useCssVar): support `initialValue`

jd-solanki:
### Description

We might want to set an initial value for CSS var based on some variable if not preset on it.

This will allow us writing safe CSS where value will be always present.

### Additional context

I haven't added this to watch because I think changing the initial value once it's initialized doesn't matter.

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useCssVar/index.ts
+++ b/packages/core/useCssVar/index.ts
@@ -17,19 +17,18 @@ import { unrefElement } from '../unrefElement'
 export function useCssVar(
   prop: MaybeRef<string>,
   target?: MaybeElementRef,
-  initialValue?: MaybeRef<string>,
+  initialValue?: string,
   { window = defaultWindow }: ConfigurableWindow = {},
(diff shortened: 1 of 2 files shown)


packages/core/useCssVar/index.ts
```diff
@@ -10,22 +10,27 @@ import { unrefElement } from '../unrefElement'
...
+  const _initialValue = unref(initialValue)
 
   watch(
     [elRef, () => unref(prop)],
     ([el, prop]) => {
-      if (el && window)
-        variable.value = window.getComputedStyle(el).getPropertyValue(prop)
+      if (el && window) {
+        const value = window.getComputedStyle(el).getPropertyValue(prop)
+        variable.value = value == '' ? _initialValue : value
```
antfu, line 32: Should we unref here?

PR #1622, opened 2022-05-21
feat(useScrollLock): adds new option `avoidShake`

0x-jerry:
### Description

When I use `useScrollLock`, there is a slight shake when switching between lock-on/lock-off.

Then, I think, maybe we can provide an option for users to avoid this shake.

And, here is an example of how it works:

https://user-images.githubusercontent.com/14226737/169635252-c06218cf-0a76-4da1-91c0-1f1b1210ac70.mp4

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useScrollLock/demo.vue
+++ b/packages/core/useScrollLock/demo.vue
@@ -9,7 +9,7 @@ useScroll(el)
 const isMobile = /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent)
 
 const isLocked = useScrollLock(el, false, {
-  avoidShake: !isMobile
+  placeholder: !isMobile
 })
 const toggleLock = useToggle(isLocked)
 </script>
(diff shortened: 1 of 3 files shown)


wheatjs:
I don't know if we should be modifying the padding of the container element. This might produce undesirable results for the user? Does this take into account existing padding on the element?

0x-jerry:
@wheatjs Yes, this may produce an undesirable result, that's why this option is not enabled by default. And thanks for pointing out we should take care of the existing padding. I have fixed it by using the `getComputedStyle` method to get the correct padding value.

wheatjs:
LGTM, only thing I think is maybe there could be a better name for it although I can't think of one myself 😅

0x-jerry:
How about `fixScrollbarPlaceholder`? or `tryFixScrollbarPlaceholder`? or just `scrollbarPlaceholder`?

antfu:
Thanks for working on it. This seems to be a good-to-have feature. However, the implementation feels a bit too hacky for me. I think this fall into CSS's domain, which I'd consider out-of-scope of VueUse.

PR #1626, opened 2022-05-22
feat(useInterval): Improve random selection from array

aurium:
### Description

Just a small improvement on the demo.

A solution using `Math.random()` and `length-1` will give the half probability of any other item to the first and the last items. [`Math.random()` will never return `1`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random), so don't wore to use `Math.floor()`.

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/shared/useIntervalFn/demo.vue
+++ b/packages/shared/useIntervalFn/demo.vue
@@ -7,7 +7,7 @@ const word = ref('Hello')
 const interval = ref(500)
 
 const { pause, resume, isActive } = useIntervalFn(() => {
-  word.value = greetings[Math.round(Math.random() * (greetings.length - 1))]
+  word.value = greetings[Math.floor(Math.random() * greetings.length)]
 }, interval)
 </script>
 


packages/shared/useIntervalFn/demo.vue
```diff
@@ -7,7 +7,7 @@ const word = ref('Hello')
 const interval = ref(500)
 
 const { pause, resume, isActive } = useIntervalFn(() => {
-  word.value = greetings[Math.round(Math.random() * (greetings.length - 1))]
+  word.value = greetings[Math.floor(Math.random() * greetings.length)]
```
okxiaoliang4, line 10: or maybe we can `import { rand } from '@vueuse/shared` first, and:
```suggestion
  word.value = greetings[rand(0, greetings.length - 1)]
```

aurium: Done!

PR #1627, opened 2022-05-22
feat(useColorMode): add option to return `auto` mode

dfreier:
### Description

Some applications need to know if the preferred color mode was used, e.g. when showing a dropdown menu with the three options `light`, `dark` or `system`. With the current implementation, there is no way to figure out if the user selected `auto` mode. That means it's not possible to highlight the `system` option in the dropdown as active.

I added an option `emitAuto` which defaults to `false`. When enabled, it includes the `auto` mode in the returned state variable.

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useColorMode/index.ts
+++ b/packages/core/useColorMode/index.ts
@@ -62,13 +62,14 @@ export interface UseColorModeOptions<T extends string = BasicColorSchema> extend
   storage?: StorageLike
 
   /**
-   * Omit `auto` mode from state
+   * Emit `auto` mode from state
    *
-   * When set to `true`, preferred mode is translated into `light` or `dark`
+   * When set to `true`, preferred mode won't be translated into `light` or `dark`.
+   * This is useful when the fact that `auto` mode was selected needs to be known.
    *
-   * @default true
+   * @default false
    */
-  omitAuto?: boolean
+  emitAuto?: boolean
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/core/useColorMode/index.ts
```diff
@@ -60,6 +60,15 @@ export interface UseColorModeOptions<T extends string = BasicColorSchema> extend
...
   storage?: StorageLike
+
+  /**
+   * Omit `auto` mode from state
+   *
+   * When set to `true`, preferred mode is translated into `light` or `dark`
+   *
+   * @default true
+   */
+  omitAuto?: boolean
```
antfu, line 71: Instead of "omit" and defaulting to true, I think it's better to have it as opt-in, like "emitAuto: false"

dfreier: Makes sense, feels more natural 👍 I pushed another commit to change it.

packages/core/useColorMode/index.ts
```diff
@@ -77,6 +87,7 @@ export function useColorMode<T extends string = BasicColorSchema>(options: UseCo
     storageKey = 'vueuse-color-scheme',
     listenToStorageChanges = true,
     storageRef,
+    emitAuto = false,
```
antfu, line 90:
```suggestion
    emitAuto,
```

We don't need to assign the default since it's already falsy

PR #1647, opened 2022-05-30
feat(useAsyncState): Add throw error option

### LINKED ISSUE #1629: useAsyncState's execute will not throw exceptions
### Describe the bug

I think sometimes I just use the execute of useAsyncState to call manually, at this time I prefer to handle the error myself from try catch instead of taking `error.value`, `execute` should keep the function of throwing error

### Reproduction

```js
import { useAsyncState } from '@vueuse/core'
const request = () => {
  throw Error('...')
}
const { state, execute } = useAsyncState(() => request(), [], { immediate: false })

try {
  await execute()
} catch (error) {
  // did not enter
}
```

### System Info

```Shell
none
```

### Used Package Manager

pnpm

### Validations

TuiMao233:
fix #1629

### DIFF
--- a/packages/core/useAsyncState/index.ts
+++ b/packages/core/useAsyncState/index.ts
@@ -34,6 +34,13 @@ export interface AsyncStateOptions {
    * @default true
    */
   resetOnExecute?: boolean
+  /**
+   *
+   * An error is thrown when executing the execute function
+   *
+   * @default false
+   */
+  throw?: boolean
 }
 
 /**
@@ -84,6 +91,7 @@ export function useAsyncState<T>(
     catch (e) {
       error.value = e
       onError(e)
+      throw error
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/core/useAsyncState/index.ts
```diff
@@ -34,6 +34,13 @@ export interface AsyncStateOptions {
    * @default true
    */
   resetOnExecute?: boolean
+  /**
+   *
+   * An error is thrown when executing the execute function
+   *
+   * @default false
+   */
+  throw?: boolean
```
okxiaoliang4, line 43:
```suggestion
  throwError?: boolean
```

packages/core/useAsyncState/index.ts
```diff
@@ -84,6 +91,7 @@ export function useAsyncState<T>(
     catch (e) {
       error.value = e
       onError(e)
+      throw error
```
okxiaoliang4, line 94:
```suggestion
      if (options.throwError) throw error
```

antfu:
Closing due to stale. Feel free to bring it back once you can work on it. Thanks.

PR #1657, opened 2022-06-02
fix(unrefElement): don't return the Vue instance when `$el` is `null`/`undefined`

DrJume:
When using `unrefElement` on a template ref of a Vue component which exposes a custom `$el` ([HeadlessUI](https://github.com/tailwindlabs/headlessui)]) that points to a not rendered inner ref (`v-if="false"`), `unrefElement` returns the Vue instance object with `{ $el: null }`. It should return `null` as it should unref the Vue instance ref.

The cause of this bug is how `plain` is checked to be a `VueInstance`. Rather than using the value of `$el`, it should check if it has a property named `$el`.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

---

_Update 1: added tests_

### DIFF
--- a/packages/core/unrefElement/index.test.ts
+++ b/packages/core/unrefElement/index.test.ts
@@ -11,7 +11,7 @@ describe('unrefElement', () => {
   it('works with regular html elements', () => {
     const refKey = 'target'
 
-    const vm = mount(defineComponent({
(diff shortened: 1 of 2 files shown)


DrJume:
added tests

PR #1665, opened 2022-06-05
feat(useHistoryState): new function

NoelDeMartin:
### Description

This PR introduces a new function called `useHistoryState` which binds state into the current page's [History.state](https://developer.mozilla.org/en-US/docs/Web/API/History/state). I just implemented it in a project and I thought it may be useful to others as well.

### Additional context

In my use-case, I'm using this to keep state when navigating back in an app. I think this is a similar use-case to keeping the scroll position, but using a generic interface. I'm not sure if there is a function that already does this in the library, but if there is I couldn't find it.

Also, I'm aware that the demo is not working at the moment. I think it has something to do with the way the scroll position is stored that overwrites the state written by this function, but I'm not entirely sure. I have tried the same code in a vanilla Vue app and it works, so it has to be something specific to VitePress.

I'm also aware that this PR is lacking tests. In both cases, I'd rather have some confirmation that there is interest in getting this merged before proceeding. I've seen in the contributing guidelines that it's suggested to open an issue first before starting to work on a new function, so that's essentially what this is. It's just with some real code and not just a "feature request" :).

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


NoelDeMartin:
Thanks for checking it out.

Seeing what you said I was curious to see how `useLocalStorage` works, because I thought it'd have the same limitation. But I was surprised to see that [there is actually some tracking going on](https://github.com/vueuse/vueuse/blob/main/packages/core/useStorage/index.ts#L156) :D. I didn't even know that this was possible, but after further inspection it seems like this `storage` event [is only triggered for changes in other pages](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event). So in most cases, it will have the same limitation, right? If I call `window.localStorage.setItem`, existing refs won't be updated.

Although to be fair, it's more likely that this is a problem for the history state because it changes during page navigations. This is not a problem in my use-case because I'm using it on a component that only shows up in a specific route, but it could be an issue for components that persist page navigations (like headers or footers).

I guess one way to improve it would be to listen for navigation events and update the value. But I won't work on this because it's not necessary for my use-case and it complicates the logic too much, I just wanted to share the code I already had.

So I'm closing the PR for the time being, anyone's welcome to pick up the code and improve upon it!

Issue #1168, opened 2022-01-23
reactive RTL mode.


AtefR:
Yo, 
I'm trying to use vueuse to switch direction from ltr to rtl, the only way I could think of now is using [@vueuse/head](https://github.com/vueuse/head) and injecting html tag attributes.

```
useHead({
  htmlAttrs: [
    {
      dir: "rtl",
      lang: "ar-sa",
    },
  ],
})
```
I can code custom vueuse function but I thought it'll be better to open an issue to discuss it first.

harmyderoman:
I think this is a very useful feature.

PR #1678, opened 2022-06-10
feat(useTextDirection): new function

iChengbo:
### Description

The dir [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes) is an enumerated attribute that indicates the directionality of the element's text.

Some languages are written from *right to left*, such as `Arabic` and `Hebrew`, So, when using i18n, it needs `useDir` to get the `ref(dir)` of the element.

reactive RTL mode. https://github.com/vueuse/vueuse/issues/1168

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


michealroberts:
I would perhaps rename this to `useRTLTextDirection()` or `useTextDirection()`.

iChengbo:
@antfu @michealroberts 
The `useDir` is named based on the name of the property,, but it is really a bit confusing. And compared to `useRTL` and `useRTLTextDirection`, I think `useTextDirection` is more appropriate, so I will rename it.

antfu (approved):

PR #1679, opened 2022-06-10
feat(useStepper): new function

innocenzi:
# useWizard

### Description

This is a utility function that helps building step-based wizard components. This is something that I found myself doing over and over in our apps (at least 3 times), and I think it's quite useful. 

The declaration of the steps is done via an array of ~~strings~~ steps of arbitrary types. The composable returns a few functions and computed variables that are useful for navigating through the steps.

For instance, the `goToPrevious` and `goToNext` function check if we can actually go back or forward. Computed variables like `isFirst` or `isLast` or `current` help determining what to display.  To display the component for the current step, one can use `currentStepIs('step name')`.

I'll happily review the wording of every function and variable, but they are all useful ones that I actually used in the real world.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


michealroberts:
@innocenzi I'd maybe rename this to "useStepper" or "useStepperWizard" ... what do you think?

innocenzi:
I think `useWizard` is more widely used and would be easier to find for someone looking for this composable, `useStepper` is also fine by me so I went ahead and renamed it

packages/core/useStepper/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    index.value++
+  }
+
+  function previous() {
+    if (isFirst.value)
+      return
+
+    index.value--
+  }
```
antfu, line 30: I somehow prefer these to be computed.

packages/core/useStepper/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    return currentStepIsAfter(step)
+  }
+
+  function active(step: T) {
+    return currentStepIs(step)
+  }
+
+  function todo(step: T) {
+    return currentStepIsBefore(step)
+  }
```
antfu, line 62: I not sure if I understand what's these three are for.

packages/core/useStepper/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    index.value--
+  }
+
+  function isNext(step: T) {
+    return steps.indexOf(step) === index.value + 1
+  }
+
+  function isPrevious(step: T) {
+    return steps.indexOf(step) === index.value - 1
+  }
```
antfu, line 38: If we provide `next` and `previous` as computed, I think we don't need this as ppl could just compare `step === next.value`

packages/core/useStepper/index.ts
antfu: Oh, I misunderstand them. Probably having both action and ref, like `next()` and `nextStep.value`, or `goNext()`?

antfu (commented):
I like this idea. But I would try to make it minimal as possible.

packages/core/useStepper/index.ts
innocenzi: These are aliases, which are useful in some contexts when using the literal function name (eg. `currentStepIsBefore`) makes things more confusing than they should (eg. `todo`).

I would understand if you'd rather not have them. They helped make the code clean when I needed to determine if a step was yet to be done, I could use `todo` instead of having to wrap my head around what was `currentStepIsBefore` doing. What do you think?

packages/core/useStepper/index.ts
innocenzi: Good idea!

packages/core/useStepper/index.ts
innocenzi: If think `isNext('Step name')` is more practical than `next.value === 'Step name'`. I like the function syntax better.

I would rather keep that function, but if you really want it gone I'll remove it

antfu: Sure. I am fine with this.

packages/core/useStepper/index.ts
```diff
@@ -0,0 +1,83 @@
...
+export function useStepper<T extends string>(steps: readonly T[], initial?: T) {
+  const index = ref(steps.indexOf(initial ?? steps[0]))
+  const current = computed(() => steps[index.value])
+  const isFirst = computed(() => index.value === 0)
+  const isLast = computed(() => index.value === steps.length - 1)
+
+  function backTo(step: T) {
+    if (currentStepIsAfter(step))
+      goTo(step)
+  }
```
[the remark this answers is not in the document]
antfu: It would feel a bit weird to me that:

```ts
const s = useStepper([1,2,3,4])
s.goTo(2)
// ...
s.backTo(3)
// and s stays on 2
```

In your case, the `disabled` already serve as a guard for it.

packages/core/useStepper/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    return steps.indexOf(step) === index.value
+  }
+
+  function currentStepIsAfter(step: T) {
+    return index.value > steps.indexOf(step)
+  }
+
+  function currentStepIsBefore(step: T) {
+    return index.value < steps.indexOf(step)
+  }
```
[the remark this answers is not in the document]
antfu: Maybe `isStepAfter` and `isStepBefore`?

packages/core/useStepper/index.ts
```diff
@@ -0,0 +1,83 @@
+import { computed, ref } from 'vue-demi'
+
+export function useStepper<T extends string>(steps: readonly T[], initial?: T) {
```
antfu, line 3:
```suggestion
export function useStepper<T>(steps: readonly T[], initial?: T) {
```

Can we have it arbitrary? I might want to use objects.

packages/core/useStepper/index.ts
innocenzi: Good point. I removed them. 👍

packages/core/useStepper/index.ts
innocenzi: Done! Though I'm not sure when I would use them, it's nice to have them.

packages/core/useStepper/index.ts
innocenzi: Yes I don't see a reason why it should only be strings. I use that mostly but you're right that objects can be useful. Also added basic tests for that.

packages/core/useStepper/index.ts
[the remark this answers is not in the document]
innocenzi: How about `isCurrentStepAfter` and `isCurrentStepBefore`? Same length and explicitness, but it does read better than what I did. 

I'd be fine with `isStepAfter` but it sounds like the function name is missing context to what it compares to.

packages/core/useStepper/index.ts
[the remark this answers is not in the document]
antfu: Yeah, what I mean is the guard should be done on userland when calling `goTo`. Since there is only one line of addition in `backTo`, the trade-off of saving one line but introducing an additional abstraction layer that users/readers need to understand is a bit less worthy to me. I'd suggest we remove it, and let users define it when needed.

packages/core/useStepper/index.ts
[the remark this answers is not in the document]
antfu: We may use JSDocs to explain it better, WDYT? I am ok to be explicit, but the longer names mean a larger bundle and more user code shipped, and the price is paid for every app using it (sorry for a bit nitpicking)

packages/core/useStepper/index.ts
```diff
@@ -0,0 +1,72 @@
+import { computed, ref } from 'vue-demi'
+
+export function useStepper<T>(steps: readonly T[], initial?: T) {
+  const index = ref(steps.indexOf(initial ?? steps[0]))
+  const current = computed(() => steps[index.value])
+  const isFirst = computed(() => index.value === 0)
+  const isLast = computed(() => index.value === steps.length - 1)
+  const nextStep = computed(() => isLast.value ? undefined : steps[index.value + 1])
+  const previousStep = computed(() => isFirst.value ? undefined : steps[index.value - 1])
```
antfu, line 9:
```suggestion
  const nextStep = computed<T | undefined>(() => steps[index.value + 1])
  const previousStep = computed<T | undefined>(() => steps[index.value - 1])
```

outbound access returns undefined already.

packages/core/useStepper/index.ts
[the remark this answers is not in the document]
innocenzi: That's a good compromise, I'll do that 👍 

Don't worry about nitpicking, that's what makes you a great developer

packages/core/useStepper/index.ts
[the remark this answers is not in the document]
innocenzi: I understand, I still think it's worth it since I used it in every wizard I made, but I'll remove it since it's not that bad to create another function for that

innocenzi:
I followed-up in https://github.com/vueuse/vueuse/pull/1754 with a better implementation. Closing this because the review is stale.

PR #1680, opened 2022-06-10
feat(useInternalClock): new function

michealroberts:
### Description

This is a simple utility function for displaying an internal ticking clock that ticks at a rate of one second per second.

### Additional context

I will add configurable options for tick rate, and also accept some starting date.

---

- [ ] Bug fix
- [X] New Feature
- [X] Documentation update
- [ ] Other


okxiaoliang4:
Actually we already have [useNow](https://vueuse.org/core/usenow/)

okxiaoliang4:
Thanks your contribute. I will close it.

PR #1687, opened 2022-06-13
fix(useThrottleFn): `trailing` option should be `false` by default

### LINKED ISSUE #1684: useThrottleFn 快速点击会触发两次
### Describe the bug

如果在设定时间内，快速点击多次，useThrottleFn 内的方法会触发两次，第一次在点击的时候触发，第二次触发是在 useThrottleFn 定时结束后触发。不知是原先就如此设计，还是因为 bug。如果设计如此是否有违 throttle 方法在短时间内只触发一次的原则。希望官方可以给个解答。

### Reproduction

https://github.com/vueuse/vueuse/blob/main/packages/shared/useThrottleFn/demo.vue

### System Info

```Shell
System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
    Memory: 7.20 GB / 15.92 GB
  Binaries:
    Node: 16.15.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.5.5 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (102.0.1245.33)
  npmPackages:
    @vueuse/core: ^5.3.0 => 5.3.0
    vue: ^3.2.33 => 3.2.33
```

### Used Package Manager

pnpm

### Validations

webfansplz:
### Description

Close #1684 

Follow the design principles of `throttle function`,I think we shouldn't  call fn again by default after the time is up .

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


septatrix:
I just saw this in the BREAKING CHANGES section of the release notes.

The correct behaviour should be to only call the function after expiration of the timeout if any throttled call occured since the last call which was forwarded to the wrapped function. This is also how underscore implements it IIRC.
In most cases (e.g. resize, scroll, input) you want to run the function one last time if a change occurred between the last time the wrapped function was called.

I am not sure what you refer to by "design principles of throttle function". The event will still only be called at most once per interval. And those which would like the other behaviour can use `trailing: false` though the cases where this is the correct behaviour are rare in comparison to the indented use cases like the ones given in the docs.

It remember there was a way to search all of github how some APIs were used which could be used to confirm that most people use the previous default though I am not sure how to do this (the normal github search is not really suited for this)

septatrix:
@antfu I see that v9 is released now. Could this be reconsidered for v10?

septatrix:
Or at least add a TIP/WARNING box like it is done for some other function about this behaviour

septatrix:
Also all other throttle related helpers `watchThrottled`, `refThrottled`, `throttleFilter` default to `trailing = true` affirming my statement about it being the expected and more reasonable default in most cases.

antfu:
Yeah I think it's reasonable to have it default to `true`, refer to loash https://lodash.com/docs#throttle

@septatrix You can send a PR to change it back and we will ship if in v10

septatrix:
> @septatrix You can send a PR to change it back and so we could ship it in v10

Sure thing!

septatrix:
Sorry I currently cannot say when I will have time to work on this

Issue #1685, opened 2022-06-12
Infinite scroll not working


francoism90:
### Describe the bug

I'm trying to use [infinite scroll](https://vueuse.org/core/useinfinitescroll/#useinfinitescroll=), however the provided examples don't seem to work.

```
<template>
  <div
    class="flex flex-col gap-2 p-4 w-300px h-300px m-auto overflow-y-scroll bg-gray-500/5 rounded"
    ref="el"
  >
    <div v-for="item in data" :key="item">
      {{ item }}
    </div>
  </div>
</template>

<script setup lang="ts">
import { ref } from "vue";
import { useInfiniteScroll } from "@vueuse/core";

const el = ref<HTMLElement | null>(null);

useInfiniteScroll(el, () => console.log("fetch"), { distance: 100 }); // nothing logged to console
</script>
```

### Reproduction

https://stackblitz.com/edit/vitejs-vite-srqqu3

### System Info

```Shell
System:
    OS: Linux 5.18 Arch Linux
    CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor
    Memory: 19.16 GB / 31.27 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.3.0 - /usr/bin/node
    Yarn: 1.22.19 - /usr/bin/yarn
    npm: 8.5.5 - /usr/bin/npm
  npmPackages:
    @vueuse/components: ^8.6.0 => 8.6.0 
    @vueuse/core: ^8.6.0 => 8.6.0 
    vue: ^3.2.36 => 3.2.36
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
[...]

PR #1701, opened 2022-06-17
docs(useInfiniteScroll): add usage note

luvejo:
### Description

It took me some time to realize that I needed to care about the content's height, as this is not how popular libraries like [vue-infinite-loading](https://www.npmjs.com/package/vue-infinite-loading) work.

Also, I'm not the first person to have this issue. See https://github.com/vueuse/vueuse/issues/1685#issuecomment-1153501021.

- [ ] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useInfiniteScroll/index.md
+++ b/packages/core/useInfiniteScroll/index.md
@@ -6,6 +6,8 @@ category: Sensors
 
 Infinite scrolling of the element. 
 
+> **Note**: You should ensure that the container element's height < content's height.
+
 ## Usage
 
 


webfansplz:
Can we just provide a `forceFire` option  to support it ?

Force triggered the `onLoadMore` callback once when the container element's height >= content's height.

webfansplz:
> Just wondering if other libs could support it, any blockers for us to support it automatically?

I didn't find a good way to support it  automatically  because that for the browser native behavior ,the scroll event’ll never triggered when the container element's height >= content's height  😅

The [vue-infinite-loading](https://github.com/PeachScript/vue-infinite-loading) library seems triggered callback once when init, See:

https://github.com/PeachScript/vue-infinite-loading/blob/4baed2bb078f076d3bff48c783ed324236630ed6/src/components/InfiniteLoading.vue#L160

PR #1706, opened 2022-06-19
feat(useBase64): accept objects

### LINKED ISSUE #1677: Allow useBase64 to take an object

I have a use case where I want to base64 an object (e.g. like on vue SFC playground). The current version of this composable doesn't seem to support the `object` type.

### Suggested solution

Allow the `object` type, then just `JSON.stringify(object)` and then handle it like you would a string.

### Alternative

_No response_

### Additional context

_No response_

### Validations

chaii3:
### Description

Allow to add objects in useBase64

fixes #1677 

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other


packages/core/useBase64/index.ts
```diff
@@ -39,8 +48,8 @@ export function useBase64(
     promise.value = new Promise<string>((resolve, reject) => {
       try {
         const _target = unref(target)
-        // undefined or null
-        if (_target === undefined || _target === null) { resolve('') }
+        // undefined
+        if (_target === undefined) { resolve('') }
```
antfu, line 52: What's the reason that null is being excluded here?

antfu (approved):

PR #1714, opened 2022-06-23
feat(useDialog): new function

innocenzi:
### Description

This is a new function similar to the existing `useConfirmDialog` but more generic. It supports passing properties when revealed and returning a response, both typed.

```ts
const { isRevealed, reveal, close, on, properties } = useDialog<DialogProperties, DialogResult>()

reveal(/** DialogProperties **/)
close(/** DialogResult **/)

on.reveal(({ data }) => {})
on.close(({ result }) => {})
```

While `useDialog` is useful, the interesting function is `makeDialogComposable`, that allows for making specific composable functions for a specific dialog.

### Description

This is a new function based on `useDialog`, that simply returns a composable function for that dialog. This is super useful when you have multiple, application-wide, typed dialogs:

```ts
export const useLoginDialog = makeDialogComposable<PrefilledLoginForm, LoginFormResult>()
export const useRegisterDialog = makeDialogComposable<RegisterDialogResult>()
export const use2faDialog = makeDialogComposable<ChallengeData, ChallengeResult>()
```

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


jd-solanki:
Can I bind reveal with `modelValue` to create a component using this composable?

jd-solanki:
Moreover, I was also looking for something like this. However, one major thing missing is body scroll lock.

Do you have any idea how one can achieve that? _(even without composable)_

innocenzi:
> However, one major thing missing is body scroll lock.

I think body scroll lock is out of scope? The composable just gives state and functions to help creating and managing dialogs. I suggest using [Headless UI](https://headlessui.dev/vue/dialog) for implementing the actual dialogs, it has scroll lock built-in 👍 

> Can I bind reveal with `modelValue` to create a component using this composable?

Can you show me with pseudo-code what you mean?

innocenzi:
Is there a way to get this in VueUse v10?

innocenzi:
I can explore that for the use case I was trying to solve — it's a bit of a different paradigm and it's lower-level, but it does seem like a potential solution 👍 

I'll mark this PR as draft in the meantime

PR #1730, opened 2022-07-01
fix(useIntersectionObserver): 🐞

codehzy:
useIntersectionObserver不适配ios12.1以下版本

### UseIntersectionObServer code does not exist in the browser in Window's browsers that will execute Noop in the browser in Window, hoping to throw an error on the user on the console. It is recommended that users install Resize-Observer to achieve Polyfill.So that the program is running normally.

---

- [ X ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useIntersectionObserver/index.ts
+++ b/packages/core/useIntersectionObserver/index.ts
@@ -75,7 +75,9 @@ export function useIntersectionObserver(
       },
       { immediate: true, flush: 'post' },
     )
-    : noop
+    : () => {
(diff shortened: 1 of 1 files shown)


okxiaoliang4:
I think you can use `isSupported` to solve this problem
```ts
const { isSupported } = useIntersectionObserver(xxx)

if (isSupported) {
  // do something
} else {
  // not supported
}

```

PR #1741, opened 2022-07-04
fix(useCssVar): remove spaces at both ends

### LINKED ISSUE #1738: `useCssVar` should trim the value on first read
### Describe the bug

According to the [CSSWG](https://drafts.csswg.org/css-syntax/#consume-declaration), when reading a CSS value the white spaces between the colon and the first character should be ignored.

But some browser doesn't follow this and returns the spaces at the beginning of the value. (I'm using Edge 103)

For example, this causes problem in case of an `input type="color"`. The color format **must** be `#RRGGBB` but on first load is ` #RRGGBB` so the input show a black default color.

I think we should trim the left side of the value.

```js
const value = window.getComputedStyle(el).getPropertyValue(prop).trimStart()
```

### Reproduction

https://stackblitz.com/edit/vitejs-vite-dhwvrq/?file=src%2FApp.vue,index.html

### System Info

```Shell
System:
    OS: macOS 12.5
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 543.20 MB / 16.00 GB
    Shell: 3.4.1 - /usr/local/bin/fish
  Binaries:
    Node: 16.15.1 - /usr/local/opt/node@16/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.11.0 - /usr/local/opt/node@16/bin/npm
  Browsers:
    Firefox: 92.0
    Firefox Developer Edition: 86.0
    Safari: 15.6
  npmPackages:
    @vueuse/core: ^8.7.5 => 8.7.5 
    vue: ^3.2.37 => 3.2.37
```

### Used Package Manager

yarn

### Validations

murongg:
fix #1738

### Description

When I define a css variable, `useCssVar` contains spaces in variable value. 
e.g.
In body: 
```css
body {
 --my-aaa:bold ;
 --my-bbb: bold;
}
```
```ts
const aaa = useCssVar('--my-aaa', document.body);
console.log(aaa.value) // `bold ` note: `bold` has a space after.

const bbb = useCssVar('--my-bbb', document.body);
console.log(bbb.value) // ` bold` note: `bold` has a space before.
```
So we maybe need remove spaces at both ends for ensure the correct value .

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useCssVar/index.ts
+++ b/packages/core/useCssVar/index.ts
@@ -30,7 +30,7 @@ export function useCssVar(
     [elRef, () => unref(prop)],
     ([el, prop]) => {
       if (el && window) {
-        const value = window.getComputedStyle(el).getPropertyValue(prop)
+        const value = window.getComputedStyle(el).getPropertyValue(prop)?.trim()
         variable.value = value || initialValue
       }
     },


packages/core/useCssVar/index.ts
```diff
@@ -30,7 +30,7 @@ export function useCssVar(
     [elRef, () => unref(prop)],
     ([el, prop]) => {
       if (el && window) {
-        const value = window.getComputedStyle(el).getPropertyValue(prop)
+        const value = window.getComputedStyle(el).getPropertyValue(prop).trimStart().trimEnd()
```
antfu, line 33:
```suggestion
        const value = window.getComputedStyle(el).getPropertyValue(prop).trim()
```

murongg: Thank you for your advice, done!

antfu (approved):

PR #1754, opened 2022-07-05
feat(useStepper): new function

innocenzi:
# useStepper

### Note

This is an improved version of https://github.com/vueuse/vueuse/pull/1679, with which I was not happy. I've continued building wizards (steppers) in my application, and this composable is the result of what feels to me is perfect and flexible for most use cases.

This new implementation has better types and support both array and objects as steps.

### Description

This new utility helps building step-based wizard user interfaces. This is something that I did repeatedly in one of our applications, and this is the actual composable I'm using.

The declaration of the steps is done via an array of strings, which serve as identifiers, or via an object, which keys serve as identifiers. The composable returns a set of computed values and functions that are very useful to navigate through the steps.

For instance, the `goToPrevious` and `goToNext` function move. Computed variables like `isFirst`, `isLast` or `current` help determining what to display. To display the component for the current step, one can use `isCurrent('step name')`.

## Example

The demo is a full example, and tests cover all functionalities, but here is what it looks like at a glance.

```ts
const stepper = useStepper({
  'user-information': {
    title: 'User information',
    isValid: () => form.firstName && form.lastName,
  },
  'billing-address': {
    title: 'Billing address',
    isValid: () => !!form.billingAddress,
  },
  'terms': {
    title: 'Terms',
    isValid: () => form.contractAccepted === true,
  },
  'payment': {
    title: 'Payment',
    isValid: () => ['credit-card', 'paypal'].includes(form.payment),
  },
})
```

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/useStepper/index.md
```diff
@@ -0,0 +1,54 @@
...
+  /** Checks whether the given step is the next step. */
+  isNext,
+  /** Checks whether the given step is the previous step. */
+  isPrevious,
+  /** Checks whether the given step is the current step. */
+  isCurrent,
+  /** Checks if the current step is before the given step. */
+  isBefore,
+  /** Checks if the current step is after the given step. */
+  isAfter,
```
antfu, line 48: Can we move the comments to TSDocs instead?

antfu: Oh you already did, then I would prefer we remove the comments here to have avoid duplications

packages/core/useStepper/index.ts
```diff
@@ -0,0 +1,139 @@
+import type { MaybeRef } from '@vueuse/shared'
+import type { ComputedRef, Ref } from 'vue-demi'
+import { computed, ref } from 'vue-demi'
+
+type Primitive = string | number
+
+interface UseStepperReturn<StepName, Steps, Step> {
```
antfu, line 7:
```suggestion
export interface UseStepperReturn<StepName, Steps, Step> {
```

packages/core/useStepper/index.ts
```diff
@@ -0,0 +1,139 @@
...
+  isPrevious: (step: StepName) => boolean
+  /** Checks whether the given step is the current step. */
+  isCurrent: (step: StepName) => boolean
+  /** Checks if the current step is before the given step. */
+  isBefore: (step: StepName) => boolean
+  /** Checks if the current step is after the given step. */
+  isAfter: (step: StepName) => boolean
+}
+
+export function useStepper<T extends Primitive>(steps: MaybeRef<T[]>, initialStep?: T): UseStepperReturn<T, T[], T>
```
antfu, line 48:
```suggestion
export function useStepper<T extends string | number>(steps: MaybeRef<T[]>, initialStep?: T): UseStepperReturn<T, T[], T>
```

If it's the only place using it I think it's better to use literals directly.

packages/core/useStepper/index.md
```diff
@@ -0,0 +1,37 @@
...
+  isPrevious,
+  isCurrent,
+  isBefore,
+  isAfter,
+} = useStepper([
+  'billing-address',
+  'terms',
+  'payment',
+])
+```
```
antfu, line 37: Should we also have a short example of object usage?

antfu (approved):

packages/core/useStepper/index.md
innocenzi: Added it 👍

antfu (approved):

innocenzi:
What do you think of disallowing mixed object keys instead, as a compromise? That might prove easier to type.

darrylnoakes:
Even if it proves easier to type (I doubt it), it is still a compromise, disallowing real use cases.

Any reasons the potential fix I suggested would not work?

innocenzi:
It might work, but I have no clue how to type that, it's more boilerplate and it removes one potential key from the object too.

PR #1755, opened 2022-07-05
fix(useMagicKeys): reset refs on target blur

### LINKED ISSUE #1350: useMagicKeys: Alt is left after switching windows using Alt + Tab
### Describe the bug

Open page in browser, then switch windows using Alt + Tab and then again Alt + Tab to switch back to the page,
the Alt key is still there, which should be disappear

This bug can be recreate on demo page:
https://vueuse.org/core/usemagickeys/
chrome-capture-2022-2-4 (1)

### Reproduction

https://vueuse.org/core/usemagickeys/

### System Info

```Shell
System:
    OS: Windows 10 10.0.19042
    CPU: (12) x64 Intel(R) Core(TM) i5-10500 CPU @ 3.10GHz
    Memory: 2.83 GB / 15.72 GB
  Binaries:
    Node: 16.13.2 - D:\software\nodejs_16.13.2\node.EXE
    npm: 8.1.2 - D:\software\nodejs_16.13.2\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (98.0.1108.62)
    Internet Explorer: 11.0.19041.1202
```

### Used Package Manager

pnpm

### Validations

lavolpecheprogramma:
### Description

On macos when some keys are pressed in combo(cmd+p for example) open a dialog and the keyup is not triggered and the refs keep a wrong state.
I add a blur event listener so when the target lost the focus, the key refs are resetted to false.

fixes #1350 

- [x ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useMagicKeys/index.ts
+++ b/packages/core/useMagicKeys/index.ts
@@ -135,6 +135,10 @@ export function useMagicKeys(options: UseMagicKeysOptions<boolean> = {}): any {
       updateRefs(e, false)
       return onEventFired(e)
     }, { passive })
+    useEventListener(target, 'blur', (e: KeyboardEvent) => {
+      updateRefs(e, false)
+      return onEventFired(e)
+    }, { passive })
   }
 
   const proxy = new Proxy(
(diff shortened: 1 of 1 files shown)


antfu:
https://github.com/vueuse/vueuse/pull/1506 already fixing it, I think using `blur` would sometime cause indented behaviour that might confuse developers.

PR #1756, opened 2022-07-05
feat(useTextareaAutosize): new function

innocenzi:
### Description

This function makes it easy to make a textarea automatically grow depending on its content. This is something not easily achievable when not knowing the underlying trick.

```vue
<script setup lang="ts">
const { textarea, input } = useTextarea()
</script>

<template>
  <textarea
    ref="textarea"
    v-model="input"
    class="resize-none"
    placeholder="What's on your mind?"
  />
</template>
```

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/.vitepress/theme/styles/demo.css
```diff
@@ -101,8 +101,8 @@
     min-width: 20rem;
     font-size: 1.05rem;
     padding: 0.5em 1em 0.4em 1em;
-    border: 1px solid var(--vt-c-divider);
     border-radius: 4px;
+		box-shadow: var(--vt-c-divider) 0px 0px 0px 1px;
```
antfu, line 105:
```suggestion
    box-shadow: var(--vt-c-divider) 0px 0px 0px 1px;
```

packages/core/useTextareaAutosize/demo.vue
```diff
@@ -0,0 +1,17 @@
+<script setup lang="ts">
+import { useTextarea } from '@vueuse/core'
+
+const { textarea, input } = useTextarea()
```
antfu, line 4:
```suggestion
const { textarea, input } = useTextareaAutoResize()
```

antfu:
Could we also have a directive version of it?

packages/core/useTextareaAutosize/demo.vue
innocenzi: Good catch, renamed it in a few other places I missed too

innocenzi:
> Could we also have a directive version of it?

That would be nice, but I didn't manage to make one, I have no experience with directives. I may try to PR one in the future though, if that's okay?

packages/core/useTextareaAutosize/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  watch([input, textarea], triggerResize, { immediate: true })
+
+  if (options?.watch)
+    watch(options.watch, triggerResize, { immediate: true, deep: true })
+
+  return {
+    textarea,
+    input,
```
antfu, line 37:
```suggestion
    input,
    triggerResize,
```

antfu (approved):

PR #1769, opened 2022-07-07
docs(useNetwork):Add Navigator.connection.type compatibility hint and format my pr

freezepluto:
### Description
Navigator.connection.type  is only supported in Chrome Os.

### Additional context
This is Compatibility Notes.
https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/type

---

- [ ] Bug fix
- [ ] New Feature
- [X] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useNetwork/index.md
+++ b/packages/core/useNetwork/index.md
@@ -4,7 +4,7 @@ category: Sensors
 
 # useNetwork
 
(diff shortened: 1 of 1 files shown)


antfu:
Thanks but I think the link we pointed to already explains it, which makes it out of the duty of VueUse to explain and keep the info up to date.

PR #1815, opened 2022-07-10
feat(usePow): new function

webfansplz:
### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


antfu (approved):

antfu:
But I am not sure if we should have it. I can think useRound or useMax have their use cases, but what would the use case be for this function?

webfansplz:
> But I am not sure if we should have it. I can think useRound or useMax have their use cases, but what would the use case be for this function?

I didn't had the use case too , just want to cover the basic math function. (I think it may used to related to the some graphical formulas)

antfu:
Let's not have it for now, since when ppl really need this they probably need more than this.

PR #1825, opened 2022-07-10
feat(useAbs): new function

LittleSound:
### Description

Reactively [Math.abs(value)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/abs).

### Usage

```ts
import { useAbs } from '@vueuse/math'

const value = ref(-23)
const absValue = useAbs(value)

// absValue.value: 23
```

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/math/useAbs/index.md
+++ b/packages/math/useAbs/index.md
@@ -1,0 +1,17 @@
+---
+category: '@Math'
+alias: abs
+---
+
+# useAbs
+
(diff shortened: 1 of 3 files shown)


packages/math/useAbs/index.md
```diff
@@ -0,0 +1,17 @@
+---
+category: '@Math'
+alias: abs
```
antfu, line 3:
```suggestion
```

Issue #1812, opened 2022-07-10
New `@vueuse/math` package


antfu:
We plan to ship a new `@vueuse/math` package in v9.0, where you can see it in the `next` branch already.

We planned to adding those new functions:

- [ ] `useMax`
- [ ] `useMin`
- [x] [`useRound`](https://github.com/vueuse/vueuse/pull/1820)
- [x] [`useFloor`](https://github.com/vueuse/vueuse/pull/1817)
- [x] [`useCeil`](https://github.com/vueuse/vueuse/pull/1818)
- [ ] `useSum`
- ...

Open for contributions

LittleSound:
For some of the single parameter `Composables` maybe we can support `setter` ? 
For example `useRound`:
```ts
const round = useRound(0.5)

watchSyncEffect(() => console.log(value)) // log: 1

round.value = 0.1 // log: 0

round.value = 2.5 // log: 3
```

antfu:
@LittleSound Yes we could. But how should we handle this case?:

```ts
const round = useRound(0.5)

watchSyncEffect(() => console.log(value)) // log: 1

round.value = 0.1 // log: 0

round.value = 2.5 // log: 3

LittleSound:
@antfu  😂 I made a typo, it should be `watchSyncEffect(() => console.log(round.value))`

antfu:
No problem with the typo. My question is, how should we handle the write back if the value is provided as a ref?

LittleSound:
> @LittleSound是的，我们可以。但是我们应该如何处理这种情况呢？：
> 
> ```ts
> const value = ref(0.5)
> const round = useRound(value)
> 
> watchSyncEffect(() => console.log(value)) // log: 1
> 
> round.value = 0.1 // log: 0 // value?
> 
> round.value = 2.5 // log: 3
> ```

It seems to me that "write-back" should probably be disabled, or needs to be turned on via options

zojize:
For associative binary operations (i.e. `Math.min, Math.max, +, *` etc.), I'm wondering if we could provide overloads for both array input and multiple arguments input?

e.g.
```ts
function min(arr: MaybeComputedRef<MaybeComputedRef<number>[]>): ComputedRef<number>
function min(...args: MaybeComputedRef<number>[]): ComputedRef<number>
function min(...args: [MaybeComputedRef<MaybeComputedRef<number>[]>] | MaybeComputedRef<number>[]): ComputedRef<number> {
[...]

PR #1831, opened 2022-07-10
feat(useSum): reactively adds the values together

LittleSound:
### Description
```ts
import { useSum } from '@vueuse/math'

const value1 = ref(10)
const value2 = ref(20)

const sum = useSum([value1, value2])
// sum.value: 30
```

```ts
import { useSum } from '@vueuse/math'

const list = reactive([10, 20])

const sum = useSum(list)
// sum.value: 30
```

#### issues #1812 

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/math/useSum/index.test.ts
+++ b/packages/math/useSum/index.test.ts
@@ -1,0 +1,7 @@
+import { useSum } from '.'
+
+describe('useSum', () => {
+  test('should be defined', () => {
+    expect(useSum).toBeDefined()
+  })
+})
--- a/packages/math/useSum/index.ts
+++ b/packages/math/useSum/index.ts
@@ -1,0 +1,3 @@
+export function useSum() {
+
+}


antfu:
Interesting, but I think this might be better to introduce a more general `useArrayReduce` in `@vueuse/share` first and reuse it.

LittleSound:
> Interesting, but I think this might be better to introduce a more general `useArrayReduce` in `@vueuse/share` first and reuse it.

I see that someone is already working on [`useArrayReduce`](https://github.com/vueuse/vueuse/pull/1874) and has a merged [`useSum`](https://github.com/vueuse/vueuse/pull/1837) method. Is there anything I need to do now?

antfu:
Thanks for the contribution. I think #1874 is based on your implementation.

PR #1838, opened 2022-07-10
feat(useTrunc): new function

hanpei:
### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/math/useTrunc/index.test.ts
```diff
@@ -0,0 +1,44 @@
+import { ref } from 'vue-demi'
+import { useTrunk } from '.'
+
+// 返回：
```
antfu, line 4:
```suggestion
// Returns:
```

packages/math/useTrunc/index.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { MaybeComputedRef } from '@vueuse/shared'
+import { resolveUnref } from '@vueuse/shared'
+
+/**
+ * Reactively Math.trunc(value).
+ *
+ * @see https://vueuse.org/useTrunc
+ * @param value
+ */
+export function useTrunk(value: MaybeComputedRef<number>): ComputedRef<number> {
```
antfu, line 12:
```suggestion
export function useTrunc(value: MaybeComputedRef<number>): ComputedRef<number> {
```

packages/math/useTrunc/index.ts
```diff
@@ -0,0 +1,19 @@
...
+  return computed<number>(() => {
+    if (resolveUnref(value) > 0)
+      return Math.floor(resolveUnref(value))
+    else
+      return Math.ceil(resolveUnref(value))
```
antfu, line 17: Why don't we use `Math.trunc` directly?

antfu (changes requested):

hanpei:
i've push the changes, what should i do next?

antfu (approved):

PR #1841, opened 2022-07-10
feat(useMath): new function

FrankFang:
### Description

```ts
import { useMath } from '@vueuse/math'

const value = ref(45.95)
const rounded = useMath('round', value)
// rounded.value is 46
const floored = useMath('floor', value)
// floored.value is 45
```

### Additional context

https://github.com/vueuse/vueuse/issues/1812

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other


zojize:
ideally, I'd love to use this function like this:

```ts
const {
  ceiled, // 43
  floored, // 42
  // better names?
  absed, // 42.42
  sqrted, // 6.513063795173513
  sign, // 1
} = useMath(42.42)
```

antfu:
For reference: https://github.com/vueuse/vue-chemistry#usage

FrankFang:
@antfu  你的意思是这样吗？

```ts
import {floor} from 'vuevue/math'

const {floor} = useMath()
const value1 = ref(1.1)
const result = floor(value1)
//result.value is 1
```

FrankFang:
https://github.com/vueuse/vueuse/pull/1844 is better. close pr.

FrankFang:
@zojize  It looks good, But it's a little strange that the type of result depends on the number of params:
```
const {floored} = useMath(1.1)
const {min} = useMath(1, 2)
```

zojize:
> @zojize It looks good, But it's a little strange that the props of result depends on the number of params:
> 
> ```
> const {floored} = useMath(1.1) // no powed
> const {powed} = useMath(2, 2) // no floored
> ```

IMO in case of multiple arguments, it should be returned as a function.

e.g.
```ts
const { pow } = useMath(3)
const squared = pow(2) // ref(9)
const cubed = pow(3) // ref(27)
```

PR #1858, opened 2022-07-12
feat(useFileDownload): new function

LadyChatterleyLover:
### Description

download file with ease.

### Additional context

---

- [ ] Bug fix
- [x ] New Feature
- [ ] Documentation update
- [ ] Other

- [√] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [√ ] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [√ ] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [√] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [√ ] Ideally, include relevant tests that fail without this PR but pass with it.


packages/core/useFileDownload/index.ts
```diff
@@ -0,0 +1,38 @@
+export interface DownloadOptions {
+  href: string | Blob
+  fileName: string
```
userquin, line 3: how about `content-disposition`header? 

there are situations where you don't know the filename, for example using an api to download some doc for some id...

Firefox will handle attachments in a diferent way chromium based browsers, I use this code when the response is an attachement:
```ts
      const contentDisposition = headers.get("content-disposition");
      // const inline = contentDisposition.indexOf("inline;") > -1;
      const matches = /filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/.exec(
        contentDisposition
      );
      const fileName = matches && matches[1].replace(/['"]/g, "");
      // FIX: force download the document: the browser can open it or not (ff does but chrome doesn't)
      const contentType = "application/octet-stream"; // headers.get("content-type");
      link.setAttribute("download", fileName);
      await nextTick();
      link.click();     
```

packages/core/useFileDownload/index.md
```diff
@@ -0,0 +1,28 @@
...
+# useFileDownload
+
+download file with ease.
+
+## Usage
+
+``` ts
+import { useFileDownload } from '@vueuse/core'
+
+const { download } = useFileDialog({
```
ntnyq, line 13:
```suggestion
const { download } = useFileDownload({
```

xuzuodong:
This PR provides exactly what I wanted, will there be any further progress on this PR? 👀

PR #1882, opened 2022-07-13
refactor(useSpeechSynthesis)!: remove `voiceInfo`, allow `voice` as ref

sibbng:
### Description

[useSpeechSynthesis ](https://vueuse.org/core/usespeechsynthesis/) demo is not playing the right voice when multiple voices are available for the same language, because we are passing language code instead of a `SpeechSynthesisVoice` object. It wouldn't work even if we pass it as `voice` option, because `voice` option is static. I changed it to `MaybeRef` and this required me to make `voiceInfo` a `ref`, and this was a breaking change. Then I thought it was pointless to return the same input that we received from users. So, I removed it.

### Additional context

---

- [x] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useSpeechSynthesis/demo.vue
+++ b/packages/core/useSpeechSynthesis/demo.vue
@@ -2,22 +2,22 @@
 import { ref } from 'vue'
 import { useSpeechSynthesis } from '@vueuse/core'
 
-const lang = ref('en-US')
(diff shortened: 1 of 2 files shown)


antfu:
Changed target to `next` branch, can you rebase it and resolve the conflicts? Thanks

PR #1911, opened 2022-07-14
feat(all): use MaybeComputedRef

gumingWu:
### Description

use MaybeComputedRef instead of MaybeRef

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other

### DIFF
--- a/packages/core/useTitle/index.ts
+++ b/packages/core/useTitle/index.ts
@@ -1,4 +1,4 @@
-import type { MaybeComputedRef, MaybeRef } from '@vueuse/shared'
+import type { MaybeComputedRef } from '@vueuse/shared'
 import { isFunction, isString, resolveRef } from '@vueuse/shared'
 import type { ComputedRef, Ref } from 'vue-demi'
 import { unref, watch } from 'vue-demi'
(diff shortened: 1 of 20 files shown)


packages/core/useTitle/index.ts
```diff
@@ -18,11 +18,11 @@ export interface UseTitleOptions extends ConfigurableDocument {
    *
    * @default '%s'
    */
-  titleTemplate?: MaybeRef<string> | ((title: string) => string)
+  titleTemplate?: MaybeComputedRef<string> | ((title: string) => string)
 }
 
 export function useTitle(
-  newTitle?: MaybeRef<string | null | undefined>,
+  newTitle?: MaybeComputedRef<string | null | undefined>,
```
antfu, line 25: This one is indented (it's an overload). `MaybeComputedRef` should only be used when we are sure we only read the value (in this case it's retuned as a mutable ref)

antfu:
Thank you for your contribution. But the task is not only about replacing all `MaybeRef` but we also need to make corresponding changes to the logic to support it. Closing this for now, but feel free to open another PR if can you take care of the logic.

PR #1918, opened 2022-07-14
fix(useStyleTag): fix hydration mismatch issue

sibbng:
### Description

The buttons in the [useStyleTag](https://vueuse.org/core/usestyletag/) demo are not working if this page is visited with full-reload.  There is a hydration mismatch error on console. I thought changing the value of `isLoaded` immediately on client side is might be the cause of the issue. I wrapped it with `nextTick`. The hydration mismatch error is still there but at least the buttons are working now.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
I think instead, we should wrap `load` with `tryOnMounted`

sibbng:
> I think instead, we should execute the immediate `load()` in `tryOnMounted`

It's worked 👍

PR #1925, opened 2022-07-15
docs(codeCopy): update doc

genffy:
### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [X] Documentation update
- [ ] Other

Add code copy function. Inspired and copy from `vitepress`.
The latest `vitepress` already have this function, but upgrade it may need more affords. 😅


genffy (commented):
nothing

antfu:
Thanks but we probably should adapt to new VitePress theme later directly.

PR #1929, opened 2022-07-15
fix: UseInfiniteScroll direction type

sand4rt:


sand4rt (commented):
fix: UseInfiniteScroll direction type

PR #1933, opened 2022-07-16
fix(useDropZone): improve argument type

aki77:
```vue
<script setup lang="ts">
import { useDropZone } from '@vueuse/core'

const dropZoneRef = ref<HTMLElement>()

function onDrop(files: File[] | null) {
  // called when files are dropped on zone
}

const { isOverDropZone } = useDropZone(dropZoneRef, onDrop)
</script>

<template>
  <div ref="dropZoneRef">
    Drop files here
  </div>
</template>
```

This code is a type error.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useDropZone/demo.vue
+++ b/packages/core/useDropZone/demo.vue
@@ -15,7 +15,7 @@ function onDrop(files: File[] | null) {
   }
 }
 
-const dropZoneRef = ref<HTMLElement | null>(null)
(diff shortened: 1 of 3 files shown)


antfu (approved):

aki77:
Thanks!

PR #1947, opened 2022-07-17
docs: remove navbar transparency

userquin:
### Description

imagen

imagen

- [ ] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other


antfu:
Is that FF because does not support backdrop blur?

userquin:
It doesn't work also on Chrome, I will check it again in a few minutes. We have without it also on Vitest and vite plugin pwa.

userquin:
in chrome some navbar entries don't look quite right with blur:

imagen

Issue #1939, opened 2022-07-17
CreateGlobalState: does not support Set() when using UseStorage


LoganTann:
### Describe the bug

Hello, 
I implemented in my project the example code for CreateGlobalState (https://vueuse.org/shared/createGlobalState/)

However, one of my objects stored is initialized as a `new Set()`. Using a set instead of an array is better in my use case.

When trying to mutate a set using the methods `.add()` or `.delete()`, the set is not edited.

When reloading my browser tab, I can no longer use the methods (will result into a `state.mySet.has` is not a function)

I suspect the UseStorage function not properly serializing and deserializing the set. We can see in the local storage that it is stingified as an empty object (`{}`)

### Reproduction

https://stackblitz.com/edit/vitejs-vite-b3qc2h?file=src/Item.vue

### System Info

```Shell
-- my real config

NuxtJS v3.0.0-rc.5 and Firefox 102.0.1

-- nxp envinfo
  System:
    OS: Windows 10 10.0.19043
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 4.61 GB / 15.74 GB
  Binaries:
    Node: 16.15.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.18 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.11.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (103.0.1264.62)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @vueuse/core: ^8.9.2 => 8.9.2
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
[...]

PR #1948, opened 2022-07-18
fix: support complex Object type for useStorge

azaleta:
### Description
fix for issue #1939 

### Additional context

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useStorage/index.test.ts
+++ b/packages/core/useStorage/index.test.ts
@@ -143,24 +143,58 @@ describe('useStorage', () => {
     const store = useStorage(KEY, {
       name: 'a',
       data: 123,
+      cDate: new Date('2000-01-02T00:00:00.000Z'),
+      cLogic: true,
+      cSet: new Set([1, 2, 3]),
+      cMap: new Map<number, string | number>([
+        [1, 'a'],
+        [2, 2],
+      ]),
     }, storage)
 
-    expect(storage.setItem).toBeCalledWith(KEY, '{"name":"a","data":123}')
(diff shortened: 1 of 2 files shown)


sibbng:
This way they won't be deserialised back to Set, Map etc. Methods like `.has()`, `.set()` won't be available after reload and it will throw error again.

azaleta:
@[sibbng](https://github.com/sibbng)
yes, this use case has some limitation right now.

suppose we init store as 
```ts
      a: new Set<number>(),
      b: new Map<string,string>()
```
after write to the store, they will all be the same value `[]`
in this case we could not know the type when we reload.

Now I convert all of them to Set.

sibbng:
@azaleta We could store metadata in another property like `__usestorage__metada__`. Or directly in properties itself, like:

`{ a: new Set([1]) }` will be `"{ a: '__usestorage:set__[1]' }"` and we will remove `__usestorage:set__` prefix before deserializing and use this information to make it Set.

Also, currently we are serializing only the first level of `object`. We should deeply serialize-deserialize properties. But I'm not sure this is worth the added complexity.

antfu:
I would suggest not doing this, since there is no perfect and straightforward way to solve it. If we can't correctly parse it, which could make the app fail silently, I would prefer to keep the current approach so the user aware they need to provide the custom serializer/deserializer them self and responsible for the caveats.

antfu:
- Nested
- Recursive
- Map/Set
- Custom classes
- Dynamic (different types between the initial and serialized)
- Functions or other non-serializable data
- ...

Data serializing/deserializing is a very complex topic and should be handled by a standalone lib with its caveat. That's why the options are provided for customization. Thanks but this is out-of-scope of VueUse.

PR #1957, opened 2022-07-19
feat(useStorage): `mergeDefaults` option

### LINKED ISSUE #1954: useStorage: difference between Object.keys(rawValue) and Object.keys(rawInit)
### Describe the bug

if an object exists in localStorage with given key:
```javascript
test: {
  a:1
}
```

oops:
```javascript
const testState = useLocalStorage('test', {
  a: 1,
  b: 2,
})
console.log(testState.value.b) // returns undefined
```

### Reproduction

dev

### System Info

```Shell
--npmPackages {
  "@vueuse/core": "^8.9.3"
}
```

### Used Package Manager

pnpm

### Validations

webfansplz:
### Description

Closes #1954 

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/useStorage/index.ts
```diff
@@ -75,6 +75,13 @@ export interface UseStorageOptions<T> extends ConfigurableEventFilter, Configura
    */
   writeDefaults?: boolean
 
+  /**
+   * Merge the default value to the storage
+   *
+   * @default false
+   */
+  mergeDefaults?: boolean
```
antfu, line 83:
```suggestion
  mergeDefaults?: boolean | ((storage: T, defaults: T) => T)
```

We should allow them to provide a custom function for deep merge. We should also note that we only do shallow merge when this is set to `true`

webfansplz: Nice suggest, I always learn something from your review .

packages/core/useStorage/index.ts
```diff
@@ -195,10 +197,11 @@ export function useStorage<T extends(string | number | boolean | object | null)>
         return rawInit
       }
       else if (!event && mergeDefaults) {
-        if (type === 'object') {
-          const value = serializer.read(rawValue)
+        const value = serializer.read(rawValue)
+        if (isFunction(mergeDefaults))
+          return mergeDefaults(storage!, value)
```
antfu, line 202:
```suggestion
          return mergeDefaults(value, rawInit)
```

antfu (dismissed):

antfu (approved):

PR #1962, opened 2022-07-19
fix(refDebounced): allow debouncing objects

DamianGlowala:
### Description
Currently refs holding an object are updated immediately. This PR deeply watches the original ref value and clones it to avoid copying by reference in case the ref value is an object.

### Additional context
TODO: choose, install and import the library for deep cloning the original ref value (here with an example of lodash)

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/shared/refDebounced/index.ts
+++ b/packages/shared/refDebounced/index.ts
@@ -12,13 +12,15 @@ export function refDebounced<T>(value: Ref<T>, ms = 200, options: DebounceFilter
   if (ms <= 0)
     return value
 
-  const debounced = ref(value.value as T) as Ref<T>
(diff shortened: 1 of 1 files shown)


antfu:
Thanks, but I think this is out-of-scope of this function. Deep clone comes with a lot of caveats and not everyone using this function wants this behavior. I'd suggest you have it directly in your codebase to solve your needs.

DamianGlowala:
I fully get you on the above, thank you. I'll make a modified version of it in my codebase. Do you think that recognising whether the ref's value is a primitive one or an object and then applying cloning and deep watcher conditionally is still not an idea that is worth considering?

If not, shall there be an option in the docs briefly mentioning the above behaviour to avoid any confusion in the future?

PR #1969, opened 2022-07-20
feat(useDark): added initial value.

### LINKED ISSUE #1967: Enable default color scheme value in `useDark`

By default, `useDark` uses `usePreferredDark` to set up the color scheme value by default. As a result, we cannot force  `light` as the default value.

### Suggested solution

This problem could be solved by implementing a `defaultValue` parameter in the `UseDarkOptions` payload, as follow:

```
import { useDark, useToggle } from '@vueuse/core'

const isDark = useDark({ defaultValue: "light" })
const toggleDark = useToggle(isDark)
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

chaii3:
closes #1967 
### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useDark/index.ts
+++ b/packages/core/useDark/index.ts
@@ -26,6 +26,8 @@ export interface UseDarkOptions extends Omit<UseColorModeOptions<BasicColorSchem
    * @default undefined
    */
   onChanged?: (isDark: boolean) => void
+
+  initialValue?: 'light' | 'dark'
 }
 
 /**
@@ -54,6 +56,9 @@ export function useDark(options: UseDarkOptions = {}) {
       light: valueLight,
     },
   })
+
+  if (options.initialValue)
+    mode.value = options.initialValue
 
   const preferredDark = usePreferredDark({ window })
 


packages/core/useDark/index.ts
```diff
@@ -55,6 +57,9 @@ export function useDark(options: UseDarkOptions = {}) {
     },
   })
 
+  if (options.initialValue)
+    mode.value = options.initialValue
```
antfu, line 61: If you are doing so, there will be no persistent at all

orionna319:
Add an initial value for useColorMode?

packages/core/useDark/index.ts
chaii3: Sorry I didn't understand

tripflex:
```javascript
import { useColorMode, useToggle, useDark } from '@vueuse/core'

useColorMode({initialValue: 'dark'})

export const isDark = useDark()

export const toggleDark = useToggle(isDark)
```

PR #1994, opened 2022-07-26
docs: add focus outline to checkboxes for accessibility

curtgrimes:
### Description

Previously, when tabbing through elements on the page, the checkboxes would not change in appearance on focus. [Common web accessibility guidance](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-visible.html) typically says that elements like checkboxes should have some keyboard focus indicator.

This also changes the focus indicator to match the primary color.

These changes only affect when navigating through the VueUse docs with a keyboard. These particular changes probably don't make sense to contribute upstream to the Vitepress project because this PR is focused on primarily styling the checkboxes that are implemented only in the VueUse docs.

### Before (checkboxes do not have focus outline)

https://user-images.githubusercontent.com/3038600/180933187-8427c285-3e51-450f-900c-cd01762665ce.mov

### After (checkboxes have focus outline, outline color matches theme)

https://user-images.githubusercontent.com/3038600/180933209-dd95b9d1-85af-41cd-8371-3e0f9e5d8e5c.mov

### After (light mode)

Screen Shot 2022-07-26 at 12 55 18 AM

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/packages/.vitepress/theme/components/FunctionsList.vue
+++ b/packages/.vitepress/theme/components/FunctionsList.vue
@@ -177,12 +177,13 @@ input {
 }
 
 .checkbox {
-  @apply inline-flex items-center my-auto cursor-pointer select-none;
+  @apply inline-flex items-center my-auto cursor-pointer select-none rounded px-2;
 }
 
 .checkbox input {
   appearance: none;
   padding: 0;
+  margin: 0;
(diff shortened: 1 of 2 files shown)


antfu:
Nice!

PR #2017, opened 2022-07-31
feat(useObjectTemplate): new function

### LINKED ISSUE #1859: new composable: `usePropClone`

When creating a dialog with form inside, we need to clone the passed form data to the dialog component and emit that data on submit and reset data on cancel

### Suggested solution

We can create a new function `usePropClone` which will clone the prop value.

```ts
const { data, reset } = usePropClone(toRef(props.formData))
```

### Alternative

Building components without this composable, writing reset and cloning in each dialog.

### Additional context

I am happy to make PR _(if this function should be allowed to be added in vueuse)_ but I would like to know should I clone the prop data using:
```js
JSON.stringify(JSON.parse(propToClone))
```

Thanks.

### Validations

chaii3:
### Description

closes #1859

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other


antfu:
I am not sure about this. Maybe a `useCloned` that watches the source ref and applies a clone when the source changes, with the capability to customize the clone function.

PR #2022, opened 2022-08-01
feat(useVModel): support `clone` option

edumudu:
### Description

Fixes side effect when using `useVModel` with `passive` flag and with an object as value.
Added a clone option to clone the obj before set to `ref`

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


antfu (changes requested):
It would be better to be a separate `clone` option that accepts `boolean` and custom functions. Refer to https://github.com/vueuse/vueuse/blob/98104e172cd23a6f212fe2985f8a8036176d80db/packages/core/useManualRefHistory/index.ts#L18-L24

edumudu:
done

packages/core/useVModel/index.test.ts
```diff
@@ -178,7 +178,7 @@ describe('useVModel', () => {
     expect(emitValue instanceof SomeClass).toBeTruthy()
   })
 
-  it('should be side effect free when using objects', async () => {
+  it('should clone object', async () => {
```
antfu, line 181: We might better to keep the original test and start as a new one

packages/shared/utils/index.ts
```diff
@@ -104,20 +101,3 @@ export function objectPick<O, T extends keyof O>(obj: O, keys: T[], omitUndefine
     return n
   }, {} as Pick<O, T>)
 }
-
-/**
- * Simple recursive deep clone
- *
- * @category Object
- */
-export const cloneDeep = (obj: any) => {
```
antfu: Please revert this change

edumudu: you wanna to use this as default function instead `JSON.parse(JSON.stringfy(...))`?

antfu:
Ah, nvm. it seems I was reviewing the wrong diff. This looks good to me in general but I need to hold it for a while long to make sure it behave consistent with #2045

antfu (approved):

antfu (dismissed):

antfu (approved):

PR #2023, opened 2022-08-01
feat(useColorMode): overloads for more returning options

Waleed-KH:
### Description

Added `emit` argument (for `useColorMode`) that can be used to control what should be returned when using the function with the following options:
- `mode` (default): return a `ref` to get or set the current mode (or when `emitAuto` is true you get the `auto` setting) (same as current behavior)
- `state`: return an object with props: `{ mode, preference, isDark }`

The `state` option is useful when you want to know the mode setting (`auto`) and at the same time have access to the current applied mode (`dark`/`light` ...).

### Additional context

I've updated the demo to use the `state` emit option and added a small example in the docs.

Ideally the `state` emit option would be the default, but we can't do that right-now for backward-compatibility, I also think the `emitAuto` option should be deprecated.

continuation for #2004

---

- [x] Bug fix
- [x] New Feature
- [x] Documentation update
- [x] Other


antfu:
This adds too much complexity and I don't understand the motivation behind it. Can you explain more why the current approach does not work for you and require such a change?

Waleed-KH:
Currently, there are only 2 ways to use the `useColorMode` function:
- `emitAuto: false` (default): then you will get the current applied mode
- `emitAuto: true`: then you will get the `auto` setting, but then you lost the ability to know the current applied mode (`dark`/`light`)

So what if you want to have access to both (the `auto` setting and the current applied mode)? and like I mentioned before

> The `state` option is useful when you want to know the mode setting (`auto`) and at the same time have access to the current applied mode (`dark`/`light` ...).

Waleed-KH:
Actually, because of this missing use case, I wasn't able to use the function and had to write my own mini version that look like this:
```ts
import { computed, watch } from 'vue';
import { useStorage, usePreferredDark } from "@vueuse/core";

type Modes = 'dark' | 'light' | 'auto';

const preferredDark = usePreferredDark();
const preferredMode = computed(() => preferredDark.value ? 'dark' : 'light');
const setting = useStorage<Modes>('theme', 'auto');
const currentMode = computed(() => setting.value === 'auto' ? preferredMode.value : setting.value);
const isDark = computed(() => currentMode.value === 'dark');

function applyMode(dark: boolean) {
	if (dark) {
		document.documentElement.classList.add('dark');
	} else {
		document.documentElement.classList.remove('dark');
	}
}

watch(isDark, applyMode, { immediate: true });

export { setting, currentMode, isDark };
```

if this PR got merged, it will look like this
```ts
import { computed } from 'vue';
import { useColorMode } from "@vueuse/core";

const { setting, currentMode } = useColorMode({}, 'state');
const isDark = computed(() => currentMode.value === 'dark');

export { setting, currentMode, isDark };
```

now I can import this file from anywhere and easily predict the current applied mode (usually I use isDark to apply some custom class/attributes) or have full control over the mode setting (usually I use `setting` for a dropdown in the header to switch the mode)

antfu:
I am still not sure about the current design. I guess maybe we could use an approach similar to [@nuxt/color-mode](https://github.com/nuxt-modules/color-mode).

```ts
const mode = useColorMode({ emitAuto: true })

console.log(mode.value) // 'auto'
console.log(mode.preference) // 'dark'
```

Waleed-KH:
I'm working on a better approach and will open new PR soon

Waleed-KH:
I've updated the PR to match the approach you mentioned @antfu :
```ts
const { mode, preference } = useColorMode({}, 'state') // in v.10 just useColorMode()
console.log(mode.value) //'auto'
console.log(preference.value) //'light'
```
The method call is still different to avoid breaking changes, maybe in v.10 just `useColorMode()` by default, if you have a better idea for it, please let me know.

I've also set the `initialValue` to `MaybeRef` to achieve the same goal:
```ts
const mode = ref('auto')
const preference = useColorMode({ initialValue: mode })
console.log(mode.value) //'auto'
console.log(preference.value) //'light'
```

antfu:
Sorry for the long delay. Now implemented in https://github.com/vueuse/vueuse/commit/d150ca221b88887e86c4d0283f8c0ce71b5c3e50, you don't need to pass any addition args, `system` (preference) and `store` (mode) are directly available.

PR #2028, opened 2022-08-01
feat(useAsyncFunction): new function

### LINKED ISSUE #1992: useAsyncFunction

I would love a hook, that makes getting the pending state of an async function easier.

Currently I find myself writing this kind of boilerplate a lot:

```ts
import { logout } from '@/store/auth'

const logoutPending = ref(false)

async function handleLogout() {
  logoutPending.value = true
  await logout()
  logoutPending.value = false
}
```

Therefore I created a use hook for that and would like to submit a PR to implement this hook for VueUse. It can be used as follows:

```ts
import { logout } from '@/store/auth'
import { useAsyncFunction } from '@/utils'

const { pending: logoutPending, trigger: handleLogout } = useAsyncFunction(logout)
```

## Simple Implementation

In the core module, we could provide this implementation:

```ts
import { Ref } from 'vue'

export function useAsyncFunction<TArgs extends any[], TReturn extends any>(
  fn: (...args: TArgs) => Promise<TReturn>
): {
  pending: Ref<boolean>
  trigger: (...args: TArgs) => Promise<TReturn | undefined>
} {
  const pending = ref(false)

  async function trigger(...args: TArgs) {
    if (pending.value) {
      return
    }
    pending.value = true
    const result = await fn(...args)
    pending.value = false
    return result
  }

  return {
    pending,
    trigger,
  }
}

```

## Possible features to enhance this hook 

- `state` or `data` ref that contains the latest result
- `abort` method to cancel the running async call(s)
- allow `trigger` to be called multiple times and handle `pending` for multiple calls
- add `abortPendingOnTrigger` option to automatically abort pending calls when triggering a new one
- add `pendingCount` option to see how many calls are currently pending
- add `accumulateResults` option to make `state` be an array of results that were returned

### Alternative

Just figured out, that `useAsyncState` can achieve something similar:

```ts
const { execute, isLoading } = useAsyncState(() => logout(), null, {
  immediate: false,
})
```

It might be enough to add some examples to the docs of `useAsyncState` that show these use cases. However we could still provide a thin wrapper around `useAsyncState` that makes the usage more intuitive for the use case explained above.

### Additional context

_No response_

### Validations

chaii3:
### Description

closes #1992

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/useAsyncFunction/index.ts
```diff
@@ -0,0 +1,65 @@
+import type { ComputedRef, Ref } from 'vue-demi'
```
songjianet, line 1: import type is best placed at the bottom of all references.

songjianet (changes requested):
Can be used as a reference.

Issue #1563, opened 2022-05-04
useUserMedia should allow constraints property


MihaiVoinea:
### Clear and concise description of the problem

I'm using useUserMedia to get access to the user's camera, but the problem is that I can't use the constraints to set the desired resolution or other constraints.

I tried calling useDevicesList before useUserMedia, but even if I'm using the deviceIds returned by useDevicesList with the desired constraints, those specific constraints are not applied for useUserMedia. In particular, I can't get the camera to be in a higher resolution using VueUse.

### Suggested solution

I propose that useUserMedia allows an extra optional property called 'contraints'.

Then, in the `_start`, the user-provided constraints should be applied when calling `navigator.mediaDevices.getUserMedia`.

### Alternative

The way I'm solving the issue currently is that I'm not using VueUse's `useUserMedia`, but instead using the native `navigator.mediaDevices.getUserMedia` directly.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #2032, opened 2022-08-02
feat(useUserMedia): add constraints - remove deviceIds

mango-martin:
### Description

UseUserMedia should be able to accept a `constraints` object (also #1563)

### Additional context

This is a breaking change. At first I tried to implement this in a non-breaking way. But when trying to do it I realized that I'm trying to fit something that actually shouldn't be so difficult. The previous implementation only takes in a `deviceId` for both video and audio, but they are both properties of the constraints options. So it actually should really be only the constraints that need to be passed to the composable.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

Since it's a breaking change I don't expect this to get merged soon. But if you give me some direction to what you would like this to look like, I'd be willing to make the necessary changes. We need constraints for our project and while we can easily implement our own composable for that, I don't see why the "official" helper function shouldn't support that as well.

Edit: To be fair this implementation is so close to vanilla that it might not even be worth it to use a separate function for that. But I'll let you decide.


antfu:
We could provide the fallback compact for those options and mark them as deprecated, so we could ship it first without waiting for the next major. Then we could remove the deprecated options later.

mango-martin:
@antfu 
I've updated to allow both old behavior and passing `constraints` at the same time - without marking as deprecated. It's possible to have both co-existing at the same time. The problem now is that it's an either/or situation. Constraints will always be applied first. I think the best approach would be to handle a constraints ref internally in the composable and update it if a new deviceId was provided. But I'd like to confirm if that's desired or not - or in other words, which method should be prioritized? Do constraints trump individual ids or the other way around?

mango-martin:
Updated.

I tried to merge deviceIds and constraints but it gets a bit messy when you have not provided any of those on initial call of the composable.

So I opted for a simpler approach to just always use constraints if they exist and if not it will check for deviceIds with same logic as before.

packages/core/useVolumeReader/index.ts
```diff
@@ -0,0 +1,76 @@
+import { ref } from 'vue-demi'
+import { tryOnBeforeUnmount } from '@vueuse/core'
+
+export interface UseVolumeReaderOptions {
+  track: MediaStreamTrack | undefined
+  context?: AudioContext
+  interval?: number
+}
+
+export function useVolumeReader(options: UseVolumeReaderOptions) {
```
antfu, line 10: This file seems belong to another PR

antfu (changes requested):

antfu (approved):

mango-martin:
@antfu 
Admittedly, to work with this change is a bit awkward, since you'd have to keep track of a deeply nested reactive value if all you want to do is change the device by changing the id. It's been a while since I worked on this. If you have ideas on how to improve the design, let me know and I'll have another look.

PR #2044, opened 2022-08-04
fix(useDevicePixelRatio): more efficient mechanism

Waleed-KH:
### Description

I implemented a more efficient mechanism for the `useDevicePixelRatio` function that inspired by [this example](https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio#monitoring_screen_resolution_or_zoom_level_changes)

### Additional context

Currently the `useDevicePixelRatio` function create 12*2 watchers/eventListeners (+1 window resize) to achieve the same goal, this mechanism use only 1 event listener to check if the current `resolution` changed (once) then update the value and recursively check the new value

---

- [x] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useDevicePixelRatio/index.ts
+++ b/packages/core/useDevicePixelRatio/index.ts
@@ -20,14 +20,25 @@ export function useDevicePixelRatio({
   const pixelRatio = ref(1)
 
   let mqResolution: MediaQueryList
-  tryOnScopeDispose((function MqResolutionObserve(): Fn {
+  const cleanups: Fn[] = []
+
+  const cleanup = () => {
+    cleanups.map(i => i())
+    cleanups.length = 0
+  }
+
+  const observe = () => {
     pixelRatio.value = window.devicePixelRatio
+    cleanup()
     mqResolution = window.matchMedia(`(resolution: ${pixelRatio.value}dppx)`)
-    mqResolution.addEventListener('change', MqResolutionObserve, { once: true })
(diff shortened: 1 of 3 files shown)


wheatjs (dismissed):

packages/core/useDevicePixelRatio/index.ts
```diff
@@ -34,21 +17,17 @@ export function useDevicePixelRatio({
...
-    const mqlMax = useMediaQuery(`screen and (max-resolution: ${dppx}dppx)`)
-
-    watch([mqlMin, mqlMax], handleDevicePixelRatio)
-  })
+    mqResolution = window.matchMedia(`(resolution: ${pixelRatio.value}dppx)`)
+    mqResolution.addEventListener('change', MqResolutionObserve, { once: true })
+    return () => {
+      mqResolution.removeEventListener('change', MqResolutionObserve)
+    }
+  })())
```
antfu, line 30: What's the reason for wraping the function like this?

Waleed-KH: Just a short way to immediately invoke the function and passing `removeEventListener` to `tryOnScopeDispose`

antfu (approved):

Waleed-KH:
Hi @antfu 
can you please explain the changes you did to the code , like what is the need for an array of `cleanups`?

PR #2045, opened 2022-08-04
feat(useCloned): new function

### LINKED ISSUE #1859: new composable: `usePropClone`

chaii3:
closes #1859 

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/useCloned/index.ts
```diff
@@ -0,0 +1,43 @@
...
+export function useCloned<T extends Record<any, any> = Record<any, any>>(source: Partial<T>): { cloned: Ref<T>; sync: () => void; stop: WatchStopHandle }
+export function useCloned<T extends Record<any, any> = Record<any, any>>(source: MaybeRef<Partial<T>>, options: UseClonedOptions = {}) {
+  const cloned = ref<T>({} as T)
+
+  const { manual, cloneFunction } = options
+
+  let stopWatcher: undefined | WatchStopHandle
+
+  if (!manual && isRef(source))
+    stopWatcher = watch(source, sync, { immediate: true, deep: true })
```
antfu, line 27: We should make the watch options configurable.

packages/core/useCloned/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface UseClonedOptions<T extends Record<any, any> = Record<any, any>> {
+  /**
+   * sync source only by function
+   */
+  manual?: boolean
+  /**
+   * Custom clone function should return new value for cloned data
+   */
+  cloneFunction?: (source: Partial<T>, cloned: Partial<T>) => Partial<T> | T
```
antfu, line 13:
```suggestion
  cloneFunction?: (source: T) => | T
```

And we could default it to `JSON.parse(JSON.stringify(source))`

packages/core/useCloned/index.ts
chaii3: > We should make the watch options configurable.

Should we set immediate: true and deep: true by defaults ? I think we should

packages/core/useCloned/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  function defaultCloning() {
+    return JSON.parse(JSON.stringify(unref(source)))
+  }
+
+  function sync() {
+    cloned.value = cloneFunction ? cloneFunction(unref(source), cloned.value) : defaultCloning()
+  }
+
+  return { cloned, sync, ...(stopWatcher && { stop: stopWatcher }) } as any
```
antfu, line 45:
```suggestion
  return { cloned, sync }
```

We don't expose the stop function, ppl could use `effectScope` to capture it if needed.

packages/core/useCloned/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  const cloned = ref<T>({} as T)
+
+  const { manual, cloneFunction, watchOptions = { immediate: true, deep: true } } = options
+
+  if (!manual && isRef(source))
+    watch(source, sync, watchOptions)
+  else
+    sync()
+
+  function defaultCloning() {
```
vaakian, line 32: actually there's a native deep clone api called: [structuredClone](https://caniuse.com/?search=structuredClone), and supported quite well on different browsers. Use it if exists on `window`.

chaii3: > actually there's a native deep clone api called: [structuredClone](https://caniuse.com/?search=structuredClone), and supported quite well on different browsers. Use it if exists on `window`.

Hm, I like it, I'll add this one. Thank you!

packages/core/useCloned/index.md
```diff
@@ -0,0 +1,60 @@
+---
+category: Utilities
+---
+
+# useCloned
+
+Reactive partial clone.
```
antfu, line 7:
```suggestion
Reactive clone of a ref.
```

packages/core/useCloned/index.md
```diff
@@ -0,0 +1,60 @@
...
+## Custom clone function usage
+```ts
+import { useCloned } from '@vueuse/core/useCloned'
+
+const data = ref({
+  key: 'value'
+})
+
+const { cloned, sync } = useCloned(data, {
+  cloneFunction: (source, cloned) => ({ ...source, isCloned: true })
```
antfu, line 53: We should show a case of using `deepclone` or something instead of show the example for adding extra properties (which kinda violate the idea of `cloneing`)

packages/core/useCloned/index.ts
antfu: `structuredClone` is not available in all browsers. We should let users pass it explicitly instead of using the current fallback, which will lead to different behavior on different browsers implicitly.

packages/core/useCloned/index.ts
```diff
@@ -0,0 +1,43 @@
+import type { MaybeRef } from '@vueuse/shared'
+import type { WatchOptions } from 'vue-demi'
+import { isRef, ref, unref, watch } from 'vue-demi'
+
+export interface UseClonedOptions<T extends Record<any, any> = Record<any, any>> {
```
antfu, line 5:
```suggestion
export interface UseClonedOptions<T> extends WatchOptions {
```

We'd better directly support watch options instead of nesting, to align with other functions

packages/core/useCloned/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  cloneFunction?: (source: T, cloned: T) => T
+  /**
+   * Options for watcher
+   *
+   * @default { immediate: true, deep: true }
+   */
+  watchOptions?: WatchOptions
+}
+
+export function useCloned<T extends Record<any, any> = Record<any, any>>(source: MaybeRef<T>, options: UseClonedOptions = {}) {
```
antfu, line 22:
```suggestion
export function useCloned<T>(source: MaybeComputedRef<T>, options: UseClonedOptions = {}) {
```

antfu (changes requested):

packages/core/useCloned/index.md
```diff
@@ -0,0 +1,42 @@
...
+category: Utilities
+---
+
+# useCloned
+
+Reactive clone of a ref.
+
+## Usage
+```ts
+import { useCloned } from '@vueuse/core/useCloned'
```
antfu, line 11:
```suggestion
import { useCloned } from '@vueuse/core'
```

packages/core/useCloned/index.ts
chaii3: > `structuredClone` is not available in all browsers. We should let users pass it explicitly instead of using the current fallback, which will lead to different behavior on different browsers implicitly.

removed `structuredClone`

PR #2046, opened 2022-08-05
feat(useOnce): new function

icooooler:
### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


songjianet (approved):
LGTM

packages/core/useOnce/index.ts
```diff
@@ -0,0 +1,24 @@
+import { ref } from 'vue-demi'
+
+export const useOnce = (cb: Function) => {
+  const canUse = ref(true)
+
+  // 执行
+  const action = () => {
```
sxzz, line 7: I think it should be a verb. How about `execute`?

packages/core/useOnce/index.ts
```diff
@@ -0,0 +1,24 @@
+import { ref } from 'vue-demi'
+
+export const useOnce = (cb: Function) => {
+  const canUse = ref(true)
+
+  // 执行
```
sxzz, line 6: We should use English in an international project.

packages/core/useOnce/index.ts
```diff
@@ -0,0 +1,24 @@
+import { ref } from 'vue-demi'
+
+export const useOnce = (cb: Function) => {
+  const canUse = ref(true)
```
sxzz, line 4: `executed`

packages/core/useOnce/index.md
```diff
@@ -0,0 +1,30 @@
+---
+category: Utilities
+---
+
+# useOnce
+
+make a function can only be used once.
```
sxzz, line 7:
```suggestion
A function that can only be executed once.
```

sxzz (changes requested):

antfu:
Thanks but I don't think this is related to VueUse - you can achieve this without using Vue's reactivity. This is considered as out-of-scope.

PR #2049, opened 2022-08-05
fix(useTitle): prevent `observe` and `titleTemplate` been specified at the same time

huynl-96:
### Description

When setting `observe` to `true`, it only works well with default `titleTemplate` (which returns the exact input title e.g. `'%s'`,  `(title) => title`).

If changing `titleTemplate` into any other string template. e.g.

```js
const title = useTitle('Vite', { observe: true, titleTemplate: '%s - %s' });
```

When we try to input the first character, `title` ref changes which then leads to the change of `document.title`. But since `document.title` and `title` are not the same due to `titleTemplate`, it again reassigns the new value from `document.title` to `title` ref.

The above process keeps repeating and freezes the entire page.

https://user-images.githubusercontent.com/15758406/183067426-3901c4b4-9c8b-4df9-baba-7c16820ce8af.mov

### Additional context

---

- Prevent using `observe` with `titleTemplate`
- Improve types for `useTitle`
- Replace all matches instead of only the first match for `titleTemplate` string
- Add unit tests

- [x] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other


antfu:
Maybe instead we should throw an error to now allow using `observe` with `titleTemplate`

huynl-96:
@antfu I've updated the implementation to throw error. Please kindly check it when you're available.
Thank you

antfu:
In VueUse we don't throw errors as the message would increase the bundle size and impossible to do i18n. Changed to do it on type level, and force observe disabled when occurred at the same time.

antfu (dismissed):

packages/core/useTitle/index.md
```diff
@@ -40,9 +41,23 @@ const title = computed(() => {
...
+```js
+/* ✅ Will work */
+const title = useTitle('New Title', { observe: true })
+
+/* ✅ Will work */
+const title = useTitle('New Title', { titleTemplate: '%s - %s' })
+
+/* ❌ Will throw an error */
+const title = useTitle('New Title', { observe: true, titleTemplate: '%s - %s' })
+```
```
antfu, line 63:
```suggestion
```

antfu (approved):

PR #2050, opened 2022-08-05
docs(onKeyStroke): allow typing into search on demo page

curtgrimes:
### Description

Previously, on the onKeyStroke demo page, the user could not type into
the search box because of a `preventDefault()` that was present in the
demo.

After this fix, the search box can be typed into on the onKeyStroke
demo page.

fix #1951

https://user-images.githubusercontent.com/3038600/183084586-d5a3cfa5-4129-447d-a037-65fa1a06eba0.mov

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/onKeyStroke/demo.vue
+++ b/packages/core/onKeyStroke/demo.vue
@@ -7,22 +7,18 @@ const translateY = ref(0)
 
 onKeyStroke(['w', 'W', 'ArrowUp'], (e: KeyboardEvent) => {
   translateY.value -= 10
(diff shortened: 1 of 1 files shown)


antfu:
Or maybe just remove the preventDefault would do?

curtgrimes:
Thanks, that is a good point -- it does make the purpose of the demo less clear. I updated it to only remove the preventDefault() now. Thank you!

PR #2051, opened 2022-08-06
fix(useScroll): account for rounding errors when calculating arrivedState

### LINKED ISSUE #2047: useScroll - right and bottom arrived not working in Desktop mode
### Describe the bug

Tested on your demo, and during my app development.

When the DIV uses standard desktop scrollbars, the two flags for `right` and `bottom` `arrived` do not work properly.
image

If I switch to "Responsive" in Chrome's developer tool, it now works as intended.
image

Note how the coordinates are slightly different between the two, probably due to the scrollbar thickness?

To reproduce:
- Open Chrome, your demo website for useScroll
- Go to the bottom-right, and notice how the `arrived` flags are still false
- Open the devtools, enable the "device" mode (mobile/responsive)
- Now the flag works properly, (and the scrollbars are thin)

### Reproduction

https://vueuse.org/core/usescroll/#demo

### System Info

```Shell
System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
    Memory: 21.96 GB / 31.95 GB
  Binaries:
    Node: 16.14.2 - C:\Programming\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Programming\nodejs\yarn.CMD
    npm: 8.5.0 - C:\Programming\nodejs\npm.CMD
  Browsers:
    Chrome: 103.0.5060.134
    Edge: Spartan (44.19041.1266.0), Chromium (103.0.1264.77)
  npmPackages:
    @vueuse/core: 9.0.0 => 9.0.0
    vue: 3.2.37 => 3.2.37
```

### Used Package Manager

yarn

### Validations

curtgrimes:
### Description

Sometimes when zooming the page to sizes other than 100% and scrolling a
container all the way to the bottom right, the calculation for
`arrivedState.right` and `arrivedState.bottom` would incorrectly say
`false`. This is due to a [rounding error][] because `scrollTop` is a
non-rounded number, while `scrollHeight` and `clientHeight` are rounded.

Using the MDN documentation on `Element.scrollHeight` as guidance,
this commit factors in a 1px margin of error to more accurately
calculate `arrivedState.right` and `arrivedState.bottom`.

[rounding error]: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#determine_if_an_element_has_been_totally_scrolled

fix #2047

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useScroll/index.ts
+++ b/packages/core/useScroll/index.ts
@@ -51,6 +51,14 @@ export interface UseScrollOptions {
 }
 
 /**
+ * We have to check if the scroll amount is close enough to some threshold in order to
+ * more accurately calculate arrivedState. This is because scrollTop/scrollLeft are non-rounded
+ * numbers, while scrollHeight/scrollWidth and clientHeight/clientWidth are rounded.
+ * https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#determine_if_an_element_has_been_totally_scrolled
+ */
+const ARRIVED_STATE_THRESHOLD_PIXELS = 1
+
+/**
  * Reactive scroll.
  *
  * @see https://vueuse.org/useScroll
@@ -114,7 +122,7 @@ export function useScroll(
     directions.right = scrollLeft > x.value
     arrivedState.left = scrollLeft <= 0 + (offset.left || 0)
     arrivedState.right
-          = scrollLeft + eventTarget.clientWidth >= eventTarget.scrollWidth - (offset.right || 0)
+      = scrollLeft + eventTarget.clientWidth >= eventTarget.scrollWidth - (offset.right || 0) - ARRIVED_STATE_THRESHOLD_PIXELS
     x.value = scrollLeft
 
     let scrollTop = eventTarget.scrollTop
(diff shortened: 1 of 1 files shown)


packages/core/useScroll/index.ts
```diff
@@ -114,7 +114,7 @@ export function useScroll(
     directions.right = scrollLeft > x.value
     arrivedState.left = scrollLeft <= 0 + (offset.left || 0)
     arrivedState.right
-          = scrollLeft + eventTarget.clientWidth >= eventTarget.scrollWidth - (offset.right || 0)
+      = scrollLeft + eventTarget.clientWidth >= eventTarget.scrollWidth - (offset.right || 0) - 1
```
antfu, line 117: Can we make the magic `1` as an internal const and reference to the MDN docs?

curtgrimes: Absolutely, good idea -- I pushed https://github.com/vueuse/vueuse/commit/638ceebcc01d531b1e72aa228ee40e66d05e247c. Thanks!

packages/core/useScroll/index.ts
```diff
@@ -109,12 +109,18 @@ export function useScroll(
       e.target === document ? (e.target as Document).documentElement : e.target
     ) as HTMLElement
 
+    // We have to check if the scroll amount is close enough to some threshold in order to
+    // more accurately calculate arrivedState. This is because scrollTop/scrollLeft are non-rounded
+    // numbers, while scrollHeight/scrollWidth and clientHeight/clientWidth are rounded.
+    // https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#determine_if_an_element_has_been_totally_scrolled
+    const arrivedStateThresholdPixels = 1
```
antfu, line 116: A bit nitpicking but can we move it to top-level with UPPER_CASE? Thanks.

curtgrimes: How does this look? Thanks! https://github.com/vueuse/vueuse/commit/a0d28a784659647a19c68eaedb6d0241d5cad4f9

antfu (approved):

PR #2056, opened 2022-08-08
feat(watchPromise): new function

### LINKED ISSUE #1501: feat(watchPromise): watch callback with async

sth. liks this:
```js
watch(xxx, async () => {
  await awaitFn()
})
```
if the callback toggles twice and the first one takes longer to execute than the second one, it will overwrite the sceond callback's result.

### Suggested solution

how to solve it?
it can?
```typescript

function watchPromise(...args: Parameters<typeof watch>) {
  const [source, cb, options] = args
  let promiseLink : Promise<unknown> = Promise.resolve()
  watch(source, (..._args) => {
    promiseLink = promiseLink.then(() => cb(..._args))
  }, options)
}

watchPromise(xxx, async () => {
  await awaitFn()
})
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

chaii3:
### Description

closes #1501

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other


packages/core/watchPromise/index.md
```diff
@@ -0,0 +1,25 @@
+---
+category: Reactivity
+---
+
+# watchPromise
+
+Watch with saving async functions call order.
+
+## Usage
```
wheatjs, line 9: Can you give a more clear example of how this is supposed to work? It is not very clear from this example. `asyncFn` isn't ever used, `cleanQueue` isn't ever given an example of being used. What is current vs source?

chaii3: > Can you give a more clear example of how this is supposed to work? It is not very clear from this example. `asyncFn` isn't ever used, `cleanQueue` isn't ever given an example of being used. What is current vs source?

thank you for you review. My docs was really awful. Please check it. Is that better now ? I thought about making a small demo but I found it useless. Do you have any idea for a good demo for this function ?

PR #2076, opened 2022-08-10
fix(useScriptTag): remove tag only if be not used

### LINKED ISSUE #2027: `useScriptTag` child component `unload` removes parent component tag when same `src`.
### Describe the bug

`load` get the existing element
https://github.com/vueuse/vueuse/blob/8b21ca203adcc7e3bbc6e31e49d07803cb0b19b7/packages/core/useScriptTag/index.ts#L101

`unload` deletes the element
https://github.com/vueuse/vueuse/blob/8b21ca203adcc7e3bbc6e31e49d07803cb0b19b7/packages/core/useScriptTag/index.ts#L174-L176

If there is two component using the same script, the child component will remove the parent script.
if the component is self referencing, on child component unmounted, the parent script will be removed.

It can be addressed by managing a list of usages, the last unmounted removes the tag.

### Reproduction

as above.

### System Info

```Shell
System:
    OS: Linux 5.15 Ubuntu 22.04 LTS 22.04 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Xeon(R) CPU @ 2.20GHz
    Memory: 1.58 GB / 3.83 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.7.0 - ~/.nvm/versions/node/v18.7.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.7.0/bin/yarn
    npm: 8.15.1 - ~/.nvm/versions/node/v18.7.0/bin/npm
  Browsers:
    Firefox: 103.0
  npmPackages:
    @vueuse/components: ^9.0.2 => 9.0.2
    @vueuse/core: ^9.0.2 => 9.0.2
    @vueuse/motion: ^2.0.0-beta.12 => 2.0.0-beta.12
    @vueuse/nuxt: ^9.0.2 => 9.0.2
```

### Used Package Manager

yarn

### Validations

chaii3:
### Description

closes #2027

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useScriptTag/index.ts
+++ b/packages/core/useScriptTag/index.ts
@@ -46,6 +46,8 @@ export interface UseScriptTagOptions extends ConfigurableDocument {
   attrs?: Record<string, string>
 }
 
+const store = new Map<string, number>()
+
 /**
  * Async script tag loading.
  *
@@ -98,14 +100,18 @@ export function useScriptTag(
     // Local variable defining if the <script> tag should be appended or not.
     let shouldAppend = false
 
-    let el = document.querySelector<HTMLScriptElement>(`script[src="${resolveUnref(src)}"]`)
+    const srcResolved = resolveUnref(src)
+
+    let el = document.querySelector<HTMLScriptElement>(`script[src="${srcResolved}"]`)
+
+    store.set(resolveUnref(srcResolved), (store.get(srcResolved) || 0) + 1)
 
     // Script tag not found, preparing the element for appending
     if (!el) {
       el = document.createElement('script')
       el.type = type
       el.async = async
-      el.src = resolveUnref(src)
+      el.src = srcResolved
 
       // Optional attributes
       if (defer)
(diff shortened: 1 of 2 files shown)


songjianet (approved):
LGTM

packages/core/useScriptTag/index.ts
```diff
@@ -46,6 +46,8 @@ export interface UseScriptTagOptions extends ConfigurableDocument {
   attrs?: Record<string, string>
 }
 
+const store = new Map<string, number>()
```
antfu, line 49: I am not sure about this internal global map. This will break tree-shaking and make the fix not applicable if the user has multiple versions of VueUse installed.

packages/core/useScriptTag/index.ts
```diff
@@ -171,9 +177,20 @@ export function useScriptTag(
     if (scriptTag.value)
       scriptTag.value = null
 
-    const el = document.querySelector<HTMLScriptElement>(`script[src="${resolveUnref(src)}"]`)
-    if (el)
-      document.head.removeChild(el)
```
antfu, line 176: Instead of querySelector, maybe we could keep the reference of `el` from creation.

packages/core/useScriptTag/index.ts
chaii3: To be honest I didn't think about it. Do you have any ideas about it ?

chaii3: > I am not sure about this internal global map. This will break tree-shaking and make the fix not applicable if the user has multiple versions of VueUse installed.

To be honest I didn't think about it. Do you have any ideas about it ?

antfu:
Closing this for now as I think the current implementation (using global map) obeys the design of VueUse. Feel free to open new PR if you find a better approach. Thanks

PR #2082, opened 2022-08-12
docs: added export size in FunctionInfo

stephenjason89:
### Description
Added export size information inside FunctionInfo

- [ ] Bug fix
- [ ] New Feature
- [✓] Documentation update
- [ ] Other


packages/.vitepress/theme/components/FunctionInfo.vue
```diff
@@ -1,15 +1,19 @@
...
 
 const props = defineProps<{ fn: string }>()
-
+const exportSizeMD = ref('')
 const info = computed(() => functions.find(i => i.name === props.fn))
 const format = (ts: number) => ago(-1, 'day')
 const lastUpdated = useTimeAgo(new Date(info.value?.lastUpdated || 0))
 const link = computed(() => `/functions\#category=${encodeURIComponent(info.value.category)}`)
 
+axios.get('/export-size.md').then(({ data }) => exportSizeMD.value = data)
```
wheatjs, line 14: We probably don't want to be fetching this data on each page. Ideally it would work like the other function metadata and be fetched during build time. Maybe instead we can include function size inside of `@vueuse/metadata`

stephenjason89: @wheatjs Thank you for pointing that out, I placed it in metadata.ts and used fs.readFileSync.

Would you happen to know why the document is not rendering?
I can see that console logging it in metadata shows me the correct data that I am getting.

stephenjason89: I guess it's because code is being called in server and on client. There's no fs on client. 
How do i put a if(process.client) ?
process here doesn't have server or client prop

stephenjason89: I wanted to create a json file together with the export-size.md
but everytime i run pnpm size

image

I get this error.

antfu (changes requested):
1. We should update `scripts/export-size.ts` to generate a JSON format report, instead of reading the markdown file.
2. This shouldn't be part of `@vueuse/metadata` but could be a doc-only json.

packages/.vitepress/theme/components/FunctionInfo.vue
antfu: You should run build first and then analyse the size

stephenjason89:
> 1. We should update `scripts/export-size.ts` to generate a JSON format report, instead of reading the markdown file.
> 2. This shouldn't be part of `@vueuse/metadata` but could be a doc-only json.

Hello @antfu, let me know if I still need to change something.

antfu:
Looks good. Thanks.

PR #2093, opened 2022-08-16
fix(useActiveElement): use `computedWithControl` instead of `counter` ref

vaakian:
### Description

instead of creating a pointless `counter` ref to do the `trigger` and `track`, \
this change is to reuse `computedWithControl` to trigger the `document.activeElement` update.

**no any functionality change, just a better approach.**

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/packages/core/useActiveElement/index.ts
+++ b/packages/core/useActiveElement/index.ts
@@ -1,4 +1,4 @@
-import { computed, ref } from 'vue-demi'
+import { computedWithControl } from '@vueuse/shared'
 import { useEventListener } from '../useEventListener'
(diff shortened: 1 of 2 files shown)


antfu (approved):

adrianjost:
This broke the `useActiveElement` implementation for Vue 2 applications, because the [`trigger` of `computedWithControl` is only available in Vue3 environments](https://github.com/vueuse/vueuse/blob/main/packages/shared/computedWithControl/index.md#manual-triggering). 😢

PR #2104, opened 2022-08-19
Update index.md

sykes10:
Add information about the auto removal of the listener and how to manually remove it yourself

### Description
I think that composables that are adding listeners needs a bit of info about when and how they are being unregistered.

---

- [ ] Bug fix
- [ ] New Feature
- [x ] Documentation update
- [ ] Other


antfu:
Thanks but tbh I feel this is redundant. Almost all VueUse functions behave like this.

sykes10:
That's a fair point. Thanks for the review

PR #2107, opened 2022-08-20
feat(useRouteQuery, useRouteHash): add `preserveHash`/`preserveQuery` option

jeongtae:
### Description

In the current version, updating the query will lose the hash and vice versa.

```ts
const query = useRouteQuery('q')
const hash = useRouteHash()
query = 'foo' // http://some.url/?q=foo
hash.value = 'bar' // http://some.url/#bar  *query is removed*
```

This PR enables the following approach:

```ts
const hashPreservesQuery = useRouteHash(undefined, { preserveQuery: true })
query.value = 'foo' // http://some.url/?q=foo
hashPreservesQuery.value = 'bar' // http://some.url/?q=foo#bar
```

### Additional context

vue-router removes a query(or hash) unless explicitly specified. The maintainer said it is intentional behavior. (https://github.com/vuejs/vue-router/issues/2868)
But since these hooks abstract vue-router push/replace, I thought it should be able to preserve a query/hash via options.

I'm not good at English. Any comments about English grammar on JSDocs are welcome.

Thanks for providing nice hooks.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


mercmobily:
Is this not going to be merged?

jeongtae:
@mercmobily 

This was addressed in PR #2350  via a different method.

And it was deployed in v9.4.0, so if you're experiencing issues like the one in the text, I recommend upgrading your library version.

mercmobily:
Thank you and sorry for the noise

PR #2108, opened 2022-08-20
feat(useMagicKeys): Add keyboard sequences support

### LINKED ISSUE #427: Feature request: useMagicKeys() sequence support
Sequence support like that of Gmail would be really awesome (such as `g` + `n`, or `g` + `p`). That would mean not having to hold down `g` while pressing `n` (which is already supported), but simply hitting those keys one by one in the correct order.

lookasc:
### Description

Currently `useMagicKeys` supports only keyboard shortcuts (`A+S`, `Meta+S`...) - to trigger something keys need to be pressed at the same time. This PR changes add support for keyboard sequences (`A then S`...) by extending current functionality.

PR fixes #427

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useMagicKeys/index.md
+++ b/packages/core/useMagicKeys/index.md
@@ -55,6 +55,24 @@ watch(Ctrl_A_B, (v) => {
 })
 ```
 
+### Sequences
+
+You can magically use sequences of keys by connecting keys with `>`.
+
+```ts
+import { useMagicKeys } from '@vueuse/core'
+
+const keys = useMagicKeys()
+const sequence = keys['A>B>C']
+
+watch(sequence, (v) => {
+  if (v)
+    console.log('Sequence A then B then C has been pressed')
+})
+```
+
+### `whenever`
+
 You can also use `whenever` function to make it shorter
 
 ```ts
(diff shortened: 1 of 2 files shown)


antfu:
I am not very sure about this as this adds quite some weight to the function - I would suggest to implement it on the user land, or as a higher level composable

hartmut-co-uk:
How about adding this feature as a separate composable `useMagicSequence` or `useMagicKeySequence`?

lookasc:
> How about adding this feature as a separate composable `useMagicSequence` or `useMagicKeySequence`?

@hartmut-co-uk In my opinion it is not ideal - at least for me 🤷 
What about adding `enableHistory` option which may enable/disable sequences feature based on user requirements? Disabled feature will not trigger unnecessary computations and timeouts. What do you think @antfu?

hartmut-co-uk:
> > How about adding this feature as a separate composable `useMagicSequence` or `useMagicKeySequence`?
> 
> @hartmut-co-uk In my opinion it is not ideal - at least for me 🤷 What about adding `enableHistory` option which may enable/disable sequences feature based on user requirements? Disabled feature will not trigger unnecessary computations and timeouts. What do you think @antfu?

The composable should (could) be able to detect (`enableHistory`) by itself, based on if it's a sequence or not? 🤔 

Btw: I really like the sequence feature + API + implementation! 👍 👍

PR #2111, opened 2022-08-21
feat(useMark): new function

Jclong98:
### Description

New `@vueuse/integrations` function for [`mark.js`](https://markjs.io/)

### Additional context

I was unsure how to get the type for `Mark` so I made a small mock for the parts that were used in [`useMark`](https://github.com/Jclong98/vueuse/blob/main/packages/integrations/useMark/index.ts#L17). Is this acceptable practice?

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/integrations/useMark/index.ts
```diff
@@ -0,0 +1,81 @@
+import { computed, ref } from 'vue-demi'
+import type { Ref } from 'vue-demi'
```
songjianet, line 2: import type is best placed at the bottom of all references.

songjianet (changes requested):
Can be used as a reference.

packages/integrations/useMark/index.ts
Jclong98: I've moved the type imports under the rest of the imports, is that what you meant?

antfu:
Thank you. Unfortunately we no longer accept new functions. Feel free to publish a standalone package on your own.

PR #2120, opened 2022-08-24
feat(usePictureInPicture): new function

michealroberts:
### Description

Reactive handler for the [Picture-in-Picture API](https://developer.mozilla.org/en-US/docs/Web/API/Picture-in-Picture_API).

The Picture-in-Picture API allow websites to create a floating video window always on top of other windows so that users may continue consuming media while they interact with other content sites, or applications on their device.

### Additional context

---

https://www.loom.com/share/91d5dc7a6b974fd79c7f912777bae1ca

- [ ] Bug fix
- [X] New Feature
- [X] Documentation update
- [ ] Other


webfansplz:
https://github.com/vueuse/vueuse/blob/23d186e2da529b1b81f741b54ebee2aa1aaa56bd/packages/core/useMediaControls/index.ts#L453

Thanks for the PR, but we already have it in [useMediaControls](https://github.com/vueuse/vueuse/blob/main/packages/core/useMediaControls/index.ts) .

michealroberts:
@webfansplz There's no room for a silmmer version? That media controls one is fairly hefty ...

packages/core/usePictureInPicture/index.ts
```diff
@@ -0,0 +1,43 @@
...
+import type { MaybeElementRef } from '../unrefElement'
+
+import { useEventListener } from '../useEventListener'
+
+export function usePictureInPicture(
+  el: MaybeElementRef<HTMLVideoElement | undefined>,
+) {
+  const video = ref(el)
+
+  const isSupported = computed(() => {
```
webfansplz, line 12: use `useSupported` instead and maybe we should add document && .. for friendly SSR support

packages/core/usePictureInPicture/index.md
```diff
@@ -0,0 +1,32 @@
+---
+category: Browser
+---
+
+# usePictureInPicture
+
+Reactive handler for the [Picture-in-Picture API](https://developer.mozilla.org/en-US/docs/Web/API/Picture-in-Picture_API).
```
webfansplz, line 7: To ensure consistency of Documentation (as with other functions):

Reactive  [Picture-in-Picture API](https://developer.mozilla.org/en-US/docs/Web/API/Picture-in-Picture_API).

packages/core/usePictureInPicture/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    return 'pictureInPictureEnabled' in document && !video.value?.disablePictureInPicture
+  })
+
+  const pictureInPictureWindow = ref<PictureInPictureWindow | null>(null)
+
+  const isInPictureInPictureMode = computed(() => {
+    return !!pictureInPictureWindow.value
+  })
+
+  useEventListener(video, 'leavepictureinpicture', () => pictureInPictureWindow.value = null)
```
webfansplz, line 22: Seems we need add `enterpictureinpicture` event

packages/core/usePictureInPicture/index.ts
```diff
@@ -0,0 +1,43 @@
+import { computed, ref } from 'vue-demi'
+
+import type { MaybeElementRef } from '../unrefElement'
+
+import { useEventListener } from '../useEventListener'
+
+export function usePictureInPicture(
+  el: MaybeElementRef<HTMLVideoElement | undefined>,
+) {
+  const video = ref(el)
```
webfansplz, line 10: Why do we need to add it instead of using `el` with `resolveUnref` directly

packages/core/usePictureInPicture/index.ts
michealroberts: Done ✅

packages/core/usePictureInPicture/index.md
michealroberts: Done ✅

packages/core/usePictureInPicture/index.ts
michealroberts: What would be the need? We're using the `computed` value to detect if we're in Picture in Picture mode ...

packages/core/usePictureInPicture/index.ts
michealroberts: So we could use unrefElement() ... ?

michealroberts: The only issue I have found is the unrefElement doesn't provide reactivity, and as such the `video.requestPictureInPicture()` call doesn't work....

packages/core/usePictureInPicture/index.ts
webfansplz: We should capture it when users toggle it without `togglePictureInPicture` function

PR #2136, opened 2022-08-28
feat(useContextMenu): new function

vaakian:
### Description

add [`contextMenu(s)`](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event)  to the vue app with ease. 

image

### Additional context

----

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


vaakian:
oh, maybe I should provide a renderless component as well. 🤔

vaakian:
> oh, maybe I should provide a renderless component as well. 🤔

done! 🌙

packages/core/useContextMenu/component.ts
```diff
@@ -0,0 +1,29 @@
+import { defineComponent, h, reactive, ref } from 'vue-demi'
+import type { RenderableComponent, UseContextMenuOptions } from '@vueuse/core'
```
songjianet, line 2: import type is best placed at the bottom of all references.

packages/core/useContextMenu/component.ts
```diff
@@ -0,0 +1,29 @@
...
+import { useContextMenu } from '@vueuse/core'
+export interface useContextMenuProps extends UseContextMenuOptions, RenderableComponent {
+
+}
+
+export const UseContextMenu = defineComponent<useContextMenuProps>({
```
songjianet, line 8: Can useContextMenuProps be written directly as UseContextMenuOptions | RenderableComponent.

packages/core/useContextMenu/index.ts
```diff
@@ -0,0 +1,139 @@
+import type { MaybeComputedRef } from '@vueuse/shared'
```
songjianet, line 1: import type is best placed at the bottom of all references.

packages/core/useContextMenu/index.ts
```diff
@@ -0,0 +1,139 @@
...
+    accessMenuElementIfExists((el) => {
+      el.style.left = `${position.value.x}px`
+      el.style.top = `${position.value.y}px`
+    })
+  })
+
+  return {
+    visible,
+    position,
+
```
songjianet, line 135: Please remove useless newlines.

packages/core/useMouseInElement/index.ts
```diff
@@ -1,5 +1,6 @@
+import { resolveRef } from '@vueuse/shared'
 import { ref, watch } from 'vue-demi'
-import type { MaybeElementRef } from '../unrefElement'
+import type { MaybeComputedElementRef } from '../unrefElement'
```
songjianet, line 3: import type is best placed at the bottom of all references.

songjianet (changes requested):
Can be used as a reference.

packages/core/useContextMenu/index.ts
vaakian: Actually the newline is for separating the `state` and `method` since there might be a lot more 
 property to return. \
see [core/useManualRefHistory/index.ts#L201](https://github.com/vueuse/vueuse/blob/08cbd6d9f4b72954eca1865ef1f136f61daba711/packages/core/useManualRefHistory/index.ts#L201)

but it's okay to remove it as well, thanks for the suggestions!

vaakian:
## TODO
Fix the edge case: **If the menu is clicking on the edge of the viewport**.

- [ ] case 1: click on the `bottom edge`, put it on the `left` of the cursor.
- [ ] case 2: click on the `right edge`, put it on the `top` of the cursor.
- [ ] case 3: click on the `bottom right corner`, put it on the `top & left` of the cursor.

Issue #2137, opened 2022-08-28
`useCycleList` bug


ap32:
### Describe the bug

```
const { next } = useCycleList([0, 1, 2, 0, 3, 4], { initialValue: 4 });
console.log(Array.from({ length: 6 }, () => next()));

// expected [0, 1, 2, 0, 3, 4]
// recieved [0, 1, 2, 0, 1, 2]
```

`useCycleList` calculates `index` from `state` using  `indexOf`. https://github.com/vueuse/vueuse/blob/08cbd6d9f4b72954eca1865ef1f136f61daba711/packages/core/useCycleList/index.ts#L34

### Reproduction

https://stackblitz.com/edit/vitejs-vite-tvsa1h/?file=src%2Fmain.js

### System Info

```Shell
npmPackages:
  @vueuse/core: ^9.1.1 => 9.1.1 
  @vueuse/head: ^0.7.9 => 0.7.9 
  @vueuse/rxjs: ^9.1.1 => 9.1.1 
  @vueuse/shared: ^9.1.1 => 9.1.1
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

runyasak:
Hi, 😁.
I created a pull request to solve this issue.
Please, checkout on my PR: #2141

ap32:
[const stateIndex = shallowRef(0)](https://github.com/runyasak/vueuse/blob/fix/use-cycle-list-index/packages/core/useCycleList/index.ts#L30)
should be
`const stateIndex = shallowRef(list.indexOf(state.value))`
because of `options.initialValue`.
[...]

PR #2141, opened 2022-08-29
feat(useCycleList): support non-unique list

runyasak:
### Description

This PR is related and solving #2137.

There has a mistake with `useCycleList` when some values are identical.
The index of these values will always be the same as the first one.

So, I implemented `stateIndex` as state variable to store own index, when `getIndexOf` is not defined`

### Additional context

* I use `shallowRef` for `stateIndex`. I'm not sure that should I use `ref` instead?
* This is my first PR for VueUse. If you have any feedback with my code or I missed some guidelines, please do not hesitate to comment on this PR.
* And thank you so much. 😁

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useCycleList/index.ts
+++ b/packages/core/useCycleList/index.ts
@@ -27,11 +27,13 @@ export interface UseCycleListOptions<T> {
 export function useCycleList<T>(list: T[], options?: UseCycleListOptions<T>): UseCycleListReturn<T> {
   const state = shallowRef(options?.initialValue ?? list[0]) as Ref<T>
 
+  const stateIndex = shallowRef(0)
+
   const index = computed<number>({
     get() {
       let index = options?.getIndexOf
         ? options.getIndexOf(state.value, list)
-        : list.indexOf(state.value)
+        : stateIndex.value
 
       if (index < 0)
         index = options?.fallbackIndex ?? 0
@@ -48,6 +50,7 @@ export function useCycleList<T>(list: T[], options?: UseCycleListOptions<T>): Us
     const index = (i % length + length) % length
     const value = list[index]
     state.value = value
+    stateIndex.value = index
     return value
   }
 
--- a/packages/core/useCycleList/index.ts
+++ b/packages/core/useCycleList/index.ts
@@ -27,7 +27,7 @@ export interface UseCycleListOptions<T> {
 export function useCycleList<T>(list: T[], options?: UseCycleListOptions<T>): UseCycleListReturn<T> {
   const state = shallowRef(options?.initialValue ?? list[0]) as Ref<T>
 
-  const stateIndex = ref(0)
+  const stateIndex = ref(list.indexOf(state.value))
 
   const index = computed<number>({
     get() {
--- a/packages/core/useCycleList/index.ts
+++ b/packages/core/useCycleList/index.ts
@@ -1,4 +1,5 @@
-import { type Ref, computed, ref, shallowRef } from 'vue-demi'
+import type { Ref } from 'vue-demi'
+import { computed, ref, shallowRef } from 'vue-demi'
 import { type MaybeRef } from '@vueuse/shared'
 
 export interface UseCycleListOptions<T> {
--- a/packages/core/useCycleList/index.ts
+++ b/packages/core/useCycleList/index.ts
@@ -1,4 +1,4 @@
-import { type Ref, computed, shallowRef } from 'vue-demi'
+import { type Ref, computed, ref, shallowRef } from 'vue-demi'
 import { type MaybeRef } from '@vueuse/shared'
 
 export interface UseCycleListOptions<T> {
@@ -27,7 +27,7 @@ export interface UseCycleListOptions<T> {
 export function useCycleList<T>(list: T[], options?: UseCycleListOptions<T>): UseCycleListReturn<T> {
   const state = shallowRef(options?.initialValue ?? list[0]) as Ref<T>
 
-  const stateIndex = shallowRef(0)
+  const stateIndex = ref(0)
 
   const index = computed<number>({
     get() {
--- a/packages/core/useCycleList/index.test.ts
+++ b/packages/core/useCycleList/index.test.ts
@@ -26,4 +26,19 @@ describe('useCycleList', () => {
       next()
     })
   })
+
+  it('should cycle list objects with `initialValue` and `getIndexOf` orderly', () => {
+    const listItems = [{ name: 'Dog' }, { name: 'Cat' }, { name: 'Lizard' }]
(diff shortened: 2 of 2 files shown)


ap32:
[const stateIndex = shallowRef(0)](https://github.com/runyasak/vueuse/blob/fix/use-cycle-list-index/packages/core/useCycleList/index.ts#L30)
should be
`const stateIndex = shallowRef(list.indexOf(state.value))`
because of `options.initialValue`.
And maybe `options` needs `initialIndex` because of possible duplicate values.

chris-zhu:
> [const stateIndex = shallowRef(0)](https://github.com/runyasak/vueuse/blob/fix/use-cycle-list-index/packages/core/useCycleList/index.ts#L30) should be `const stateIndex = shallowRef(list.indexOf(state.value))` because of `options.initialValue`. And maybe `options` needs `initialIndex` because of possible duplicate values.

I agree with you, and `indexOf` return a plain value, `shallowRef` is unnecessary.

ap32:
Maybe something like
```ts
const stateIndex = shallowRef(
  options?.initialIndex ?? options?.initialValue === undefined
    ? 0
    : list.indexOf(options.initialValue),
);

if (stateIndex.value < 0) {
  stateIndex.value = options?.fallbackIndex ?? 0;
}

const state = shallowRef(list[stateIndex.value]) as Ref<T>;
```

chris-zhu:
I have also thought about this question. Someone needs to decide the priority of value and index🤔

runyasak:
Hi, @ap32 @chris-zhu.
Thank you, for reviewing my PR.

I agree with ideas default value of `stateIndex` and `shallowRef` is unnecessary.
I will update my PR soon. 😁

runyasak:
> I have also thought about this question. Someone needs to decide the priority of value and index🤔

If someone has decided, please let me know and I will contribute as PR. 😂

antfu:
The current approach has a downside in that if you want to update the list for the cycle dynamically the index will break. I am more leaning toward having the input list unique as an requirement (or provide your down getIndex function to overcome it).

runyasak:
I have updated my PR.
Please @ap32 @chris-zhu, review my code again. 😁

ap32:
@antfu but documentation says nothing about the uniqueness of the list :thinking:. I think that more universal approach will be better in this case.

packages/core/useCycleList/index.ts
```diff
@@ -1,4 +1,4 @@
-import { type Ref, computed, shallowRef } from 'vue-demi'
+import { type Ref, computed, ref, shallowRef } from 'vue-demi'
```
songjianet, line 1: It is better to import the type type separately.

songjianet (changes requested):
Can be used as a reference.

packages/core/useCycleList/index.ts
runyasak: Thank you. I have updated PR. 😁

antfu:
Can you also initiate some tests for it? Thanks.

runyasak:
> Can you also initiate some tests for it? Thanks.

Great. I will make create some tests soon, maybe tomorrow.
Because today is Vue.js Forge day 2. 😂

runyasak:
I have initiate tests for `useCycleList` with some cases, please review my tests.
However, I didn't make `getIndexOf` case. I'm not sure about input and expected result.

@antfu Could you give me some test case for `getIndexOf`?

runyasak:
I have updated `getIndexOf` test case.
Please, review my code again. 😁

songjianet (approved):
LGTM

hirschadway:
I think, its not need to get index from state, getting index from index is better

runyasak:
> I think, its not need to get index from state, getting index from index is better

That's a good idea. It looks strange with a duplicate index, maybe confused.
I will try to update by using index from computed. 😁

hirschadway:
> > I think, its not need to get index from state, getting index from index is better
> 
> That's a good idea. It looks strange with a duplicate index, maybe confused. I will try to update by using index from computed. 
yes that right, I mean using index from computed.

packages/core/useCycleList/index.ts
```diff
@@ -27,11 +28,13 @@ export interface UseCycleListOptions<T> {
 export function useCycleList<T>(list: T[], options?: UseCycleListOptions<T>): UseCycleListReturn<T> {
   const state = shallowRef(options?.initialValue ?? list[0]) as Ref<T>
 
+  const stateIndex = ref(list.indexOf(state.value))
```
hirschadway, line 31: this no effect on current problem.only reacting indexOf(state.value).
best work is focus on self index instead of state

hirschadway (approved):
please note on this comment. your change  is not reasonable.

runyasak:
Hi, @hirschadway.

I'm not sure that I follow you.
I try to remove `stateIndex` and set value to `index` computed instead.
Then, I've got errors with `Maximum call stack size`.

Actually, I have separate `stateIndex` because this errors.
By the way, I really want to improve this duplicated index too.

Could you give me some advice or solution to solve this error ?
It would be appreciated with example code. 😁

# Code

```diff
diff --git a/packages/core/useCycleList/index.ts b/packages/core/useCycleList/index.ts
index 536db3e7..712b961d 100644
--- a/packages/core/useCycleList/index.ts
+++ b/packages/core/useCycleList/index.ts
@@ -28,13 +28,11 @@ export interface UseCycleListOptions<T> {
 export function useCycleList<T>(list: T[], options?: UseCycleListOptions<T>): UseCycleListReturn<T> {
   const state = shallowRef(options?.initialValue ?? list[0]) as Ref<T>
 
-  const stateIndex = ref(list.indexOf(state.value))
-
   const index = computed<number>({
     get() {
       let index = options?.getIndexOf
         ? options.getIndexOf(state.value, list)
-        : stateIndex.value
+        : list.indexOf(state.value)
 
       if (index < 0)
         index = options?.fallbackIndex ?? 0
@@ -48,10 +46,10 @@ export function useCycleList<T>(list: T[], options?: UseCycleListOptions<T>): Us
 
   function set(i: number) {
     const length = list.length
-    const index = (i % length + length) % length
-    const value = list[index]
+    const newIndex = (i % length + length) % length
+    const value = list[newIndex]
     state.value = value
-    stateIndex.value = index
+    index.value = newIndex
     return value
   }
```

# Errors Screenshot
image

hirschadway:
set function is depepend on index and index is depend on set(by computed) so, this is infintly doing and causes "Maximum call stack size exceeded"

hirschadway:
my friend @runyasak
The **main problem** is not primarily related to the code, but related to the basic concepts used in the components.
In an array, indexes are unique, but values are not necessarily. Therefore, there is only one value for each index, but not vice versa.

Subsequently, this problem is not solved with the code

Issue #2135, opened 2022-08-27
[Feature] Add isLoadingDelay to useAsyncState


Bobakanoosh:
### Clear and concise description of the problem

When using [useAsyncState](https://vueuse.org/core/useasyncstate/), I feel its a common use case to display some loading symbol/icon when `isLoading` is true.

However, if the endpoint takes < 100ms to respond, this loading symbol actually ends up looking bad because it flickers in and out of visibility for < 100ms.

### Suggested solution

It would be nice to able to pass `isLoadingDelay: 250` or something similar as an option to this function.

I would imagine implementation would be simple:
```ts
const {
   isLoadingDelay = 0,
} = options ?? {}

// in execute function
setTimeout(() => {
   // If the request finished while we were waiting, we dont want to set loading to true
   if(!isReady.value && !error.value) {
         isLoading.value = true;
   }
}, isLoadingDelay);
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
https://github.com/vueuse/vueuse/pull/2155#issuecomment-1235093468

PR #2155, opened 2022-09-01
feat(useAsyncState): add isLoadingDelay option to 

Bobakanoosh:
Resolves https://github.com/vueuse/vueuse/issues/2135

Implementation poses no breaking changes, as the default `isLoadingDelay` is `0`, so it will be set instantly.


antfu:
Thanks. But I think might be a bit confusing for ppl out-of-context and it does not work with manually triggered executing. I'd suggest you do it in your component, or use `useDebounded`  to compose.

Issue #1638, opened 2022-05-25
proposal: useOverlap


damienroche:
### Clear and concise description of the problem

Know when a HTMLElement Overlap one or multiple other HTMLElement

https://user-images.githubusercontent.com/1529058/170302736-4e0c1f43-a432-4197-b68d-8a146e26d584.mp4

#### Example using useOverlap() with useDragglable()

```typescript
const origin = ref()
const disabled = ref()

const { x, y } = useDraggable(origin, {...})
const { isOverlaping } = useOverlap(origin, disabled, { watchers: [x, y], allowEdge: true })
```

### Suggested solution

```typescript
// useOverlap/index.ts

interface UseOverlapOptions {
  /**
   * Set isOverlaping to false when elements are overlaping from edges
   *
   * @default true
   */
  allowEdge: boolean
   /**
   * Array of ref that recalculate overlap
   *
   * @default undefined
   */
  watchers: Ref[]
}

const overlap = (r1: DOMRect, r2: DOMRect, offset = 0) => !(r1.right - offset < r2.left || r1.left + offset > r2.right || r1.bottom - offset < r2.top || r1.top + offset > r2.bottom)

export const useOverlap = (origin: Ref<HTMLELement>, target: Ref<HTMLElement> | Ref<HTMLElement>[], options: UseOverlapOptions) => {
  const { watchers = ref(undefined), allowEdge = false } = options
  const isOverlaping = ref(false)

  const checkOverlap = () => {
    const targetUnref = unref(target)
    const targets = Array.isArray(targetUnref) ? targetUnref : [targetUnref]
    const rect = unref(origin)!.getBoundingClientRect()
    const targetRects = targets.map(value => value.getBoundingClientRect())
    isOverlaping.value = targetRects.some(r => overlap(rect, r, +allowEdge))
  }

  watch(watchers, () => {
    checkOverlap()
  })

  onMounted(() => {
    checkOverlap()
  })

  return {
    isOverlaping,
    checkOverlap
  }
}
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

JMaylor:
[...]

PR #2156, opened 2022-09-02
feat(useElementCross): new function

JiatLn:
### Description

Reactive two element whether cross state.

### Additional context

---

- [ ] Bug fix
- [X] New Feature
- [X] Documentation update
- [ ] Other


packages/core/useElementCross/index.ts
```diff
@@ -0,0 +1,21 @@
+import { computed } from 'vue-demi'
+import type { MaybeComputedElementRef } from '../unrefElement'
```
songjianet, line 2: import type is best placed at the bottom of all references.

songjianet (approved):
Can be used as a reference.

chris-zhu:
I think we should improve the demo. Like [`useDraggable`](https://github.com/vueuse/vueuse/blob/main/packages/core/useDraggable/demo.vue). Move those two elements to observe more values of `isCross`

wheatjs:
A couple of thoughts:

1. This seems a little specific, I think we would probably want something more generic that would accept more elements.
2. If we did add this function, I think something like `useElementIntersection` might be a better name

Tbh though, I feel like this type of function is more of an edge case and could be left up to the user. Otherwise we might provide something that isn't flexible enough for the cases where it might be used and not common enough for general use.

darrylnoakes:
How does this relate to #1638?

PR #2163, opened 2022-09-03
chore(useColorMode): use strict equal

Sepush:
Signed-off-by: Sepush <sepush@outlook.com>

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other


antfu:
Emm, why?

Sepush:
> Emm, why?

`undefined == null` will get `true` make it explicit will be better
but `storageKey` has defualt value so it also OK

antfu:
I think it's intended to check both `undefined` and `null`. The same patterns have been used in many places so I'd think this is necessary.

PR #2168, opened 2022-09-04
feat(useBrowserLocation): two-way binding

sxzz:
closes #1125

### Description

Two-way binding of `window.location`. It's readonly before. After this PR, it can be changed by user.

### Additional context

I think the returning value of `useBrowserLocation` should be a ref map. But ATM let it be a reactive (maybe it can be refactored in the next major version.)

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useBrowserLocation/demo.vue
+++ b/packages/core/useBrowserLocation/demo.vue
@@ -7,5 +7,7 @@ const text = stringify(location)
 </script>
 
 <template>
+  Input and hash will be changed:
+  <input v-model="location.hash" type="text" placeholder="Hash">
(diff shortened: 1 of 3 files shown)


songjianet (approved):
LGTM

antfu:
I think we should make it return as refs object in the next major version. The current approach of wrapping `reactive` with `ref` is not ideal to me.

sxzz:
Yeah, as I mentioned in `Additional context`

sxzz:
I think we can have it now, and refactor it before the next major released.

Issue #2160, opened 2022-09-03
error when use ref in Vue component instance(should use unrefElement)


mymx2:
### Describe the bug

error when use ref in Vue component instance. same error in
// 1. useElementSize
// 2. useVirtualList
// 3. useScrollLock
// 4. usePointerSwipe
// 5. useInfiniteScroll
// 6. useElementVisibility
// 7. useDraggable
fix: use unrefElement rather than resolveUnref
![image](https://user-images.githubusercontent.com/32814920/188263067-f37ab029-9b9b-4b22-add2-4760bd5b2192.png)

### Reproduction

[...]

PR #2169, opened 2022-09-04
feat(useElementVisibility): support watch the real element

sxzz:
relate #2160

### Description

Cannot get the real HTMLElement, we should use `unrefElement`.

Also, if `element` is changed, it should also be updated.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useElementVisibility/index.ts
+++ b/packages/core/useElementVisibility/index.ts
@@ -50,8 +50,11 @@ export function useElementVisibility(
     { immediate: true, flush: 'post' },
   )
 
-  if (window)
-    useEventListener(scrollTarget || window, 'scroll', testBounding, { capture: false, passive: true })
+  if (window) {
+    useEventListener(scrollTarget || window, 'scroll', testBounding, {
+      capture: false, passive: true,
+    })
+  }
 
   tryOnScopeDispose(() => stop())
 
--- a/packages/core/useElementVisibility/index.ts
+++ b/packages/core/useElementVisibility/index.ts
@@ -33,7 +33,7 @@ export function useElementVisibility(
       elementIsVisible.value = false
     }
     else {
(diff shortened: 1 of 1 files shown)


songjianet (dismissed):
LGTM

packages/core/useElementVisibility/index.ts
```diff
@@ -43,12 +44,19 @@ export function useElementVisibility(
...
 
   if (window) {
-    tryOnMounted(() => useEventListener(
-      () => resolveUnref(scrollTarget) || window, 'scroll', testBounding, { capture: false, passive: true }))
+    useEventListener(scrollTarget || window, 'scroll', testBounding, {
+      capture: false, passive: true,
+    })
   }
 
+  tryOnScopeDispose(() => stop())
```
antfu, line 59: this is redundant, `watch` already did it.

sxzz: Oops, just updated.

mymx2:
also remove (el as HTMLElement)
`
const rect = (el as HTMLElement).getBoundingClientRect()
`
to
`
const rect = el.getBoundingClientRect()
`

mymx2:
and what if use watch instead of tryOnMounted. also Considering use v-if both on the Vue component instance and component.
```
const el = resolveUnref(element)
if (!el) {
  elementIsVisible.value = false
}
```
el instanceof Element
```
const el = unrefElement(element)
if (!el || !(el instanceof Element)) {
  elementIsVisible.value = false
}
```

sxzz:
@mymx2 `MaybeComputedElementRef` it must be a Element, so I think we needn't have repeat judgement in runtime code.

antfu (approved):

Issue #2096, opened 2022-08-17
WebSocket: Wrong implementation of heartbeat


miili:
### Describe the bug

The heartbeat is currently sending a `'ping'`, this ought to be `0x9` and pong is `0xA`.

Also I do not see the acknowledgement of the Pong. I can work on this if required.

See the reference on MDN:
https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#pings_and_pongs_the_heartbeat_of_websockets

### Reproduction

https://github.com/vueuse/vueuse/blob/dc3ebe80bc61ee56004fdcd417cb234ee749e41e/packages/core/useWebSocket/index.ts#L235

### System Info

```Shell
generic
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

miili:
@antfu ping, I think this should be high prio.

ilyaliao:
Closed via #2170

PR #2170, opened 2022-09-05
feat(useWebSocket): enhence heartbeat

azaleta:
### Description
issue: #2096

1. a new option `pongTimeout` -> after sending ping, could not recieve any response in pongTimeout, then close the socket
2. skip the pong message in ws.onmessage()

### Additional context

---

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useWebSocket/index.md
+++ b/packages/core/useWebSocket/index.md
@@ -71,6 +71,7 @@ const { status, data, close } = useWebSocket('ws://websocketurl', {
   heartbeat: {
     message: 'ping',
     interval: 1000,
+    pongTimeout: 1000,
   },
 })
 ```
--- a/packages/core/useWebSocket/index.ts
+++ b/packages/core/useWebSocket/index.ts
@@ -6,6 +6,8 @@ import { useEventListener } from '../useEventListener'
 
 export type WebSocketStatus = 'OPEN' | 'CONNECTING' | 'CLOSED'
 
+const DEFAULT_PING_MESSAGE = 'ping'
+
 export interface UseWebSocketOptions {
   onConnected?: (ws: WebSocket) => void
   onDisconnected?: (ws: WebSocket, event: CloseEvent) => void
@@ -31,6 +33,13 @@ export interface UseWebSocketOptions {
      * @default 1000
      */
     interval?: number
+
+    /**
+     * Heartbeat response timeout, in milliseconds
+     *
+     * @default 1000
+     */
+    pongTimeout?: number
   }
 
   /**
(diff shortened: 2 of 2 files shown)


antfu (approved):

volarname:
useWebsocket stopeed working in version v9.2.0, so something is broken now

azaleta:
@volarname can you describe the situation or offer some repo?
did some easy test and don't see stop working

```
<script setup lang='ts'>
import { useWebSocket } from '@vueuse/core'

const { status, data, send, open, close } = useWebSocket('wss://ws.postman-echo.com/raw', {
  onMessage: (ws: WebSocket, event: MessageEvent) => {
    console.log('get message')
  },
})
const i = ref(1)
function dodo() {
  console.log(status.value)
  send(`test${i.value}`)
  i.value++
}

function dodo2() {
  console.log(data.value)
}
</script>

<template>
  <div>
    <button m-1 p-1 @click="dodo">
      123
    </button>

    <button m-1 p-1 @click="dodo2">
      223
    </button>
  </div>
</template>
```

image

kellerza:
From this code a "pong" seems to be mandatory... and probably why it broke. My server also don't send pongs and have some quiet times

azaleta:
@kellerza 
heartbeat is an option, if the sever don't support it, I think you should not trigger ping.
if you do send ping to sever of this type, it will close socket because could not get pong response -> this will consider as server down.

My concept is 
1.  send heartbeat is you enabled it in option (also start a timer `pongTimeoutWait` to close socket after a period of time)
https://github.com/vueuse/vueuse/blob/main/packages/core/useWebSocket/index.ts#L260-L277

2.every time you get a server response, `pongTimeWait` will be clear by `resetHeartbeat`
Also if the response is PONG, the response will be skipped.
https://github.com/vueuse/vueuse/blob/main/packages/core/useWebSocket/index.ts#L244-L258

kellerza:
Here we are talking about an app layer heartbeat, so there is no right/wrong on how to implement it.

Up to 9.2.0 you can freely enable the heartbeat without any support from the server.
In 9.2.0 the server now needs to respond immediately (or a pong)

Not necessarily wrong, but the behavior changed enough to bring down my and @volarname's websockets after the upgrade

kellerza:
fyi, not arguing about the behavior. I already added the pong to my server and will go to 9.2.0 eventually...

kellerza:
This also affects `autoReconnect`. Once the pong closes the websocket you need to perform an explicit `open`

ssickles:
One thing I ran into is that it doesn't appear that your ping message is being sent with the correct opcode. It is coming across as a TextMessage (0x1) for me so my server wasn't recognizing the ping. According to the websocket spec, you should use an opcode of 0x9. And you might want to also check the pong message for an opcode of 0xA.

https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2

kellerza:
@ssickles this it seems that the browser/JS API only exposes TextMessages. So this is an application layer heartbeat, rather than the Websocket ping/pong that uses other opcodes from the RFC

MatiasHiltunen:
I'm having a problem with firefox "Firefox can’t establish a connection to the server at ws://127.0.0.1:8080/ws.". Connection status is open but running send function gives the previous error. If set to autoReconnect status wont reach the open state at all and will continue trying to establish the connection. With chrome based browsers there is no issue.

Issue #2159, opened 2022-09-02
[useArrayMap] Allow output type different to input type


mandoway:
### Clear and concise description of the problem

As a dev using VueUse, I want to map an array to a different type so that I can map API result lists to another object.

The useArrayMap function currently limits the output type to be the same as the input type.
This does not reflect the behaviour of the native Array.map() function.

### Suggested solution

In module shared/useArrayMap there should be a second generic argument to the function useArrayMap which is then used as the return type of the passed mapper function.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #2172, opened 2022-09-05
fix(useArrayMap): allow return type matches the mapper function

GODLiangCY:
fix [#2159](https://github.com/vueuse/vueuse/issues/2159)

### Description

fix [#2159](https://github.com/vueuse/vueuse/issues/2159)

### Additional context

let the `useArrayMap()` function to take the second generics argument, which should be used in the return type

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/shared/useArrayMap/index.ts
+++ b/packages/shared/useArrayMap/index.ts
@@ -12,7 +12,7 @@ import { resolveUnref } from '../resolveUnref'
  *
  * @returns {Array} a new array with each element being the result of the callback function.
  */
-export function useArrayMap<T, U>(
+export function useArrayMap<T, U = T>(
   list: MaybeComputedRef<MaybeComputedRef<T>[]>,
   fn: (element: T, index: number, array: T[]) => U,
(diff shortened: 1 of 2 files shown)


packages/shared/useArrayMap/index.ts
```diff
@@ -12,9 +12,9 @@ import { resolveUnref } from '../resolveUnref'
  *
  * @returns {Array} a new array with each element being the result of the callback function.
  */
-export function useArrayMap<T>(
+export function useArrayMap<T, U>(
```
chris-zhu, line 15:
```suggestion
export function useArrayMap<T, U = T>(
```

chris-zhu (changes requested):

packages/shared/useArrayMap/index.ts
GODLiangCY: Done

sxzz (approved):
LGTM

chris-zhu (approved):

PR #2192, opened 2022-09-11
docs(onClickOutside): Mention unregister onClickOutside listener in the doc

### LINKED ISSUE #2188: Mention «close» function in the «Usage» section of «onClickOutside» documentation

Right now `onClickOutside` function returns [`close`](https://github.com/vueuse/vueuse/blob/main/packages/core/onClickOutside/index.ts#L91) that seems to remove all related listeners, but documentation does not mention it. 
Therefore, It's not evident that there's such functionality.

### Suggested solution

Show functionality in the [usage example](https://vueuse.org/core/onclickoutside/#usage).

For example, instead of:
```js
  setup() {
    const target = ref(null)

    onClickOutside(target, (event) => console.log(event))

    return { target }
  }
```
Describe it as:
```js
  setup() {
    const target = ref(null)

    const stop = onClickOutside(target, (event) => console.log(event))

    return { target, stop }
  }
```

It might also be a good idea to describe it more, but I think it should be mentioned at least in the way I showed it above.

### Alternative

_No response_

### Additional context

I would like to make this change if one of the members confirms it's needed.

### Validations

edumudu:
### Description

fixes #2188

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/onClickOutside/index.md
+++ b/packages/core/onClickOutside/index.md
@@ -34,6 +34,15 @@ export default {
 </script>
 ```
 
+You can also call the returned to unregister the listener.
+```ts
+import { onClickOutside } from '@vueuse/core'
+
+const cleanup = onClickOutside(target, event => console.log(event))
+
+cleanup() // This will unregister the listener.
+```
+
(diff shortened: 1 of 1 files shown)


antfu:
Thanks but 56ce7e63b00c5b2375890f2aee924cd090d2f1e2

PR #2194, opened 2022-09-12
fix(useElementVisibility): add update condition

vaakian:
### Description

**1. Fixed two cases:**
- On window `resize`(as shown in the video).
- On element itself `resize`, when element's `height` / `width` changes, it may become invisible.

**2. Re-implemented the condition of `scrollTarget`：**
 in the previous implementation, the visibility is still relative to the `window`, I think the expected behavior should to be relative to the `scrollTarget`.

https://user-images.githubusercontent.com/30516060/189672096-ff2f7d92-0861-47aa-b70b-cb10aa867c8e.mov

### Additional context

Thanks to the former authors, but I refactored it since there's a chance to reuse existing composables right now, so the **resize** issue and other **edge cases** were easily handled.

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


vaakian:
Seemed the conficts is because of the refactoring?

packages/core/useElementVisibility/index.ts
```diff
@@ -19,40 +19,44 @@ export interface UseElementVisibilityOptions extends ConfigurableWindow {
...
+        && elementBounding.bottom >= 0
+        && elementBounding.right >= 0
+    )
+  })
 
-  if (window) {
-    useEventListener(scrollTarget || window, 'scroll', testBounding, {
-      capture: false, passive: true,
-    })
+  if (scrollTarget) {
```
antfu, line 53: The check here seems to be not right. `scrollTarget` can be a template ref that is fulfilled later. We was checking for `window` for SSR.

vaakian: I thought `scrollTarget` can be `undefined` instead of `Ref<undefined>`, but now I checked `useEventLister` and it seems already SSR friendly, so I can just remove that check? (don't need to check the `window` either before?)

vaakian:
just accidentally closed this PR, sorry for the disturbance.

PR #2198, opened 2022-09-14
fix(onClickOutside): put ignore logic on `pointerdown` event

### LINKED ISSUE #2196: onClickOutside: listen only to mousedown event for ignore elements
### Describe the bug

Hi! 👋

If I click inside an ignore element  then release my mouse outside of it and the target, the handler function is called.

In a previous version of vueUse, it was possible to specifiy which events should trigger the handler ( see https://github.com/vueuse/vueuse/pull/508/commits/e080ec61f8b99527f45458b4d7d9de3ee378f3b8)

I'm not sure if this was also applied to ignore elements, but the option was then removed in this commit https://github.com/vueuse/vueuse/commit/2ff6ca9b0b1cf3f3ea885c9130d3b417f680b5cd

Why the option has been removed? Could it be possible to bring it back? How can I do to not trigger the handler function?

### Reproduction

https://stackblitz.com/edit/vitejs-vite-jkw1v5/?file=src%2FApp.vue

### System Info

```Shell
vueuse stackblitz
```

### Used Package Manager

npm

### Validations

GODLiangCY:
### Description

fix #2196 

### Additional context

put ignore logic on `pointerdown` event, making sure the proper behavior.

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -54,22 +54,28 @@ export function onClickOutside<T extends OnClickOutsideOptions>(
     if (!el || el === event.target || composedPath.includes(el) || !shouldListen.value)
       return
 
-    if (ignore && ignore.length > 0) {
-      if (ignore.some((target) => {
-        const el = unrefElement(target)
-        return el && (event.target === el || composedPath.includes(el))
-      }))
-        return
-    }
-
     handler(event)
+  }
+
+  const shouldIgnore = (event: PointerEvent): boolean => {
+    if (!ignore || !ignore.length)
+      return false
+
+    const composedPath = event.composedPath()
+    if (ignore.some((target) => {
+      const el = unrefElement(target)
+      return el && (event.target === el || composedPath.includes(el))
+    }))
+      return true
+
+    return false
   }
 
   const cleanup = [
     useEventListener(window, 'click', listener, { passive: true, capture }),
     useEventListener(window, 'pointerdown', (e) => {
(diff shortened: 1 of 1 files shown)


packages/core/onClickOutside/index.ts
```diff
@@ -54,22 +54,28 @@ export function onClickOutside<T extends OnClickOutsideOptions>(
...
+
+    const composedPath = event.composedPath()
+    if (ignore.some((target) => {
+      const el = unrefElement(target)
+      return el && (event.target === el || composedPath.includes(el))
+    }))
+      return true
+
+    return false
+  }
```
sibbng, line 72:
```suggestion
  const shouldIgnore = (event: PointerEvent) => {
    return ignore && ignore.some((target) => {
      const el = unrefElement(target)
      return el && (event.target === el || event.composedPath().includes(el))
    })
  }
```

packages/core/onClickOutside/index.ts
```diff
@@ -54,22 +54,28 @@ export function onClickOutside<T extends OnClickOutsideOptions>(
     if (!el || el === event.target || composedPath.includes(el) || !shouldListen.value)
```
sibbng, line 54:
```suggestion
    if (!el || el === event.target || event.composedPath.includes(el) || !shouldListen.value)
```

sibbng (changes requested):

sibbng (approved):

Issue #1757, opened 2022-07-06
`templateRef` type argument does not support vue copmponent


Holi0317:
### Describe the bug

Currently the type argument for [`templateRef`](https://vueuse.org/core/templateref/#templateref) is `<T extends HTMLElement | SVGElement | null>`. If I pass in a vue component type in the type argument, typescript will fail as the type does not extend `HTMLElement` or `SVGElement`.

(Run `npm run build` in the given stackblitz below)

### Reproduction

https://stackblitz.com/edit/vitejs-vite-sthn2a?file=src/App.vue

### System Info

```Shell
❯ npx envinfo --system --npmPackages '{vue,@vueuse/*}' --binaries --browsers
success Install finished in 1.656s


  System:
    OS: Linux 5.0 undefined
    CPU: (4) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.10 - /bin/yarn
    npm: 7.17.0 - /bin/npm
  npmPackages:
    @vueuse/core: ^7.6.2 => 7.7.1 
    vue: ^3.2.25 => 3.2.37
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

Holi0317:
Side comment/suggestion: I think providing a stackblitz link in the issue template that contains vite+vue2/3+vueuse would make reproducing issues easier.

Holi0317:
[...]

PR #1760, opened 2022-07-06
fix(templateRef): add Component type

murongg:
fix #1757

### Description

`templateRef` not supported vue componet type

### Additional context

---

### What is the purpose of this pull request? 

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.

PR #2203, opened 2022-09-15
fix(templateRef): add Component type

Holi0317:
CC: @murongg

Fixes https://github.com/vueuse/vueuse/issues/1757, replace PR #1760

### Description

Allow `templateRef` to support vue component type

### Additional context

This replaces PR #1760 by rebasing the latest main branch, then add a test case for the new use case for templateRef function.

---

- [x] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

- [x] ~~Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.~~ Replacing another PR

### DIFF
--- a/packages/core/templateRef/index.test.ts
+++ b/packages/core/templateRef/index.test.ts
@@ -72,4 +72,31 @@ describe('templateRef', () => {
 
     expect(vm.targetEl).toBe(null)
   })
+
+  it('support vue component as ref', async () => {
+    const ChildComponent = defineComponent({
+      name: 'ChildComponent',
+      render() {
+        return null
+      },
+    })
+
+    const vm = mount(defineComponent({
+      components: {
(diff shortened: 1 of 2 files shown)


packages/core/templateRef/index.test.ts
```diff
@@ -72,4 +72,31 @@ describe('templateRef', () => {
...
+          targetEl,
+        }
+      },
+      render() {
+        return h(ChildComponent, { ref: 'target' })
+      },
+    }))
+
+    expect(vm.targetEl).toBeDefined()
+    expect(vm.targetEl.$options.name).toBe('ChildComponent')
```
songjianet, line 100: It's best to use unmount when you're done.

songjianet (changes requested):
for reference only.

PR #2206, opened 2022-09-15
fix(useWebSocket): don't hide pong setTimeout reference

kellerza:
### Description

In #2170 the capability was added to monitor pong's (or server replies) as part of the existing heartbeat/ping functionality.

Currently if the pong gets delayed/dropped for any reason (e.g. server restart) and "pong reply time">pingInterval; the ping function would execute again

The ping function creates a pong setTimout and saves the reference (this timeout will close the connection if allowed to expire). 

If a previous setTimeout has not been cancelled yet, the reference is lost and the  connection will be closed.

### Additional context

By adjusting the timers appropriately you can now be more leanient and allow for "dropped" pongs. E.g if you require 1 pong (or server message) for every 3 pings you could set pongTimeout >= 3x ping interval

It's ok to clear and invalid setTimeout reference - https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout#notes

---

- [x] Bug fix #2170 
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


kellerza:
Anything still missing here?

packages/core/useWebSocket/index.ts
```diff
@@ -267,6 +268,8 @@ export function useWebSocket<Data = any>(
     const { pause, resume } = useIntervalFn(
       () => {
         send(message, false)
+        if (typeof(pongTimeoutWait) !== 'undefined')
```
antfu, line 271:
```suggestion
        if (pongTimeoutWait != null)
```

antfu (approved):

Issue #1467, opened 2022-03-30
[bug]useVirtualList 在ts中使用v-bind containerProps时类型错误


yanbowe:
### Describe the bug

![image](https://user-images.githubusercontent.com/43030980/160732785-7152c5f4-7929-409a-89d2-71f9f85ef733.png)

复制demo至个人项目后 ts报错containerProps与div类型不重合

### Reproduction

https://github.com/vueuse/vueuse/blob/main/packages/core/useVirtualList/demo.vue

### System Info

```Shell
System:
    OS: Windows Server 2016 10.0.14393
    CPU: (4) x64 Intel(R) Core(TM) i5-4460  CPU @ 3.20GHz
    Memory: 9.34 GB / 15.94 GB
  Binaries:
    Node: 16.13.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.1.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Internet Explorer: 11.0.14393.2007
  npmPackages:
    @vueuse/core: ^7.7.1 => 7.7.1
    vue: ^3.2.31 => 3.2.31
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

yukecs:
可以显式声明containerProps为any，规避ts报错问题

JMaylor:
I ran into this issue as well. It seems to occur whenever a class is applied to the container element, as well as binding the `containerProps`

i.e.

```html
<!-- setting height with inline style works -->
[...]

PR #2215, opened 2022-09-16
fix(useVirtualList) - containerProp styles typing fix

JMaylor:
This fixes https://github.com/vueuse/vueuse/issues/1467.

Before this fix, we run into a TypeScript error when running the given example in the docs, on the html element where `containerProps` is bound.

```
error TS2322: Type '{ ref: Ref<HTMLElement | null>; onScroll: () => void; style: CSSStyleDeclaration; }' is not assignable to type 'ElementAttrs<HTMLAttributes>'.  Type '{ ref: Ref<HTMLElement | null>; onScroll: () => void; style: CSSStyleDeclaration; }' is not assignable to type 'HTMLAttributes'.
    Types of property 'style' are incompatible.
      Type 'CSSStyleDeclaration' is not assignable to type 'StyleValue | undefined'.        Type 'CSSStyleDeclaration' is missing the following properties from type 'StyleValue[]': pop, push, concat, join, and 26 more.
```

### Additional context

This bug was also causing the following error in the return statement of useVirtualList/index.ts

```
Expression produces a union type that is too complex to represent.ts(2590)
```

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


StepanMynarik:
Hi, this fix is exactly what I need!
Why was this PR just closed?

JMaylor:
I ran into some TS issues and didn't have the time to fix them, to be honest.
I'll reopen it in case you can resolve them.

Surof1n:
hi i fixed this error in [PR](https://github.com/vueuse/vueuse/pull/2310),

PR #2221, opened 2022-09-18
fix(useStorage): fixed data desynchronization in case of clearing storage

yuryrudey:
### Description

Fixes data desynchronization in case of clearing storage(using `storage.clear()` or manually)
Condition to detect storage clearing is taken from [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event#event_properties)(pay attention to description of `key` property)

### Additional context

Using `useStorage` I noticed that after clearing storage in Chrome DevTools, my computed property wasn't recomputed. The cause was condition in the storage event handler that ignored this case

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/useStorage/index.ts
```diff
@@ -218,7 +218,15 @@ export function useStorage<T extends(string | number | boolean | object | null)>
   }
 
   function update(event?: StorageEvent) {
-    if (event && event.key !== key)
+    if (event?.storageArea !== storage)
+      return
```
yuryrudey, line 222: This condition also fixes bug with using `useStorage` on `sessionStorage` and `localStorage` with the same key

packages/core/useStorage/index.ts
```diff
@@ -218,7 +218,15 @@ export function useStorage<T extends(string | number | boolean | object | null)>
...
+
+    if (event?.key === null) {
```
yuryrudey, line 224: According to documentation, when `key` is `null`, it means that storage has been cleared

packages/core/useStorage/index.ts
```diff
@@ -218,6 +215,14 @@ export function useStorage<T extends(string | number | boolean | object | null)>
   }
 
   function update(event?: StorageEvent) {
+    if (event && event.storageArea !== storage)
+      return
```
yuryrudey, line 219: This condition also fixes overlapping bug when using `useStorage` on `sessionStorage` and `localStorage` with the same key

packages/core/useStorage/index.ts
```diff
@@ -218,6 +215,14 @@ export function useStorage<T extends(string | number | boolean | object | null)>
...
+
+    if (event && event.key === null) {
+      data.value = rawInit
+      return
+    }
```
yuryrudey, line 224: According to documentation, when key is null, it means that storage has been cleared

antfu (approved):

PR #2224, opened 2022-09-19
fix(until): `.not` returns new instance

### LINKED ISSUE #2223: `until.not` is not implemented correctly
### Describe the bug

Currently `until().not` is a getter which toggles the internal state and returns self. It will cause problems if the user store the result of `until()` in a variable and use `.not` more than once, e.g.:
```ts
const a = ref(false);
const u = until(a);
u.not.toBeTruthy().then(() => console.log('ok1'));
u.not.toBeTruthy().then(() => console.log('ok2'));
```

We expect both `ok1` and `ok2` will be output but in fact `ok2` is not.

### Reproduction

https://stackblitz.com/edit/vue-zhcbnm?file=src%2Fcomponents%2FHelloWorld.vue,src%2FApp.vue

### System Info

```Shell
System:
    OS: macOS 12.0.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  Binaries:
    Node: 16.14.2
    npm: 8.5.0
  Browsers:
    Chrome: 105.0.5195.125
    Safari: 15.1
  npmPackages:
    @vueuse/core: ^8.6.0 => 8.6.0 
    @vueuse/shared: ^8.6.0 => 8.6.0 
    vue: 2.7.10 => 2.7.10
```

### Used Package Manager

npm

### Validations

lsdsjy:
### Description

Fixes #2223 

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/shared/until/index.ts
+++ b/packages/shared/until/index.ts
@@ -66,7 +66,7 @@ export interface UntilArrayInstance<T> extends UntilBaseInstance<T> {
   toContains(value: MaybeComputedRef<ElementOf<ShallowUnwrapRef<T>>>, options?: UntilToMatchOptions): Promise<T>
 }
 
-function _until<T>(r: any, isNot = false): any {
+function createUntil<T>(r: any, isNot = false) {
   function toMatch(
     condition: (v: any) => boolean,
     { flush = 'sync', deep = false, timeout, throwOnTimeout }: UntilToMatchOptions = {},
@@ -184,7 +184,7 @@ function _until<T>(r: any, isNot = false): any {
       changed,
       changedTimes,
       get not() {
-        return _until(r, !isNot)
+        return createUntil(r, !isNot) as UntilArrayInstance<T>
       },
     }
     return instance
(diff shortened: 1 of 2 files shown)


antfu:
Can we have some tests for it? Thanks

packages/shared/until/index.ts
```diff
@@ -66,24 +66,7 @@ export interface UntilArrayInstance<T> extends UntilBaseInstance<T> {
...
- *
- * alert('Counter is now larger than 7!')
- * ```
- */
-export function until<T extends unknown[]>(r: WatchSource<T> | MaybeComputedRef<T>): UntilArrayInstance<T>
-export function until<T>(r: WatchSource<T> | MaybeComputedRef<T>): UntilValueInstance<T>
-export function until<T>(r: any): any {
-  let isNot = false
-
+export function _until<T>(r: any, isNot = false): any {
```
antfu, line 69: We should not expose it

antfu (approved):

PR #2229, opened 2022-09-20
feat(toReadonly): new function

vaakian:
### Description

**Unexpected mutation** to a `ref` that returned from some `composables` should be prevented, same as **props** in **Vue**.

for example:
```ts
const { height, width } = useElementBounding(...)

// accidental mutation
width.value = -999
```

Also, there are already some functions trying to do this, but only a fraction:
- [usePointerSwipe](https://github.com/vueuse/vueuse/blob/dc3ebe80bc61ee56004fdcd417cb234ee749e41e/packages/core/usePointerSwipe/index.ts#L157)
- [useStyleTag](https://github.com/vueuse/vueuse/blob/af3b34ddb533f393cf7000e44fb5f772bf6eea1c/packages/core/useStyleTag/index.ts#L114)
- ...

### Additional context

This can be useful to a lot of existing composables, I'm willing to check all functions and fix them if this proposal is considered to be feasible.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


sxzz (approved):
LGTM in general /cc @antfu

antfu:
I am not sure about the naming am not sure if the extra runtime cost is worth it. I think it might be better to improve our types to mark those props readonly instead. Thanks for your contribution.

PR #2231, opened 2022-09-20
fix(useFetch): fix incorrect `chainCallbacks` behavior

### LINKED ISSUE #2225: useFetch: chainCallbacks wrong use of forEach with async-await
### Describe the bug

https://github.com/vueuse/vueuse/pull/1545
https://github.com/vueuse/vueuse/blob/a1d93d3b7068ad9e08ee38b75504d8cbf0c16e0b/packages/core/useFetch/index.ts#L217-L225

The current code immediately executes all callbacks in parallel and does not wait for the results to be returned, and causes errors thrown in the intercept method to not be caught correctly

This will result in the following errors：
```js
const useMyFetch = createFetch({
  options: {
    async beforeFetch(ctx) {
      await delay(2000)
      return ctx
    },
    afterFetch(ctx){
      throw new Error('afterFetch Error')
      return ctx
    }
  }
})

const { data: data_chainCallbacks, error: error_chainCallbacks } = useMyFetch('https://httpbin.org/get', {refetch:false})
//If the useFetchOptions parameter is passed, the intercept method will be wrapped by chainCallbacks and the error will not be caught.
//The delay in beforeFetch doesn't work either

const { data, error } = useMyFetch('https://httpbin.org/get')
//The error will be caught correctly when not wrapped by the chainCallbacks
//The delay in beforeFetch will work properly
```

Using for ... of loop or Promise.all() will get the correct behavior
```ts
function chainCallbacks<T = any>(...callbacks: (((ctx: T) => void | Partial<T> | Promise<void | Partial<T>>) | undefined)[]) {
  return async (ctx: T) => {
    for(const callback of callbacks){
      if (callback)
        ctx = { ...ctx, ...(await callback(ctx)) }
    }
    return ctx
  }
}
```

### Reproduction

 https://play.vueuse.org/#N4IgDghgxg1hDmBTAziAXAbVAOwgW0XRADcBXRAWgBNE8BLEAGhGQHtSAnKQtEU7MDHhMQNZFA50wAFzqtsRAGrkABABFadFQAoAFhAA2AMxV1sKgGIdE2KLoCUp5CogqaxRAdZgz8FaVkDOmkATxcDLwB3ZxD2FWlWFUjJaUQVAFVsOg8OZEMVZTSAGToAIw4ISRQVI1YOAtUAJhUAMhUAZhFOAyJdaWkwZDQAemH+QXgAOihWPGGyShp6YaDS4bMaAA9JvAArVABfRhx8HhAAAQXSZERh5H1rKhE2Tm4icaERMQkpWXkiADKD0QVAaiHSN38gWCdBQky6HB6vD6AyGow+Uxmc0u5Gut3ulRB5wAnJMAAzk9bYLY7fYgI4nAhEHGIPHDGbWZ7sLhnDFfFA-GRyBS8ADCrAiiCgf3MrBMKBu2Fk+UKKnFeDArGQMPkGWhshQCKRIBRgxGYwEQmms2GLLZHMQJPJlI2iG2e0OAF1mEY6AZDWhQABBMBgSYLdCgVIagwQVJEAA8VGyAD5EBwOHUAPp2CBmUWGAylaAwIYqYDAdOZjg5-T5wvF2DIA4HBPDZPEFMAHQ4XewSdTVbqaHLlYzdRbbY73d72B7fYHnaocYgtbz2ALEUbpZHFeX0lXufrW5Lzdb7dT8-70-3EF3wFvk4vnavz0F0iIdA1dWk5ZUEkQONEAsRBpDsRh-BuECwN0FQDhqTM8BUAByO0bnZOpEGQvs+xmbBkF-GhYzCABeHRZAIdhpHsEiU2wRBIhUAAFRC6BubRrDYAwPFom5pAAFS-RAqI4lAJQ8RgKOEgJ7HsHDbHkAjIMQABZEJoLsFQyIAoCNN0bRgD7FQVG8GUy0M8xjJcZAQlsFRSkQWprD07RpU2RwLKsqyIEiPNCM8CAQm0RoyVCuTLK86xpE4cw3KMqyjni4yICMVIOBctz7E8rz4l0TNGPoxiAFFxw4bRkJStK9JUErq2Q8KcpUKKYv-aRNiSuD4oOPsDnC3DFN-YA3BXEdbzXY8i1PCChw4EcZvGjcG1POCtOUtSXOQ000WGU1SjMSY6ngYYkGkZCIOAawjFAuw0CMQwbl6vtRgASRMaRdDSPE9IAeSFRSVEgCoCDSpwAYgZAbioCD3rSMw0u4GQVGB3RWFBSI-QMey0mSCBQxBeywiPRaTybFxqVytIZqSDGVGwVhfwc-8IFIeA+kmJ7hn4j63ACsIzCxpzgOu2CqFYFBsGQ39IjqGAVEQYIPtnfr8MG4aD2m0qVrIvF1uF8qtvNXb9sO47QPqjmucpzX0YiLGmZZvp-zqaxpQMMJIg+8w6alio8dBUowhh-862Jyamwt7miMC0xzAcwXqptzHpY4WWwEzMB0zd+kWyAA

### System Info

```Shell
The bug is not related to the system.
```

### Used Package Manager

pnpm

### Validations

GODLiangCY:
### Description

fixes #2225, #1310 

### Additional context

~~Like @imaverickk metioned in #2225, there's another solution: use `Promise.all()`. I'm not sure which one is better~~

> Transforming for-await loops to the configured target environment ("es2017") is not supported yet. Thus `Promise.all()` seems to be the only solution?
---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useFetch/index.test.ts
+++ b/packages/core/useFetch/index.test.ts
@@ -337,6 +337,48 @@ describe('useFetch', () => {
     })
   })
 
+  test('async chained beforeFetch and afterFetch should be executed in order', async () => {
+    const sleep = (delay: number) => new Promise(resolve => setTimeout(resolve, delay))
+
+    const useMyFetch = createFetch({
+      baseUrl: 'https://example.com',
+      options: {
+        async beforeFetch({ options }) {
+          await sleep(50)
+          options.headers = { ...options.headers, title: 'Hunter X Hunter' }
+          return { options }
+        },
+        async afterFetch(ctx) {
+          await sleep(50)
+          ctx.data.message = 'Hunter X Hunter'
+          return ctx
+        },
+      },
+    })
+
+    const { data } = await useMyFetch(
+      'test?json',
+      { method: 'GET' },
+      {
+        async beforeFetch({ options }) {
+          await Promise.resolve()
+          options.headers = { ...options.headers, title: 'Hello, VueUse' }
+          return { options }
+        },
+        async afterFetch(ctx) {
+          await Promise.resolve()
+          ctx.data.message = 'Hello, VueUse'
+          return ctx
+        },
+      },
+    ).json()
+
+    await retry(() => {
+      expect(fetchSpyHeaders()).toMatchObject({ title: 'Hello, VueUse' })
+      expect(data.value).toEqual(expect.objectContaining({ message: 'Hello, VueUse' }))
+    })
+  })
+
   test('should run the onFetchError function', async () => {
     const { data, statusCode } = useFetch('https://example.com?status=400&json', {
       onFetchError(ctx) {
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -216,11 +216,10 @@ function headersToObject(headers: HeadersInit | undefined) {
 
 function chainCallbacks<T = any>(...callbacks: (((ctx: T) => void | Partial<T> | Promise<void | Partial<T>>) | undefined)[]) {
   return async (ctx: T) => {
-    await callbacks.reduce(async (prevCallback, callback) => {
-      await prevCallback
+    await callbacks.reduce((prevCallback, callback) => prevCallback.then(async () => {
       if (callback)
         ctx = { ...ctx, ...(await callback(ctx)) }
-    }, Promise.resolve())
+    }), Promise.resolve())
     return ctx
   }
 }
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -216,10 +216,11 @@ function headersToObject(headers: HeadersInit | undefined) {
 
 function chainCallbacks<T = any>(...callbacks: (((ctx: T) => void | Partial<T> | Promise<void | Partial<T>>) | undefined)[]) {
   return async (ctx: T) => {
-    await Promise.all(callbacks.map(async (callback) => {
+    await callbacks.reduce(async (prevCallback, callback) => {
+      await prevCallback
       if (callback)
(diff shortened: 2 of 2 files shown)


imaverickk:
Promise.all() is parallel, and to ensure that the order of execution is predictable, it should probably look like this.
```ts
function chainCallbacks<T = any>(...callbacks: (((ctx: T) => void | Partial<T> | Promise<void | Partial<T>>) | undefined)[]) {
  return async (ctx: T) => {
    await callbacks.reduce((prevCallback, callback)=>{
      return prevCallback.then(async ()=>{
        if (callback)
          ctx = { ...ctx, ...(await callback(ctx)) }
      })
    }, Promise.resolve())
    return ctx
  }
}
```

GODLiangCY:
> Promise.all() is parallel, and to ensure that the order of execution is predictable, it should probably look like this.
> 
> ```ts
> function chainCallbacks<T = any>(...callbacks: (((ctx: T) => void | Partial<T> | Promise<void | Partial<T>>) | undefined)[]) {
>   return async (ctx: T) => {
>     await callbacks.reduce((prevCallback, callback)=>{
>       return prevCallback.then(async ()=>{
>         if (callback)
>           ctx = { ...ctx, ...(await callback(ctx)) }
>       })
>     }, Promise.resolve())
>     return ctx
>   }
> }
> ```

Oh, you're right! And it makes sense that executing callback in order. Really appreciate for your review!

GODLiangCY:
Sure

wheatjs (approved):
Seem okay to me

PR #2236, opened 2022-09-21
docs(useDark): not about default localStorage key

kcole93:
### Description

This PR clarifies the default behavior of the isDark composable, which persists the current value to localStorage under the key `vueuse-color-scheme`, and fixes a few grammatical points in the remainder of the page for clarity.

### Additional context

While the docs say that this composable reads "the value from localStorage/sessionStorage (the key is configurable) to see if there is a user configured color scheme," it's not illustrated how this key can be configured. I'm not sure how that is done, so I just added information about the default behavior.

---

- [ ] Bug fix
- [ ] New Feature
- [X] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useDark/index.md
+++ b/packages/core/useDark/index.md
@@ -21,7 +21,7 @@ const toggleDark = useToggle(isDark)
 
 ## Behavior
 
-`useDark` combines with `usePreferredDark` and `useStorage`. On start up, it reads the value from localStorage/sessionStorage(the key is configurable) to see if there is a user configured color scheme, if not, it will use users' system preferences. When you change the `isDark` ref, it will update the corresponding element's attribute and then store the preference to storage for persistence.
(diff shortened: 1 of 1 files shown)


packages/core/useDark/index.md
```diff
@@ -37,7 +37,7 @@ By default, it uses [Tailwind CSS favored dark mode](https://tailwindcss.com/doc
 <html class="dark"> ... </html>
 ```
 
-While you can customize it and make it work for most of the CSS frameworks.
+Stillm you can also customize it to make it work with most CSS frameworks.
```
sxzz, line 40: Is `Stillm` a typo?

kcole93: Yes, sorry about that. It should read: "Still,"

sxzz (approved):

PR #2258, opened 2022-09-26
feat(useElementHover): support multiple target

Dunqing:
### Description

Support multiple EventTarget

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useElementHover/index.ts
+++ b/packages/core/useElementHover/index.ts
@@ -1,13 +1,47 @@
 import type { Ref } from 'vue-demi'
-import { ref } from 'vue-demi'
+import { computed, ref, watch } from 'vue-demi'
 import type { MaybeComputedRef } from '@vueuse/shared'
 import { useEventListener } from '../useEventListener'
+import type { MaybeElementRef } from '../unrefElement'
+import { unrefElement } from '../unrefElement'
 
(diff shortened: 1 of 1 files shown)


sxzz (commented):
First of all, it will cause a breaking change.

Also, I don't think this feature is necessary, please let each function be simple and have only one main feature, then you can compose them.

For example, you can write this function by `useElementHover` and [`logicOr`](https://vueuse.org/math/logicOr) by yourself.

Dunqing:
Thank for your review!

> First of all, it will cause a breaking change.

I don't think so, it still supports non-arrays.

> 
> Also, I don't think this feature is necessary, please let each function be simple and have only one main feature, then you can compose them.

This would actually be useful in a dropdown and popover component, making it support array targets would reduce duplicate code

> For example, you can write this function by `useElementHover` and [`logicOr`](https://vueuse.org/math/logicOr) by yourself.

Thanks! Let me see how I can simplify this logic.

antfu:
I don't think the complexity is necessary here, while you can simply call it multiple times when needed. We don't have such a convention to support multiple targets, and this will cause users to ship extra code even if they don't need it. Thanks but I'd reject this.

PR #2265, opened 2022-09-27
feat(useUrlSearchParams): allow custom params delimiter

yassilah:
### Description

This PR allows setting a custom params delimiter instead of the default "&". 

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useUrlSearchParams/index.test.ts
+++ b/packages/core/useUrlSearchParams/index.test.ts
@@ -98,6 +98,34 @@ describe('useUrlSearchParams', () => {
 
         expect(params.customFoo).toEqual(42)
       })
+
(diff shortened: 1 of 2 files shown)


yassilah:
Hi! If it was only up to me I would say no to be honest aha But this has been a requirement for some apps we've built as part of larger ones, when a client wants to ensure the query params don't mess with theirs (which are using default "&"). I just thought I might share my tweak in case that can be useful to others! 😉

antfu (changes requested):
I am not sure about the motivation, it feels a bit too niche. I would prefer to open an custom getter function to pass the params, so user get the chance to transform as want before pass in.

PR #2275, opened 2022-09-29
Make capture mode configurable for useDraggable

mvdbeek:
### Description

I wanted to use multiple nested draggable element, where the selected element is the only element that should move.
I didn't manage to do this because `useDraggable` sets the `useCapture` option to `true`.
I've made the capture mode configurable and added an example to the demo.vue file that lets users drag a child component without also dragging the parent component (please let me know if I missed another way to do this).

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useDraggable/index.ts
+++ b/packages/core/useDraggable/index.ts
@@ -142,7 +142,7 @@ export function useDraggable(target: MaybeComputedRef<HTMLElement | SVGElement |
   }
 
   if (isClient) {
-    const useCapture = options.useCapture || options.useCapture === null
+    const useCapture = options.useCapture ?? true
     useEventListener(draggingHandle, 'pointerdown', start, useCapture)
     useEventListener(draggingElement, 'pointermove', move, useCapture)
     useEventListener(draggingElement, 'pointerup', end, useCapture)
--- a/packages/core/useDraggable/index.ts
+++ b/packages/core/useDraggable/index.ts
@@ -142,7 +142,7 @@ export function useDraggable(target: MaybeComputedRef<HTMLElement | SVGElement |
   }
 
   if (isClient) {
-    const useCapture = options.useCapture ?? true
+    const useCapture = resolveUnref(options.useCapture) ?? true
     useEventListener(draggingHandle, 'pointerdown', start, useCapture)
     useEventListener(draggingElement, 'pointermove', move, useCapture)
     useEventListener(draggingElement, 'pointerup', end, useCapture)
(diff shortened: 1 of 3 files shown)


packages/core/useDraggable/index.ts
```diff
@@ -135,9 +142,10 @@ export function useDraggable(target: MaybeComputedRef<HTMLElement | SVGElement |
   }
 
   if (isClient) {
-    useEventListener(draggingHandle, 'pointerdown', start, true)
-    useEventListener(draggingElement, 'pointermove', move, true)
-    useEventListener(draggingElement, 'pointerup', end, true)
+    const useCapture = options.useCapture || options.useCapture === null
```
ElectronicBlueberry, line 145: This defaults to `false` because `undefined !== null`
Try this instead:
```suggestion
    const useCapture = options.useCapture ?? true
```

packages/core/useDraggable/index.ts
```diff
@@ -135,9 +142,10 @@ export function useDraggable(target: MaybeComputedRef<HTMLElement | SVGElement |
...
+    useEventListener(draggingHandle, 'pointerdown', start, useCapture)
+    useEventListener(draggingElement, 'pointermove', move, useCapture)
+    useEventListener(draggingElement, 'pointerup', end, useCapture)
```
ElectronicBlueberry, line 148: Normalizing `useCapture` fixes the typescript error which occurs because `useEventListener` does not accept a `MaybeComputedRef` as an `options` parameter
```suggestion
    useEventListener(draggingHandle, 'pointerdown', start, resolveUnref(useCapture))
    useEventListener(draggingElement, 'pointermove', move, resolveUnref(useCapture))
    useEventListener(draggingElement, 'pointerup', end, resolveUnref(useCapture))
```


Though note, that this is not reactive. You could either make it reactive here, or make `options` reactive in `useEventListener`, or use a plain optional boolean instead of a `MaybeComputedRef` for the `useCapture` parameter.

mvdbeek: Thanks! I'll go with your suggestion, that seems consistent with how the other optional booleans are handled here.

ElectronicBlueberry: The other optional booleans in this composable are however reactive, as they are normalized within callback functions. I worry that using `MaybeComputedRef` for an option without it being reactive may set false expectations when using the composable.

mvdbeek: I don't think you can change the capture mode of an event listener, so should I go with a plain optional boolean ?

ElectronicBlueberry: Since `useEventListener`s capture mode isn't reactive, I think this composables shouldn't be as well for now. So that approach should be fine.

PR #2276, opened 2022-09-29
feat(useFirestore): support delay for autoDispose, fixes #2252

### LINKED ISSUE #2252: Delay before disconnecting useFirestore

In the case you are using vue-plugin-pages you can have a case that you load the same documents multiple times in case you came back to a page you already loaded once.
But when you change the page the connection is closed at the moment you left the page thanks to:
```javascript
    tryOnScopeDispose(() => {
      close()
    })
```
So you will be charged by one read each time.

### Suggested solution

Using a little delay could alow you to use the cached document instead of get one new.
```javascript
    tryOnScopeDispose(() => {
      useTimeoutFn(() => {
        close()
      }, 30000)
    })
```
You don't need more code because the firestore sdk handle it by default. Even if the first `useFirestore` is closed I still get updates on the new `useFirestore`

### Alternative

An alternative is to store your `useFirestore` in a pinia store but it's lead to many tracks of what you put there.

### Additional context

I can make a PR for that but I don't know about the delay, should it's be lower ? configurable ?
Maybe a force close option ?
Maybe a close if I get an update during the timeout ?

### Validations

Zehir:
### Description
Allow to add a delay on the autoDispose option to allow user to load the same document on a new page just after the document was previously loaded and disposed. This will allow to use the firestore document cache and avoid a Firestore read that cost around €0.06 for 100 000 document's reads and some network usage.

### Additional context
We previously discuss about this feature on the issue #2252 

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other


kiyopikko:
I want the following test 🙏

1. collection or doc dispose
2. `autoDispose: false`
3. `autoDispose: timeout`
4. `autoDispose: timeout` close immediately when new listener comes

Zehir:
@kiyopikko I added test and fix one bug

Zehir:
Do you think we should make a setting for the dispose on read during timeout ?

kiyopikko:
oh sorry, I imagined `B` of the following picture, but this PR is `C`...?

スクリーンショット 2022-10-02 16 35 41

if true, `D` can happen so [this code](https://github.com/vueuse/vueuse/pull/2276/files#diff-c1c3f04ee8dc57220aaea96f3eb292ded6a03ed040024193269da826141da724R115-R120) may not be very effective(and may be less predictable to users).

Zehir:
Yes it's implemented like C but the Firebase SDK is re-using the data from the cache so it's like B.

Yes D can happen that why I think we should add a setting for that that is not enabled by default ?

kiyopikko:
It seems too much configuration so I think this would be sufficient just to be able to set a timeout. (removing [this lines](https://github.com/vueuse/vueuse/pull/2276/files#diff-c1c3f04ee8dc57220aaea96f3eb292ded6a03ed040024193269da826141da724R117-R128))

1. I'm wondering if how many users want to do this optimization.
2. That option is complex and difficult to explain.

Simple and configurable is trade-off, `useFirestore` is **one of the vueuse utility collection**, I don't know if it's a good idea to have such a variety of options. Maybe a library like `vue-use-firestore`(fictional) will support that.

This is just my opinion and I'm just a contributor. 
Only the vueuse members have the philosophy. It seems better to let them be the judge

Zehir:
Or we add a `control` option like in `useTimestamp` to expose the stop option ?

kiyopikko:
@Zehir 

> Or we add a control option like in useTimestamp to expose the stop option ?


https://github.com/vueuse/vueuse/pull/2276/files#diff-c1c3f04ee8dc57220aaea96f3eb292ded6a03ed040024193269da826141da724R101

Zehir:
In the `useTimestamp` have an option that modify the returned data. Instead of having only the data we will have an object with the data in a data key, a stop method, and anything else we add down the road.

Yes I will make a PR to fix the issue

Zehir:
@kiyopikko any chances to get this merge ?

kiyopikko:
@Zehir 

As I wrote in https://github.com/vueuse/vueuse/pull/2276#issuecomment-1264963235, please remove `autoDisposingReadLimit`?

> It seems too much configuration so I think this would be sufficient just to be able to set a timeout. (removing [this lines](https://github.com/vueuse/vueuse/pull/2276/files#diff-c1c3f04ee8dc57220aaea96f3eb292ded6a03ed040024193269da826141da724R117-R128))

Zehir:
@kiyopikko I removed the autoDisposingReadLimit option

kiyopikko (approved):
LGTM!

Zehir:
Anybody can merge this PR please ?

antfu (approved):

PR #2294, opened 2022-10-05
docs(useFocusTrap): improve the documentation

agenordebriat:
Even though `useFocusTrap` is categorized as an integration and described to be a "reactive wrapper for [focus-trap](https://github.com/focus-trap/focus-trap).", I think it'd be better to specify that you need to install `focus-trap` as a dependency.
Also, other integrations do this, just before the "Usage" section.


antfu (approved):

antfu:
We could probably do this also for other functions in the integrations package.

agenordebriat:
I can take a look at the others and update my PR.

agenordebriat:
Did a similar update on `useNProgress`. All the integrations now follow the same documentation pattern.

sxzz (approved):

PR #2295, opened 2022-10-06
feat(useTimeOfDay): function to determine time of day reactively

therealokoro:
### Description

This pull request proposes a new function that helps calculate the time of day reactively. It determines if the current time is `Morning`, `Afternoon`, `Noon`, `Evening` or `Night`.

It also takes an optional first parameter which can be a ref or a plain instance of the `Date` object. This will then be used as the base of computing the time of the day. If not provided it falls back to the current time which is computed using `useNow` function from the core library, meaning that the output is always reactive.

I find this as a useful addition to the library as it could help developers who like to make personified greetings do this a little bit easier.

Thank you🙂.

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


therealokoro:
Very well understood @antfu. I thought about i18n support but felt I should this out and probably get help on that and most especially locale support. So what do you suggest sir.

antfu:
I consider this is not general enough. I would suggest leaving it to users to implement on their side to fit their needs. Thanks.

PR #2309, opened 2022-10-10
feat(createInjectionState): return state when providing injection state

### LINKED ISSUE #1997: createInjectionState provide should return composable result

When using `createInjectionState`, there is no way of using the provided data within the provider. If `useProvidingState` would return the state it provides, it can be used in the provider as well.

```ts
const [useProvideState, useState] = createInjectionState(() => {
  const myState = {}
  return myState
})

// App.vue
defineComponent({
  setup () {
    useProvideState() // I want to use myState now...
  }
})

```

### Suggested solution

In [this file](https://github.com/vueuse/vueuse/blob/main/packages/shared/createInjectionState/index.ts#L14) the code could look as follows:

```ts
  const useProvidingState = (...args: Arguments) => {
    const result = composable(...args)
    provide(key, result)
    return result
  }
```

### Alternative

My workaround for now is (my code includes automatically asserted injections):

```ts
import { createInjectionState } from '@vueuse/core'

export function assertInjection<Return> (injection: () => Return): NonNullable<Return> {
  const value = injection()
  if (value == null) {
    throw new Error('Injection was called before the value was provided by a parent.')
  }
  return value!
}

export function assertCreateInjectionState<Arguments extends Array<any>, Return> (
  composable: (...args: Arguments) => Return,
): readonly [
  useProvidingState: (...args: Arguments) => Return,
  useInjectedState: () => Return
  ] {
  let context: Return
  const [provide, inject] = createInjectionState((...args: Arguments) => {
    context = composable(...args)
    return context
  })
  return [
    (...args) => {
      provide(...args)
      return context
    },
    () => assertInjection(inject),
  ]
}
```

### Additional context

_No response_

### Validations

Tanimodori:
### Description

Returns the `state` created in `useProvidingState`, fix #1997

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/shared/createInjectionState/index.ts
+++ b/packages/shared/createInjectionState/index.ts
@@ -8,7 +8,7 @@ import { type InjectionKey, inject, provide } from 'vue-demi'
  */
 export function createInjectionState<Arguments extends Array<any>, Return>(
   composable: (...args: Arguments) => Return,
-): readonly [useProvidingState: (...args: Arguments) => void, useInjectedState: () => Return | undefined] {
+): readonly [useProvidingState: (...args: Arguments) => Return, useInjectedState: () => Return | undefined] {
   const key: string | InjectionKey<Return> = Symbol('InjectionState')
   const useProvidingState = (...args: Arguments) => {
     const state = composable(...args)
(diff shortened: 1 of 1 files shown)


packages/shared/createInjectionState/index.ts
```diff
@@ -11,7 +11,9 @@ export function createInjectionState<Arguments extends Array<any>, Return>(
 ): readonly [useProvidingState: (...args: Arguments) => void, useInjectedState: () => Return | undefined] {
```
lsdsjy, line 11:
```suggestion
): readonly [useProvidingState: (...args: Arguments) => Return, useInjectedState: () => Return | undefined] {
```

lsdsjy (changes requested):

xiaoxiangmoe:
LGTM

PR #2312, opened 2022-10-11
feat(useTextareaAutosize): added `styleTarget` option to style other element

jd-solanki:
### Description

When `detachStyles` is `true` styles won't get applied to the textarea. You can use `textareaScrollHeight` ref returned by this composable to apply the height instead. This is useful in cases where we have custom UI for textarea as in [Anu](https://anu-vue.netlify.app/guide/components/textarea.html) or any other textarea customization.

Related PR in Anu: https://github.com/jd-solanki/anu/pull/37

With this PR merged we can set the height of textarea wrapper based on the height of the actual textarea content.

### Additional context

This is just not specific to Anu, this is useful even if someone has a custom UI for textarea.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


jd-solanki:
Hi @antfu 
According to your [comment](https://discord.com/channels/937808017016119440/937969993964978197/1034707254961782814) in discord, I have provided `styleTarget` option instead of `detachStyles`.

This was a nice improvement that reduces # of lines. Thanks for the suggestion.

`demo.vue` code if you want to play:
```vue
<script setup lang="ts">
import { useTextareaAutosize } from '@vueuse/core'
import { ref } from 'vue'

// const { textarea, input } = useTextareaAutosize()

// --

const wrapper = ref()
const textareaCustom = ref()
const { input: wrapperInput } = useTextareaAutosize({
  element: textareaCustom,
  styleTarget: wrapper,
})
</script>

<template>
  <div>
    <span>Type, the textarea will grow:</span>
    <div ref="wrapper" class="textarea-container outline outline-color-[tomato]">
      <textarea ref="textareaCustom" v-model="wrapperInput" class="!m-0 w-full h-full !shadow-none" />
    </div>
    {{ wrapperInput }}
    <!-- <textarea
      ref="textarea"
      v-model="input"
      class="resize-none"
      placeholder="What's on your mind?"
    /> -->
  </div>
</template>

```

packages/core/useTextareaAutosize/index.ts
```diff
@@ -11,18 +11,32 @@ export interface UseTextareaAutosizeOptions {
...
+    textareaScrollHeight.value = textarea.value?.scrollHeight
+
+    // If style target is provided update its height
+    if (options?.styleTarget)
+      unref(options.styleTarget).style.height = `${textareaScrollHeight.value}px`
+
+    // else update textarea's height by updating height variable
+    else height = `${textareaScrollHeight.value}px`
+
+    textarea.value!.style.height = height
```
antfu, line 39:
```suggestion
    textareaScrollHeight.value = textarea.value?.scrollHeight

    const styleTarget = unref(options?.styleTarget) || textarea.value
    styleTarget!.style.height = height
```

Do you mean by this? In the previous implementation `textarea`'s style seems always applied (L39)

jd-solanki:
Hi @antfu 

Thanks for your review. I tried your suggested changes but it isn't working because we have to reset the style of textarea so textarea can be rendered as normal (_like height isn't modified_).

Screenshot 2022-11-07 at 6 18 46 PM

PR #2323, opened 2022-10-14
feat(useElementHover): support delay for hover

WX-DongXing:
### Description

Support delay for useElementHover, sometimes we don't want the levitation event to trigger too sensitively, so it may be useful to add a delay option.

**[Demo](https://codesandbox.io/p/sandbox/useelementhover-viy481)**
In this scnece, if want to select pane C, the mouse may move over pane B and causing display. After adding an appropriate delay time, the pane you want to select can be displayed accurately.

https://user-images.githubusercontent.com/25719069/195824046-07ca076c-b119-40df-8e80-ac1dc66ebb4c.mov

**Use**
```ts
const el = ref<HTMLButtonElement>()

const isHovered = useElementHover(el, { delay: 500 })
```

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useElementHover/demo.vue
+++ b/packages/core/useElementHover/demo.vue
@@ -3,11 +3,18 @@ import { ref } from 'vue'
 import { useElementHover } from '@vueuse/core'
 
 const el = ref<HTMLButtonElement>()
+const elOptions = ref<HTMLButtonElement>()
+
 const isHovered = useElementHover(el)
+const isHoveredOptions = useElementHover(elOptions, { delay: 500 })
 </script>
 
 <template>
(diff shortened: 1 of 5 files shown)


antfu (commented):
How about we introduce a more general `delayedRef` so we could "compose" them as needed, while keep this function simple and light?

WX-DongXing:
@antfu Hi, Just like the delayed behavior trigger needs to remain suspended above the element for a period of time, I am a little confused about how the general `delayedRef` meets different conditions.

PR #2325, opened 2022-10-14
fix(useDevicePixelRatio): remove unnecessary cleanup

Waleed-KH:
### Description

In `useDevicePixelRatio` function, we don't have to remove the listener with every invoke, because we already set `once: true`, we only need to remove the last listener observed on dispose

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/useDevicePixelRatio/index.ts
```diff
@@ -19,25 +19,17 @@ export function useDevicePixelRatio({
...
-    cleanups.map(i => i())
-    cleanups.length = 0
-  }
-
+  let media: MediaQueryList
   const observe = () => {
     pixelRatio.value = window.devicePixelRatio
-    cleanup()
-    const media = window.matchMedia(`(resolution: ${pixelRatio.value}dppx)`)
+    media = window.matchMedia(`(resolution: ${pixelRatio.value}dppx)`)
```
antfu, line 25: Maybe we could do a clean up for the previous event then?

PR #2333, opened 2022-10-18
feat(useFetch): allow configure lifecycle handler behavior

climba03003:
### Description

Allow to configure the lifecycle handler behavior before `createFetch` and `useFetch`.

Address my https://github.com/vueuse/vueuse/pull/1545#issuecomment-1118193827

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -1,6 +1,6 @@
-import type { ComputedRef, Ref } from 'vue-demi'
 import type { EventHookOn, Fn, MaybeComputedRef, Stoppable } from '@vueuse/shared'
 import { containsProp, createEventHook, resolveRef, resolveUnref, until, useTimeoutFn } from '@vueuse/shared'
+import type { ComputedRef, Ref } from 'vue-demi'
 import { computed, isRef, ref, shallowRef, watch } from 'vue-demi'
 import { defaultWindow } from '../_configurable'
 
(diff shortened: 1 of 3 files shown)


packages/core/useFetch/index.ts
```diff
@@ -141,8 +142,8 @@ export interface UseFetchOptions {
 
   /**
    * Will automatically refetch when:
-   * - the URL is changed if the URL is a ref
-   * - the payload is changed if the payload is a ref
+   * - the URL is combineCallbacksnged if the URL is a ref
```
wheatjs, line 145: Is `combineCallbacksnged` a typo?

packages/core/useFetch/index.ts
```diff
@@ -90,6 +90,7 @@ export interface UseFetchReturn<T> {
 
 type DataType = 'text' | 'json' | 'blob' | 'arrayBuffer' | 'formData'
 type HttpMethod = 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH' | 'HEAD' | 'OPTIONS'
+type Combination = 'overwrite' | 'combineCallbacksin'
```
wheatjs, line 93: I'm wondering if `chain` would not be a better word here than `combineCallbacksin`

climba03003: I means to be combine. It turns out pushed a strange word.

packages/core/useFetch/index.ts
```diff
@@ -187,6 +188,12 @@ export interface CreateFetchOptions {
    */
   baseUrl?: MaybeComputedRef<string>
 
+  /**
+   * Determine the inherit behavior for beforeFetch, afterFetch, onFetchError
+   * @default 'combineCallbacksin'
+   */
+  combination?: Combination
```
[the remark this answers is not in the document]
climba03003: Yes, I hope it should be `overwrite` but I want to prevent breaking changes.

wheatjs (approved):

PR #2335, opened 2022-10-18
feat(useIDBKeyval): new integration - Idb-keyval wrapper

### LINKED ISSUE #230: useIndexedDb / useIDB
Any plans for adding a IndexedDb utility ?

Harmony222:
https://user-images.githubusercontent.com/55414687/196748412-f2e15e1b-09c4-41c9-b4fd-e1b9012ea0a0.mov

### Description

Provides a composition wrapper around [idb-keyval](https://www.npmjs.com/package/idb-keyval), which is a key value store api on top of Indexed DB. Gives you typed keys and values without the need of a serializer, adds web worker support, and lets you use more storage space when compared to using the local storage variant.

### Additional context

This partially addresses [issue #230](https://github.com/vueuse/vueuse/issues/230), but does not add full support for indexed DB (just for typed keys and values).
closes #230 

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/integrations/useIDBKeyval/demo.vue
```diff
@@ -0,0 +1,24 @@
...
+
+<template>
+  <div>
+    <input v-model="state.name" type="text">
+    <input v-model="state.color" type="text">
+    <input v-model="state.size" type="text">
+    <input v-model.number="state.count" type="range" min="0" step="0.01" max="1000">
+
+    <pre lang="json">{{ text }}</pre>
+  </div>
```
JessicaSachs, line 23: BTW you don't need this `div` wrapper in Vue 3.

packages/integrations/useIDBKeyval/index.ts
```diff
@@ -0,0 +1,87 @@
...
+   * Default log error to `console.error`
+   */
+  onError?: (error: unknown) => void
+
+  /**
+   * Use shallow ref as reference
+   *
+   * @default false
+   */
+  shallow?: boolean
```
JessicaSachs, line 27: Can we remove this option and the associated code by having the user pass in a computed, ref, or shallowRef?

JessicaSachs (commented):
Needs tests still. I'm happy to write them.

packages/integrations/package.json
```diff
@@ -151,6 +156,7 @@
     "drauu": "^0.3.2",
     "focus-trap": "^7.0.0",
     "fuse.js": "^6.6.2",
+    "idb-keyval": "^6.2.0",
```
sxzz, line 159: It also requires to be added to peer deps

sxzz (commented):
I think we could reuse `useStorage` and implement [`StorageLike`](https://github.com/vueuse/vueuse/blob/d0dffc6f514f8c0f4b163d6c8c2f8835d1f68913/packages/core/ssr-handlers.ts#L9) to avoid repeating code. Just like `useLocalStorage` and `useSessionStorage`.

https://github.com/vueuse/vueuse/blob/d0dffc6f514f8c0f4b163d6c8c2f8835d1f68913/packages/core/useLocalStorage/index.ts#L26

sxzz:
@JessicaSachs I think we could mock functions of `idb-keyval`. Since there are unit tests on `idb-keyval`, we can believe it always works fine. It's easier than mocking `IndexDB`.

ping @antfu Is this function necessary and acceptable?

sxzz:
By the way, the keyword `closes #pr_number` needs to be added to the PR body for referencing the related issues.

packages/integrations/useIDBKeyval/index.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    expect(state.value.name).toBe('Apple')
+    expect(state.value.color).toBe('Red')
+    expect(state.value.size).toBe('Giant')
+    expect(state.value.count).toBe(1)
+  })
+
+  it('del', async () => {
+    const state = useIDBKeyval(KEY, defaultState)
+    state.value = null
+    await nextTick()
```
JessicaSachs, line 47: Did you end up needing to do this? Just curious.

packages/integrations/useIDBKeyval/index.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  read()
+
+  async function write() {
+    try {
+      if (data.value == null)
+        await del(key)
+      else
+        await update(key, () => ({ ...data.value }))
```
JessicaSachs, line 77: Question, did this have the same issue that you were messaging me about? The issue with `...someString` etc?

packages/integrations/useIDBKeyval/index.test.ts
Harmony222: yes! it did not work without it

packages/integrations/useIDBKeyval/index.ts
Harmony222: yeah looks like that one is having an issue too. This is what jimmy and I added when we worked through the bug the other day, it had something to do with proxy objects not able to be stored in index db if I remember correctly.

packages/integrations/package.json
Harmony222: added to peer deps

packages/integrations/useIDBKeyval/index.ts
Harmony222: changed this logic to only make copy with spread syntax if data.value is an object or array

Harmony222:
hi @sxzz the contributor generation script is failing

sxzz:
There are some TS type error

Harmony222:
Thank you @sxzz, I have addressed the TS type error.

JessicaSachs (approved):
LGTM and the types are now fixed. Once we start using Vitest's browser mode, I think this test will be a good candidate for the browser environment and we can remove the `fake-indexeddb` module + polyfill

sxzz:
@JessicaSachs @Harmony222 
https://github.com/vueuse/vueuse/pull/2335#pullrequestreview-1149759772

packages/integrations/package.json
Harmony222: Regarding the first comment, thank you for your patience as I looked into this.
I looked through `useLocalStorage` and `useSessionStorage` to understand how they reuse `useStorage`. I believe you are suggesting that we could use indexDB storage in place of `window?.localStorage`, the problem I see is that `useStorage` converts all values to strings, and a big reason our team wanted to use indexDB is so we can store values other than just strings. 
I also took a closer look at `StorageLike`, its get and set functions get and return strings, so I am not sure that using indexDB in this way would provide any gains to using `localStorage` as is.
Please let me know if I am misunderstanding something.

Harmony222:
@sxzz 
Reposting my comment here as it may have gotten lost. 
This is in regards to reusing useStorage. 
I believe you are suggesting that we could use indexDB storage in place of `window?.localStorage`, the problem I see is that `useStorage` converts all values to strings, and our team wants to use indexDB so we can store values other than just strings.
I also took a closer look at `StorageLike`, its get and set functions get and return strings, so I am not sure that using indexDB in this way would provide any gains to using localStorage as is.
Please let me know if I am misunderstanding something.

antfu (approved):

antfu:
I am fine to have this as-is for now. Later we could see how we can refactor this with `useStorageAsync`

PR #2336, opened 2022-10-19
feat(useClipboard): support legacy copy

lumdzeehol:
### Description
Add `legacy: boolean` option. 
It will fallback to legacy methods `document.execCommand('copy')` if ClipboardApi is not allowed, for scenes like hosting website with IP adress, especially during development.
Relate to #1276

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useClipboard/index.ts
+++ b/packages/core/useClipboard/index.ts
@@ -2,8 +2,8 @@
 
 import type { MaybeComputedRef } from '@vueuse/shared'
 import { resolveUnref, useTimeoutFn } from '@vueuse/shared'
-import { computed, ComputedRef, Ref } from 'vue-demi'
-import { ref } from 'vue-demi'
+import type { ComputedRef, Ref } from 'vue-demi'
+import { computed, ref } from 'vue-demi'
 import type { WindowEventName } from '../useEventListener'
 import { useEventListener } from '../useEventListener'
 import { useSupported } from '../useSupported'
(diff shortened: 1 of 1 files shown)


sxzz (commented):
IMHO, I don't think this feature is necessary. `document.execCommand` is supported in all modern browsers. https://caniuse.com/document-execcommand
Also, it will cause a breaking change if `isSupported` is changed.

If Clipboard API is not supported, then you can handle fallback by yourself.

lumdzeehol:
> IMHO, I don't think this feature is necessary. `document.execCommand` is supported in all modern browsers. https://caniuse.com/document-execcommand Also, it will cause a breaking change if `isSupported` is changed.
> 
> If Clipboard API is not supported, then you can handle fallback by yourself.

I think you're right. useClipboard should just focus on clipboardApi.
but what if i want a compatible use function to handle copy? is it necessary to create a use like `useCopy` or sth ? for that execCommand needs lots of template code everytime i code it.😢

sxzz:
Yeah, I think it's a good idea, to combine Clipboard API and `document.execCommand` into one function called `useCopy`. WDYT @antfu

antfu (approved):

antfu:
LGTM. Do you mind to send another PR to the document mention about this? Thanks

PR #2337, opened 2022-10-19
feat(refDebounced): use MaybeComputedRef for ms

dominikklein:
### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other (Type improvement)

### DIFF
--- a/packages/shared/refDebounced/index.ts
+++ b/packages/shared/refDebounced/index.ts
@@ -1,14 +1,14 @@
 import type { Ref } from 'vue-demi'
 import { ref, watch } from 'vue-demi'
 import { useDebounceFn } from '../useDebounceFn'
-import type { DebounceFilterOptions } from '../utils'
+import type { DebounceFilterOptions, MaybeComputedRef } from '../utils'
 
 /**
  * Debounce updates of a ref.
  *
  * @return A new debounced ref.
  */
-export function refDebounced<T>(value: Ref<T>, ms = 200, options: DebounceFilterOptions = {}): Readonly<Ref<T>> {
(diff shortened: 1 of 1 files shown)


packages/shared/refDebounced/index.ts
```diff
@@ -1,14 +1,14 @@
...
+import type { DebounceFilterOptions, MaybeComputedRef } from '../utils'
 
 /**
  * Debounce updates of a ref.
  *
  * @return A new debounced ref.
  */
-export function refDebounced<T>(value: Ref<T>, ms = 200, options: DebounceFilterOptions = {}): Readonly<Ref<T>> {
+export function refDebounced<T>(value: Ref<T>, ms: MaybeComputedRef<number> = 200, options: DebounceFilterOptions = {}): Readonly<Ref<T>> {
   if (ms <= 0)
```
wheatjs, line 12: In this case, we should change this line to be `if(unref(ms) <= 0))` or just remove it all together. Currently if the initial value is `0`, you won't be able to change it later which is probably not the desired behavior.

dominikklein: Ah, correct, I think when currently a `0` is used the complete usage of the `refDebounced` make no sense or?
So I think it makes sense to remove this check and support a reactive value. At least it's already working, only the type will not support it.

wheatjs: Yep, so I suggest we just remove the check all together

wheatjs:
Looks good to me, let's get a review from Anthony before merging

antfu (approved):

Issue #2345, opened 2022-10-21
Adding v-visible directive (and component/hook)


Blackfaded:
### Clear and concise description of the problem

Hey there!

Sometimes v-if and v-show are not enough. The v-if directive removes an element from the DOM and the v-show directive sets the `display` value to `none`. Both of them come with the caveat that the element does not take any space in the DOM anymore.
As a developer I often need the flexibility to hide an element, but keep their space in the DOM.

### Suggested solution

In module useVisible we could provide the following implementation:

As a hook
```ts
useVisible(elementRef, flag)
```

As a directive
```vue
<div v-visible="flag">Hello World!</div>
```

As a component
```vue
<Visible :show="flag">Hello World!</Visible>
```

### Alternative

The only alternative to implement this so far is to give the DOM element a class and set the css to `visibility: hidden`

### Additional context

![Bildschirmfoto 2022-10-21 um 09 09 33](https://user-images.githubusercontent.com/19570581/197134752-b2b86528-0e4e-4ec1-8f91-ecc83a474ec6.png)

I would love to open a PR if this feature is needed in this library

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

ywenhao:
I think this is great!

PR #2354, opened 2022-10-24
feat(useVisible): implement useVisible functionality

Blackfaded:
### Description

Implementation of [this](https://github.com/vueuse/vueuse/issues/2345) proposal

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
How is this defer from `useElementVisibility`?

Blackfaded:
`useElementVisibility` indicates if an element is inside the window or scrollTarget specified. It does not allow to set the visibility of an element.

This PR adds support for hiding an element, which has nothing to do with the viewport.

antfu:
I am hesitant about:
- If the main usage is directive like `v-visible, it might be better to be provided as a separate package instead of in VueUse
- We should have a more distinguishable name compare to `useElementVisibility`

Blackfaded:
Valid points. I always found the name `useElementVisibility` misleading, because it does not state the scope like the viewport. But changing this name would break backwars compatability, ao it is no option.


But how about the reverse approach:


`useHideElement`

`v-hidden`

`<Hidden :value="true" />`

?

Issue #2278, opened 2022-09-30
[useScrollLock] cause enexpected behavior on Safari besides other browsers


Mister-Hope:
### Describe the bug

Related to https://github.com/vueuse/vueuse/issues/1558

My case is that I am locking body to prevent page scroll, while trying to like an fix modal be able to scroll.

I am building a feature extactly like vitepress. See NavScreen Component:

![image](https://user-images.githubusercontent.com/33315834/193261090-acc000d6-7ca6-4e1d-91d3-ef7ab10cf480.png)

This `NavScreen` component will lock body scroll, and try to let the modal scrollable. And vitepress itself is not using `@vueuse`, instead it use `body-scroll-lock` package. (See: https://github.com/vuejs/vitepress/blob/main/src/client/theme-default/components/VPNavScreen.vue)

While I am trying to build a similar layout with `useScrollLock`, I am getting the same behavior with Chrome. But on Safari, the NavScreen Component is not scrollable:

![image](https://user-images.githubusercontent.com/33315834/193261435-0d70952a-9b24-409b-a268-49ae611a3b73.png)

You can try https://vuepress-theme-hope.github.io/v2/

I am sure the issue only appears in safari, and:

- Switching vitepress NavScreen component to `useScrollLock` has same issues on Safari
- Using `body-scroll-lock` in my theme works well on Safari like vitepress default theme

---

Or more specificly:

- when in Chrome based browers, the `useScrollLock` only lock that elements
- when in safari, the `useScrollLock` locks elment and any elements in it.

I think this should be consider as a bug, and if this is not fixable, I do think we'd better add some warnings in docs.

### Reproduction

https://github.com/vuepress-theme-hope/vuepress-theme-hope

### System Info

```Shell
I can provide a real "minimal reproduction" and system info if you do need this.
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #2362, opened 2022-10-25
fix(useScrollLock): fix iOS touchmove bug

robinscholz:
### Description

This PR solves iOS touchmove issues that prevent any children elements of the locked element from scrolling. I believe this will solve #2278.

### Additional context

I believe there are no breaking changes, however it'd be good to confirm this with a few different devices.

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


robinscholz:
Not sure why the test is failing, should have nothing to do with the proposed changes

webfansplz (dismissed):

robinscholz:
@webfansplz After some more tests I ran into problems with my original solution. I believe the only way to truly solve this is by checking for `overflow: scroll` within the `event.target` and its’ parentNodes. 

Caveat is that the user needs to explicitly set the overflow to scroll to keep elements scrollable on iOS.

robinscholz:
@antfu would be happy to get your opinion on this implementation!

dpschen (dismissed):

David-Tsui:
@robinscholz This problem still exists at `9.10.0`, test by IOS devices in chrome emulator. I found a way to fix is removing the touchmove event listener added to IOS.

Issue #2361, opened 2022-10-25
feat(useSafeArrayIndex): new function


okxiaoliang4:
### Clear and concise description of the problem

Make sure the index does not exceed the range of the array.

### Suggested solution

```ts
export function useSafeArrayIndex<T>(arr: MaybeComputedRef<T[]>, initinal: MaybeRef<number> = 0) {
  return useClamp(initinal, 0, () => resolveUnref(arr).length - 1)
}
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #2366, opened 2022-10-26
feat(useSafeArrayIndex): new function

okxiaoliang4:
### Description

Make sure the index does not exceed the range of the array.

### Additional context

closes: #2361
---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
I am not so sure. It's hard to define what is "Safe" for the expected behavior. Like should `index.value = -1` be the first one or the last one? I would refer to not having this and letting users to compose this with `useClamp`.

PR #2380, opened 2022-10-30
fix: using proper variable names

Amorites:


Amorites:
BTW, I noticed demos in `rxjs` are named `_demo.vue` and would not show on the docs site. What are the obstacles to showing the demo?

PR #2386, opened 2022-10-31
fix(useColorMode): remove `tryOnMounted`

### LINKED ISSUE #2035: useColorMode bugs
### Describe the bug

Hi, i found few bugs in useColorMode. 

First bug on page refresh flashes white background even i on Light or Dark or Auto mode.  
You can test on page [https://vueuse.org/core/usecolormode/](https://vueuse.org/core/usecolormode/) try set system to dark mode set page to Light mode and refresh page, page will flash Dark mode then loads Light mode and this work too vice versa

Second bug as i see in LocalStorage do not changing **vueuse-color-scheme** only auto do not changing. 

In older update if i remember about 8.8.0 version working fine as LocalStorage and no Flashing screen. 
Tested on all browsers as in System info

Laravel v 9.22.1

My code 
```
import { useColorMode } from "@vueuse/core";

const mode = useColorMode();

const getTheme = ref(localStorage.getItem("vueuse-color-scheme"));

const themes = [{ mode: "light" }, { mode: "dark" }, { mode: "auto" }];

function setTheme(id) {
    if (getTheme.value === themes[id].mode) {
        return;
    }
    mode.value = themes[id].mode;
    getTheme.value = themes[id].mode;
}

```

https://user-images.githubusercontent.com/42121280/182602557-8b9ba2f5-8241-43b0-a750-19c9585132b9.mov

https://user-images.githubusercontent.com/42121280/182603022-58eeea8b-03ab-4651-b79d-7670f7f9b691.mov

### Reproduction

https://vueuse.org/core/usecolormode/

### System Info

```Shell
System:
    OS: macOS 12.4
    CPU: (8) arm64 Apple M1
    Memory: 140.13 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    npm: 8.5.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 103.0.5060.134
    Firefox: 103.0.1
    Safari: 15.5
    Safari Technology Preview: 16.0
  npmPackages:
    @vueuse/components: ^9.0.2 => 9.0.2 
    @vueuse/core: ^9.0.2 => 9.0.2 
    @vueuse/electron: ^9.0.2 => 9.0.2 
    @vueuse/firebase: ^9.0.2 => 9.0.2 
    @vueuse/integrations: ^9.0.2 => 9.0.2 
    @vueuse/metadata: ^9.0.2 => 9.0.2 
    vue: ^3.2.31 => 3.2.37
```

### Used Package Manager

npm

### Validations

GODLiangCY:
### Description

fixes #2035 

See #2035, when page refreshed, `useColorMode` cannot work immediately. This could be solved by removing `  tryOnMounted(() => onChanged(state.value))`

### Additional context

NO

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useColorMode/index.ts
+++ b/packages/core/useColorMode/index.ts
@@ -1,6 +1,5 @@
 import type { Ref } from 'vue-demi'
 import { computed, ref, watch } from 'vue-demi'
-import { tryOnMounted } from '@vueuse/shared'
 import type { StorageLike } from '../ssr-handlers'
 import { getSSRHandler } from '../ssr-handlers'
 import type { UseStorageOptions } from '../useStorage'
@@ -162,8 +161,6 @@ export function useColorMode<T extends string = BasicColorSchema>(options: UseCo
   watch(state, onChanged, { flush: 'post', immediate: true })
   if (emitAuto)
     watch(preferredMode, () => onChanged(state.value), { flush: 'post' })
-
-  tryOnMounted(() => onChanged(state.value))
 
   return state
 }


BenaXskope:
If you'll try to build docs and serve it, you'll notice, that bug is still reproducing.
I think it's caused when SSR/SSG is enabled in the project. Because on the server side you may not to have (if u use SSR) or exactly doesn't have any information about user selected theme, so you generate the page with default theme.
When the client gets the page, it starts hydrating. And the time between HTML rendered and the components hydrated (at this moment user selected theme is applied) is the reason of this flashing

Arecsu:
> If you'll try to build docs and serve it, you'll see that this "bug" is still reproducing. I think it's caused when SSR/SSG is enabled in the project. Because on the server side you may not to have (if u use SSR) or exactly doesn't have any information about user selected theme, so you generate the page with default theme. When the client gets the page, it starts hydrating. And the time between HTML rendered and the components hydrated (at this moment user selected theme is applied) is the reason of this flashing So i don't think it's problem of the composable

I think I have this exact problem when using useDark or useColorMode. I have SSR enabled. No matter where I use those functions (in a header component or app.vue), it will always produce a flash with the initial default colors of the web, and then the saved theme value in local storage. 

In the inspector DevTools, I can see it sets the value from useDark/useColorMode a bit too late and what you've mentioned seems to be the case. Is there a workaround to prevent this flashing from happening in SSR and set the correct values in the DOM before the initial render?

PR #2394, opened 2022-11-02
feat: add reactive window.scrollTo watcher to useWindowScroll

Solant:
### Description

Add watcher with `window.scrollTo` callback for exported x and y refs, so anyone can make whole page scroll with already exported refs

### Additional context

Currently, exported `x`, and `y` are simple writable refs, so I can write to them without any warning and with no side effect. It makes sense to either make them read-only (computed) or just add `scrollTo` function to handle window scroll on change

Things that are not covered in this PR:
 - scroll behavior (https://vueuse.org/core/useScroll/#smooth-scrolling)

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useWindowScroll/demo.vue
+++ b/packages/core/useWindowScroll/demo.vue
@@ -2,6 +2,11 @@
 import { useWindowScroll } from '@vueuse/core'
 
 const { x, y } = useWindowScroll()
+
+function scrollToTop() {
+  x.value = 0
+  y.value = 0
+}
 </script>
 
 <template>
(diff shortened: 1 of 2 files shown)


Solant:
@webfansplz thanks

Should I close this PR as it is duplicate/deprecated functionality (e.g. migration from `useWindowScroll` to `useScroll`), or I update PR to just `useScroll` with `window` parent, to make things work seamlessly?

webfansplz:
> @webfansplz thanks
> 
> Should I close this PR as it is duplicate/deprecated functionality (e.g. migration from `useWindowScroll` to `useScroll`), or I can update PR to call `useScroll` with `window` parent, to make things work seamlessly?

I think we can just close it, anyway, thanks for the PR.

PR #2405, opened 2022-11-06
refactor(onClickOutside): better safari workaround

sibbng:
### Description

Better fix for #1520. I didn't read the article I mentioned in #1522 carefully. It seems there is a better fix for it. 

Also, it seems this issue is fixed in the latest version of Safari.

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other


packages/core/onClickOutside/index.ts
```diff
@@ -66,18 +62,13 @@ export function onClickOutside<T extends OnClickOutsideOptions>(
   }
 
   const cleanup = [
+    [].slice.call(window.document.body.children)
+      .map(el => useEventListener(el, 'click', () => {})),
```
antfu, line 66: This feels too expensive and I am not sure about it.

sibbng: It's not. But we could optimize it. Currently, these event handlers will be attached on each use of `onClickOutside`. We could make it a one-time operation, initialize it only on iOS, and filter out `<script>`, `<link>` and `<style>` elements.

sibbng:
A quick note about why I use `addEventListener` instead of `useEventListener`. I don't want these listeners to be removed when scope disposed as they are attached only on the first use of `onClickOutside`. Since we don't need to remove listeners on scope disposal relying on `addEventListener` removes the overhead of `useEventListener` and makes the memory footprint smaller.

packages/core/onClickOutside/index.ts
```diff
@@ -43,8 +45,8 @@ export function onClickOutside<T extends OnClickOutsideOptions>(
   if (!window)
     return
 
-  if (!(window as any).__VUEUSE_ONCLICKOUTSIDE__ && /iPad|iPhone|iPod/.test(navigator.userAgent)) {
-    (window as any).__VUEUSE_ONCLICKOUTSIDE__ = true
+  if (!_iOSWorkaround && isIOS) {
```
antfu, line 48:
```suggestion
  if (isIOS && !_iOSWorkaround) {
```

antfu (approved):

packages/core/onClickOutside/index.ts
```diff
@@ -42,9 +45,13 @@ export function onClickOutside<T extends OnClickOutsideOptions>(
   if (!window)
     return
 
-  let shouldListen = true
+  if (isIOS && !_iOSWorkaround) {
+    _iOSWorkaround = true
+    Array.from(window.document.body.children)
```
antfu, line 50: Are you able to provide some reference to explain why this is needed in the comments?

PR #2416, opened 2022-11-10
fix(useIDBKeyval): incorrect value init set

sun0day:
### Description

`useIDBKeyval` has some problems in L58-L59. This PR purpose is to solve these problems below.

-  Only `undefined` and `null`  `rawInit` should be ignored to set
- Since `set` is an async method, it is unsafe without `await`(L65-L67 can't catch error when `set` error). 

https://github.com/vueuse/vueuse/blob/0a49ea4dacbd98cdeb6e4be1bba3bc9347a8845a/packages/integrations/useIDBKeyval/index.ts#L54-L70

### Additional context

Unit test covered.

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


sun0day:
@sxzz @antfu  cc

sxzz (approved):

PR #2418, opened 2022-11-10
fix(useStorage): fix useStorage sync error

linghaoSu:
### Description

通过在更新 data.value 的时候禁用 watch ，解决循环更新的问题

https://user-images.githubusercontent.com/56526981/201071421-34e08e66-68b1-462e-8154-5f4a9b672e85.mov

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useStorage/index.ts
+++ b/packages/core/useStorage/index.ts
@@ -1,6 +1,6 @@
 import type { Awaitable, ConfigurableEventFilter, ConfigurableFlush, MaybeComputedRef, RemovableRef } from '@vueuse/shared'
 import { isFunction, pausableWatch, resolveUnref } from '@vueuse/shared'
-import { ref, shallowRef } from 'vue-demi'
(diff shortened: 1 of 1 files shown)


NamesMT:
+1 This made me crazy, implementing a unique data refresher for multi-tab, randomly this happens and hangs everything.

NamesMT:
This PR fixes the circular update and therefore fixes freezing, but sometimes data still do go out of sync when multiple tabs updates at the same time.

The attached screenshots are an example of one case where the data goes unsynced, when two windows trying to update the value at the same time.

First image is the data that the tab wrote (left), and the data that the useStorage Ref holds (right), logged after a 1-sec sleep.
Second image is the actual data in the browser's localStorage.

image
image

As you can see, the right tab got out of sync with the actual value of localStorage.

PR #2431, opened 2022-11-11
fix(useFirestore): fix falsy type error

kiyopikko:
### Description

Now you can use only `false` value for dependent query, so I added `Falsy` type.

### Additional context

Also fixed to use `null` value to initial value.

---

- [x] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/firebase/useFirestore/index.ts
+++ b/packages/firebase/useFirestore/index.ts
@@ -33,25 +33,27 @@ function isDocumentReference<T>(docRef: any): docRef is DocumentReference<T> {
   return (docRef.path?.match(/\//g) || []).length % 2 !== 0
 }
 
+type Falsy = false | 0 | '' | null | undefined
+
 export function useFirestore<T extends DocumentData>(
-  maybeDocRef: MaybeRef<DocumentReference<T> | false>,
+  maybeDocRef: MaybeRef<DocumentReference<T> | Falsy>,
   initialValue: T,
   options?: UseFirestoreOptions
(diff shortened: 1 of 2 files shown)


packages/firebase/useFirestore/index.ts
```diff
@@ -33,25 +33,27 @@ function isDocumentReference<T>(docRef: any): docRef is DocumentReference<T> {
   return (docRef.path?.match(/\//g) || []).length % 2 !== 0
 }
 
+type Falsy = false | 0 | '' | null | undefined
```
kiyopikko, line 36: Adding `Falsy` fixed these errors

スクリーンショット 2022-11-12 0 53 30

packages/firebase/useFirestore/index.ts
```diff
@@ -33,25 +33,27 @@ function isDocumentReference<T>(docRef: any): docRef is DocumentReference<T> {
...
   initialValue: T[],
   options?: UseFirestoreOptions
 ): Ref<T[]>
 
 // nullable initial values
 export function useFirestore<T extends DocumentData>(
-  maybeDocRef: MaybeRef<DocumentReference<T> | false>,
-  initialValue?: T | undefined,
+  maybeDocRef: MaybeRef<DocumentReference<T> | Falsy>,
+  initialValue?: T | undefined | null,
```
kiyopikko, line 52: this line fixed this error

スクリーンショット 2022-11-12 0 53 50

PR #2432, opened 2022-11-12
feat(eventBus): listeners for defined events

### LINKED ISSUE #2375: useEventBus: allow filtering events in `on()` function

When using `useEventBus` it is possible to emit 'events' but it's not efficient to filter events on listeners. You need to filter events inside the listener for example:
```ts
const bus = useEventBus();

bus.on((event, payload)=>{
    if(event !=== 'playAnimation') return;
    [...]
})
bus.on((event, payload)=>{
    if(event !=== 'openSidebar') return;
    [...]
})

bus.emit('playAnimation');
bus.emit('openSidebar', {...});
```

or you need to create massive listener using `switch`. It's not efficient and logical to use only listener for every events and/or filter inside listeners.
I want to take your opinions and alternatives about this issue.

### Suggested solution

Allow event filtering in `on` function alongside global listeners. It won't break previous use cases but enhances the usage.

```ts
const bus = useEventBus();

// playAnimation
bus.on('playAnimation', (payload)=>{
    [...]
})
// openSidebar
bus.on((event, payload)=>{
    [...]
})

bus.emit('playAnimation');
bus.emit('openSidebar', {...});
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

chaii3:
fixes #2375

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useEventBus/index.md
+++ b/packages/core/useEventBus/index.md
@@ -17,16 +17,25 @@ const listener = (event: string) => {
   console.log(`news: ${event}`)
 }
 
+const specialEventListener = (event: string) => {
+  console.log(`special event: ${event}`)
+}
+
 // listen to an event
 const unsubscribe = bus.on(listener)
+
+// listen defined event
+const unsubscribeSpecial = bus.onSpecial('The special event', specialEventListener)
 
 // fire an event
 bus.emit('The Tokyo Olympics has begun')
 
 // unregister the listener
 unsubscribe()
+unsubscribeSpecial()
 // or
 bus.off(listener)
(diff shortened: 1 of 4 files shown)


antfu:
Thank you. But I would avoid adding new APIs. I think doing function overloads like https://github.com/vueuse/vueuse/issues/2375 suggests would be better.

chaii3:
> Thank you. But I would avoid adding new APIs. I think doing function overloads like https://github.com/vueuse/vueuse/issues/2375 suggests would be better.

Ok, I'll do overloads

PR #2442, opened 2022-11-14
fix(useStyleTag): allow use of existing node

jameswragg:
### Description

When using useStyleTag with an ID this PR ensures that if the target ID already exists, don't re-append the node to the document `<head>` causing the node to move in the DOM.

This allows an empty style node to be placed in the page template & useStyleTag will use/update that - in-place, e.g.
```
<style type="text/css" id="custom-id"></style>
```

### Additional context

When style elements move in the DOM it can cause unintended side-effects in terms of CSS specificity.

---

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other


sibbng (approved):

packages/core/useStyleTag/index.ts
```diff
@@ -76,7 +76,9 @@ export function useStyleTag(
     el.id = id
     if (options.media)
       el.media = options.media
```
sibbng, line 78: Could you fix lint errors and move this part inside the `if (!el.isConnected)`?

sibbng (changes requested):

sibbng (approved):

WebMechanic:
wonderful! 
The unintended side-effects where actually caused by CSS **_inheritance_** (source order) not so much specificity (selector "weight").

PR #2446, opened 2022-11-15
feat(createEventHook): second param detached for on function

MinatoHikari:
### Description

second param detached for on function to enable auto off in components

https://github.com/vueuse/vueuse/issues/2532

### Additional context

---

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other


MinatoHikari:
> I would like to know more context about why this is needed. Otherwise, I would suggest using `effectScope`'s detach feature to initialize the compsoables.

https://codesandbox.io/s/awesome-currying-tcemqh?file=/src/components/eventHook.js

I think this case is enough? In my project it is more complex.

PR #2459, opened 2022-11-21
feat(useTimeSpan): new function

Waleed-KH:
### Description

New `useTimeSpan` function to get reactive TimeSpan object that represents a time interval (duration of time or elapsed time) that is measured as a positive or negative number of days, hours, minutes, seconds, and fractions of a second.

### Additional context

Inspired by [.Net System.TimeSpan](https://learn.microsoft.com/en-us/dotnet/api/system.timespan)

ToDo:
- [x] Format option
- [x] Parser
- [x] Overloads to receive `days`, `hours`, `minutes`, `seconds`, `milliseconds`
- [ ] Tests

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other


Waleed-KH:
finished most of the features and ready for review, I'll add tests files soon

Waleed-KH:
@antfu Please check and let me know what you think

antfu:
Thanks a lot for your work! It looks great but I am worried about if this is adding too much complexity to VueUse. Generally I would recommend users to work with Date/Time with a better and standalone lib like `dayjs` or so. Or if we really want to be self-contained, maybe a `@vueuse/time` could be a better fit, if someone is willing to dedicate time on it. I am closing this PR for now, but open to future discussions. Thanks.

Waleed-KH:
@antfu 

> I am worried about if this is adding too much complexity to VueUse.

I don't think it add complexity at all, it has many use cases, I use it a lot in my projects, and I'm sure many developers will find it useful. I've also added some usage examples in the doc file.

> Generally I would recommend users to work with Date/Time with a better and standalone lib like dayjs or so.

VueUse already has functions (e.g. `useDateFormat`) that provide similar usage, so I think it fits good with them.

I also was planning for it to be used in [a new `useTimer` function](https://github.com/Waleed-KH/vueuse/blob/useTimer/packages/core/useTimer/index.ts), and I believe it could be used as a base for many other new functions (`useStopwatch`, `useCountdown` ...).

> maybe a @vueuse/time could be a better fit, if someone is willing to dedicate time on it.

I could dedicate some time for it, but I'm not sure it will be worth it, and will end up also being closed after 4 months of waiting  with no reply or discussion.

antfu:
I am sorry for letting you wait, but as you can see, this is the problem we have right now, the scope of VueUse has become too big, and we don't have enough resources to maintain them all or even respond to them in time. I am sure every feature could be found useful by a lot of users, but this is the trade-off we have to make to make the project move forward.

PR #2477, opened 2022-11-25
fix(useVirtualList): List sometimes missing elements

### LINKED ISSUE #2065: useVirtualList sometimes missing elements
### Describe the bug

With certain combinations of overscan and low element counts, it can happen that the virtual list is simply missing a few elements.

For example, I tweaked the demo to have an overscan of 1, and the demo list to have 3 elements. 
However, it only ends up displaying 2 elements, with no way of seeing the 3rd one.

### Reproduction

https://github.com/stefnotch/vueuse/blob/bug-virtual-list/packages/core/useVirtualList/demo.vue

### System Info

```Shell
System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 23.74 GB / 31.86 GB
  Binaries:
    Node: 16.14.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.16.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (104.0.1293.47)
```

### Used Package Manager

pnpm

### Validations

begedin:
### Description

Fixes #2065 

This fixes an issue that was closed, seemingly due to inactivity, but was not actually fixed. 

This fixes a single scenario I was able to isolate, but I'm not sure if there are actually others. One of the added tests covers the scenario.

### Additional context

The scenario I can reliably reproduce the issue in is if the original list is smaller than the capacity of the virtual list. It just doesn't render all the items in that case.

For example, render a list of 5 elements when the screen is able to show 10. With an overscan of 2,  2 + 1 elements will be rendered. With an overscan of 1, it will be 1 + 1 = 2, etc.

### Extra notes

It seems that there is still some faulty logic around how overscan works. If you set the overscan to 0, based on tests, you can't really get the first item in the original list to display, possibly the last as well.

I would love to look into that, but it's a bit difficult to unravel what all the start/end variables are intended to be. Not always clear if they are 0 or 1-indexed, inclusive or exclusive, etc. 

After this is accepted in, I can look into investigating that and opening a ticket.

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/useVirtualList/index.ts
```diff
@@ -119,21 +119,24 @@ function useVirtualListResourses<T>(list: MaybeRef<T[]>): UseVirtualListResource
...
-  return (containerHeight: number) => {
+function createGetViewCapacity<T>(
+  state: UseVirtualListResources<T>['state'],
+  source: UseVirtualListResources<T>['source'],
+  itemSize: UseVirtualListItemSize,
+) {
+  return (containerSize: number) => {
     if (typeof itemSize === 'number')
-      return Math.ceil(containerHeight / itemSize)
+      return Math.ceil(containerSize / itemSize)
```
begedin, line 129: Height naming was from when list was vertical only. Size is more appropriate, IMO

packages/core/useVirtualList/index.ts
```diff
@@ -119,21 +119,24 @@ function useVirtualListResourses<T>(list: MaybeRef<T[]>): UseVirtualListResource
...
-    for (let i = start; i < source.value.length; i++) {
-      const height = itemSize(i)
-      sum += height
-      if (sum >= containerHeight) {
-        capacity = i
+    for (let i = start; i < source.value.length + 1; i++) {
+      const size = itemSize(i)
+      sum += size
+      capacity = i
+      if (sum > containerSize)
```
begedin, line 138: Moving capacity out of the if is one of the fixes. Without this, if the original list size is smaller than what could fit in the view, we end up with a capacity of 1 + overscan

packages/core/useVirtualList/index.ts
```diff
@@ -155,7 +158,7 @@ function createGetOffset<T>(source: UseVirtualListResources<T>['source'], itemSi
         break
       }
     }
-    return offset + 1
+    return offset
```
begedin, line 161: This sounded like an off by 1. Without this, change, we can never have an offset of 0. With 0 overscan, that would mean we can never render the very start of the list.

packages/core/useVirtualList/index.ts
```diff
@@ -194,7 +198,7 @@ function createGetDistance<T>(itemSize: UseVirtualListItemSize, source: UseVirtu
     }
 
     const size = source.value
-      .slice(0, index)
+      .slice(0, index + 1)
```
begedin, line 201: Slice end index is exclusive. Without this and with 0 overscan, we never render the end of the list.

packages/core/useVirtualList/index.ts
StepanMynarik: Shouldn't the same thing be done with 'offset = i' in function 'createGetOffset'?

PR #2485, opened 2022-11-29
fix(useAxios): execute rejects on error

FRSgit:
### Description

Currently execute method doesn't really follow [the default behaviour of axios in terms of error handling](https://www.npmjs.com/package/axios#handling-errors). It might be beneficial for users to cover this case.

### Additional context

I'd like to get an answer to the question: was the error "silenced" (it's still accessible via error property) by design or not?

If it was - this PR should be closed. If not - we should consider merging it.

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/integrations/useAxios/index.test.ts
+++ b/packages/integrations/useAxios/index.test.ts
@@ -181,9 +181,24 @@ describe('useAxios', () => {
 
     await then((result) => {
       expect(result.data.value.id).toBe(1)
-      expect(isLoading.value).toBeFalsy()
-      expect(onRejected).toBeCalledTimes(0)
     }, onRejected)
+    expect(isLoading.value).toBeFalsy()
+
+    expect(onRejected).toBeCalledTimes(0)
+  })
+
+  test('execute rejects on error', async () => {
+    const { isLoading, then, execute } = useAxios(config, instance)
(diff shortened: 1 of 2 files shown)


jwrycza-tl:
Very important change! Right now there is no possibility to add catch to useAxios. You can use only error ref to stop/prevent async/promise.

YogaLin:
Why does the execute function not reject when there is error. The error ref is not elegant to use at all

laterdayi:
@FRSgit   If that's the case,


```
const test =async ()=>{

try{
  const {data,error} = await useAxios({immediate:true})
}
catch(error){
 
}
}
```

If so, try will never get an error

laterdayi:
@FRSgit  For example, if I treat code = 10001 as an error in the interceptor, how can I get this custom error in the try

laterdayi:
In that case, the error returned by useAxios doesn't make any sense. Right

laterdayi:
@antfu

laterdayi:
@FRSgit @jwrycza-tl @YogaLin @antfu

antfu:
Please don't ping ppl, and don't comment in closed issues/PR. Create a new issue with more detailed description/reproduction instead.

PR #2491, opened 2022-12-01
docs: explain in more detail how to specify a custom color scheme with useColorMode

Mintnoii:
### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useDark/index.md
+++ b/packages/core/useDark/index.md
@@ -21,7 +21,7 @@ const toggleDark = useToggle(isDark)
 
 ## Behavior
 
-`useDark` combines with `usePreferredDark` and `useStorage`. On start up, it reads the value from localStorage/sessionStorage (the key is configurable) to see if there is a user configured color scheme, if not, it will use users' system preferences. When you change the `isDark` ref, it will update the corresponding element's attribute and then store the preference to storage (default key: `vueuse-color-scheme`) for persistence.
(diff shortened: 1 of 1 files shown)


packages/core/useDark/index.md
```diff
@@ -21,7 +21,7 @@ const toggleDark = useToggle(isDark)
 
 ## Behavior
 
-`useDark` combines with `usePreferredDark` and `useStorage`. On start up, it reads the value from localStorage/sessionStorage (the key is configurable) to see if there is a user configured color scheme, if not, it will use users' system preferences. When you change the `isDark` ref, it will update the corresponding element's attribute and then store the preference to storage (default key: `vueuse-color-scheme`) for persistence.
+`useDark` combines with `usePreferredDark` and `useStorage`. On start up, it reads the value from localStorage/sessionStorage (the key is configurable) to see if there is a user configured color scheme. If so, you can specify it using the `storageKey` parameter (inherited from the configuration of `UseColorMode`). If not, it will use users' system preferences. When you change the `isDark` ref, it will update the corresponding element's attribute and then store the preference to storage (default key: `vueuse-color-scheme`) for persistence.
```
antfu, line 24: I am not sure, they make the "if not" sound like if not providing the `storageKey`, which is not the case since `storageKey` has it's default value.

I feel your changes duplicated with the existing `it reads the value from localStorage/sessionStorage (the key is configurable)`

Mintnoii:
Okay, I just pointed out that the storageKey item can be used to specify the value of custom storage. If this will cause ambiguity, I will turn this off.

packages/core/useDark/index.md
Mintnoii: Okay, I just pointed out that the storageKey item can be used to specify the value of custom storage. If this will cause ambiguity, I will turn this off.

PR #2493, opened 2022-12-02
feat(useRafFn): add `delta` and `timestamp`

innocenzi:
### Description

This pull requests adds an object containing the `delta` and `timestamp` of the current `requestAnimationFrame` to the callback function of `useRafFn`.

The `requestAnimationFrame` method is most useful when used in conjunction with its given timestamp, and the delta with the previous frame. One example is to make animations.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useRafFn/index.ts
+++ b/packages/core/useRafFn/index.ts
@@ -4,7 +4,7 @@ import { tryOnScopeDispose } from '@vueuse/shared'
 import type { ConfigurableWindow } from '../_configurable'
 import { defaultWindow } from '../_configurable'
 
-interface RafFnCallbackArguments {
+export interface UseRafFnCallbackArguments {
   /**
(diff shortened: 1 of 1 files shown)


packages/core/useRafFn/index.ts
```diff
@@ -1,9 +1,21 @@
 import { ref } from 'vue-demi'
-import type { Fn, Pausable } from '@vueuse/shared'
+import type { Pausable } from '@vueuse/shared'
 import { tryOnScopeDispose } from '@vueuse/shared'
 import type { ConfigurableWindow } from '../_configurable'
 import { defaultWindow } from '../_configurable'
 
+interface RafFnCallbackArguments {
```
antfu, line 7:
```suggestion
export interface UseRafFnCallbackArguments {
```

PR #2500, opened 2022-12-05
Make the ID property enumerable

llamington:
Would allow the ID property to show up in Vue devtools, allow copying via spread operator, etc.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/firebase/useFirestore/index.ts
+++ b/packages/firebase/useFirestore/index.ts
@@ -23,6 +23,7 @@ function getData<T>(
     Object.defineProperty(data, 'id', {
       value: docRef.id.toString(),
       writable: false,
+      enumerable: true
     })
   }
 


packages/firebase/useFirestore/index.ts
```diff
@@ -23,6 +23,7 @@ function getData<T>(
     Object.defineProperty(data, 'id', {
       value: docRef.id.toString(),
       writable: false,
+      enumerable: true
```
vaakian, line 26:
```suggestion
      enumerable: true,
```

vaakian (changes requested):

packages/firebase/useFirestore/index.ts
vaakian, line 26:
```suggestion
      enumerable: true,
```

vaakian (changes requested):
please fix the lint issue

Issue #2311, opened 2022-10-11
Warning 'Vue 2 does not support reactive collection types such as Map or Set' in useMagicKeys on Vue 2.7


MrCube42:
### Describe the bug

As stated in https://blog.vuejs.org/posts/vue-2-7-naruto.html even though Vue 2.7 fully includes the composition API there are a few caveats as e.g. `Map` and `Set` are not reactive. I am using `useMagicKeys` to listen for global _Ctrl+S_ in order to save some changes. The feature itself is working properly, however in my tests there appears a warning from inside `useMagicKeys` saying that  'Vue 2 does not support reactive collection types such as Map or Set'. In our setup we fail on warnings in our tests.

Having a look at the sources shows that `current` is defined as a reactive `Set` (https://github.com/vueuse/vueuse/blob/main/packages/core/useMagicKeys/index.ts#L83). This line of code produces the warning described above.

As _vueuse_ states to support both Vue 3 and 2(.7) (with the use of vue-demi and so one) I would not expect a warning like this.

---

For a minimal reproduction, use the repo from https://github.com/MrCube42/magickeys-warning and simply `npm run dev`. The browser will also log the warning. The code using `useMagicKeys` itself is located in the `App.vue`.

### Reproduction

https://github.com/MrCube42/magickeys-warning

### System Info

```Shell
System:
    OS: Linux 5.10 Ubuntu 20.04 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz
    Memory: 14.27 GB / 23.48 GB
    Container: Yes
    Shell: 5.0.16 - /bin/bash
  Binaries:
    Node: 16.16.0 - /mnt/wslg/runtime-dir/fnm_multishells/709_1665480869441/bin/node
    npm: 8.11.0 - /mnt/wslg/runtime-dir/fnm_multishells/709_1665480869441/bin/npm
  npmPackages:
    @vueuse/core: ^9.3.0 => 9.3.0 
    vue: ^2.7.10 => 2.7.10
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
[...]

PR #2503, opened 2022-12-05
fix(useMagicKeys): throw error on Vue 2.7

rick-you:
### Description

Fixes https://github.com/vueuse/vueuse/issues/2311

### Additional context

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


boomyao (dismissed):

packages/core/useMagicKeys/index.ts
```diff
@@ -2,6 +2,7 @@ import type { ComputedRef } from 'vue-demi'
 import { computed, reactive, ref, unref } from 'vue-demi'
 import type { MaybeComputedRef } from '@vueuse/shared'
 import { noop } from '@vueuse/shared'
+import { useReactiveSet } from '@vueuse/shared/utils/set'
```
antfu, line 5:
```suggestion
import { useReactiveSet } from '@vueuse/shared'
```

antfu:
Can you add some tests? This seems to break the demo page.

luvsic3:
I managed to fix the error on useReactiveSet, but not knowing how it is fixed, maybe you can add more detailed comment in useReactiveSet

antfu:
I am not confident about it. Adding this means every user would need to pay the cost (size and computational runtime cost) using this function regardless if they are on 2.7 or 3.0+. I feel the complexity is now worth it.

I think it would be better either:

- Get Vue 2.7 support reactive Map natively
- Explicit says that `useMagicKeys` is not supported with 2.7, and someone can release a fork version doing it.

PR #2507, opened 2022-12-05
feat(useWebSocket): handle heartbeat timeouts appropriately

### LINKED ISSUE #2506: `useWebSocket` doesn't auto-reconnect when it doesn't receive a pong within its specified timeout duration
### Describe the bug

In its current state the [heartbeat's pong timeout](https://github.com/vueuse/vueuse/blob/cfcc295261f463c74d80fcb9863cb4bdd7e673b9/packages/core/useWebSocket/index.ts#L275) calls the composable's `close()` function to close the WebSocket connection if it doesn't receive a pong within its specified timeout duration. The `close()` function unfortunately sets the composable's internal `explicitlyClosed` variable to `true`, which causes the composable to not auto-reconnect. Also, it closes the WebSocket connection gracefully and doesn't throw an error.

This implementation could be intentional, so it might now even be a bug per se, but I doubt it because the comment inside the heartbeat's pong timeout callback even states: _auto-reconnect will be trigger with ws.onclose()_
This just doesn't actually happen because of the aforementioned situation.

In my opinion the composable should throw an error that will be caught by its `onError()` function and attempt to auto-reconnect after it reached the pong timeout.

The creation of a minimal reproduction is a little bit more effort than I think would be worth it for this issue, but if you'd still require one to be absolutely sure, I'll create a repo for it.

### Reproduction

https://github.com/vueuse/vueuse/blob/cfcc295261f463c74d80fcb9863cb4bdd7e673b9/packages/core/useWebSocket/index.ts#L275

### System Info

```Shell
System:
    OS: Linux 4.19 Arch Linux
    CPU: (12) x64 Intel(R) Core(TM) i7-6800K CPU @ 3.40GHz
    Memory: 20.90 GB / 24.99 GB
    Container: Yes
    Shell: 3.5.1 - /usr/bin/fish
  Binaries:
    Node: 18.6.0 - ~/.local/state/fnm_multishells/16636_1670247225995/bin/node
    Yarn: 1.22.19 - ~/.local/state/fnm_multishells/16636_1670247225995/bin/yarn
    npm: 8.13.2 - ~/.local/state/fnm_multishells/16636_1670247225995/bin/npm
  npmPackages:
    @vueuse/core: ^9.5.0 => 9.5.0
    @vueuse/math: ^9.5.0 => 9.5.0
    @vueuse/nuxt: ^9.5.0 => 9.5.0
```

### Used Package Manager

yarn

### Validations

jasontheiler:
### Description

Fixes #2506

This PR fixes the problems that I described in #2506. It does so by replacing the invocation of the built-in `close()` function with the WebSocket instance's own `close()` method. Additionally, it dispatches a custom generic error which will then be received by the built-in `onError()` function.

To send a proper close reason to the WebSocket server, I added two new configuration keys:

- `heartbeat.pongTimeoutCloseCode?: number` (Default: `3001`)
- `heartbeat.pongTimeoutCloseReason?: string` (Default: `'Did not receive pong to previous ping'`)

I chose the default close code `3001` because the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code) states that close codes in the range of `3000-3999` are open for usage in libraries and it seemed most analogous to the generic `1001` _Going Away_ close code.

### Additional context

The default close code and reason I specified here are just my personal suggestions. If you have better ideas or just prefer different ones, please let me know.

---

- [x] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useWebSocket/index.ts
+++ b/packages/core/useWebSocket/index.ts
@@ -273,6 +273,7 @@ export function useWebSocket<Data = any>(
         if (pongTimeoutWait != null)
           return
         pongTimeoutWait = setTimeout(() => {
+          wsRef.value?.dispatchEvent(new Event('error'))
           // auto-reconnect will be trigger with ws.onclose()
           close()
         }, pongTimeout)
--- a/packages/core/useWebSocket/index.ts
+++ b/packages/core/useWebSocket/index.ts
@@ -40,6 +40,21 @@ export interface UseWebSocketOptions {
      * @default 1000
      */
     pongTimeout?: number
+
+    /**
+     * Close code when no heartbeat response is received within the timeout.
+     *
+     * @see https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code
+     * @default 3001
+     */
+    pongTimeoutCloseCode?: number
+
+    /**
+     * Close reason when no heartbeat response is received within the timeout.
+     *
+     * @default 'Did not receive pong to previous ping'
+     */
+    pongTimeoutCloseReason?: string
   }
 
   /**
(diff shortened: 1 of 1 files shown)


Rokixy:
got the same problem, autoReconnect not triggered.

can anyone reopen and review this pr?@antfu

PR #2530, opened 2022-12-12
fix(useStorage): circular & excessive updates on multiple tabs

NamesMT:
### Description

This fixes the infinity circular update and reduces the enormous amount of watch updates if you have multiple tabs open at the same time, logic:
+ Do not accept updates that are not meant for this tab
  + (the Web Storage API is dumb and if you have tabs A, B, and C open, tabs A and B will receive update events that are meant for both tabs A and B upon any change on C, this would also be circular for a bit and each tab will receive a huge amount of update requests.
+ Auto stops simple infinity circular that jumps between two values.

Tested non-stop for about 6 hours, should work fine for ~5 tabs updating at the same time, extreme cases with >= 10 tabs updating at the exact same time would require an additional `eventFilter` (both throttle & debounce works, tested 10ms-1000ms) on the useStorage to be flawless.

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

Relevant + issue video: #2418

### DIFF
--- a/packages/core/useStorage/index.ts
+++ b/packages/core/useStorage/index.ts
@@ -137,6 +137,8 @@ export function useStorage<T extends(string | number | boolean | object | null)>
   } = options
 
   const data = (shallow ? shallowRef : ref)(defaults) as RemovableRef<T>
+  let _old_data = null;
+  let _is_circular = 0;
 
   if (!storage) {
     try {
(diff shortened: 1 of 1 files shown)


NamesMT:
The current `_is_circular` simple count checking could be breaking on toggle spamming, maybe increasing the count to an unholy number or implementing a different/additional checking.
`serializer.write()` also got called twice in one update, once for `_old_data`, and once for `write()` watch.

Issue #2528, opened 2022-12-11
useCounter loop option


SwatDoge:
### Clear and concise description of the problem

Right now, `useCounter` has 2 additional options to set the minimum and maximum counting level:

```ts
useCounter(props.index, { min: 0, max: 3));
```

When the counter reaches the maximum value, it will stay at its maximum value when you try to increase it. In some it is useful to "overflow" the number back to 0, and vice versa. 

### Suggested solution

An additional option for useCounter, which enables this behaviour.

```ts
useCounter(props.index, { min: 0, max: 3, loop: true));
```

### Alternative

Not being a lazy developer.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
https://github.com/vueuse/vueuse/pull/2531#issuecomment-1354969596

PR #2531, opened 2022-12-12
feat(useCounter): add loop option

LadyChatterleyLover:
### Description

for issue https://github.com/vueuse/vueuse/issues/2528
### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/shared/useCounter/index.ts
+++ b/packages/shared/useCounter/index.ts
@@ -3,6 +3,7 @@ import { ref } from 'vue-demi'
 export interface UseCounterOptions {
   min?: number
   max?: number
+  loop?: boolean
 }
 
 /**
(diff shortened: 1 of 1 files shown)


packages/shared/useCounter/index.ts
```diff
@@ -18,10 +19,29 @@ export function useCounter(initialValue = 0, options: UseCounterOptions = {}) {
   const {
     max = Infinity,
     min = -Infinity,
+    loop = false,
   } = options
 
-  const inc = (delta = 1) => count.value = Math.min(max, count.value + delta)
-  const dec = (delta = 1) => count.value = Math.max(min, count.value - delta)
+  const inc = (delta = 1) => {
+    if (loop) {
```
webfansplz, line 26:
```js
if(count.value === max && loop){
 count.value = initialValue
}else {
 count.value = Math.min(max, count.value + delta)
}

```

antfu:
Thank you. But I think this feature is too specific. You might implement it in your own codebase if needed instead.

PR #2539, opened 2022-12-13
fix(useIntervalFn): watch computed refs instead of just refs

innocenzi:
`useIntervalFn`'s `interval` option has the ability to be watched to dynamically change the interval. However, the check to add the watcher uses `isRef`, even though `interval` is of type `MaybeComputedRef`, which allows it to be a function, which is a valid watch value.

This pull request changes the check to ensure `interval` is either a `Ref` or a `Function`, so the following is also valid:

```ts
const $props = defineProps<{
  interval?: number
}>()

useIntervalFn(() => { /** ... **/ }, () => $props.interval ?? 5000)
```


innocenzi:
CI fails but the errors seem to be unrelated to this PR

PR #2559, opened 2022-12-20
chore(useStorage): change how the type is obtained

Alfred-Skyblue:
### Description

The original nested ternary expression is modified to policy mode execution to increase code readability and maintainability

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -37,7 +37,7 @@ export function onClickOutside<T extends OnClickOutsideOptions>(
   handler: OnClickOutsideHandler<{ detectIframe: T['detectIframe'] }>,
   options: T = {} as T,
 ) {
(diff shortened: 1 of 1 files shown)


antfu:
Thank you, but I found the ternaries also quite readable and maintainable. Although it might be subtle I would be more worried about the perf impact on this change.

Alfred-Skyblue:
> Thank you, but I found the ternaries also quite readable and maintainable. Although it might be subtle I would be more Ok, maybe I'm personally used to writing code, and I think it's easier to extend this way, thanks for replying

Alfred-Skyblue:
> Thank you, but I found the ternaries also quite readable and maintainable. Although it might be subtle I would be more worried about the perf impact on this change.

Ok, maybe I'm personally used to writing code, and I think it's easier to extend this way, thanks for replying

PR #2561, opened 2022-12-20
feat(useAsyncState): Add onSuccess callbacks

### LINKED ISSUE #2560: Add "onSuccess" option for "useAsyncState" function

I'm using the `useAsyncState` function, which provides a useful `onError` option to do something on failure. Also, I often do something else after a successful execution, but I don't seem to see an option like `onSuccess` in the type definition of the UseAsyncStateOptions interface. I wonder if it would be useful to add a similar `onSuccess` option.

### Suggested solution

Add an `onSuccess` option to the `UseAsyncStateOptions` interface.

### Alternative

I achieved my purpose by using the following code, but I think providing onSuccess option can save writing these boilerplate codes.

```js
const { state, isReady, isLoading } = useAsyncState(
  axios
    .get('https://jsonplaceholder.typicode.com/todos/1')
    .then(t => t.data),
  { id: null },
)

watch(() => isReady.value, () => {
  if(isReady.value) {
    doSomething()
  }
})

```

### Additional context

_No response_

### Validations

Alfred-Skyblue:
### Description

Add an onSuccess callback in useAsyncState

### Additional context

issues [2560](https://github.com/vueuse/vueuse/issues/2560)

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #2560`, [讨论地址](https://github.com/vueComponent/ant-design-vue-pro/issues/838)).

### DIFF
--- a/packages/core/useAsyncState/index.ts
+++ b/packages/core/useAsyncState/index.ts
@@ -10,7 +10,7 @@ export interface UseAsyncStateReturn<Data, Shallow extends boolean> {
   execute: (delay?: number, ...args: any[]) => Promise<Data>
 }
 
-export interface UseAsyncStateOptions<Shallow extends boolean> {
+export interface UseAsyncStateOptions<Shallow extends boolean, D = any> {
   /**
    * Delay for executing the promise. In milliseconds.
    *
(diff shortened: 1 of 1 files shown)


packages/core/useAsyncState/index.ts
```diff
@@ -105,6 +111,7 @@ export function useAsyncState<Data, Shallow extends boolean = true>(
       const data = await _promise
       state.value = data
       isReady.value = true
+      onSuccess({ ...data })
```
Bustuk, line 114: what if data won't be an object? For example it may resolve to a number

Issue #2410, opened 2022-11-08
expiring async ref


silberGuy:
### Clear and concise description of the problem

Hello :)
I could not find an existing feature for my need, please correct me if I missed it.
My app is using a server generated JWT, that expires after 15 minutes. I use this token for certain requests to a different service, and I want to reuse the same token as long as it is alive. After expiration, I want to automatically regenerate a JWT **only** on the next usage.
More generically - I need a ref to automatically clear its value, and that an expiring value could be asynchronously generated  on demand.

### Suggested solution

I implemented for myself something similar to the following:

## Generic Composables

```ts
function useExpiringRef<T>(timeout: number) {
  const value = ref<T>()

  return computed({
    get() {
      return value.value
    },
    set(newVal: T | undefined) {
      value.value = newVal
      setTimeout(() => {
        value.value = undefined
      }, timeout)
    },
  })
}

function useAsyncExpiringRef<T>(fn: () => Promise<T>, timeout: number) {
  const value = useExpiringRef<T | Promise<T>>(timeout)

  return computed(() => {
    if (!value.value) {
      value.value = fn()
    }

    return value.value
  })
}
```

## Using in App Logic
```ts
async function createJWT(): Promise<string> {
  return await fetch('/generate-jwt').then(res => res.text())
}

const tokenAsyncRef = useAsyncExpiringRef(createJWT, 1000 * 60 * 15 // 15min)

// The function I use in the app
function doStuffWithAnotherService() {
  const token = await tokenAsyncRef.value
  doStuffWithToken(token)
}

```

### Alternative

A `setInterval` of 15 minutes that refetches the token

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

climba03003:
Feature look good, the use-case is a bit prone to error.

```ts
function checkExpire(timeout: number): boolean {
  const last = new Date()
  return {
    check() {
      return (last + timeout) > Date.now()
    },
    reset() {
      last = new Date()
    }
  }
}

function useExpiringRef<T>(timeout: number | Function) {
  const value = ref()
[...]

PR #2563, opened 2022-12-20
feat(useExpiringRef): new function

silberGuy:
### Description

solves #2410

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
How would it differencing from https://vueuse.org/shared/refautoreset/#refautoreset?

silberGuy:
> How would it differencing from https://vueuse.org/shared/refautoreset/#refautoreset?

Thank you for the response, I did not find this previously but it really the same as `useExpiringRef`, except for the ability to update the timeout after creation
Also `useAsyncExpiringRef` creates an auto-reseting ref that defaults to an async value from a function

antfu:
> except for the ability to update the timeout after creation

Well actually you can:

```ts
const timeout = ref(100)
const foo = refAutoReset('foo', timeout)

// change the timeout
timeout.value = 200
```

antfu:
Thanks for your contribution. I think introducing two similar functions with only subtle different would cause the confusion and increase the maintenance burden. It would be better to improve the existing one if you see something really missing. Thanks

PR #2578, opened 2022-12-25
keepAliveScope

sabereen:
#### keepAliveScope
Creates a [effectScope](https://vuejs.org/api/reactivity-advanced.html#effectscope) that is disposed when the component is deactivated.

We can disable composable functions that are not needed in deactivated state of "KeepAlive".

```js
import { keepAliveScope, useEventListener } from '@vueuse/core'

keepAliveScope(() => {
  // listen to scroll event only if the component is not deacivated
  useEventListener('scroll', () => {
    // ...
  })
})
```

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
Sounds like good ideas. Can you split them into two separate PRs so we can discuss the improvements we could make separately? Thank you

sabereen:
I removed one of them and will make another PR for that one.

sabereen:
https://github.com/vueuse/vueuse/pull/2615

okxiaoliang4:
LGTM.

PR #2580, opened 2022-12-25
feat(useDebounceFn): return result using promise

### LINKED ISSUE #2566: Make useDebounceFn return fetchable

When we make a debouced method with returned values, we're not allowed to get the values, which could be usefull in some cases

```js
const deboucedFn = useDebounceFn(() => 'value', 100);
console.log(deboucedFn()); // undefined
```
This will currently return undefined

### Suggested solution

We can make this method return a promise with the returned values

```js
const deboucedFn = useDebounceFn(() => 'value', 100);
console.log(deboucedFn()); // Promise<'value'>
```
This can return promise that will be resolve with 'value'

As example for this implementation, i've done in a project this method
```ts
function useDeboundeFnWithRet<Args extends unknown[], R>(fn: (...args: Args) => Promise<R>, time: number) {
  let timerId: NodeJS.Timeout;

  return function debounced(...args: Args) {
    if (timerId) {
      clearTimeout(timerId);
    }

    return new Promise<R>((resolve) => {
      timerId = setTimeout(() => resolve(fn(...args)), time);
    });
  };
}
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

vaakian:
closes #2566 

This is incremental, no breaking change. And I shall make the same change on `useThrottleFn`.
### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/shared/useDebounceFn/index.md
+++ b/packages/shared/useDebounceFn/index.md
@@ -35,7 +35,34 @@ const debouncedFn = useDebounceFn(() => {
 window.addEventListener('resize', debouncedFn)
 ```
 
+Optionally, you can get the return value of the function using promise operations.
 
+```js
+import { useDebounceFn } from '@vueuse/core'
+const debouncedRequest = useDebounceFn(() => 'value', 1000)
+
+debouncedRequest().then((value) => {
+  console.log(value) // 'value'
+})
+```
+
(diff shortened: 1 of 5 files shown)


antfu:
That's a nice feature, thank you!

PR #2581, opened 2022-12-25
fix(useActiveElement): scheduled update

### LINKED ISSUE #2214: useFocusWithin gives incorrect values
### Describe the bug

When changing focus between children of the referenced element, `focused` will always be set to `false` before being set back to `true`. This is not what I expect to happen, as the focus within the referenced element never gets lost. 

 Because of this behavior, following statement from the docs is incorrect:
>  It is meant to match the behavior of the :focus-within CSS pseudo-class.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-ifsyhn/?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: Linux 5.10 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (16) x64 AMD Ryzen 7 2700X Eight-Core Processor
    Memory: 4.61 GB / 7.74 GB
    Container: Yes
    Shell: 3.5.1 - /usr/bin/fish
 Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
 npmPackages:
    @vueuse/core: ^8.7.5 => 8.9.4 
    vue: ^3.2.25 => 3.2.37
```

### Used Package Manager

pnpm

### Validations

### LINKED ISSUE #2565: useFocus doesn't update on blur
### Describe the bug

At some point after 9.6.0 `focused` from `useFocus` is not set to false on `blur`. It might have something to do with this change https://github.com/vueuse/vueuse/pull/2540

### Reproduction

https://stackblitz.com/edit/vue-6quiul?file=src/App.vue

### System Info

```Shell
Chrome, FF
```

### Used Package Manager

yarn

### Validations

vaakian:
### Description
replaced by #2600 👍

fixes #2565
fixes #2214

The PR #2540 has actually broken the functionality.

See the demos in current docs, it doesn't update state on blur:
- https://vueuse.org/core/usefocuswithin/#usefocuswithin
- https://vueuse.org/core/usefocus/#usefocus

By the description in #2540, the `activeElement` will be the `<body />` in between(when focus from one element to another element, which will trigger an extra useless update).
This PR is fixing it by using a **schedular** that triggers the getter at next render.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useActiveElement/index.ts
+++ b/packages/core/useActiveElement/index.ts
@@ -17,11 +17,9 @@ export function useActiveElement<T extends HTMLElement>(options: ConfigurableWin
   )
 
   if (window) {
-    useEventListener(window, 'blur', (event) => {
-      if (event.relatedTarget === null)
-        return
-
-      activeElement.trigger()
+    useEventListener(window, 'blur', () => {
+      const schedular = window.requestAnimationFrame || window.setTimeout
+      schedular(() => activeElement.trigger())
     }, true)
     useEventListener(window, 'focus', activeElement.trigger, true)
   }


dpschen:
@vaakian Hey,
I'm so sorry for introducing this bug. When I wrote the PR I translated a composable that I wrote myself in my head to the one of vueuse and by doing this I ignored the `relatedTarget` if it is *not* set instead of ignoring it if *is* set.

I fixed that in https://github.com/vueuse/vueuse/pull/2600

This time I tried it out, seems like it works.
Since that doesn't need a scheduled update I think it's slightly better than your approach. What do you think?

vaakian:
> @vaakian Hey,
> 
> I'm so sorry for introducing this bug. When I wrote the PR I translated a composable that I wrote myself in my head to the one of vueuse and by doing this I ignored the `relatedTarget` if it is *not* set instead of ignoring it if *is* set.
> 
> 
> 
> I fixed that in https://github.com/vueuse/vueuse/pull/2600
> 
> 
> 
> This time I tried it out, seems like it works.
> 
> Since that doesn't need a scheduled update I think it's slightly better than your approach. What do you think?

I agree with that, Well done! 👍

PR #2588, opened 2022-12-28
feat(useAxios): add a delay option

### LINKED ISSUE #2586: useAxios

useAxios add a delay option  

### Suggested solution

```
const { data, isFinished, execute, isLoading } = useAxios('', { method: 'GET' }, instance,{
delay :1000
})
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

Alfred-Skyblue:
### Description

useAxios add a delay option

### Additional context

issues [2586](https://github.com/vueuse/vueuse/issues/2586)

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #2586`).

### DIFF
--- a/packages/integrations/useAxios/index.ts
+++ b/packages/integrations/useAxios/index.ts
@@ -3,7 +3,7 @@ import { ref, shallowRef } from 'vue-demi'
 import { isString, until } from '@vueuse/shared'
 import type { AxiosError, AxiosInstance, AxiosRequestConfig, AxiosResponse, CancelTokenSource } from 'axios'
 import axios from 'axios'
-
+const DEFAULT_DELAY = 0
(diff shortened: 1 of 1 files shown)


wheatjs:
I think it would be best to keep the function small and focused and instead encourage using other functions to achieve this functionality.

For example, we can achieve this functionality by doing

```ts
const { data, isFinished, execute, isLoading } = useAxios('', { method: 'GET' }, instance, { immediate: false })
useTimeoutFn(() => { execute }, 1000)
```

Since this seems like a rather uncommon use case I think it should probably not be included in the `useAxios` function. If you disagree feel free to reopen and we can keep discussing.

PR #2590, opened 2022-12-28
feat(usePointerLock): new function

### LINKED ISSUE #2584: usePointerLock function and useMouse enhancement

I'd be glad to submit a PR that incapsulates [Pointer Lock API](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API) functionality within a composable.

### Suggested solution

Right now I'm using this code in my project:
```js
import { ref } from "vue";
import { unrefElement, useEventListener } from "@vueuse/core";
import type { MaybeElementRef } from "@vueuse/core";

/**
 * Reactive pointer lock.
 */
export function usePointerLock(target?: MaybeElementRef<HTMLElement>) {
  const element = ref<Element | null>()

  function lock(e: MaybeElementRef<HTMLElement> | Event) {
    const target = e instanceof Event ? <Element>e.target : unrefElement(e)
    target && target.requestPointerLock()

    const cleanupChange = useEventListener(document, 'pointerlockchange', () => {
      (element.value = document.pointerLockElement) || cleanupChange()
    })

    const cleanupError = useEventListener(document, 'pointerlockerror', () => {
      cleanupChange()
      cleanupError()
    })
  }

  function unlock() {
    element.value && document.exitPointerLock()
  }

  target && lock(target)

  return {
    lock,
    unlock,
    element,
  }
}

export type UsePointerLockReturn = ReturnType<typeof usePointerLock>
```
I decided to allow passing `Event` into the `lock` function so user may directly attach it to element as event handler:
```vue
<template>
    <div @mousedown.ctrl="lock" @mouseup="unlock"></div>
</template>

<script setup lang="ts">
const { lock, unlock, element } = usePointerLock()
</script>
```
This is the most likely usage scenario as browsers blocks any attempts to lock pointer without preceding user input:
```
Uncaught (in promise) DOMException: A user gesture is required to request Pointer Lock.
```
Also, `useMouse` must be amended to accept `movement` mouse tracking type:
```js
else if (type === 'movement') {
  x.value = event.movementX
  y.value = event.movementY
}
```
It's needed as it's the only way to track mouse when pointer lock mode is activated.

### Alternative

_No response_

### Additional context

_No response_

### Validations

daniser:
### Description

Closes #2584.

**this PR depends on following:**
- https://github.com/vueuse/vueuse/pull/2589

### Additional context

I'm not sure if I should incapsulate mouse tracking into this so it'd provide `{ x, y }` among other return values.
Because like 100% of time it'll be used with `useMouse({ type: 'movement' })`.
Right now I left this as it is for better separation of concerns, but I'd like to hear your opinion.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


wheatjs (dismissed):

packages/core/usePointerLock/demo.vue
```diff
@@ -0,0 +1,91 @@
...
+        <span style="--i: 0" />
+        <span style="--i: 1" />
+        <span style="--i: 2" />
+        <span style="--i: 3" />
+      </div>
+    </div>
+  </div>
+</template>
+
+<style scoped>
```
wheatjs, line 35: We usually try to use Unocss for the styles in our demos. Can you check the other demos and follow how we do styles there?

daniser: Where should I store custom rules and shortcuts? I think it's overkill to use root `unocss.config.ts` for styling a single demo. Can I configure it in the demo itself?

wheatjs: Any custom ones you can keep as is

wheatjs (approved):

PR #2591, opened 2022-12-28
feat(useActiveElement): optionally ignore active <body> element

JMaylor:
### Description

When focus is switched from some element, `elOne` to some other element, `elTwo` to another, the activeElement ref briefly changes to `<body>`. i.e. the value goes `elOne` -> `<body>` -> `elTwo`. This causes other composables that rely on `useActiveElement` to also update. e.g. https://github.com/vueuse/vueuse/issues/2214 (useFocusWithin).

While the useActiveElement composable is accurately reflecting what is happening with [Document.activeElement](https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement), I think that more often than not, this is not the desired outcome. The developer generally thinks:

> I am switching from one input in my form to another within my form, therefore I can use `useFocusWithin` and this should stay true throughout the focus switch.

I added an additional option to `useActiveElement`, called `ignoreBody`, which when set to true, will **not update activeElement when the focus switches to a (the) body tag**. This option is then inherited by `useFocusWithin`, so that we have:

```vue
<script lang="ts" setup>
const { focusedOriginal } = useFocusWithin(target)
// switching focus from elOne to elTwo, focusedOriginal values are the following:
// true -> false -> true

const { focusedNew } = useFocusWithin(target, { ignoreBody: true })
// switching focus from elOne to elTwo, focusedNew values are the following:
// true -> true (unchanged)
</script>

<template>
<form ref="target">
  <input ref="elOne">
  <input ref="elTwo">
</form>
</template>
```

### Additional context

---

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useActiveElement/index.ts
+++ b/packages/core/useActiveElement/index.ts
@@ -1,7 +1,12 @@
+import type { MaybeRef } from '@vueuse/shared'
 import { computedWithControl } from '@vueuse/shared'
 import { useEventListener } from '../useEventListener'
 import type { ConfigurableWindow } from '../_configurable'
 import { defaultWindow } from '../_configurable'
+
+export interface UseActiveElementOptions extends ConfigurableWindow {
+  ignoreBody?: MaybeRef<boolean>
+}
 
 /**
  * Reactive `document.activeElement`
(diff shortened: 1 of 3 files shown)


vaakian:
#2581 sloves this without adding extra options.

antfu:
Thank you! Close in favor of https://github.com/vueuse/vueuse/pull/2600

PR #2592, opened 2022-12-28
feat(useActiveElement): support shadow roots

43081j:
Assuming you have a DOM structure like so:

```
body
  my-element
    #shadow-root
      input
```

When the `input` becomes active, the outer document's `activeElement` will actually be `my-element`, _not_ the input.

Each shadow root has its own `activeElement` in case you want to get hold of the inner element in these cases.

This adds support to `useActiveElement` such that you can pass a shadow root as the document to observe the active element of.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

---

let me know if there's some reason you wouldn't want this or it needs doing differently and ill update.

### DIFF
--- a/packages/core/_configurable.ts
+++ b/packages/core/_configurable.ts
@@ -11,7 +11,7 @@ export interface ConfigurableDocument {
   /*
    * Specify a custom `document` instance, e.g. working with iframes or in testing environments.
    */
-  document?: Document
+  document?: DocumentOrShadowRoot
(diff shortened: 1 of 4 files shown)


dpschen (approved):

43081j:
rebased onto master since it was out of date fyi

antfu (approved):

PR #2596, opened 2022-12-29
fix(useElementBounding): update on scroll

### LINKED ISSUE #2319: useElementBounding scroll inside element not working
### Describe the bug

When using some features like the `useElementBounding` feature, the scroll handler only triggers when the window gets scrolled. The values don't update if the scrolling happens inside an element (i.e. `overflow: scroll`).

Seems like this has worked some time in the past. While creating the reproduction I noticed that it suddenly worked (the version in the StackBlitz template is `^7.6.2`. After switching to `^9.3.0` it no longer worked)

### Reproduction

https://stackblitz.com/edit/vitejs-vite-fcit5l?file=src/Current.vue

### System Info

```Shell
❯ npx envinfo --system --npmPackages '{vue,@vueuse/*}' --binaries --browsers
success Install finished in 1.432s


  System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    npm: 7.17.0 - /usr/local/bin/npm
  npmPackages:
    @vueuse/core: ^9.3.0 => 9.3.0 
    vue: ^3.2.25 => 3.2.31
```

### Used Package Manager

yarn

### Validations

Ducz01:
### Description

Closes #2319, The Scrolling of useElementBounding didn't work when a div gets scrolled, only when the window was scrolled. Now it works.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useElementBounding/index.ts
+++ b/packages/core/useElementBounding/index.ts
@@ -93,7 +93,7 @@ export function useElementBounding(
   watch(() => unrefElement(target), ele => !ele && update())
 
   if (windowScroll)
-    useEventListener('scroll', update, { passive: true })
+    useEventListener('scroll', update, true, { passive: true })
   if (windowResize)
     useEventListener('resize', update, { passive: true })
 
--- a/packages/core/useElementBounding/index.ts
+++ b/packages/core/useElementBounding/index.ts
@@ -93,7 +93,7 @@ export function useElementBounding(
   watch(() => unrefElement(target), ele => !ele && update())
 
   if (windowScroll)
-    useEventListener('scroll', update, true, { passive: true })
+    useEventListener('scroll', update, { capture: true, passive: true })
   if (windowResize)
     useEventListener('resize', update, { passive: true })
 


packages/core/useElementBounding/index.ts
```diff
@@ -93,7 +93,7 @@ export function useElementBounding(
   watch(() => unrefElement(target), ele => !ele && update())
 
   if (windowScroll)
-    useEventListener('scroll', update, { passive: true })
+    useEventListener('scroll', update, true, { passive: true })
```
sibbng, line 96:
```suggestion
    useEventListener('scroll', update, { capture: true, passive: true })
```

sibbng (changes requested):

dpschen (approved):

sibbng (approved):

PR #2599, opened 2022-12-30
feat(usePrevious): new function

Alfred-Skyblue:
### Description

A Hook to return the previous state.

### Additional context

usePrevious

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/usePrevious/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  shouldUpdate: ShouldUpdateFunc<T> = defaultShouldUpdate,
+) {
+  const prevRef = ref<T>()
+  const curRef = ref<T>()
+  watchEffect(() => {
+    if (shouldUpdate(curRef.value, state.value)) {
+      prevRef.value = curRef.value
+      curRef.value = state.value
+    }
+  })
```
vaakian, line 29: Seems `curRef` is just a intermediate state that doesn't necessarily need to be reactive? would slightly improve the performance.
```suggestion
  let cur: T | undefined
  watchEffect(() => {
    if (shouldUpdate(cur, state.value)) {
      prevRef.value = cur
      cur = state.value
    }
  })
```

vaakian (changes requested):

packages/core/usePrevious/demo.vue
```diff
@@ -0,0 +1,21 @@
...
+</script>
+
+<template>
+  <div>
+    <p>counter current value：{{ count }}</p>
+    <p>counter previous value：{{ prevCount }}</p>
+    <button @click="inc()">
+      increase
+    </button>
+    <button style="margin-left: 8px;" @click="dec()">
```
vaakian, line 16: The docs have the [unocss](https://uno.antfu.me/?s=ml-2) pre-installed, which is more concise.
```suggestion
    <button class="ml-2" @click="dec()">
```

packages/core/usePrevious/index.ts
vaakian: sorry being verbose

wheatjs:
I'm not completely opposed to this, but it seems like it does have some functionality overlap with `useRefHistory`. We can achieve something similar with just another line of code using `useRefHistory` https://stackblitz.com/edit/vueuse-reproduction?file=src%2FApp.vue&terminal=dev

```html
<script setup lang="ts">
import { computed } from 'vue'
import { useCounter, useRefHistory } from '@vueuse/core'

const { count, inc, dec } = useCounter()
const { history, undo, redo } = useRefHistory(count, { capacity: 2 })
const previous = computed(() => history.value.length < 2 ? 0 : history.value[1].snapshot)
</script>

<template>
  <div>
    <div>Current Value: {{ count }}</div>
    <div>Previous Value: {{ previous }}</div>

    <div>
      <button @click="dec(1)">-</button>
      <button @click="inc(1)">+</button>
    </div>
  </div>
</template>
```

Another advantage is that we can also undo/redo with this. I wonder if we would want to just leave this to the user or have `usePrevious` use `useRefHistory` internally. The one argument I could think of against doing this is performance could be an issue, but I don't have any numbers to prove that.

Issue #2117, opened 2022-08-24
asyncComputed should be using a shallowRef rather than a ref


basvanmeurs:
### Describe the bug

Hello @antfu 

I noticed that the `asyncComputed` of an object produces a reactive object rather than the raw object. In the source code, I found that a `ref` was used to contain the value rather than a `shallowRef`: https://github.com/vueuse/vueuse/blob/dc3ebe80bc61ee56004fdcd417cb234ee749e41e/packages/core/computedAsync/index.ts#L63

Producing the ref is a performance hazard and in conflict with how computed is working, as that produces the raw object. See https://github.com/vuejs/core/blob/main/packages/reactivity/src/computed.ts#L29

The solution is simply to use a `shallowRef` rather than a `ref`.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-mjexxs?file=src/App.vue

### System Info

```Shell
not important
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

xiankaiqun:
This issues should not have been closed, using `ref` to return `reference types` crashes people.
In addition, the return value should also be `ShallowReadonly` to ensure compliance with `computed` ideas.

PR #2616, opened 2023-01-05
feat(computedAsync): introduce `shadow` option

xiankaiqun:
### Description

Use `shallowRef` and mark the return value as `readonly` to conform to the `computed` ideas.

issues #2117 

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
`shallowRef` should be an options, just like: https://github.com/vueuse/vueuse/blob/2851c32aaf80762d37705a740a1d1edf2cea0a07/packages/core/useAsyncState/index.ts#L88-L91

We need to make it default to `false` to avoid breaking change, and change to `true` in another PR aiming for the next major version

xiankaiqun:
hi，I've added `shallow` to the options, for which I also provided an update to the docs
code: [88de7b2](https://github.com/vueuse/vueuse/pull/2616/commits/88de7b25e24c3a63684f8842e130f1c08c5c93a0)
docs:  [b627a94](https://github.com/vueuse/vueuse/pull/2616/commits/b627a945366c7790471b5da81d2d7655a7d4906c)

xiankaiqun:
I'm also preparing the next version :  #2621

PR #2620, opened 2023-01-05
fix(useThrottleFn): check negative timeout

### LINKED ISSUE #2618: UseThrottleFn does not call function - completely broken
### Describe the bug

In version 9.9.0 this function works

  const ZoomToPointsThrottled = useThrottleFn(
    (force: boolean) => {
      ZoomToPoints(force);
    },
    1500,
    true,
    false
  );

In version 9.10.0 the ZoomToPoints function is never called.  I tried removing the "force" parameter from both the throttle and the called functions, but this made no difference.

Problem seems to have been introduced in pull request: https://github.com/vueuse/vueuse/pull/2580

Rolling back to 9.9.0 fixes the problem for me.

The same can be seen in the linked reproduction project.  The "Increment Throttled" button works in 9.9.0 but not in 9.10.0

### Reproduction

https://stackblitz.com/edit/vitejs-vite-bjrjmi/?file=package.json,src%2FApp.vue

### System Info

```Shell
System:
    OS: Windows 10 10.0.22621
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 658.50 MB / 15.69 GB
  Binaries:
    Node: 16.19.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.2.0 - ~\AppData\Roaming\npm\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.963.0), Chromium (108.0.1462.46)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @vueuse/core: 9.10.0 => 9.10.0
    vue: 3.2.45 => 3.2.45
```

### Used Package Manager

npm

### Validations

vaakian:
### Description

fixes #2618 

### Additional context

caused by a **BIG** negative timeout. Exists in the past versions as well, but only occurs at the first call.
```js
const printHello = () => console.log('hello')
// will not print `hello`
setTimeout(printHello, -1672930567173)

// will print `hello`
setTimeout(printHello, -167293056)
```

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/shared/utils/filters.ts
+++ b/packages/shared/utils/filters.ts
@@ -157,9 +157,10 @@ export function throttleFilter(ms: MaybeComputedRef<number>, trailing = true, le
 
     if (elapsed > duration && (leading || !isLeading)) {
       lastExec = Date.now()
-      invoke()
+      return invoke()
     }
     else if (trailing) {
+      const remaining = duration - elapsed
       return new Promise((resolve, reject) => {
         lastRejector = rejectOnCancel ? reject : resolve
         timer = setTimeout(() => {
@@ -167,7 +168,7 @@ export function throttleFilter(ms: MaybeComputedRef<number>, trailing = true, le
           isLeading = true
           resolve(invoke())
           clear()
-        }, duration - elapsed)
+        }, remaining > 0 ? remaining : 0)
       })
     }
 
--- a/packages/shared/utils/filters.ts
+++ b/packages/shared/utils/filters.ts
@@ -157,11 +157,11 @@ export function throttleFilter(ms: MaybeComputedRef<number>, trailing = true, le
 
     if (elapsed > duration && (leading || !isLeading)) {
       lastExec = Date.now()
-      return invoke()
+      invoke()
     }
     else if (trailing) {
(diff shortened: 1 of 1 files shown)


wheatjs (approved):

packages/shared/utils/filters.ts
```diff
@@ -160,14 +160,16 @@ export function throttleFilter(ms: MaybeComputedRef<number>, trailing = true, le
...
+
+      lastValue = new Promise((resolve, reject) => {
         lastRejector = rejectOnCancel ? reject : resolve
         timer = setTimeout(() => {
           lastExec = Date.now()
           isLeading = true
           resolve(invoke())
           clear()
-        }, duration - elapsed)
+        }, timeout > 0 ? timeout : 0)
```
webfansplz, line 172: Math.max(0,duration - elapsed)

shrpne:
@vaakian can you update your PR please?

PR #2630, opened 2023-01-08
fix(useStorage): emit a custom event to support custom storage areas

### LINKED ISSUE #2567: [useStorage] Error in console when integrated with Pinia
### Describe the bug

After updating the `@vueuse/core` to version **9.8.2** I'm getting this error everytime I try to set a value in the storage using an action defined in a Pinia store
```js
TypeError: Failed to construct 'StorageEvent': Failed to read the 'storageArea' property from 'StorageEventInit': Failed to convert value to 'Storage'.
    at write (@vueuse_core.js?v=0b11bf63:2580:62)
    at watchPausable.flush.flush (@vueuse_core.js?v=0b11bf63:2565:80)
    at filter.fn.fn (@vueuse_core.js?v=0b11bf63:99:21)
    at bypassFilter (@vueuse_core.js?v=0b11bf63:104:10)
    at eventFilter (@vueuse_core.js?v=0b11bf63:183:7)
    at wrapper (@vueuse_core.js?v=0b11bf63:99:5)
    at callWithErrorHandling (chunk-QO6ZHO7P.js?v=482669c2:1378:18)
    at callWithAsyncErrorHandling (chunk-QO6ZHO7P.js?v=482669c2:1386:17)
    at job (chunk-QO6ZHO7P.js?v=482669c2:2660:9)
    at flushPreFlushCbs (chunk-QO6ZHO7P.js?v=482669c2:1512:7)
```
image

Rolling back to version **9.8.1** of `@vueuse/core` seems to solve the issue.

### Reproduction

https://stackblitz.com/edit/usestorage-issue-pinia?file=src/App.vue

### System Info

```Shell
System:
    OS: Windows 10 10.0.19045
    CPU: (4) x64 Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
    Memory: 5.53 GB / 15.90 GB
  Binaries:
    Node: 16.17.1 - C:\Program Files\nodejs\node.EXE
    npm: 8.15.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 108.0.5359.125
    Edge: Spartan (44.19041.1266.0), Chromium (108.0.1462.54)
```

### Used Package Manager

npm

### Validations

### LINKED ISSUE #2605: useLocalStorage integrated with Pinia, the storage value just update once
### Describe the bug

appeared after 800f74f49197b11ee5691ac8f4061eb2459ad07e

@vueuse/core: 9.8.1 is ok

image

### Reproduction

https://stackblitz.com/edit/vitejs-vite-7bhe6s

### System Info

```Shell
System:
    OS: macOS 13.2
    CPU: (8) arm64 Apple M1 Pro
    Memory: 45.00 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.1 - ~/.nvm/versions/node/v16.17.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.17.1/bin/yarn
    npm: 8.15.0 - ~/.nvm/versions/node/v16.17.1/bin/npm
  Browsers:
    Chrome: 108.0.5359.124
    Edge: 108.0.1462.54
    Safari: 16.2
  npmPackages:
    @vueuse/core: 9.8.2 => 9.8.2
    vue: ^3.2.45 => 3.2.45
```

### Used Package Manager

npm

### Validations

43081j:
### Description

Fixes #2605 
Fixes #2567 

Basically when you introduced a fix to update instances on the same window using the same store (800f74f), you did this by dispatching a `StorageEvent` (such that the already existent listener would catch it and update the value as needed).

This caused a breaking change in that things like pinia stores would no longer work with it.

The reason for that is that you can only construct a `StorageEvent` if the storage area is built-in pretty much (e.g. local storage, session storage, etc). You definitely _cannot_ construct one with a custom storage area such as pinia (anything that conforms to your `StorageLike` interface).

All of the `useStorage` tests were throwing an exception because of this, which was being eaten up by the `onError` handler (which the tests don't assert wasn't called, probably a good improvement in another PR to check that).

To fix, you could just fake a `storage` event (i.e. fire a `CustomEvent` with the name `'storage'`). However, that will confuse any third party code elsewhere which relies on the `storage` event being an actual `StorageEvent`.

Even if you somehow create your own event with the same interface (so it can't be a `CustomEvent`, since that'd nest your values in `detail`), you'd then be breaking the fact that storage events are only meant to fire **when another window mutated the store**.

Because of all of the above, i opted here to create a custom event specific to vueuse which we listen for. That way we can safely dispatch it without affecting any third-party code and can do so with any `StorageLike` area.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useStorage/index.test.ts
+++ b/packages/core/useStorage/index.test.ts
@@ -1,7 +1,7 @@
 import { debounceFilter, promiseTimeout } from '@vueuse/shared'
 import { isVue3, ref } from 'vue-demi'
 import { nextTwoTick, useSetup } from '../../.test'
-import { StorageSerializers, useStorage } from '.'
+import { StorageSerializers, customStorageEventName, useStorage } from '.'
 
 const KEY = 'custom-key'
 
@@ -427,6 +427,29 @@ describe('useStorage', () => {
     window.dispatchEvent(new StorageEvent('storage', { storageArea: localStorage, key: KEY, newValue: '1' }))
     await nextTwoTick()
     expect(data.value).toBe(1)
+  })
+
+  it('emits custom storage events on change', async () => {
+    const eventFn = vi.fn()
+    const data0 = useStorage(KEY, 0, storage)
+    const data1 = useStorage(KEY, 0, storage)
+    expect(data0.value).toBe(0)
+    expect(data1.value).toBe(0)
+
+    window.addEventListener(customStorageEventName, eventFn, { once: true })
+
+    data0.value = 1
+    await nextTwoTick()
+
+    expect(data0.value).toBe(1)
+    expect(data1.value).toBe(1)
+    expect(eventFn).toHaveBeenCalled()
+    const call = eventFn.mock.calls[0] as [CustomEvent]
+
+    expect(call[0].detail.storageArea).toEqual(storage)
+    expect(call[0].detail.key).toEqual(KEY)
+    expect(call[0].detail.oldValue).toEqual('0')
+    expect(call[0].detail.newValue).toEqual('1')
   })
 
   it('handle error', () => {
--- a/packages/core/useStorage/index.ts
+++ b/packages/core/useStorage/index.ts
@@ -53,6 +53,15 @@ export const StorageSerializers: Record<'boolean' | 'object' | 'number' | 'any' 
   },
 }
 
+export const customStorageEventName = 'vueuse-storage'
+
+export interface StorageEventLike {
+  storageArea: StorageLike | null
+  key: StorageEvent['key']
+  oldValue: StorageEvent['oldValue']
+  newValue: StorageEvent['newValue']
+}
+
 export interface UseStorageOptions<T> extends ConfigurableEventFilter, ConfigurableWindow, ConfigurableFlush {
   /**
    * Watch for deep changes
@@ -160,8 +169,10 @@ export function useStorage<T extends(string | number | boolean | object | null)>
     { flush, deep, eventFilter },
   )
 
-  if (window && listenToStorageChanges)
+  if (window && listenToStorageChanges) {
     useEventListener(window, 'storage', update)
+    useEventListener(window, customStorageEventName, updateFromCustomEvent)
+  }
 
   update()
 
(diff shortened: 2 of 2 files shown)


okxiaoliang4 (approved):

43081j:
anyone i can nudge to move this forward?

happy to rework anything if needed, as long as we can fix it

GODLiangCY:
This PR hasn't fixed these issues yet and calling `useEventListener` [here](https://github.com/vueuse/vueuse/pull/2630/files#diff-015d17be17c2a02d87f6b40bc47f526a5e82a01ee334f2883d271c6d98543d21R174) caused #2782.

IMO, the core problem is that after executing `window.dispatchEvent`, `update` was called but `watch` didn't be resumed correctly, that's why localStorage didn't sync up with pinia store after the first change. Having no ideas about how to solve this and why it goes this way.

43081j:
@GODLiangCY no worries, i hadn't seen that related issue at all so thanks for pointing that out

confident the fix in this pr was needed, but does seem like there's still a bug (introduced by this or not).

i did try debug this and got a little tangled up in the watcher logic. will take another look soon and see if i can figure it out

43081j:
to be clear it also seems the issue you linked has a repro with a version _before_ this particular PR landed

this PR seems to be shipped in 9.13.0 while the referenced issue is using 9.8.2

so it does indeed look like its a bug that was in the original event changes, before this PR. will still try figure it out though

GODLiangCY:
> to be clear it also seems the issue you linked has a repro with a version _before_ this particular PR landed (so the extra `useEventListener` call did not cause that issue)

Apologize for my carelessness!! 

I agree with you that this pr is needed and solves problems you mentioned in desciptions. [800f74](https://github.com/vueuse/vueuse/commit/800f74f49197b11ee5691ac8f4061eb2459ad07e) commit do not work well with pinia's store, maybe something like `store.count++` did *not* trigger `watch`? I've made a [demo](https://stackblitz.com/edit/vitejs-vite-zkrafb?file=src/App.vue) which runs expectly, hope it could help

43081j:
i think i roughly understand it now

if you have a watcher on `storage.value`, the following will happen as of that mentioned commit:

- set `storage.value`
- `useStorage` instance triggers, sets the inner storage value (e.g. local storage)
- `useStorage` instance dispatches event
- all `useStorage` instances receive event, pause **their own watcher** then set their internal value

the instance which fired the event will also catch its own event, leading to setting it twice. though i'd hope that'd be fine since you would be setting the value to what it already is, i.e. noop.

the issue you mentioned before, too, could be a separate thing. hes mutating a key _within_ an object in a store. doing that will dispatch an event to tell other instances _the overall object_ changed, so any watcher watching anything inside that object will fire i imagine. that probably explains the behaviour they saw

no solution off the top of my head (yet) but at least we probably now understand the two issues.

PR #2651, opened 2023-01-12
feat(watch): shorthands for watch

mrhammadasif:
### Description

We often have to watch for objects with `{deep: true}` or `{immediate: true}` ~~or both `{deep: true, immediate: true}`~~. I just added few shortHands **watchDeep** and **watchImmediate** ~~and **watchDeepImmediate**~~ to avoid writing them explicitly each time

Update 20 Feb, 2023: Removed `{watchDeepImmediate}` shorthand as suggested by @antfu 

### Additional context

These shorthands are based on `shared/watchOnce`

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
I think I am ok with having `watchDeep` and `watchImmediate`. But `watchDeepImmediate` feels a bit too much for me, as we can't really introduce all the combinations.

mrhammadasif:
> I think I am ok with having `watchDeep` and `watchImmediate`. But `watchDeepImmediate` feels a bit too much for me, as we can't really introduce all the combinations.

Well, there are many cases while writing an application where you want to do both of them. Do you have any better idea to write a shortcut for the combination?

antfu:
I would honestly think 

```ts
watchDeep(s, f, {
  immediate: true
})
```

would be enough. Or otherwise, I don't think we should provide something like `watchPostDeepImmediate`

mrhammadasif:
> I would honestly think
> 
> ```ts
> watchDeep(s, f, {
>   immediate: true
> })
> ```
> 
> would be enough. Or otherwise, I don't think we should provide something like `watchPostDeepImmediate`

I have updated the PR, and removed the `watchDeepImmediate` as you suggested

mrhammadasif:
Any update on this ?

antfu (approved):

PR #2652, opened 2023-01-13
feat(onKeyStroke): support ignoring repeated stroke

### LINKED ISSUE #2639: onKeyStroke fires rapidly when OS keyrepeat is active

As a developer using VueUse I want to benefit from keyboard shortcuts / hotkeys so users have an easier time navigating my app. 

unlike vue-hotkey or vue-shortkey back in the Vue 2 world, `onKeyStroke` fires the handler rapidly when keyrepeat is active.

A workaround could be to set `{ eventName: 'keyup' }`, but makes the action feel slow and might not what you expect when you hold down the key. 

Both `keydown` and `keypress` both have the same issue.

I'm sure you could make another workaround using `useDebounceFn`, but I (and many others) always need to fiddle until it works as expected. `useDebounceFn` is not as intuitive. Debouncing is also not a real solution, just another workaround.

### Suggested solution

I suggest to add the function `onKeyStrokeOnce` which works just as `onKeyStroke` but triggers only once on keydown/keypress, then "locks the function", and only unlocks it after `keyup` happened. 

### Alternative

_No response_

### Additional context

_No response_

### Validations

vaakian:
### Description

closes #2639 

Although developer can do the same in the callback, but I think integrating this feature is essential.

### Additional context

reference: [KeyboardEvent.repeat](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/repeat)

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/onKeyStroke/index.ts
```diff
@@ -105,3 +116,15 @@ export function onKeyPressed(key: KeyFilter, handler: (event: KeyboardEvent) =>
...
+
+/**
+ * Listen for keyboard keys being stroked while ignoring [repeated strokes](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/repeat).
+ *
+ * @see https://vueuse.org/onKeyStroke
+ * @param key
+ * @param handler
+ * @param options
+ */
+export function onKeyStrokeOnce(key: KeyFilter, handler: (event: KeyboardEvent) => void, options: Omit<OnKeyStrokeOptions, 'repeat'> = {}) {
```
antfu, line 128: I think the `Once` is a bit confusing, that may indicate you listen once and dispose. Maybe let's not introduce a new function for that.

packages/core/onKeyStroke/index.ts
```diff
@@ -9,6 +9,12 @@ export interface OnKeyStrokeOptions {
   eventName?: KeyStrokeEventName
   target?: MaybeComputedRef<EventTarget | null | undefined>
   passive?: boolean
+  /**
+   * Set to `false` to ignore repeated strokes.
+   *
+   * @default true
+   */
+  repeat?: boolean
```
antfu, line 17: This is too general, to me, it's more like "we will fire the event" repeatedly.

Instead, I would call it `dedupe` and default to `false`.

vaakian: I agree with that, but I just found a conventional name `autoRepeat` that seems adopted by public:
1. https://developer.apple.com/documentation/uikit/uistepper/1624079-autorepeat
2. https://learn.microsoft.com/en-us/office/vba/api/access.commandbutton.autorepeat
3. https://doc.qt.io/qt-6/qabstractbutton.html#autoRepeat-prop

packages/core/onKeyStroke/index.ts
vaakian: agree

packages/core/onKeyStroke/index.ts
```diff
@@ -9,6 +10,12 @@ export interface OnKeyStrokeOptions {
   eventName?: KeyStrokeEventName
   target?: MaybeComputedRef<EventTarget | null | undefined>
   passive?: boolean
+  /**
+   * Set to `false` to ignore repeated events when the key is being held down.
+   *
+   * @default true
+   */
+  autoRepeat?: MaybeComputedRef<boolean>
```
antfu, line 18: I think calling it `dedupe` would be more intuitive about what this configs do

antfu (changes requested):

katerlouis:
friendly push? :)

antfu (approved):

PR #2662, opened 2023-01-19
fix(useAxios): use RawAxiosRequestConfig type

lalalazero:
### Description

Due to Axios type [refactor](https://github.com/axios/axios/pull/5420), `headers` in `AxiosRequestConfig` is now mandatory, 

`useAxios` would cause typescript error like this

image

this pr is to replace `AxiosRequestConfig` with `RawAxiosRequestConfig` as suggested [here](https://github.com/axios/axios/issues/5476#issuecomment-1386773179) 

### Additional context

this refactor is introduced in axios 1.2.3, with 1.2.2 `AxiosRequestConfig` works fine, I'm not sure how to set this tip

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/integrations/useAxios/index.ts
+++ b/packages/integrations/useAxios/index.ts
@@ -1,7 +1,7 @@
 import type { Ref, ShallowRef } from 'vue-demi'
 import { ref, shallowRef } from 'vue-demi'
 import { isString, until } from '@vueuse/shared'
-import type { AxiosInstance, AxiosRequestConfig, AxiosResponse, CancelTokenSource } from 'axios'
+import type { AxiosInstance, RawAxiosRequestConfig, AxiosResponse, CancelTokenSource } from 'axios'
 import axios, { AxiosError } from 'axios'
 
 export interface UseAxiosReturn<T, R = AxiosResponse<T>, D = any> {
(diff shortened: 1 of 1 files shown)


laterdayi:
In my opinion, do you not need to upgrade useAxios first? At present, it seems that this is a compatible solution of Axios, and Axios may fix it later

https://github.com/silx-kit/h5web/pull/1353
https://github.com/axios/axios/issues/5476

lalalazero:
thanks! I'll downgrade to 1.2.2

laterdayi:
> thanks! I'll downgrade to 1.2.2

Personally, I think the current problems can be solved through demotion, and I think Axios is also looking for a solution. If it is finally determined that `AxiosRequestConfig ` is switched to `RawAxiosRequestConfig`, I think it is better to modify useAxios

lalalazero:
> > thanks! I'll downgrade to 1.2.2
> 
> Personally, I think the current problems can be solved through demotion, and I think Axios is also looking for a solution. If it is finally determined that `AxiosRequestConfig ` is switched to `RawAxiosRequestConfig`,I think it would be better to modify useAxios at this point

totally agree, when it's determined I'll reopen this pr. let me have this contribution ;)

PR #2680, opened 2023-01-26
fix(useMediaControls): ended status not updating

ferferga:
### Description

When the media element ends playing, the ``ended`` status is set to true. However, if the media element status change, changing the ``ended`` status variable is not taken into consideration.

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


wheatjs (approved):

hexiaokang (approved):
Great!

PR #2693, opened 2023-01-28
fix(useTimeoutFn): respect callback parameters in start function

### LINKED ISSUE #2691: `start` function of `Stoppable` doesn't receive arguments
### Describe the bug

One can provide a function with arguments to `useTimeoutFn` and then use the arguments in the `start` function returned.

Everything works like a charm but typescript isn't happy with that at all due to `Stoppable` being typed wrongly. It complains that `start` function expects 0 arguments. 

image

Also, as seen on the screenshot, user can't specify arguments' types as typescript complains about `string` not being compatible with `unknown` 

### Reproduction

https://stackblitz.com/edit/vitejs-vite-izdzrq/?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 7.17.0 - /usr/local/bin/npm
  npmPackages:
    @vueuse/core: ^9.11.1 => 9.11.1 
    vue: ^3.2.25 => 3.2.45
```

### Used Package Manager

npm

### Validations

wvffle:
### Description

fixes #2691

### Additional context

On L62 of useTimeoutFn, `start()` is used without any parameters. What if user defines a callback that requires arguments to work? Maybe a new option like `defaultArguments` could be introduced?

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/shared/useTimeoutFn/index.ts
+++ b/packages/shared/useTimeoutFn/index.ts
@@ -20,11 +20,11 @@ export interface UseTimeoutFnOptions {
  * @param interval
  * @param options
  */
-export function useTimeoutFn(
-  cb: (...args: unknown[]) => any,
+export function useTimeoutFn<CallbackFn extends(...args: any[]) => any>(
+  cb: CallbackFn,
   interval: MaybeComputedRef<number>,
   options: UseTimeoutFnOptions = {},
-): Stoppable {
+): Stoppable<Parameters<CallbackFn> | []> {
   const {
     immediate = true,
   } = options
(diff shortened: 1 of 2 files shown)


wvffle:
@antfu what's your view on the additional context provided in the PR description?

Provided:
```ts
const { start } = useTimeoutFn((arg1: string, arg2: number) =>{
  // ...
})
```

Due to explicitly added [`[]`](https://github.com/vueuse/vueuse/pull/2693/files#diff-652f6c9c6c0a0913b9c37949110a15d50053b42af103dddd3858fbdf36880f98R27) to the Stoppable type users can do `start()` without arguments even though they're required.

I've added it explicitly to allow for immediate start. Maybe there should be an option for the arguments passed to the immediate start call?

PR #2706, opened 2023-01-30
docs: fix typo

Alfred-Skyblue:
### Description
Fix demo typo

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other


Alfred-Skyblue:
@antfu It should be a delay problem that caused the test case to fail. I have not modified the test case that reported the error. Please confirm

Alfred-Skyblue:
The check failure is due to the short waiting time of the test case, resulting in the callback function of watch Throttled not being executed, which has been fixed in the [PR](https://github.com/vueuse/vueuse/pull/2707)

PR #2708, opened 2023-01-31
feat(useArrayIncludes): new function

Alfred-Skyblue:
### Description

This hook is used to determine whether the array contains an element

### Additional context

useArrayIncludes

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


antfu (changes requested):
Similar to #2710, I think a `compareFn` would make it more fixiable

packages/shared/useArrayIncludes/index.ts
```diff
@@ -0,0 +1,59 @@
+import type { ComputedRef } from 'vue-demi'
+import { computed } from 'vue-demi'
+import { containsProp, isObject } from '../utils'
+import type { MaybeComputedRef } from '../utils'
+import { resolveUnref } from '../resolveUnref'
+
+type ComparatorFn<T, V> = ((element: T, value: V, index: number, array: MaybeComputedRef<T>[]) => boolean)
+const isArrayIncludesOptions = <T, V>(obj: any): obj is UseArrayIncludesOptions<T, V> => isObject(obj)
+  && containsProp(obj, 'formIndex', 'comparator')
+interface UseArrayIncludesOptions<T, V> {
```
antfu, line 10:
```suggestion
export interface UseArrayIncludesOptions<T, V> {
```

antfu (approved):

PR #2711, opened 2023-01-31
feat(useFetch): update data on success

pbrzosko:
### Description

When people use useFetch, they expect certain type to be returned by the request, ie. an array of options to be provided to the select component. But this data is only present on successfull responses. In case of error, probably a totally different payload is returned. Having this payload set as a data, causes a lot of issues, because it is no longer what client of the useFetch expected.

### Additional context

IMHO data in case of error could be provided as a separate state. But in this PR, data is set to initialData provided in options or null, when an error happens.
Payload, when error happened, is provided to the onFetchError callback, so client of the useFetch composable can do whatever is needed to transform provided payload to some meaningful error.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -442,14 +442,15 @@ export function useFetch<T>(url: MaybeComputedRef<string>, ...args: any[]): UseF
 
           responseData = await fetchResponse[config.type]()
 
-          if (options.afterFetch && statusCode.value >= 200 && statusCode.value < 300)
-            ({ data: responseData } = await options.afterFetch({ data: responseData, response: fetchResponse }))
-
-          data.value = responseData
-
           // see: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/
-          if (!fetchResponse.ok)
+          if (!fetchResponse.ok) {
+            data.value = initialData
             throw new Error(fetchResponse.statusText)
+          }
+
+          if (options.afterFetch)
+            ({ data: responseData } = await options.afterFetch({ data: responseData, response: fetchResponse }))
+          data.value = responseData
 
           responseEvent.trigger(fetchResponse)
           return resolve(fetchResponse)
@@ -458,8 +459,7 @@ export function useFetch<T>(url: MaybeComputedRef<string>, ...args: any[]): UseF
           let errorData = fetchError.message || fetchError.name
 
           if (options.onFetchError)
-            ({ data: responseData, error: errorData } = await options.onFetchError({ data: responseData, error: fetchError, response: response.value }))
-          data.value = responseData
+            ({ error: errorData } = await options.onFetchError({ data: responseData, error: fetchError, response: response.value }))
           error.value = errorData
 
           errorEvent.trigger(fetchError)
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -349,7 +349,7 @@ export function useFetch<T>(url: MaybeComputedRef<string>, ...args: any[]): UseF
   const statusCode = ref<number | null>(null)
   const response = shallowRef<Response | null>(null)
   const error = shallowRef<any>(null)
-  const data = shallowRef<T | null>(initialData)
+  const data = shallowRef<T | null>(initialData || null)
 
   const canAbort = computed(() => supportsAbort && isFetching.value)
 
@@ -444,7 +444,7 @@ export function useFetch<T>(url: MaybeComputedRef<string>, ...args: any[]): UseF
 
           // see: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/
           if (!fetchResponse.ok) {
-            data.value = initialData
+            data.value = initialData || null
             throw new Error(fetchResponse.statusText)
           }
 
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -26,7 +26,7 @@ export interface UseFetchReturn<T> {
   error: Ref<any>
 
   /**
-   * The fetch response body, may either be JSON or text
+   * The fetch response body on success, may either be JSON or text
    */
   data: Ref<T | null>
 
--- a/packages/core/useFetch/index.test.ts
+++ b/packages/core/useFetch/index.test.ts
@@ -200,20 +200,20 @@ describe('useFetch', () => {
       baseUrl: 'https://example.com',
       options: {
         onFetchError(ctx) {
-          ctx.data.title = 'Global'
+          ctx.error = 'Global'
           return ctx
         },
       },
     })
-    const { data } = useMyFetch('test?status=400&json', {
+    const { error } = useMyFetch('test?status=400&json', {
       onFetchError(ctx) {
-        ctx.data.title += ' Local'
+        ctx.error += ' Local'
         return ctx
       },
     }).json()
 
     await retry(() => {
-      expect(data.value).toEqual(expect.objectContaining({ title: 'Global Local' }))
+      expect(error.value).toEqual('Global Local')
     })
   })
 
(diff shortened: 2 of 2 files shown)


pbrzosko:
Maintainers, what it your plan for this issue? I have created a PR that solves it: https://github.com/vueuse/vueuse/pull/2711. Can you either approve / disapprove?

packages/core/useFetch/index.ts
```diff
@@ -349,7 +349,7 @@ export function useFetch<T>(url: MaybeComputedRef<string>, ...args: any[]): UseF
   const statusCode = ref<number | null>(null)
   const response = shallowRef<Response | null>(null)
   const error = shallowRef<any>(null)
-  const data = shallowRef<T | null>(initialData)
+  const data = shallowRef<T | null>(initialData || null)
```
Glandos, line 352: Be careful, if `initialData` is falsy, it will be set as null, not expected.
E.g. if I set `{ initialData: false }` as option, then `data` will be set to `null` instead of `false`. Is it possible to use `??` instead of `||`?

aparajita:
FYI this is a BREAKING CHANGE and should have been documented as such. It is quite normal for an API to return error information along with an error status. Previously we could retrieve that response in `fetchResult.data`, but that code no longer works.

aparajita:
Also note value of `error` goes completely unused, which means the return value of `onFetchError` has been rendered completely useless.

https://github.com/pbrzosko/vueuse/blob/358acd2c455020dfc7e108f14d7c80e2d2421096/packages/core/useFetch/index.ts#L463

aparajita:
I honestly don't even see a way (other than terrible hacks) to get the response when an error status is returned. If so, this change was quite unfortunate.

orkhanahmadov:
I agree with @aparajita. This is an undocumented breaking change that caused a lot of headaches for us. We downgraded `@vueuse/core` to v9 because of it.

Not sure why this change was even needed... In our use case request can receive validation errors with 422 status code and actual validation errors as payload. We receive the errors list and display to the user. Now with this change, firstly, it is broken, secondly we can no longer access to `data` for validation errors, it is always null on error... We need to implement workarounds, which I believe was not necessary at all... you could always access `response.value.status` to check if the request was successful or not, why modify `data` instead...

antfu:
Open for PR to add a options to configure this behaviour

aparajita:
> Open for PR to add a options to configure this behaviour

@antfu If you are willing to take a look, I will be happy to submit one.

antfu:
Sure, let's directly discuss in the PR. Thanks!

pbrzosko:
> Also note value of `error` goes completely unused, which means the return value of `onFetchError` has been rendered completely useless.
> 
> https://github.com/pbrzosko/vueuse/blob/358acd2c455020dfc7e108f14d7c80e2d2421096/packages/core/useFetch/index.ts#L463

Not true. Value of error is fetchError.message || fetchError.name or whatever you set it to be in onFetchError callback.

pbrzosko:
> I honestly don't even see a way (other than terrible hacks) to get the response when an error status is returned. If so, this change was quite unfortunate.

No hacks needed. You just need to set whatever error you want in onFetchError. The main reason for change was that if you expect to receive apples from an API, you want it to be apples, not oranges if there is an error. So apples (data) should. be in data ref, and error (oranges) should be in the error ref. Event data ref is typed to be T, but apparently sometimes it is something else. I don't think that is right behavior and thus this change. If you do not use typescript, but javascript only and expect anything to be there, maybe you do not have such an issue.
And in onFetchError you can get access to error response, parse it and assign to error ref if you want or you can set a custom error or whatever you like.

pbrzosko:
> > But in this PR, data is set to initialData provided in options or null, when an error happens.
> 
> No @pbrzosko, it isn't. You are only doing that in the `.then` of the fetch and not in the `.catch`.
> 
> It really would be nice if you would go back and fix this mess. You broke error handling for everyone who uses this function.

I couldn't make it in the catch block, because then for a single moment data ref would hold the error (again not what it supposed to hold). But I agree, that it also should be set in catch block to the default value.

pbrzosko:
> I agree with @aparajita. This is an undocumented breaking change that caused a lot of headaches for us. We downgraded `@vueuse/core` to v9 because of it.
> 
> Not sure why this change was even needed... In our use case request can receive validation errors with 422 status code and actual validation errors as payload. We receive the errors list and display to the user. Now with this change, firstly, it is broken, secondly we can no longer access to `data` for validation errors, it is always null on error... We need to implement workarounds, which I believe was not necessary at all... you could always access `response.value.status` to check if the request was successful or not, why modify `data` instead...

And don't you think it is weird, that sometimes you have response type in data (T) and sometimes a list of errors?

aparajita:
> > Also note value of `error` goes completely unused, which means the return value of `onFetchError` has been rendered completely useless.
> > https://github.com/pbrzosko/vueuse/blob/358acd2c455020dfc7e108f14d7c80e2d2421096/packages/core/useFetch/index.ts#L463
> 
> @aparajita Not true. Value of error is fetchError.message || fetchError.name or whatever you set it to be in onFetchError callback.

You are ignoring the possibility that the API might return an error status AND a response body to give more information about what the error was. fetchError is useless in such a case.

orkhanahmadov:
@pbrzosko now trying to implement `onFetchError` I found out it somehow breaks certain parts of `useFetch`. If I'm setting request `headers` and also using `onFetchError` it somehow ignores and drops defined headers.

```js

useFetch('localhost:8081', {
  headers: { foo: 'bar' },
  onFetchError(ctx) { // defining `onFetchError` ignores `headers`. they won't get used
    return ctx
  }
})

useFetch('localhost:8081', {
  headers: { foo: 'bar' }, // without `onFetchError`, headers being used 
})
```

Am I missing something here? Seems like a bug. Currently with v10 if the endpoint returns a non-successful code with a body, seems like there's no way to access the body while also using headers.

I'll open a dedicated issue for this.

aparajita:
> > @orkhanahmadov And don't you think it is weird, that sometimes you have response type in data (T) and sometimes a list of errors?
> 
> 
> 
> @pbrzosko I don't think I do. If the endpoint is returning data alongside a non-successful status code, I would assume I can also access that data easily just the same way I access the status code. Take my validation errors example: yes endpoint does return non-ok code, but it also returns important validation errors that I need to access and use. Unsetting the data on error only complicates the implementation of this.

Exactly my point. There has to be a mechanism to retrieve the body when there is an error status. I'm fine with keeping it separate from data, as that actually simplifies the typing, but there should be a clean way to get the response body when response.ok === false.

danimalweb:
I've spent the last few hours puzzling over this myself. We have a 422 error being returned, and I need to access body data (validation)

Can anyone recommend a good pattern to work around this issue?  Or should I simply downgrade?

orkhanahmadov:
> I've spent the last few hours puzzling over this myself. We have a 422 error being returned, and I need to access body data (validation)
> 
> Can anyone recommend a good pattern to work around this issue? Or should I simply downgrade?

@danimalweb We tried to implement `onFetchError`, but it seems to be broken if you use it with request headers. As described [here](https://github.com/vueuse/vueuse/pull/2711#issuecomment-1526002607) and [here](https://github.com/vueuse/vueuse/issues/3040). We downgraded to v9.

aparajita:
> I've spent the last few hours puzzling over this myself. We have a 422 error being returned, and I need to access body data (validation)
> 
> Can anyone recommend a good pattern to work around this issue? Or should I simply downgrade?

I do this:

```ts
interface APIError {
  error: number
  message: string
}

type ExtendedError = Error & { apiError?: APIError }

useFetch(url, {
    onFetchError(ctx: {
      data: string | APIError
      response: Response | null
      error: ExtendedError
    }) {
      // If the response is a string and is supposed to be JSON,
      // try to parse it as JSON. We need to do this because many of our
      // endpoints return a success status and no body on success, and
      // an error status and JSON error information in the body. If you try to
      // use .json() on an empty body, useFetch throws an exception and breaks.

      const contentType = ctx.response?.headers.get('content-type') ?? ''

      if (contentType.startsWith('application/json')) {
        if (typeof ctx.data === 'string') {
          try {
            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- We're parsing JSON.
            ctx.error.apiError = JSON.parse(ctx.data)
          } catch {
            // Ignore
          }
        } else {
          // Already parsed, save it
          ctx.error.apiError = ctx.data
        }
      }

      return ctx
    },
```

The `error` that is then passed to `fetchResult.onFetchError` will have a `.apiError` property with the contents of the response body if there was an error.

climba03003:
Finally found why `10.1.2` failed.
The intention for `onFetchError` is centralize data handling.
See https://github.com/vueuse/vueuse/pull/841#issue-1023445126

Fetch is treating `400-500` status code as error even when the API is returning the data we expect.
So, it is the only hook (not right now) that allow developer to pass the correct data to `data.value`.

Yes, it is actually shipping breaking change as bugfix and also ruins the API expectation.

PR #2716, opened 2023-02-01
feat(useScroll): support scrollend event

kongmoumou:
### Description

Better scroll end event detection using native `scrollend` event.

### Additional context

See [Scrollend, a new JavaScript event](https://developer.chrome.com/blog/scrollend-a-new-javascript-event/), now shipped in Firefox 109.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useScroll/index.ts
+++ b/packages/core/useScroll/index.ts
@@ -56,6 +56,13 @@ export interface UseScrollOptions {
    * @default 'auto'
    */
   behavior?: MaybeComputedRef<ScrollBehavior>
+
+  /**
+   * trigger `onStop` on browser native `scrollend` event
+   * 
+   * @default false
+   */
+  useNativeScrollEnd?: boolean
 }
 
 /**
@@ -94,6 +101,7 @@ export function useScroll(
       passive: true,
     },
     behavior = 'auto',
+    useNativeScrollEnd = false,
   } = options
 
   const internalX = ref(0)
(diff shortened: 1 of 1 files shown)


packages/core/useScroll/index.ts
```diff
@@ -182,7 +191,7 @@ export function useScroll(
     internalY.value = scrollTop
 
     isScrolling.value = true
-    onScrollEnd(e)
+    !useNativeScrollEnd && onScrollEndDebounced(e)
```
vaakian, line 194: I think we could both have:
add the none debounced  `onScrollEnd` on `scrollend` event,
and keep the `onScrollEndDebounced` in `onScrollHandler`.

So we can get scroll end state a little bit faster if `scrollend` event is supported, otherwise it works as normal as before. And no extra option is needed.

kongmoumou: But it may be invoke `onStop` callback twice, one by scrollend, the other by debounced handler. Is it expected?

vaakian: We could do something like:
```js
const onScrollEnd = (e: Event) => {
  // dedupe the second call
  if (!isScrolling.value) return

  isScrolling.value = false
  directions.left = false
  directions.right = false
  directions.top = false
  directions.bottom = false
  onStop(e)
}
```

kongmoumou: got it and done~

PR #2725, opened 2023-02-02
feat(useDraggable): support `capture` option

### LINKED ISSUE #2724: useDraggable support trigger in bubbling phase

Currently, useDraggable trigger event in capturing phase, but somtimes there are draggable element nested in the draggable element, if we drag the one inside, it will affect the one outside.

### Suggested solution

provide a option helping user choose weather trigger in bubbling phase

### Alternative

_No response_

### Additional context

_No response_

### Validations

faga295:
closed #2724 

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


packages/core/useDraggable/index.ts
```diff
@@ -135,9 +141,10 @@ export function useDraggable(target: MaybeComputedRef<HTMLElement | SVGElement |
   }
 
   if (isClient) {
-    useEventListener(draggingHandle, 'pointerdown', start, true)
-    useEventListener(draggingElement, 'pointermove', move, true)
-    useEventListener(draggingElement, 'pointerup', end, true)
+    const useCapture = resolveUnref(options?.useCapture ?? true)
+    useEventListener(draggingHandle, 'pointerdown', start, useCapture)
+    useEventListener(draggingElement, 'pointermove', move, useCapture)
+    useEventListener(draggingElement, 'pointerup', end, useCapture)
```
vaakian, line 147:
```suggestion
    let cleanups: Fn[] = []
    watch(resolveRef(options.useCapture), (useCapture = true) => {
      cleanups.forEach(cleanup => cleanup())
      cleanups = [
        useEventListener(draggingHandle, 'pointerdown', start, useCapture),
        useEventListener(draggingElement, 'pointermove', move, useCapture),
        useEventListener(draggingElement, 'pointerup', end, useCapture),
      ]
    }, { immediate: true })
```
eventListeners should be re-registered when `useCapture` ref value changes.

vaakian (changes requested):

packages/core/useDraggable/index.ts
faga295: Thank you for your suggestion!

packages/core/useDraggable/index.ts
```diff
@@ -135,9 +141,16 @@ export function useDraggable(target: MaybeComputedRef<HTMLElement | SVGElement |
...
+    let cleanups: Fn[] = []
+    const useCapture = computed(() => resolveUnref(options?.useCapture ?? true))
+    watch(useCapture, (useCapture = true) => {
+      cleanups.forEach(cleanup => cleanup())
+      cleanups = [
+        useEventListener(draggingHandle, 'pointerdown', start, useCapture),
+        useEventListener(draggingElement, 'pointermove', move, useCapture),
+        useEventListener(draggingElement, 'pointerup', end, useCapture),
+      ]
+    }, { immediate: true })
```
antfu, line 153: I think we could make `useEventListener` able to accept ref for the options, so we don't need the workaround here.

antfu (changes requested):
Please refactor using https://github.com/vueuse/vueuse/commit/103d6c03fa5bb5f3415337b7626dc334175a0e50

Thanks

faga295:
Has refactored, please review again

artel1992:
I would recommend adding this code in case someone decides to call 'open' twice, once with 'capture' and once without it:
     if (hasOwn(_options, 'capture'))
            input.capture = _options.capture!
        else
            input.removeAttribute('capture')

PR #2731, opened 2023-02-05
Added editorRef composable

### LINKED ISSUE #2730: Editor components automatic ref for values

Imagine you have a popup window, like a dialog box, for editing a set values. In this scenario we need to have a copy of the original value (usually provided via `modelValue` or just `value`) that is one-way synchronized to the prop. This means that you can change the value locally, and it will stay that way until the prop value that it was originally retrieved from changes.

That way, we have a ref that can be used with v-model inside of the dialog box that gets initialized and automatically updated whenever that prop changes. The upside is that we can directly v-model bind it to inputs and we have full control over when that value gets propagated upwards (which will mean then the prop value will change and the local copy will get updated with the latest changes).

### Suggested solution

I think a good name for it would be `editorRef`

### Alternative

Alternatively, one uses a watch from options API and a local data field to accomplish the same result.

The following video describes how the editor components are created: https://youtu.be/RmH9_ly5Tew

### Additional context

_No response_

### Validations

padcom:
### Description

Make a ref to copy of given field of the given object that gets updated when the original object's property gets changed.

Closes #2730

### Additional context

This is used in editor components where propagation of the value needs to be controlled and needs to happen once the user decided to accept the changes.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
Thanks. But I think we already have https://vueuse.org/core/useCloned/#usecloned to do the same.

Issue #2689, opened 2023-01-28
Expose explicit transition utility


scottbedard:
### Clear and concise description of the problem

In animation projects I often find myself needing one-way transitions. That is, a transition from A to B, where A and B are explicit values.

### Suggested solution

A `transition` utility to interpolate refs from one value to another. The `to` and `from` properties should be numbers or vectors, and all options of `useTransition` should also be supported. This function should return a `Promise` to allow for async / await choreography of values. Below is an example of a chained transition.

```js
import { transition } from '@vueuse/core'

const first = ref(0)

const second = ref(0)

async function start() {
  await transition(first, { from, to })

  await transition(second, { from, to })
}
```

### Alternative

This logic can be created with [`useTransition`](https://vueuse.org/core/usetransition/#usetransition), but the mechanism to do so is complicated and prone to human error.

1. Create a `disabled` ref to manage the transition
2. Set `disabled` to `true`, and source to it's start value
3. Wait one render cycle via `nextTick`
4. Set the `disabled` transition to `false`
5. Set the source to it's new value
6. Repeat...

[...]

PR #2743, opened 2023-02-09
feat(useTransition): expose transition utility for manual control

scottbedard:
### Description

Expose underlying transition logic from `useTransition` to allow for manually transitioning a ref between values. See more information here https://github.com/vueuse/vueuse/issues/2689.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useTransition/index.ts
+++ b/packages/core/useTransition/index.ts
@@ -1,5 +1,5 @@
 import { computed, ref, unref, watch } from 'vue-demi'
-import { isFunction, isNumber, identity as linear, promiseTimeout } from '@vueuse/shared'
+import { isFunction, isNumber, identity as linear, promiseTimeout, tryOnScopeDispose } from '@vueuse/shared'
 import type { ComputedRef, Ref } from 'vue-demi'
 import type { MaybeRef } from '@vueuse/shared'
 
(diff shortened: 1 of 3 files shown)


packages/core/useTransition/index.ts
```diff
@@ -115,6 +122,68 @@ function createEasingFunction([p0, p1, p2, p3]: CubicBezierPoints): EasingFuncti
...
+
+/**
+ * Transition from one value to another.
+ *
+ * @param source
+ * @param from
+ * @param to
+ * @param options
+ */
+export function transition<T extends number | number[]>(
```
antfu, line 137: The name to be is a bit too general to me, can we have a more explicit name, like

`createTransitionPromise` or `createTransition`?

scottbedard: Personally, I like the name `transition` because it works as a compliment to `useTransition`. In other words, it's what `useTransition` is actually using. I'm open to changing though, but I don't think "create" is a good fit, since it operates on an existing ref rather than creating anything. Maybe something like `transitionRef`, `executeTransition`, `runTransition`, or `performTransition` could work...

I think of those choices, `transitionRef` is probably my favorite. I'm not sure though, what do you think? 🤔

antfu: I am ok with `performTransition` or `executeTransition`.

But `transitionRef` can be confusing as we have `deboundedRef` etc, which returns a `ref` - but in our case we didn't right?

scottbedard: I had time to try these names out over the weekend and I think `executeTransition` felt nicer, I'll make the change 🙂

antfu (approved):

Issue #2737, opened 2023-02-07
Infinite scroll doesn't work with column-reverse containers.


nablaFox:
### Describe the bug

Following the docs:

```
const target = ref<HTMLElement | null>(null)

useInfiniteScroll(
    target,
    () => { alert('ok') },
    { direction: 'top' }
)
```

with this template structure:

```
<div class="data-wrapper" ref="target">
    <div v-for="d in data" >
    	{{ d }}
    </div>
</div>
```
if `data-wrapper` has `flex-direction: column-reverse` useInfiniteScroll simply doesn't work.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-x3p5jz?file=src/App.vue

### System Info

```Shell
System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
    Memory: 1.91 GB / 7.61 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 19.5.0 - /usr/bin/node
    npm: 9.3.1 - /usr/bin/npm
  Browsers:
    Chrome: 109.0.5414.119
  npmPackages:
    @vueuse/core: ^9.11.1 => 9.11.1 
    vue: ^3.2.45 => 3.2.45
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2577, opened 2022-12-24
fix(useScroll): add support for row-reverse and column-reverse

OneLoneFox:
### Description

Fixes #2576 

also fixes incorrect calculation of directions.right mentioned in the same issue

---

### What is the purpose of this pull request? 

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Before submitting the PR, please make sure you do the following

- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [X] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [X] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [] Ideally, include relevant tests that fail without this PR but pass with it.

PR #2745, opened 2023-02-09
fix(useScroll): handle reverse flex direction

joris-gallot:
### Description

fixes https://github.com/vueuse/vueuse/issues/2737

When defining the `flex-direction` style of the html element as `column-reverse` or `row-reverse`, the `arrivedState` property must change the way its properties are calculated

### Additional context

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


antfu:
Close in favor of #2577, thanks!

PR #2748, opened 2023-02-09
fix(useEventBus,createEventHook)!: dedupe registration of the same function

### LINKED ISSUE #2747: bug: event fired multiple times
### Describe the bug

I think in the `EventBus`, the same event, the same listener, if it is listened to multiple times, should only be triggered once.

At present, it is not deduplicated in useEventBus.
```js
function fn() {
      console.log('hello world')
}
const { on, emit } = useEventBus('test')
on(fn)
on(fn)
emit()
// hello world x 2
```

### Reproduction

https://codesandbox.io/s/peaceful-kalam-yp9bg9?file=/src/App.vue

### System Info

```Shell
System:
    OS: macOS 13.2
    CPU: (8) arm64 Apple M1 Pro
    Memory: 45.00 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.0 - ~/.nvm/versions/node/v16.18.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.18.0/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v16.18.0/bin/npm
  Browsers:
    Chrome: 109.0.5414.119
  npmPackages:
    @vueuse/core: 9.12.0
    vue: ^3.2.45
```

### Used Package Manager

yarn

### Validations

Alfred-Skyblue:
### Description

closes #2747

When the same function is bound to the same event multiple times, it will be triggered multiple times.

### Additional context

`useEventBus、createEventHook`

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


wheatjs:
Makes sense to me, but I'd suggest we ship it in v10 in case it ends up being a breaking change in ways we don't see.

wheatjs (approved):

PR #2750, opened 2023-02-10
feat(useFetch): cancel previous request

### LINKED ISSUE #2677: useFetch - options for gracefully refetching

When using useFetch, If you are writing to the url ref fairly quickly, there's no guarantee that the last url value in useFetch will be the one returned. You can test this fairly easily by quickly changing the url parameter on the useFetch page with refetch on:
https://vueuse.org/core/useFetch/

Since the requests can return out of order, you can run into issues where you get a 404 on a correct url or have the error become out of sync:

image
image

### Suggested solution

Could this perhaps be addressed by allowing previous requests to be automatically aborted if the url value changes?

### Alternative

_No response_

### Additional context

_No response_

### Validations

Alfred-Skyblue:
### Description

closes #2677 

When executing requests continuously or monitoring url ref changes, the previous request will be automatically terminated to ensure that the obtained data is the most current requested data.

### Additional context

useFetch

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useFetch/index.test.ts
+++ b/packages/core/useFetch/index.test.ts
@@ -1,5 +1,5 @@
 import { until } from '@vueuse/shared'
-import { ref } from 'vue-demi'
+import { nextTick, ref } from 'vue-demi'
 import type { SpyInstance } from 'vitest'
 import { retry } from '../../.test'
 import { createFetch, useFetch } from '.'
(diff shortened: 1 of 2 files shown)


wheatjs:
Makes sense, although I wonder if we should make this an option. I could think of a scenario where perhaps you are logging data to a server and making multiple rapid requests to the endpoint. In that instance I don't think we would want to cancel the requests.

Alfred-Skyblue:
@wheatjs I think the cancellation request is unilateral, the server does not stop execution just because the front end cancels the request, it cancels the request just to ensure that the data obtained is up to date

wheatjs:
Ah you are right, I was thinking maybe on slow networks before the request was sent, but wasn't able to actually reproduce that scenario

wheatjs (approved):

dsolerh:
This change breaks my code, I'm using the `createFech` utility and now all my requests are being canceled for the first time.

s-prechtl:
I'm using `refetch: true` and the reqeusts are cancelled, but I get the error state: `AbortError: The user aborted a request.`, even though the last request sent ends in status code `200` and sends data. Am I missing something, or is this a bug?

PR #2751, opened 2023-02-10
fix(useFetch): fix fetch error data update

### LINKED ISSUE #2674: useFetch should update data only if request was successfull

I have an API, which returns error message, when a request fails. So imagine an endpoint which returns list of options that are supplied to the dropdown component. This component expects an Array. If I execute such a request with useFetch and it fails because of an error, API returns not an array but an error message. And then this error message is in the data state of the useFetch, so I can not use it as options for the combobox.

### Suggested solution

I think data state should only be changed in 2xx responses. Returned data in errorneous requests should be stored in a different state - maybe errorData? For errorneous request data should probably contain initialData config value.

### Alternative

Now I have to do workaround, add computed property over data ref and return it only if status is 2xx, otherwise an empty array.

### Additional context

_No response_

### Validations

Alfred-Skyblue:
### Description

fixed #2674 

useFetch should update data only if request was successfull, but I'm not sure if such a modification would cause issues with history usage

### Additional context
useFetch

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useFetch/index.test.ts
+++ b/packages/core/useFetch/index.test.ts
@@ -552,11 +552,7 @@ describe('useFetch', () => {
 
   test('should run the onFetchError function when network error', async () => {
     const { data, statusCode } = useFetch('https://example.com?status=500&text=Internal%20Server%20Error', {
-      onFetchError(ctx) {
-        ctx.data = { title: 'Hunter x Hunter' }
-
-        return ctx
-      },
+      initialData: { title: 'Hunter x Hunter' },
     }).json()
 
     await retry(() => {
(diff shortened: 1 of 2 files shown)


pbrzosko:
@Alfred-Skyblue changes you did are not enough. Because when you have an error response (like not 2xx), you also set the response to data.value.  Additionally I think that data should be set to initialValue if it is present on error.

Please have a look at the PR here (somehow I wasn't able to connect it to the issue #2674):
https://github.com/vueuse/vueuse/pull/2711

Alfred-Skyblue:
@pbrzosko I think that whether the change responds successfully or responds incorrectly, the changes to this hooks are destructive, so I closed this PR

pbrzosko:
@Alfred-Skyblue so I have no idea, what your plan regarding #2711 is, but please have a look here from lines 149 to 168: 
https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/asyncData.ts
Data is only set on success, on error it is set to initialValue or null. I still think that ticket #2674 is valid and it doesn't make sense, that you expect some type of the response but suddenly you get something else. But I do understand, that this is a breaking change.

PR #2770, opened 2023-02-15
feat: add useLottiePlayer

Zourunfa:
>

### Description

    I usually use lottie animation in projects, and also use vueuse in projects. I hope to integrate lottie-web into vueuse. lottie animation has good cross-platform features, and I have encapsulated some new methods that are not available in the native api of lottie, such as controlling the number and time of animation cycle playback, which is conducive to controlling lottie animation in combination with requests by segments. If I can accept this commitment, I will add it later
### Additional context

---

- [ ] Bug fix
- [1 ] New Feature
- [ ] Documentation update
- [ ] Other

- [ 1] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [ 1] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [ 1] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [1 ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ 1] Ideally, include relevant tests that fail without this PR but pass with it.


antfu:
Thank you - but I think this belongs more to your userland instead of VueUse. It probably not worth it to maintain so many things and docs for just 3 lines of implementation.

Zourunfa:
> Thank you - but I think this belongs more to your userland instead of VueUse. It probably not worth it to maintain so many things and docs for just 3 lines of implementation.
Thank you for your reply. I think it's reasonable. I'm a vegetable chicken. This is just a small attempt. I may improve more functions and submit it next time

Issue #2497, opened 2022-12-03
UseAsyncValidator doesn't work for SSR


theruziev:
### Describe the bug

useAsyncValidator doesn't work when SSR is enabled. 

I found a discussion and several issues about it:
https://github.com/vueuse/vueuse/discussions/2402
https://github.com/yiminghe/async-validator/issues/312

### Reproduction

https://stackblitz.com/edit/github-61yzgr?file=app.vue

### System Info

```Shell
System:
    OS: macOS 13.0.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 39.98 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.0 - /usr/local/bin/node
    Yarn: 3.2.3 - /usr/local/bin/yarn
    npm: 8.18.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 108.0.5359.94
    Firefox: 107.0
    Safari: 16.1
  npmPackages:
    @vueuse/core: ^9.6.0 => 9.6.0 
    @vueuse/integrations: ^9.6.0 => 9.6.0 
    @vueuse/nuxt: ^9.6.0 => 9.6.0
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

kevohagan:
have the same problem with Astro too!

theruziev:
@kevohagan 

I copy the source code of useAsyncValidation and store it in the project. I think you can use this workaround for now

theruziev:
[...]

PR #2788, opened 2023-02-18
fix(useAsyncValidator): incorrect condition for choosing constructor

theruziev:
### Description

I made a stupid mistake on #2761 😔, my solution is based on my comment on the issue https://github.com/vueuse/vueuse/issues/2497#issuecomment-1416852748. 
image

I try to get the constructor from the default attribute (it exists on SSR mode), if doesn't exist use Schema like before PR. 
Nothing broke because it didn't work before, but it is good to fix one. 

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/integrations/useAsyncValidator/index.ts
+++ b/packages/integrations/useAsyncValidator/index.ts
@@ -6,7 +6,7 @@ import type { Ref } from 'vue-demi'
 import { computed, ref, watchEffect } from 'vue-demi'
 
 // @ts-expect-error Schema.default is exist in ssr mode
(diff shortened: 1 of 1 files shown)


theruziev:
@okxiaoliang4 sorry for mentioning but you are in context because you review #2761. 
I can't provide of example on stackblitz for checking this fix work because stackblitz doesn't permit installation dependencies from github
I checked by editing a file in node_modules and run a project.

PR #2790, opened 2023-02-19
feat(createGlobalState): allow passing initial args

jd-solanki:
### Description

[`createGlobalState`](https://vueuse.org/shared/createGlobalState/) is nice and I wanted to use it for configuring [Anu](https://anu-vue.netlify.app/) instance when vue uses it as a plugin. However, `createGlobalState` doesn't accepts any initial config for initializing global state based on provided config.

### Additional context

This is really simple change but don't forget to review it 👀 

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/shared/createGlobalState/index.ts
+++ b/packages/shared/createGlobalState/index.ts
@@ -1,6 +1,6 @@
 import { effectScope } from 'vue-demi'
 
-export type CreateGlobalStateReturn<T> = () => T
+export type CreateGlobalStateReturn<T> = (...args: any[]) => T
 
 /**
  * Keep states in the global scope to be reusable across Vue instances.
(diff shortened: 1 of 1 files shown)


jd-solanki:
Still, I'm not satisfied with the API it provides. I can initialize the composable with options and don't want to require them later. However, when we use it, every time it gives `...args: any[]` as a type of argument for composable.

Moreover, We can include a generic type of accepting the type of params we can pass to avoid errors while defining options:
```ts
// Nothing, I can pass anything I want as options
useAnu({
  initialTheme: config.initialTheme,
  themes: config.themes,
})

// We can add nice autocompletion and add restrictions for error detection
useAnu<AnuOptions>({
  initialTheme: config.initialTheme,
  themes: config.themes,
})
```

Feel free to update the PR 😇

antfu (changes requested):
In general it's ok, but can we make it type-safe? Thanks

jd-solanki:
Unfortunately, I'm not good enough with typescript to improve this, I'm still improving my TS game. I referenced [createSharedComposable](https://github.com/vueuse/vueuse/blob/main/packages/shared/createSharedComposable/index.ts).

I hope you will understand, Thanks.

antfu (approved):

PR #2800, opened 2023-02-21
feat(useCssVar): introduce `observe` option

### LINKED ISSUE #2799: useCssVar - css variables out of sync
### Describe the bug

When I use useCssVar to get css variables, I can get the value of the variable, but if I use setProperty to modify the value of the variable, then the variable value obtained by useCssVar will not change.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-tkmlit?file=src%2FApp.vue&terminal=dev

### System Info

```Shell
all
```

### Used Package Manager

pnpm

### Validations

Alfred-Skyblue:
>

### Description

fixed #2799 

Fix css variables out of sync.

I rewrote the setProperty method for HTMLElement, so that the variable.value can be updated when the css variable is set, but I don't know if this is the best solution, or useMutationObserver to monitor DOM property changes, but I think performance will be worse.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useCssVar/index.ts
+++ b/packages/core/useCssVar/index.ts
@@ -10,10 +10,10 @@ import { unrefElement } from '../unrefElement'
 export interface UseCssVarOptions extends ConfigurableWindow {
   initialValue?: string
   /**
-   * Whether to update synchronously when CSS variables are modified by other means
+   * Use MutationObserver to monitor variable changes
    * @default false
    */
-  sync?: boolean
+  observe?: boolean
 }
 
 /**
@@ -29,7 +29,7 @@ export function useCssVar(
   target?: MaybeElementRef,
   options: UseCssVarOptions = {},
 ) {
-  const { window = defaultWindow, initialValue = '', sync = false } = options
+  const { window = defaultWindow, initialValue = '', observe = false } = options
   const variable = ref(initialValue)
   const elRef = computed(() => unrefElement(target) || window?.document?.documentElement)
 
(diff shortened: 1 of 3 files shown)


packages/core/useCssVar/index.ts
```diff
@@ -27,6 +27,18 @@ export function useCssVar(
...
+  tryOnMounted(() => {
+    if (elRef.value) {
+      const oldSetProperty = elRef.value.style.setProperty
+      elRef.value.style.setProperty = function (...args) {
+        const [key, value] = args
+        if (key === resolveUnref(prop) && value !== variable.value)
+          variable.value = value!
+        return oldSetProperty.apply(this, args)
+      }
+    }
```
antfu, line 39: This is probably not a good idea to monkey patch the native functions, we either need a way to listen to the change, or we just list as a limitation and discourage the usage of directly setProperty

antfu (changes requested):

packages/core/useCssVar/index.ts
Alfred-Skyblue: We can't prevent other libraries from using setProperty to modify css var. I saw in the [vue3 source code](https://github.com/vuejs/core/blob/main/packages/runtime-dom/src/helpers/useCssVars.ts) that it uses MutationObserver to monitor changes in css var. Can we also use MutationObserver to monitor?

antfu: Yeah we should use `MutationObserver`, and adding a flag to enable the watcher so ppl don't need it do pay the cost

Alfred-Skyblue: I added a sync option to enable MutationObserver, which is disabled by default .

antfu (approved):

PR #2807, opened 2023-02-22
fix(useAxios)!: `error` should return type `unknown`

jbaubree:
>

### Description

fix(useAxios): add typing for onError param

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


Alfred-Skyblue:
In the response interceptor, what we throw out is not necessarily AxiosError type.

jbaubree:
@Alfred-Skyblue i don't understand why error is typed with AxiosError ?
Capture d’écran 2023-02-23 à 09 01 04

antfu:
Yeah, I think instead, the return `error` ref should be typed as `unknown`

Alfred-Skyblue:
> Yeah, I think instead, the return `error` ref should be typed as `unknown`

I thought it was a breaking change, but it's right

antfu (approved):

PR #2821, opened 2023-02-27
fix(useWebNotification)!: should be exposed as EventHooksOn

Alfred-Skyblue:
>

### Description

In the documentation, I found that if we use `onClick/onShow/onError/onClose` these methods, we need to pass a callback function, but `useWebNotification` exports `EventHook` type, I think we should export `EventHookOn`, so we can just use onClick instead of onClick.on, but this is a destructive change. I think it should be changed in version 10.0. In order to be compatible with the original `onClick.on` listening event, a layer of conversion is made here to convert it to the type of `EventHookOn & EventHook` .

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

useWebNotification


Alfred-Skyblue:
In the latest submission, in order to be compatible with the original onClick.on listening event, I made a layer of conversion `onClick/onShow/onError/onClose` and converted them into `EventHookOn<any> & EventHook<any>` types for compatibility with historical versions.

PR #2824, opened 2023-02-28
feat(createEventHook): allow awaiting on `trigger`

Akryum:
### Description

Allows:

```js
const hook = createEventHook()

async function submit () {
  // ...
  await hook.trigger()
}

return {
  submit,
  onAfterSubmit: hook.on,
}
```

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


6XGate:
The `on` function should accept a function that return `void | Promise<void>`. This will trigger ESLint warnings with the TypeScript or promise plugin when passing an async function to `on`.

PR #2834, opened 2023-03-02
Fix misleading documentation on use of refetch option

pzoghbi:
Refetch option works only if `url` parameter is a callable.

After spending some time reading docs and trying to refetch on url change i've noticed it wouldn't work until i found [an issue ](https://github.com/nuxt/nuxt/issues/14243) where someone wrote a solution which fixed it.

Otherwise, refetch won't work.

>

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useFetch/index.md
+++ b/packages/core/useFetch/index.md
@@ -37,12 +37,12 @@ const { isFetching, error, data } = await useFetch(url)
 
 ### Refetching on URL change
 
-Using a `ref` for the url parameter will allow the `useFetch` function to automatically trigger another request when the url is changed.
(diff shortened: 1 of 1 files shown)


antfu:
I think it's not related. The issue you linked is Nuxt's `useFetch`, not VueUse's.

pzoghbi:
You're right, the solution and the issue I mentioned were related to Nuxt.
However, I'm not using nuxt, I'm using `createFetch` from @vueuse/core.
It just happens to be the exact solution for both. 

Would you consider looking into this and maybe re-opening the merge request?

PR #2835, opened 2023-03-02
feat: `immediateCallback` for `useTimestamp` & `useRafFn`

Waleed-KH:
### Description

`immediateCallback` option for `useTimestamp` & `useRafFn` functions

### Additional context

---

- [x] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useRafFn/index.ts
+++ b/packages/core/useRafFn/index.ts
@@ -23,6 +23,13 @@ export interface UseRafFnOptions extends ConfigurableWindow {
    * @default true
    */
   immediate?: boolean
+
+  /**
+   * Execute the callback immediate after calling this function
+   *
+   * @default false
+   */
+  immediateCallback?: boolean
 }
 
 /**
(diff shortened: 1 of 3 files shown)


antfu:
Thanks for contribution. I understand the usage but I feel the complexity is not worth for this as it's easy to workaround on the the userland. I'd prefer to now have them for now. Thanks.

Waleed-KH:
the workaround has a down side which is the inability of using inline callback:

```ts
// with the immediateCallback option 
useTimestamp({ immediateCallback: true, callback() {
	// ...
}})

// workaround 
function myCb() {
	// ...
}

useTimestamp({ callback: myCb });
myCb();
```

also it's just an option to invoke the function immediately, so I don't think it add any complexity, and `useIntervalFn` already has it.

antfu:
Yes, but you see, every callback option has this problem and we certainly don't want to add `immediateCallback` to every single function and every single callback. On top of that, I can say with `immediateCallback` make its inability to pass arguments, so maybe we should add `immediateCallbackArgs` as well? We have to justify "how much it cost" and "how much it helps".

I am not saying your usage is not legal. But please understand that one of the maintainers' jobs is to guard designs and try to prevent [Slippery slope](https://en.wikipedia.org/wiki/Slippery_slope) to happen. 

> useIntervalFn already has it

That is a classic one :P - Actually thanks for mentioning it, I would actually consider removing it instead of expanding this idea to every function.

Waleed-KH:
good point 👍

Issue #2814, opened 2023-02-24
🚀 Transferring useful composables from Anu UI lib to VueUse


jd-solanki:
### Clear and concise description of the problem

Hi @antfu,

As we will be releasing v10 of VueUse which will have major features I want to contribute some composables from [Anu](https://github.com/jd-solanki/anu).

Following "Slowing down new functions" warning I'm creating this issue to ask for approval for adding these composables.

Below is list of composables I'm willing to transfer to VueUse:
1. [`useParentElement`](https://github.com/jd-solanki/anu/blob/main/packages/anu-vue/src/composables/useParentEl.ts): Get parent element of component
2. [`useInternalBooleanState`](https://github.com/jd-solanki/anu/blob/main/packages/anu-vue/src/composables/useInternalBooleanState.ts)*: Inspired from Michael Thiessen's [newsletter](https://twitter.com/MichaelThiessen/status/1617969242992680961) this creates internal state for prop. (_*existing implementation in anu is for boolean prop and will be generic in VueUse_)
3.  (& 4) [`useSearch`](https://github.com/jd-solanki/anu/blob/main/packages/anu-vue/src/composables/useSearch.ts) & [`useSort`](https://github.com/jd-solanki/anu/blob/main/packages/anu-vue/src/composables/useSort.ts): Both composables can be great addition to VueUse as this will allow creating complex things like autocomplete & tables. Anu already uses it internally for table and in needed areas. However, I need some help from VueUse community to improve it further for general usage.
5. [`useGroupModel`](https://github.com/jd-solanki/anu/blob/main/packages/anu-vue/src/composables/useGroupModel.ts): My most favorite and super useful composable that let's you create v-model like binding for group. [Here's](https://anu-vue.netlify.app/guide/composables/usegroupmodel.html) demo on what it can offer. It is also used internally for components like select.

I will be more than happy to see these composables in VueUse itself if you allow. Definitely I will need some help with TypeScript to improve it. I guess @okxiaoliang4 can help me with getting this composables to VueUse if you allow this.

Please let me know your thoughts on this, Thanks ❤️  

### Suggested solution

n/a

### Alternative

n/a

### Additional context

I'm also planning for a new release of Anu and now thinking of making a release with the update of VueUse v10. May I know what will be ETA of v10 to better organize Anu release?

Regards,
JD

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
[...]

PR #2855, opened 2023-03-09
feat(useParentElement): new function

jd-solanki:
### Description

This function allows getting the parent element of the current element or passed `ref`.

### Additional context

Porting `useParentElement` from [Anu](https://github.com/jd-solanki/anu) to VueUse as discussed here: https://github.com/vueuse/vueuse/issues/2814

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


vaakian:
BTW, I am wondering if it's possible to aware that a `Element` was moved to another `ParentElement`?
```html
<div id="container1">
  <div id="target"></div>
</div>
<div id="container2">
  <!-- move #target here -->
</div>

<script>
    const container2 = document.querySelector('#container2')
    const target = document.querySelector('#target')
    // move #target to #container2
    container2.appendChild(target)
</script>
```

packages/core/useParentElement/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  tryOnMounted(() => {
+    const el = unrefElement(resolveUnref(element) || getCurrentInstance()?.proxy)
+    if (el)
+      parentElement.value = el.parentElement
+  })
+
+  return {
+    parentElement,
+  }
```
antfu, line 17:
```suggestion
  return parentElement
```

Do you foresee any other things to return in the future?

packages/core/useParentElement/index.ts
```diff
@@ -0,0 +1,18 @@
+import type { MaybeComputedRef } from '@vueuse/shared'
+import { resolveUnref, tryOnMounted } from '@vueuse/shared'
+import { getCurrentInstance, ref } from 'vue-demi'
+import { unrefElement } from '../unrefElement'
+
+export const useParentElement = (element?: MaybeComputedRef<HTMLElement | SVGElement | null | undefined>) => {
```
antfu, line 6: Please use `function` on top-level

antfu (approved):

antfu (approved):

PR #2856, opened 2023-03-09
Fix misleading documentation using refetch option

pzoghbi:
Hello. I noticed that the documentation has left out that the url ref has to be called when using `refetch: true`.

I've tried this in two different projects, and it worked the same in both.  
I'm using `useFetch` from `@vueuse/core`  
If url ref isn't passed as a callable function, no refetching occurs.

Additionally, I've checked the tests, and there isn't a test for `refetch: true` that checks if refetch was called with the new url.

### DIFF
--- a/packages/core/useFetch/index.md
+++ b/packages/core/useFetch/index.md
@@ -37,12 +37,12 @@ const { isFetching, error, data } = await useFetch(url)
 
 ### Refetching on URL change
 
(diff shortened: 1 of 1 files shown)


antfu:
I don't think this is the case. It might be a bug, please open an issue with reproduction first. Thanks.

PR #2859, opened 2023-03-10
Docs: Add examples to demonstrate useFetch requests that depend on previous requests.

dflock:
>

### Description

Add examples to `useFetch` docs to demonstrate requests that depend on a previous requests.

### Additional context

It took me a while to figure out an idiomatic way to do this - and it would have saved me a lot of time if this was documented, so I added my solution.

If there's a better way to do chained/nested requests with `useFetch`, please feel free to modify this - but it would be very useful to document this, as it's a very common use case.

---

- [ ] Bug fix
- [ ] New Feature
- [X] Documentation update
- [ ] Other


antfu:
Thank you, but I think this requirement is a bit too specific. And btw, there is a usePrevious utility: https://vueuse.org/core/usePrevious/

PR #2864, opened 2023-03-13
feat(useCycleList): allow receiving reactive list

hjJunior:
### Description

Allow useCycleList to receive a "ref" list value, this will allow the cycle list to be reactive

### Additional context

This also adds unit tests for useCycleList

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [x] Other

### DIFF
--- a/packages/core/useCycleList/demo.vue
+++ b/packages/core/useCycleList/demo.vue
@@ -1,7 +1,8 @@
 <script setup lang="ts">
 import { useCycleList } from '@vueuse/core'
+import { ref } from 'vue'
 
-const list = [
+const list = ref([
   'Dog',
   'Cat',
   'Lizard',
@@ -10,7 +11,7 @@ const list = [
   'Dolphin',
   'Octopus',
   'Seal',
-]
+])
 
 const { state, next, prev } = useCycleList(list)
 </script>
--- a/packages/core/useCycleList/index.md
+++ b/packages/core/useCycleList/index.md
@@ -13,7 +13,7 @@ Cycle through a list of items.
 ```ts
 import { useCycleList } from '@vueuse/core'
 
-const list = ref([
+const { state, next, prev } = useCycleList([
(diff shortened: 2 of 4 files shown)


antfu (approved):

chrisspiegl:
I just tested v10 beta and noticed that my previous normal array `[]` of list items did no longer work with the new update.

I then made my list a `ref([])` and it did work again.

So, to me, this looks like a breaking change. And as it stands it only supports `ref([])` and no longer supports standard arrays?

hjJunior:
> I just tested v10 beta and noticed that my previous normal array `[]` of list items did no longer work with the new update.
> 
> I then made my list a `ref([])` and it did work again.
> 
> So, to me, this looks like a breaking change. And as it stands it only supports `ref([])` and no longer supports standard arrays?

Hey @chrisspiegl, sorry to hear about that, this was supposed to work with standard arrays and with references, would you mind providing a small demo on how to reproduce it? 

I created a small example here, it seems to be working fine: https://playcode.io/1311657

antfu:
@chrisspiegl what do you mean "no longer work"? Can you add your usage to https://github.com/vueuse/vueuse/blob/b65c2ccdb7db7e3e54d9e7145aec4ed74fe4f855/packages/core/useCycleList/index.test.ts so we could see how we can fix it?

chrisspiegl:
@antfu & @hjJunior, I updated the Cloned and adjusted the PlayCode example.

I learned two things while doing so:

- I am using `useCycleList` with an array of objects (sortOrders).
- Using an array of strings seems to work with both the `@latest` and `@beta`.
- With a `ref(sortOrders) with sortOrders being an array of objects it also works in both version.

**Expectation:**

An array of objets should also work when it is not a `ref()`? Just like the array of strings does.

Here is my example which does not work: https://playcode.io/1314120

PR #2866, opened 2023-03-14
feat(useDark): passthrough default handler from `useColorSchema`

godismyjudge95:
### Description

Sometimes you want to hook into the onChanged functionality when the mode is changed, but you still want the default handler to happen.  Currently that is not possible. 

### Additional context

An example of what I was trying to do that caused me to need this functionality:
```js
const isDark = useDark({
    onChanged(dark?: boolean) {
        cookies.set('darkmode', dark ? 1 : 0, { path: '/', sameSite: true });
    },
});
```

Currently, this would cause dark mode to not function because the onChanged removes the default functionality of adding a class to the html tag.

Instead I had to do the following:
```js
const isDark = useDark({
    onChanged(dark?: boolean) {
        if (dark) {
            document.querySelector('html')?.classList.add('dark');
        } else {
            document.querySelector('html')?.classList.remove('dark');
        }
        cookies.set('darkmode', dark ? 1 : 0, { path: '/', sameSite: true });
    },
});
```

However, this is a little terse especially when this functionality is already done behind the scenes with the default hook.  Instead this pull request adds the ability to do the following:
```js
const isDark = useDark({
    onChanged(dark?: boolean, defaultHook, mode) {
        cookies.set('darkmode', dark ? 1 : 0, { path: '/', sameSite: true });
        defaultHook(mode);
    },
});
```

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other

### DIFF
--- a/packages/core/useDark/index.ts
+++ b/packages/core/useDark/index.ts
@@ -25,7 +25,7 @@ export interface UseDarkOptions extends Omit<UseColorModeOptions<BasicColorSchem
    *
    * @default undefined
    */
-  onChanged?: (isDark: boolean) => void
+  onChanged?: (isDark: boolean, defaultHandler: ((mode: T | BasicColorSchema) => void), mode: T | BasicColorSchema) => void
 }
 
 /**
--- a/packages/core/useDark/index.ts
+++ b/packages/core/useDark/index.ts
@@ -25,7 +25,7 @@ export interface UseDarkOptions extends Omit<UseColorModeOptions<BasicColorSchem
    *
    * @default undefined
    */
-  onChanged?: (isDark: boolean, defaultHandler: ((mode: T | BasicColorSchema) => void), mode: T | BasicColorSchema) => void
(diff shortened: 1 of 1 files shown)


antfu:
I think the types need to be updated as well

godismyjudge95:
> I think the types need to be updated as well

Good point, I added them.

packages/core/useDark/index.ts
```diff
@@ -25,7 +25,7 @@ export interface UseDarkOptions extends Omit<UseColorModeOptions<BasicColorSchem
    *
    * @default undefined
    */
-  onChanged?: (isDark: boolean) => void
+  onChanged?: (isDark: boolean, defaultHandler: ((mode: T | BasicColorSchema) => void), mode: T | BasicColorSchema) => void
```
Waleed-KH, line 28: there is no `T` here

```suggestion
  onChanged?: (isDark: boolean, defaultHandler: ((mode: BasicColorSchema) => void), mode: BasicColorSchema) => void
```

antfu (approved):

PR #2883, opened 2023-03-20
feat(useIntersectionObserver): support for `Pausable` interface

Mini-ghost:
### Description

Sometimes we want to conditionally enable or disable `useIntersectionObserver`. In the past, we could only use a watch function to enable it when `enabled` is `true` and clean it up when `enabled` is `false`.

```vue
<script setup lang="ts">
import { useIntersectionObserver } from '@vueuse/core';
import { computed, ref, watch } from 'vue';

const elRef = ref<HTMLElement>();
const enabled = computed(() => {
  // Conditional
  return true; // or false
});

let stop: ReturnType<typeof useIntersectionObserver>['stop'];
watch(enabled, enabled => {
  if (!enabled) {
    stop?.();
  } else {
    ({ stop } = useIntersectionObserver(elRef, () => {
      // Do something
    }));
  }
});
</script>
```

With the introduction of support for enabled: MaybeRef<boolean>, we can simplify our code:

```vue
<script setup lang="ts">
import { useIntersectionObserver } from '@vueuse/core';
import { computed, ref } from 'vue';

const elRef = ref<HTMLElement>();
const enabled = computed(() => {
  // Conditional
  return true; // or false
});

useIntersectionObserver(
  elRef,
  () => {
    // Do something
  },
  {
    enabled,
  }
);
</script>
```

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useIntersectionObserver/index.ts
+++ b/packages/core/useIntersectionObserver/index.ts
@@ -1,5 +1,6 @@
-import { watch } from 'vue-demi'
-import { noop, tryOnScopeDispose } from '@vueuse/shared'
+import { unref, watch } from 'vue-demi'
+import { isBoolean, noop, tryOnScopeDispose } from '@vueuse/shared'
+import type { MaybeRef } from '@vueuse/shared'
 import type { ConfigurableWindow } from '../_configurable'
 import { defaultWindow } from '../_configurable'
 import type { MaybeElementRef } from '../unrefElement'
@@ -7,6 +8,12 @@ import { unrefElement } from '../unrefElement'
 import { useSupported } from '../useSupported'
 
 export interface UseIntersectionObserverOptions extends ConfigurableWindow {
+  /**
+   * Set to `false` if you want to enable only when needed.
+   * @default undefined
+   */
+  enabled?: MaybeRef<boolean>
+
   /**
    * The Element or Document whose bounds are used as the bounding box when testing for intersection.
    */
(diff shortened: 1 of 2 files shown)


packages/core/useIntersectionObserver/index.ts
```diff
@@ -1,12 +1,19 @@
...
 import type { MaybeElementRef } from '../unrefElement'
 import { unrefElement } from '../unrefElement'
 import { useSupported } from '../useSupported'
 
 export interface UseIntersectionObserverOptions extends ConfigurableWindow {
+  /**
+   * Set to `false` if you want to enable only when needed.
+   * @default undefined
+   */
+  enabled?: MaybeRef<boolean>
```
antfu, line 15: I would think it might be better to implement the `Pausiable` interface with a `immediate` option 

https://github.com/mini-ghost/vueuse/blob/3c728cb2630b3afb609da4e350cd5de17686af63/packages/shared/utils/types.ts#L78-L94

antfu (changes requested):

packages/core/useIntersectionObserver/index.ts
Mini-ghost: Perhaps using `Stoppable` would be closer to the current interface?

Mini-ghost: But if we implement the `Pausable` interface with the `immediate` option, developer using directive will not be able to call `pause` and `resume`. If the developer sets immediate: false then he will never be able to resume the IntersectionObserver.

antfu: Sorry I am not sure if I got what you mean. To have a `Pausable` interface is to make toggling that behavior more explicit. I think it could work the same as `enabled.value`.

antfu (approved):

PR #2892, opened 2023-03-22
fix(useEventBus)!: fix emit parameters type

Alfred-Skyblue:
>

### Description

I found that when using useEventBus, we enforced a limit of 2 parameters for emit, but in actual use, we may need to pass more parameters, so I changed the generic type of useEventBus to pass a function type or Tuple to limit The parameter type of emit.

### Additional context

Expect:

```ts
const { on, emit } = useEventBus<string, number>('foo')

function listener(name: string, age: number, callback: Function) {
  console.log(name, age)
}
on(listener)

//  success
emit('vi', 18, () => {
  console.log('callback')
})

```

Actual:
```ts

const { on, emit } = useEventBus<string, number>('foo')

function listener(name: string, age: number, callback: Function) {
  callback()
}

// TS2345: Argument of type '(name: string, age: number, callback: Function) => void' is not assignable to parameter of type 'EventBusListener '.
on(listener)

// TS2554: Expected 0-2 arguments, but got 3.
emit('vi', 18, () => {
  console.log('callback')
})
```

After repair:

```ts

// transfer function type Tuple type
const { on, emit } = useEventBus<typeof listener>('foo')

function listener(name: string, age: number, callback: Function) {
  callback()
}
//  success
on(listener)

//  success
emit('vi', 18, () => {
  console.log('callback')
})
```

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useEventBus/demo.vue
+++ b/packages/core/useEventBus/demo.vue
@@ -2,7 +2,8 @@
 import { ref } from 'vue'
 import { useEventBus } from '@vueuse/core'
 
-const { on, emit } = useEventBus<string>('vue-use-event-bus')
+const { on, emit } = useEventBus<typeof listener>('vue-use-event-bus')
+
 const message = ref('')
 const news = [
   'Su Bingtian broke the Asian record and entered the Olympic 100-meter race finals as the first person in China-RTHK',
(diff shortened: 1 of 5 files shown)


antfu:
This seems to break many use cases, and I don't think it's worth it. I would rather keep it as-is to make it simple. This is a simple implementation of the bus anyway, for more complex usage I would suggest using a dedicated event library.

PR #2893, opened 2023-03-22
feat(useFileDialog): add listener for file change events

Alfred-Skyblue:
>

### Description

Add listener for file changes.

I need to trigger some operations when files change at work, and currently I can only monitor its changes through watch, which is a bit cumbersome for me, so I added an onChange method in this PR for monitoring The changes in files hope to be adopted, thank you! 

### Additional context

Expect: 
```ts
import { watch } from 'vue'
import { useFileDialog } from '@vueuse/core'

const { files, open, reset } = useFileDialog()

watch(files,(files) => {
  /** do something with files */
})
```
Actual:
```
import { useFileDialog } from '@vueuse/core'

const { files, open, reset, onChange } = useFileDialog()

onChange((files) => {
  /** do something with files */
})
```

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/core/useFileDialog/demo.vue
+++ b/packages/core/useFileDialog/demo.vue
@@ -1,7 +1,10 @@
 <script setup lang="ts">
 import { useFileDialog } from '.'
 
-const { files, open, reset } = useFileDialog()
+const { files, open, reset, onChange } = useFileDialog()
+onChange((files) => {
+  /** do something with files */
+})
 </script>
 
 <template>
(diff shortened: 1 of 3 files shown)


melishev:
@Alfred-Skyblue The onChange method does not work if I try to load the same file twice.

The method will work the first time, but the second time it will be silent

Alfred-Skyblue:
@melishev When the same file is selected, the onchange of the input will not be triggered. This is a feature of the input. You can pass reset as true to clear the value of the input every time it is opened.

PR #2957, opened 2023-04-12
feat(createTemplatePromise): new function

antfu:
Migrated from https://github.com/antfu/vue-template-promise


antfu:
TODO: test if this would affect vue 3 support

packages/core/createTemplatePromise/index.md
```diff
@@ -0,0 +1,175 @@
...
+Once `resolve` or `reject` is called in the template, the promise will be resolved or rejected, returning the value you passed in. Once resolved, the slot will be removed automatically.
+
+### Passing Arguments
+
+You can pass arguments to the `start` with arguments.
+
+```ts
+import { createTemplatePromise } from '@vueuse/core'
+
+const TemplatePromise = createTemplatePromise<boolean, [string, number]>()
```
xanderbarkhatov, line 84: Maybe it would be better to use an object instead of an array for arguments?
This way it would be more aligned with how we use props with regular components.

antfu: The reason here is that we need to support passing multiple arguments for the start function

packages/core/createTemplatePromise/index.md
```diff
@@ -0,0 +1,175 @@
...
+```ts
+const result = await dialog.open({
+  title: 'Hello',
+  contentSlot: () => h(MyComponent, { content }),
+})
+```
+
+This is like reinventing a new DSL in the script to express the UI template.
+
+So this library is introduce to **express the UI in Vue's template instead of the script**, as where they are supposed to be, and being able to be called programmatically.
```
innocenzi, line 175:
```suggestion
So this function allows **expressing the UI in templates instead of scripts**, where it is supposed to be, while still being able to be manipulated programmatically.
```

packages/core/createTemplatePromise/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+```ts
+const dialog = useDialog()
+const result = await dialog.open({
+  title: 'Hello',
+  content: 'World',
+})
+```
+
+This could work nicely by sending those infomation to top level component and let it render the dialog. However, it limits the flexibility you could express in the UI. For example, if you want the title to be red, if you want extra buttons, etc. You might end up with a lot of options like:
```
innocenzi, line 148:
```suggestion
This would work by sending these information to the top-level component and let it render the dialog. However, it limits the flexibility you could express in the UI. For example, you could want the title to be red, or have extra buttons, etc. You would end up with a lot of options like:
```

Issue #2947, opened 2023-04-08
Component usage creates excessive `div` element


Heniker:
### Clear and concise description of the problem

Using components from `@vueuse/components` creates a `div` element. This is probably not a desired behavior, as most of those component do not display anything and only provide some kind of functionality (e.g. `OnClickOutside`).

### Suggested solution

Use [Renderless Components](https://vuejs.org/guide/components/slots.html#renderless-components) instead.

Sample code that does not create a `div`:
```js
import * as v from 'vue'

/**
 * Component that resends all props passed to it as writeable v-slot arguments
 * 
 * @example <provide-value :a="42" v-slot="slot"><button v-on:click ="slot.a++">{{slot.a}}</button></provide-value>
 */
export const ProvideValue = v.defineComponent({
  inheritAttrs: false,
  setup(_, { attrs, slots }) {
    const writeableProps = v.reactive(Object.assign({}, attrs))

    return () => v.renderSlot(slots, 'default', v.normalizeProps(writeableProps) || {})
  },
})
```

### Alternative

Mention `div` creation in docs.

### Additional context

https://vueuse.org/guide/components.html

Can provide a PR if desired.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
Please provide concrete examples of which component is creating a div for you.

Heniker:
https://github.com/vueuse/vueuse/blob/main/packages/core/onClickOutside/component.ts

Here is a [sandbox](https://codesandbox.io/p/sandbox/sad-sea-buj7ul?file=%2Fsrc%2FApp.vue) that shows `OnClickOutside` component creates a `div` element.

If you inspect HTML elements you will see 2 `<section>` elements and 1 `<div>` element created by `OnClickOutside`.

`OnClickOutside` was chosen purely for showcase. I am pretty sure all vueuse components use the same code that produces a `div` element.

smallnine9:
I checked [sandbox](https://codesandbox.io/p/sandbox/sad-sea-buj7ul?file=%2Fsrc%2FApp.vue), and I found there is indeed an extra div:
[...]

PR #2966, opened 2023-04-14
feat(OnClickOutSide) use slot scope to avoid add excessive element

smallnine9:

- [√ ] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [ √] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [√ ] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [√ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ √] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<br>

>

</details>

---

### Description

This PR is to solve the issue #2947 
When we use OnClickOutSide to wrap an element, the OnClickOutSide will auto generate a 'div' element( or we specify an custom tag using prop.as)
Maybe we would think this is an insignificant  point, but in some case, I found it will cause bad dx, first, we can wirte `demo.vue` like that:
```
<script lang="ts">
import * as v from 'vue'
import { OnClickOutside } from './component'

export default v.defineComponent({
  components: { OnClickOutside },
  setup() {
    const color = v.ref('green')
    return { color }
  },
})
</script>

<template>
  <OnClickOutside @trigger="color = 'red'">
    <section class="base" :class="`bg-${color}`" @click="color = 'green'">
      &nbsp;click
    </section>
  </OnClickOutside>
</template>

<style>
.base {
  display: flex;
  justify-content: center;
  align-items: center;
  width: 100px;
  height: 100px;
}

.bg-green {
  background-color: green;
}
.bg-green::before {
  content: "inside";
}

.bg-red {
  background-color: red;
}
.bg-red::before {
  content: "outside";
}
</style>
```
and then we run `nr dev`, we open the website and choose onClickOutSide, we can find the problem:
image

For users ，we just want to make the `section` element be  **onClickOutSideable**, but because the extra div element is larger than the section elment, even we click outside of section, it could not change color of section element.

so, this is an an annoying problem。

### Additional context

I just passed the target ref to user, and user can bind this target to the element **_directly which they really want to make onClickOutSideable_**

But it will change the way how users using onClickOutSide, and maybe there exists a better solution。
This is my first PR to vueuse, so I am very glad to obtain feedback from maintainers.

Can I ask how you test your changes?
For me, this PR, I just modify the demo.vue in onClickOutSide folder, but I feel it's a little inconvenient.
when I open playground, I found it can't just import OnClickOutSide from @vueuse/components, maybe we can add the config in vite.config.js?
image

---

### <samp>🤖 Generated by Copilot at 66f5243</samp>

Refactored the `onClickOutside` component to remove unnecessary props and simplify its slot usage. The component now allows customizing the click outside element through the `target` ref.

### <samp>🤖 Generated by Copilot at 66f5243</samp>

*  Simplify component rendering to use default slot as a render function ([link](https://github.com/vueuse/vueuse/pull/2966/files?diff=unified&w=0#diff-3c599ae86e0d1c36cb38cc7f76591f0e19ff5aa4132a25b338920aca4c0ff484L12-R12), [link](https://github.com/vueuse/vueuse/pull/2966/files?diff=unified&w=0#diff-3c599ae86e0d1c36cb38cc7f76591f0e19ff5aa4132a25b338920aca4c0ff484L22-R22))


sibbng:
Couldn't we keep `as` option and also support default slot? 🤨

smallnine9:
so, can we continue discussion or just deal with it?

smallnine9:
@wheatjs what's your opinion? :)

antfu:
This will be a breaking change, and would not work them the slot has multiple components. We should do it in a non-destructive and braking way. 

Maybe support `:as="null"` to bypass creating an element.

smallnine9:
Yes, I didn't consider it thoroughly before. We should make it compatible with the previous users' habits. I added `:as="null"` to indicate that the user wants to use OnClickOutside without adding extra elements. 
For cases where there are multiple components in the slot, users can bind the target to multiple components.
like that:
```
  <OnClickOutside v-slot="{ target }" :as="null" @trigger="color = 'red'">
    <Foo :ref="target" :color="color" @change-color="color = 'green'" />
    <Foo :ref="target" :color="color" @change-color="color = 'green'" />
  </OnClickOutside>
```
 However, is there any issue with this?

smallnine9:
delete branch carelessly， Could this open again？

PR #3004, opened 2023-04-20
feat(useAsyncState): add support directly await

Hfutsora:
Add initialized in the [useAsyncState](https://vueuse.org/core/useAsyncState/).

### Description

Sometimes, using only `isReady` or `isLoading`  is not very convenient. 

For example, in scenarios where multiple events rely on state being triggered, and the `execute` may not have been completed yet.

We need to use `await initialized` to ensure that subsequent operations are executed only after the state has been initialized.

### Example

```vue


<script lang="ts" setup>
const { initialized, state } = useAsyncState(p1, 0, { immediate: true })

const eventA = async () => {
  await initialized.value
  // do something with the state...
}
const eventB = async () => {
  await initialized.value
  // do something with the state ...
}
</script>
```

### DIFF
--- a/packages/core/useAsyncState/index.test.ts
+++ b/packages/core/useAsyncState/index.test.ts
@@ -27,6 +27,13 @@ describe('useAsyncState', () => {
     expect(state.value).toBe(2)
   })
 
+  it('should work with await', async () => {
+    const asyncState = useAsyncState(p1, 0, { immediate: true })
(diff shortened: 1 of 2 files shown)


antfu:
I think instead, we could return a `then` method to proactively wait for current promises. You could refer to `useFetch`

Hfutsora:
Thank you for your suggestion. I have modified the original design to return a `thenable` object.

Issue #3019, opened 2023-04-23
useInfiniteScroll v10 behavior differs from v9 - Scrolling to the edge triggers an infinite event.


sondh0127:
### Describe the bug

useInfiniteScroll v10 behavior is different with v9

https://user-images.githubusercontent.com/62163604/233847985-dd49eeef-93c5-47df-937c-973c3ec9ca9c.mov

Scrolling to the edge triggers an infinite event.

### Reproduction

https://github.com/sondh0127/my-vitesse-app-useInfiniteScroll

### System Info

```Shell
System:
    OS: macOS 13.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 59.34 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.13.0 - ~/Library/Caches/fnm_multishells/94889_1682258323298/bin/node
    Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/94889_1682258323298/bin/yarn
    npm: 8.19.3 - ~/Library/Caches/fnm_multishells/94889_1682258323298/bin/npm
  Browsers:
    Brave Browser: 112.1.50.121
    Chrome: 112.0.5615.137
    Safari: 16.4
  npmPackages:
    @vueuse/core: 10.1.0 => 10.1.0
    @vueuse/head: ^1.1.26 => 1.1.26
    vue: ^3.2.47 => 3.2.47
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sondh0127:
[...]

PR #3030, opened 2023-04-26
fix(useInfiniteScroll): re-measure arrived states when async infinite scroll resolves

scottbedard:

<details>
<br>

>

</details>

---

### Description

Issue: https://github.com/vueuse/vueuse/issues/3019

Our `useInfiniteScroll` composable is not working with async functions. This is caused by [`arrivedState[direction]`](https://github.com/vueuse/vueuse/blob/81bcaa9d99bece266bc4351af3aa5147eda0020b/packages/core/useInfiniteScroll/index.ts#L69) never being set to `false` when the container is resized.

This state becomes stuck because the measurement of these values only happens during [`onScrollHandler`](https://github.com/vueuse/vueuse/blob/81bcaa9d99bece266bc4351af3aa5147eda0020b/packages/core/useScroll/index.ts#L163), and so if the user is no longer scrolling, no re-measurement occurs and we enter the infinite loop.

The simplest way to force a measurement is just to fire a scroll event ourselves, but this may not be the fix we want. We might want to let `useScroll` handle observation of it's height, but this would require a larger refactor. Let me know if that's the route we'd rather go and I'll make the changes.

Another solution that might be a happy medium is for `useScroll` to return a function that can force a re-measurement. This might be the most flexible solution, because it would allow for measurements to happen for whatever reason we want 🤔

**Update:** I've made the refactor to return a force-measurement function, I think I like this better, but let me know what you think! If we'd rather force the measurement by just firing a scroll event, we can revert 87d675748bb16134bcd8ffb88c8e7ec66c1adccb

### Additional context

---

### <samp>🤖 Generated by Copilot at e44a129</samp>

Fix initial load bug in `useInfiniteScroll` hook. Dispatch scroll event on observed element to trigger fetch function.

### <samp>🤖 Generated by Copilot at e44a129</samp>

* Dispatch scroll event on observed element to trigger initial data load ([link](https://github.com/vueuse/vueuse/pull/3030/files?diff=unified&w=0#diff-f1e18d7fa553054c65bf5b7e4fe4daf0fbfb98fe7224964f3f46661a3770d6a1R77))

### DIFF
--- a/packages/core/useInfiniteScroll/index.ts
+++ b/packages/core/useInfiniteScroll/index.ts
@@ -58,6 +58,8 @@ export function useInfiniteScroll(
   const isLoading = computed(() => !!promise.value)
 
   function checkAndLoad() {
+    state.measure()
+
     const el = toValue(element) as HTMLElement
     if (!el)
       return
@@ -74,7 +76,6 @@ export function useInfiniteScroll(
         ])
           .finally(() => {
             promise.value = null
-            state.measure()
             nextTick(() => checkAndLoad())
           })
       }
--- a/packages/core/useInfiniteScroll/index.ts
+++ b/packages/core/useInfiniteScroll/index.ts
@@ -74,7 +74,7 @@ export function useInfiniteScroll(
         ])
           .finally(() => {
             promise.value = null
-            el.dispatchEvent(new Event('scroll'))
+            state.measure()
             nextTick(() => checkAndLoad())
           })
       }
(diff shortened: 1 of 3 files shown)


packages/core/useInfiniteScroll/index.ts
```diff
@@ -74,6 +74,7 @@ export function useInfiniteScroll(
         ])
           .finally(() => {
             promise.value = null
+            state.measure()
```
antfu, line 77: Maybe we could put this at the beginning of `checkAndLoad`? Here I think it's better to do after nextTick

PR #3036, opened 2023-04-27
ci: ensure test coverage for node 19

michealroberts:
ci: ensure test coverage for node 19

---

### Description

Node v.19 has now been available since October 2022, we should look to add stability to this package by extending the test strategy for Node to includes version 19.x


antfu:
Thanks but no. I don't feel it's necessary and usually we should only test again even number versions.

antfu:
Sure :)

PR #3061, opened 2023-05-04
docs(useInfiniteScroll): demonstrate stopping when there is no more data

scottbedard:

<details>
<br>

>

</details>

---

### Description

Our `useInfiniteScroll` docs don't demonstrate how to stop fetching data, which can lead to infinite loops if you aren't careful ([see conversation here](https://github.com/vueuse/vueuse/issues/3019)). The underlying issue for this has already been fixed by https://github.com/vueuse/vueuse/pull/3030.

### Additional context

---

### <samp>🤖 Generated by Copilot at 85c21f8</samp>

This pull request improves the documentation of `useInfiniteScroll` and `useScroll` by adding examples and explanations for some common use cases. It shows how to stop infinite scrolling by unsetting the `target` element, and how to re-measure the scroll position and size of the container element using the `measure` function.

### <samp>🤖 Generated by Copilot at 85c21f8</samp>

*  Add `useInfiniteScroll` function to enable infinite scrolling on a target element ()

### DIFF
--- a/packages/core/useInfiniteScroll/index.md
+++ b/packages/core/useInfiniteScroll/index.md
@@ -41,7 +41,7 @@ To stop infinite scrolling, unset the target element. This will prevent the call
 const isComplete = ref(false)
 
 useInfiniteScroll(
-  computed(() => !isComplete.value && el.value),
+  computed(() => !isComplete.value ? el.value : null),
   async () => {
     if (hasLoadedEverything)
       isComplete.value = true
--- a/packages/core/useInfiniteScroll/index.md
+++ b/packages/core/useInfiniteScroll/index.md
@@ -35,6 +35,22 @@ useInfiniteScroll(
 </template>
 ```
 
+To stop infinite scrolling, unset the target element. This will prevent the callback from firing, and will turn off all scroll listeners.
+
+```js
+const isComplete = ref(false)
+
+useInfiniteScroll(
+  computed(() => !isComplete.value && el.value),
+  async () => {
+    if (hasLoadedEverything)
+      isComplete.value = true
+    else
+      await loadMoreStuff()
+  }
+)
+```
+
 ## Directive Usage
 
 ```html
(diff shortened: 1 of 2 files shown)


scottbedard:
> Maybe we could make it return `false` to stop?

We're actually discussing that [over here](https://github.com/vueuse/vueuse/issues/3019#issuecomment-1534814105), it might be a nice addition to `useInfiniteScroll`. I think we should still include this example though to steer users towards ending the scroll listeners when they're no longer needed, and avoiding extra calls to `onLoadMore`. Let me know what you think, and I'll adjust this PR accordingly.

packages/core/useInfiniteScroll/index.md
```diff
@@ -35,6 +35,22 @@ useInfiniteScroll(
 </template>
 ```
 
+To stop infinite scrolling, unset the target element. This will prevent the callback from firing, and will turn off all scroll listeners.
+
+```js
+const isComplete = ref(false)
+
+useInfiniteScroll(
+  computed(() => !isComplete.value && el.value),
```
wsamoht, line 44: The current example code is returning a `bool`. It should be something like this.

```suggestion
  computed(() => !isComplete.value ? el.value : null),
```

scottbedard: These are essentially the same thing, the only difference is that yours returns `null` whereas mine returns `false` when it is complete. We're just doing [a truthiness check](https://github.com/vueuse/vueuse/blob/d52524441ae08c1ef7a9d3bc86c7fa32bbb0cd75/packages/core/useInfiniteScroll/index.ts#L63-L65) though, these would behave identically. They're both good options for turning off an infinite scroller 🙂

wsamoht: Scott, I saw your comment in my email but appears you maybe deleted it here. I should have provided a little more context for my reasoning. I am running v10.1.2 and was having issues so I went looking and came across this example from looking at https://github.com/vueuse/vueuse/pull/3030.

I tried the example but it throws a type error because it needs the element and not a `bool`.

```
Type 'ComputedRef<false>' is not assignable to type 'Ref<Window | HTMLElement | SVGElement | Document | null | undefined>'
```

scottbedard: Ya I deleted it because I was incorrect. Our could _could_ work with the original syntax, but our type signatures don't support booleans. I've committed your change! 🙂

PR #3072, opened 2023-05-07
fix(useMediaControls): better representation for "waiting" value

ferferga:

---

### Description

"waiting" only fires when there's a lack of data in the middle of the playback. However, it doesn't take into account the lack of data when the source is attached to the element, as "loadeddata" does. I think it's safe to assume people want from "waiting" all possible buffering states. In the probably scarce cases where the user wants more control over the native event handlers, native event handlers can still be used.

"playing" also represents when a current playback is happening now.

---

### <samp>🤖 Generated by Copilot at b092cfe</samp>

Improved media controls state handling by `useMediaControls`. Added support for more media events and states, such as `waiting` and `playing`.

### <samp>🤖 Generated by Copilot at b092cfe</samp>

*  Update `useMediaControls` function to handle more media events and states ([link](https://github.com/vueuse/vueuse/pull/3072/files?diff=unified&w=0#diff-ecbf7bfd67f67f25217f8c5d36cfba3734d608a1bccdb3a49acba1f17a59401eL391-R399))

### DIFF
--- a/packages/core/useMediaControls/index.ts
+++ b/packages/core/useMediaControls/index.ts
@@ -388,10 +388,15 @@ export function useMediaControls(target: MaybeRef<HTMLMediaElement | null | unde
   useEventListener(target, 'progress', () => buffered.value = timeRangeToArray((toValue(target))!.buffered))
   useEventListener(target, 'seeking', () => seeking.value = true)
   useEventListener(target, 'seeked', () => seeking.value = false)
-  useEventListener(target, 'waiting', () => waiting.value = true)
+  useEventListener(target, ['waiting', 'loadstart'], () => {
+    waiting.value = true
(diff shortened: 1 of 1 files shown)


antfu (approved):

emmanuelgeoffray:
@ferferga thank you for working on this !
I agree with you, we just want that 'waiting' represents *all* buffering state.

I'm having similar issues with `waiting` not returning to false. Even with your fix.
From what I debugged, clearing `waiting` on the `seeked` event, fixes the issue.
I think it is safe, but I may not know all the details behind the `seeked` and `waiting` events.

ferferga:
@emmanuelgeoffray Can you give a playground I can test? Something with a large enough audio file would be enough I guess, given with DevTools I can mimick slow network conditions.

PR #3244, opened 2023-07-20
fix(useMouse): position won't be changed on page scroll when `type` is `page`, closes #2922

### LINKED ISSUE #2922: useMouseInElement is not updating correctly on scroll for static element
### Describe the bug

While useMouse is totally correct to not update x or y value on scroll, this is not what should happen for the value of elementX, elementY and isOutside on useMouseInElement when the element is not sticky.

If I scroll on a page the position of the mouse is not changing which is correct, but scrolling move elements in the page thus changing the position of the mouse in this element.
Since useMouseInElement is not updating elementX, elementY and isOutside, on scroll they end up been false.

To reproduce the problem you can use the [doc](https://vueuse.org/core/useMouseInElement/) : 

place your cursor in the element then scoll, the elementX, elementY are not updating and if you scroll until the cursor is out of the element isOutside is still at false.

### Reproduction

https://vueuse.org/core/useMouseInElement/

### System Info

```Shell
System:
    OS: macOS 13.2.1
    CPU: (8) arm64 Apple M1
    Memory: 119.64 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v18.12.1/bin/npm
  Browsers:
    Chrome: 111.0.5563.146
    Firefox: 111.0
    Safari: 16.3
  npmPackages:
    @vueuse/core: ^9.13.0 => 9.13.0 
    vue: ^3.2.47 => 3.2.47
```

### Used Package Manager

pnpm

### Validations

CatsJuice:

<details>
<br>

>

</details>

---

### Description

When `type` is set to `page` (which is default), it's expected to update `x`, `y` when page is scrolled(related to issue #2922 ). I've added a `scrollTarget` option to enable scroll detection when `type` is `page`.

The mouse position of page should be updated when page scroll by recalculate the position in `scrollTarget`'s `scroll` event callback.

### Additional context

I wonder if the new option `scrollTarget` that I added is right. Maybe just add a `scroll?: boolean` will be better? Because the `page` type should related to window `scroll` event, and it's unnecessary to customize `scrollTarget`.

---

### <samp>🤖 Generated by Copilot at b410832</samp>

Add `scrollTarget` option to `useMouse` hook to track mouse or touch position relative to the page. Update `useMouse/index.ts` to handle scroll events and adjust position accordingly.

### <samp>🤖 Generated by Copilot at b410832</samp>

*  Add `scrollTarget` option to `UseMouseOptions` interface to customize the element for `scroll` events ([link](https://github.com/vueuse/vueuse/pull/3244/files?diff=unified&w=0#diff-acac08f31319ce15064e72bec03f921b9731164f2022b1a17ed4c09ba2864540R35-R42))
*  Declare `_prevMouseEvent` variable to store the last mouse or touch event on the `target` element ([link](https://github.com/vueuse/vueuse/pull/3244/files?diff=unified&w=0#diff-acac08f31319ce15064e72bec03f921b9731164f2022b1a17ed4c09ba2864540L72-R85))
*  Assign `event` parameter to `_prevMouseEvent` in `handler` function for `mousemove` or `touchmove` events ([link](https://github.com/vueuse/vueuse/pull/3244/files?diff=unified&w=0#diff-acac08f31319ce15064e72bec03f921b9731164f2022b1a17ed4c09ba2864540R96))
*  Define `scrollHandler` function to update `x` and `y` reactive values based on `_prevMouseEvent` and `window` scroll position ([link](https://github.com/vueuse/vueuse/pull/3244/files?diff=unified&w=0#diff-acac08f31319ce15064e72bec03f921b9731164f2022b1a17ed4c09ba2864540R114-R122))
*  Define `scrollHandlerWrapper` function to wrap `scrollHandler` with `eventFilter` option if provided ([link](https://github.com/vueuse/vueuse/pull/3244/files?diff=unified&w=0#diff-acac08f31319ce15064e72bec03f921b9731164f2022b1a17ed4c09ba2864540R136-R139))
*  Add `scroll` event listener on `scrollTarget` element using `useEventListener` hook with `scrollHandlerWrapper` as callback ([link](https://github.com/vueuse/vueuse/pull/3244/files?diff=unified&w=0#diff-acac08f31319ce15064e72bec03f921b9731164f2022b1a17ed4c09ba2864540R149-R150))


packages/core/useMouse/index.ts
```diff
@@ -31,6 +31,14 @@ export interface UseMouseOptions extends ConfigurableWindow, ConfigurableEventFi
    */
   touch?: boolean
 
+  /**
+   * Customize scrollTarget to listen to `scroll` events for type `page`
+   * Set to `null` or `undefined` to disable
+   *
+   * @default null
```
antfu, line 38: This does not align with the actual implementation. We could say this is enabled by default when type is `page`, and is only effective on type `page`?

packages/core/useMouse/index.ts
```diff
@@ -99,6 +111,15 @@ export function useMouse(options: UseMouseOptions = {}) {
     }
   }
 
+  const scrollHandler = () => {
+    if (!_prevMouseEvent || !window || type !== 'page')
```
antfu, line 115:
```suggestion
    if (!_prevMouseEvent || !window)
```

packages/core/useMouse/index.ts
CatsJuice: I'm not pretty sure that adding a scrollTarget is appropriate. Since `pageX` and `pageY` may not change if the `scrollTarget` is not `window`, it may be better to replace it with `scroll?: boolean`, which defaults to true when `type = "page"`.

When the scroll target is not a `window` but a custom element (e.g. `#app`) , user may need to customize `extractors` in order to get the mouse distance from the top coordinate.

In this case, there's a limit to what vueuse can handle, and it may only be able to cover the case where the `scrollTarget` is `window` and the `type` is `page`. 

I'm more confused about the API design here, so I wonder if I can change the options to `scroll?: boolean` and only handle the case where the `window` is the `scrollTarget`.

antfu: Yes, I think `scroll?: boolean` makes sense

CatsJuice: When the user really needs to use reactive mouse position in a custom element with the help of `useMouse`, it won't be possible. If we use `scrollTarget` with a user-defined `type` extractor, we can achieve it, but it's a bit tricky for the user to implement.

In this situation, use `useElementBounding` and `useMouse` together can also resolve the problem(like the solution in this [PR](https://github.com/CatsJuice/AFFiNE.pro/commit/a5d56d799e1494ba03431a99ab13f9245fc6c95d))

I'll change the option to `scroll?: boolean` first in current PR!

:)

CatsJuice:
@antfu  By the way, I don't know why but when I pull the latest changes, it failed to run dev server.

I've found the reason but I'm not sure why and what happened as described in #3273

antfu (approved):

PR #3253, opened 2023-07-24
fix(createReusableTemplate): camelize props

94726:

<details>
<br>

>

</details>

---

### Description

This PR allows to mix attribute cases, similar to native Vue behavior.

Currently, attributes passed to reusableTemplates are passed to v-slot, as is. This leads to issues when trying to mix cases on define and reuse.

People might end up writing the following, which doesn't work
```vue
<script setup lang="ts">
import { createReusableTemplate } from '@vueuse/core'

const [DefineFoo, ReuseFoo] = createReusableTemplate<{ someMsg: string }>()
</script>

<template>
  <DefineFoo v-slot="{someMsg}">{{ someMsg }}</DefineFoo>

  <ReuseFoo some-msg="bar" />
</template>
```
 
### Additional context

Technically vue exports `camelize` which could be used here, but for some reason `hyphenate` is not.
So I decided to copy these functions from vue for now.

---

### <samp>🤖 Generated by Copilot at 453731b</samp>

Added prop name flexibility and improved code quality for `createReusableTemplate`. This function creates a Vue component from a template string and a set of props.

### <samp>🤖 Generated by Copilot at 453731b</samp>

*  Copy `hyphenate` and `camelize` functions from vue core package to convert attribute keys between cases ([link](https://github.com/vueuse/vueuse/pull/3253/files?diff=unified&w=0#diff-2ab8a265afffefa3f7f94474668e63c4eb38974e01686c2c439ffb672275b2ebL5-R25))
*  Use `keysToCamelKebabCase` function to create `attrs` object with both camelCase and kebab-case keys for reusable template component props ([link](https://github.com/vueuse/vueuse/pull/3253/files?diff=unified&w=0#diff-2ab8a265afffefa3f7f94474668e63c4eb38974e01686c2c439ffb672275b2ebL64-R79), [link](https://github.com/vueuse/vueuse/pull/3253/files?diff=unified&w=0#diff-2ab8a265afffefa3f7f94474668e63c4eb38974e01686c2c439ffb672275b2ebR89-R99))
*  Format type parameters for `DefineTemplateComponent`, `ReuseTemplateComponent`, `ReusableTemplatePair`, and `createReusableTemplate` to fit in one line for readability and consistency ([link](https://github.com/vueuse/vueuse/pull/3253/files?diff=unified&w=0#diff-2ab8a265afffefa3f7f94474668e63c4eb38974e01686c2c439ffb672275b2ebL13-R31), [link](https://github.com/vueuse/vueuse/pull/3253/files?diff=unified&w=0#diff-2ab8a265afffefa3f7f94474668e63c4eb38974e01686c2c439ffb672275b2ebL20-R37), [link](https://github.com/vueuse/vueuse/pull/3253/files?diff=unified&w=0#diff-2ab8a265afffefa3f7f94474668e63c4eb38974e01686c2c439ffb672275b2ebL37-R53))


antfu:
Weird, I wonder why Vue does not handle them. I'd like to see if there is a way to solve this better without needing to port code from Vue

94726:
I was wondering too. Seems like Vue transforms slot attribute cases only at compile-time, see
- [Vue playground example](https://play.vuejs.org/#eNp9UrFOwzAQ/RXLS5c2HbqVUAlQBxgAAaOXkF5CiuOzbCdEivLv3DlNmwG6+e7ee36+517eWZu0DcitTH3uKhuEh9BYoTNT3ioZvJI7ZaraogviAWsrCoe1WCRrLpi6uFEmXY9kglIRoLY6C0CVEGkktSuvMZCgdWhZs+/jaRjSKETQdD3jySVdnaMpqjI5ejTkr2c1JXNCVxrciw0VGpLaijjhWaY1/jzFXnANLKd+/gX59x/9o++4p+SrAw+uBSXPs5C5EsI43r8/Q0fn87DGQ6MJfWX4Bh51wx5H2H1jDmR7hotuH+NmK1N++H0XwPjpUWyUkUPEK0mb5kX99/SL3U2yiTxlBtrilBIHfNmvG2kphyIKxNVn5jht8CSxozSpH1HzUKiMrfk/OYFO6cvhF24mx1M=), if you go to the js output of `Comp.vue`, you can see that the hyphenated attribute was camelized.
- [related Vue compiler source](https://github.com/vuejs/vue/blob/main/src/compiler/codegen/index.ts#L605-L631) (I think)

Based on that, slot-props are never accessible in hyphenated format. So I think we could only camelize the attributes.
`camelize()` itself can also be importet from vue-demi, to avoid porting any vue code. But it is still marked as private so I'm unsure about that.

wvffle:
> Technically vue exports camelize which could be used here, but for some reason hyphenate is not.

In my opinion, the methods from Vue should be used instead of porting them. If anything changes in Vue,  VueUse wouldn't need to be updated unless it was a breaking change. Also, if somebody wanted to use `camelize`/`hyphenate` from VueUse, they'd probably have exactly the same code twice in their production build. One from Vue used internally and one from VueUse used manually.

If `hyphenate` is not exported, maybe it's a good idea to create a PR to export it rather than re-creating it?

wvffle:
Fixes #3253 3261

antfu:
In practice, it would be hard to push this to happen, as it would require alignment across all Vue versions (2.6, 2.7, 3+), while Vue has been cautious about adding new public APIs. If you want, you can try to push it in Vue core. But note that even if it gets shipped, it will only be available in **the next version**, meaning we still need to port to support the older version.

In that sense, I'd think porting is the fast solution that is approachable. Maybe we could remove the cache, as they are probably not very perf-sensitive in VueUse.

wvffle:
> it would require alignment across all Vue versions (2.6, 2.7, 3+)

You're right, I've missed that in my reasoning. Vue 2 does not export `camelize`. I guess, porting it is the right thing to do then.

packages/shared/utils/port.ts
```diff
@@ -0,0 +1,18 @@
+// copied from vue: https://github.com/vuejs/core/blob/3be4e3cbe34b394096210897c1be8deeb6d748d8/packages/shared/src/general.ts#L90-L112
```
wvffle, line 1:
```suggestion
import { camelize as camelize_ } from 'vue-demi'

// copied from vue: https://github.com/vuejs/core/blob/3be4e3cbe34b394096210897c1be8deeb6d748d8/packages/shared/src/general.ts#L90-L112
```

packages/shared/utils/port.ts
```diff
@@ -0,0 +1,18 @@
...
+  }) as T
+}
+
+const hyphenateRE = /\B([A-Z])/g
+export const hyphenate = cacheStringFunction((str: string) =>
+  str.replace(hyphenateRE, '-$1').toLowerCase(),
+)
+
+const camelizeRE = /-(\w)/g
+export const camelize = cacheStringFunction((str: string): string => {
```
wvffle, line 16:
```suggestion
export const camelize = camelize_ ?? cacheStringFunction((str: string): string => {
```

wvffle (commented):
How about using Vue's `camelize` if available and if not, then port?

packages/shared/utils/port.ts
94726: Wouldn't this result in an error if someone is on a Vue version that doesn't export it?

I don't think it's worth it.

wvffle: `vue-demi` returned an `undefined` on vue 2 when I was testing it

94726: Unfortunately not. That would require `vue-demi` to include such exports but it only re-exports everything from vue, meaning that you basically try to import something non-existant from Vue

I created an example for you https://stackblitz.com/edit/vitejs-vite-eikher?file=src%2FApp.vue

PR #3298, opened 2023-08-08
fix(useClipboard): "Read" mode is not working on Firefox and some other changes 

oliver139:

<details>
<br>

>

</details>

---

### Description

6 things are done in this PR

#### (1) Read mode support on Firefox

According to [MDN Clipboard API Docs](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText#browser_compatibility), Firefox is only able to write text into the clipboard, but not read from it. So that legacy method should be used for reading the clipboard in Firefox.

To do so, I tried to check if `navigator.clipboard.readText()` is available, where Firefox does support Clipboard API but not `read()` & `readText()`. If not and `legacy` is set to be `true`, legacy reading method would be used.

#### (2) Wrong type is using for the returns variable

Previously:
```ts
isSupported: Ref<boolean>
text: ComputedRef<string>
copied: ComputedRef<boolean>
```

Which should be:
```ts
isSupported: ComputedRef<boolean>
text: Ref<string>
copied: Ref<boolean>
```

#### (3) Exporting 2 more variables

While we are facing some different behavior about Clipboard API compatibility between browsers, for example:

|  | Clipboard API | Write | Read |
| -------- | :------: | :------: | :------: |
| Chrome     | ✅     | ✅     | ✅     |
| Firefox     | ✅     | ✅     | 🚫     |
| IE     | 🚫     | 🚫     | 🚫     |

I think it is good to export some more flag for users to check the compatibility, so that apart from `isSupported`, I also want to export `isClipboardApiSupported` and `isClipboardReadSupported` (Newly added) for detail checking purpose.

#### (4) Throw error on fail copy with legacy copy

I tried to check for the return value of `document.execCommand("copy")` and throw an error if it returns `false`.

#### (5) JSDoc for the returning variables

I tried to write some JSDoc for them.

#### (6) Update for the use Clipboard document

I wrote a sentence about the use of `read` mode and also include new checking in the demo.

### Additional context

I am not so sure if (4) would be working or not, while according to [MDN execCommand Document](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand#return_value), `false` means unsupported or disabled, but not meaning fail.

---

### <samp>🤖 Generated by Copilot at 800ec42</samp>

Added `read` option and browser compatibility checks to `useClipboard` function. The function can now read the clipboard text using the Clipboard API or the legacy method, and exposes two refs to indicate the support level. The demo and the documentation were updated accordingly.

### <samp>🤖 Generated by Copilot at 800ec42</samp>

*  Add two new computed refs to `useClipboard` to indicate browser support for Clipboard API and `navigator.clipboard.readText()` ([link](https://github.com/vueuse/vueuse/pull/3298/files?diff=unified&w=0#diff-104a20ea8a35b01c36721b36a81d3138312e27a997280d66290ce7e8a9bc466aL41-R73),[link](https://github.com/vueuse/vueuse/pull/3298/files?diff=unified&w=0#diff-104a20ea8a35b01c36721b36a81d3138312e27a997280d66290ce7e8a9bc466aR95),[link](https://github.com/vueuse/vueuse/pull/3298/files?diff=unified&w=0#diff-104a20ea8a35b01c36721b36a81d3138312e27a997280d66290ce7e8a9bc466aL71-R107),[link](https://github.com/vueuse/vueuse/pull/3298/files?diff=unified&w=0#diff-104a20ea8a35b01c36721b36a81d3138312e27a997280d66290ce7e8a9bc466aR146-R147),[link](https://github.com/vueuse/vueuse/pull/3298/files?diff=unified&w=0#diff-f6f0edcfde360bae35a2945d8afe122f29bcb8ea4f5d3fa60283b33d91c3b245L7-R7),[link](https://github.com/vueuse/vueuse/pull/3298/files?diff=unified&w=0#diff-01b6125e1ab05d5dc5311c2a032284dbd17cd32e724eba71e6bbd4b1c512af98L17-R17))
*  Update the demo and the documentation to show the new return values and explain the `read` option ([link](https://github.com/vueuse/vueuse/pull/3298/files?diff=unified&w=0#diff-f6f0edcfde360bae35a2945d8afe122f29bcb8ea4f5d3fa60283b33d91c3b245L15-R31),[link](https://github.com/vueuse/vueuse/pull/3298/files?diff=unified&w=0#diff-01b6125e1ab05d5dc5311c2a032284dbd17cd32e724eba71e6bbd4b1c512af98R37-R38))
*  Improve error handling and feedback for the legacy copy method by throwing an error if `document.execCommand('copy')` fails ([link](https://github.com/vueuse/vueuse/pull/3298/files?diff=unified&w=0#diff-104a20ea8a35b01c36721b36a81d3138312e27a997280d66290ce7e8a9bc466aL104-R137))
*  Add a condition to check the `legacy` option before using the legacy read method ([link](https://github.com/vueuse/vueuse/pull/3298/files?diff=unified&w=0#diff-104a20ea8a35b01c36721b36a81d3138312e27a997280d66290ce7e8a9bc466aL71-R107))

### DIFF
--- a/packages/core/useClipboard/index.ts
+++ b/packages/core/useClipboard/index.ts
@@ -93,7 +93,7 @@ export function useClipboard(options: UseClipboardOptions<MaybeRefOrGetter<strin
 
   const isClipboardApiSupported = useSupported(() => (navigator && 'clipboard' in navigator))
   const isClipboardReadSupported = useSupported(() => (navigator && 'clipboard' in navigator) && navigator.clipboard.readText)
-  const isSupported = computed(() => (isClipboardApiSupported.value && isClipboardReadSupported.value) || legacy)
+  const isSupported = computed(() => isClipboardApiSupported.value || legacy)
   const text = ref('')
   const copied = ref(false)
   const timeout = useTimeoutFn(() => copied.value = false, copiedDuring)
@@ -104,7 +104,7 @@ export function useClipboard(options: UseClipboardOptions<MaybeRefOrGetter<strin
         text.value = value
       })
     }
-    else {
+    else if (legacy) {
       text.value = legacyRead()
     }
   }
--- a/packages/core/useClipboard/index.ts
+++ b/packages/core/useClipboard/index.ts
@@ -38,6 +38,8 @@ export interface UseClipboardOptions<Source> extends ConfigurableNavigator {
 }
 
 export interface UseClipboardReturn<Optional> {
+  isClipboardApiSupported: ComputedRef<boolean>
+  isClipboardReadSupported: ComputedRef<boolean>
   isSupported: ComputedRef<boolean>
   text: Ref<string>
   copied: Ref<boolean>
@@ -110,6 +112,8 @@ export function useClipboard(options: UseClipboardOptions<MaybeRefOrGetter<strin
   }
 
   return {
+    isClipboardApiSupported,
+    isClipboardReadSupported,
     isSupported,
     text: text as ComputedRef<string>,
     copied: copied as ComputedRef<boolean>,
--- a/packages/core/useClipboard/index.ts
+++ b/packages/core/useClipboard/index.ts
@@ -69,9 +69,9 @@ export interface UseClipboardReturn<Optional> {
    * The function for Copying the text
    *
    * @param {string} text to be copied
-   * @returns {Promise<void>}
+   * @returns {Promise<boolean>}
    */
-  copy: Optional extends true ? (text?: string) => Promise<void> : (text: string) => Promise<void>
+  copy: Optional extends true ? (text?: string) => Promise<boolean> : (text: string) => Promise<boolean>
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/core/useClipboard/index.ts
```diff
@@ -101,7 +132,9 @@ export function useClipboard(options: UseClipboardOptions<MaybeRefOrGetter<strin
     ta.style.opacity = '0'
     document.body.appendChild(ta)
     ta.select()
-    document.execCommand('copy')
+    const copyResult = document.execCommand('copy')
+    if (!copyResult)
+      throw new Error('[useClipboard] Fail to copy with legacy method')
```
antfu, line 137: In VueUse we avoid throwing errors, which increase the bundle size and hard to do i18n.

Instead, we could return a boolean indicating if the copy is supported.

PR #3394, opened 2023-09-12
fix(useAxios): bail out on request abort

jahnli:
When the `request aborted`, the `response `  is `undefined`, it causes Cannot read properties of undefined (reading '`data`')


jahnli:
@antfu  This issue needs to be reviewed

jahnli:
cc @antfu

antfu (approved):

yuantongkang:
如果请求是 Aborted 是走不到 then 这里的

PR #3400, opened 2023-09-14
feat(useRafFn): add interval and isLoop options

Simon-He95:
I think useRafFn could be used as a setTimeout or setInterval function with better performance.
I can control interval times to call the function or just run one time and auto-stop.


antfu:
https://github.com/vueuse/vueuse/pull/3459 serves a similar purpose. I am not sure if `isLoop` as you can always pause it in the callback. Thanks.

PR #3464, opened 2023-10-07
fix: vue 2 support for provideLocal and injectLocal

xiaoxiangmoe:

---

### Description

fix vue 2.7  support for provideLocal and injectLocal

See
https://github.com/vuejs/vue/blob/49b6bd4264c25ea41408f066a1835f38bf6fe9f1/src/v3/currentInstance.ts#L5-L14

Every time call getCurrentInstance in vue 2.7 will create a new Object，so we should use getCurrentInstance()?.proxy other than getCurrentInstance() as WeakMap key.

### Additional context

---

### <samp>🤖 Generated by Copilot at 9bfc23c</samp>

This pull request fixes bugs and adds tests for the `provideLocal` and `injectLocal` functions, which enable local state management across components. It uses the `proxy` property of the component instance to access and set the local state in the `localProvidedStateMap` weak map.

### <samp>🤖 Generated by Copilot at 9bfc23c</samp>

*  Fix bugs in `provideLocal` and `injectLocal` functions that use the wrong instance object to access and set the local state ([link](https://github.com/vueuse/vueuse/pull/3464/files?diff=unified&w=0#diff-ba81180701e963bbf6dcf1e15f6ef59f90f788179faa05cb16b6624494193925L16-R18), [link](https://github.com/vueuse/vueuse/pull/3464/files?diff=unified&w=0#diff-ebc0176920d2f721b2765598ee0d5f39d9b1eb81174861056ff465591b0d21b2L14-R14))
*  Update the type of the `localProvidedStateMap` variable to use the `proxy` property of the instance as the key ([link](https://github.com/vueuse/vueuse/pull/3464/files?diff=unified&w=0#diff-0aa9ceee4c7b48c3ad9c5bf0d5768e8e69ef2b87ee0e3524a43ff30bc676e8c0L3-R3))
*  Add test cases for `provideLocal` and `injectLocal` functions using `vitest` and `vue-demi` ([link](https://github.com/vueuse/vueuse/pull/3464/files?diff=unified&w=0#diff-7522ecb4b07e9cec47f5b8f739fc0ff0fa9413bcb936c757aa0e2349832f9955R1-R48))

### DIFF
--- a/packages/shared/injectLocal/index.test.ts
+++ b/packages/shared/injectLocal/index.test.ts
@@ -1,0 +1,61 @@
+import { type InjectionKey, defineComponent, getCurrentInstance, h, nextTick } from 'vue-demi'
+import { injectLocal, provideLocal } from '@vueuse/shared'
+import { describe, expect, it } from 'vitest'
+import { mount, useSetup } from '../../.test'
+
+describe('provideLocal injectLocal deps', () => {
+  it('depends on getCurrentInstance().proxy === getCurrentInstance().proxy', async () => {
+    const vm = useSetup(() => {
+      const instance1 = getCurrentInstance()
+      const instance2 = getCurrentInstance()
+      const instanceProxyStable = instance1?.proxy === instance2?.proxy
+      return { instance1, instance2, instanceProxyStable }
+    })
+    await nextTick()
+    expect(vm.instance1).not.toBeNull()
+    expect(vm.instance2).not.toBeNull()
+    expect(vm.instance1).toBeTypeOf('object')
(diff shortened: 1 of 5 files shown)


packages/shared/injectLocal/index.test.ts
```diff
@@ -0,0 +1,48 @@
+import { type InjectionKey, defineComponent, h } from 'vue-demi'
+import { injectLocal, provideLocal } from '@vueuse/shared'
+import { describe, expect, it } from 'vitest'
+import { mount } from '../../.test'
+
+{
```
antfu, line 6: Why not do it inside describe?

PR #3480, opened 2023-10-17
feat(useMutationObserver): add `takeRecords` function

Alfred-Skyblue:

<details>
<br>

>

</details>

---

### Additional context

---

### <samp>🤖 Generated by Copilot at a4d76ea</samp>

This pull request adds a `takeRecords` function to the `useMutationObserver` hook, which lets users access the mutation records without affecting the observer. It also adds a test case for this feature in `index.test.ts`.

### <samp>🤖 Generated by Copilot at a4d76ea</samp>

*  Implement and expose `takeRecords` function for `useMutationObserver` hook ([link](https://github.com/vueuse/vueuse/pull/3480/files?diff=unified&w=0#diff-da50736ee00ff6293e6541e95e2518436421b4ddc33730c401b602f349be86b0R49-R52),[link](https://github.com/vueuse/vueuse/pull/3480/files?diff=unified&w=0#diff-da50736ee00ff6293e6541e95e2518436421b4ddc33730c401b602f349be86b0R63))
* Add test case for `takeRecords` function in `index.test.ts` ([link](https://github.com/vueuse/vueuse/pull/3480/files?diff=unified&w=0#diff-8d4441a9e8238cd8a892b795951468494603938a4469271654ef9d93c63bc2eaR169-R189))

### DIFF
--- a/packages/core/useMutationObserver/index.ts
+++ b/packages/core/useMutationObserver/index.ts
@@ -46,6 +46,10 @@ export function useMutationObserver(
     { immediate: true },
   )
 
+  const takeRecords = () => {
+    return observer?.takeRecords()
+  }
+
   const stop = () => {
     cleanup()
     stopWatch()
@@ -56,6 +60,7 @@ export function useMutationObserver(
   return {
     isSupported,
     stop,
+    takeRecords,
   }
 }
 
--- a/packages/core/useMutationObserver/index.test.ts
+++ b/packages/core/useMutationObserver/index.test.ts
@@ -166,4 +166,25 @@ describe('useMutationObserver', () => {
     await promiseTimeout(10)
     expect(cb).toHaveBeenCalledTimes(1)
   })
+
+  it('should work with takeRecords', async () => {
+    const target = document.createElement('div')
+    const cb = vi.fn()
+
+    const { takeRecords } = useMutationObserver(target, cb, {
+      attributes: true,
+    })
+
+    target.setAttribute('id', 'footer')
+    await promiseTimeout(10)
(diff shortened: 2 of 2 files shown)


packages/core/useMutationObserver/index.ts
```diff
@@ -56,6 +60,7 @@ export function useMutationObserver(
   return {
     isSupported,
     stop,
+    takeRecords,
```
okxiaoliang4, line 63: Maybe just return the `observer` variables.
```suggestion
    observer,
```

Alfred-Skyblue: If we return `observer`, we need to wrap it as a `Ref`.

okxiaoliang4: I think make it to `shallowRef` is more flexible

Alfred-Skyblue: Both are possible. Vue only proxies native `Object` and does not proxy `MutationObserver`, for `MutationObserver`, the behavior is consistent between `Ref` and `shallowRef `.

Alfred-Skyblue: However, I believe we should return a `takeRecords` function. Exposing the `observer` to the user would allow them to manually call `disconnect` to stop the listening, but we prefer users to use `stop` to halt the listening. This is because `stop` can also stop the `watch`. In the `MutationObserver` instance, only the `observer`, `takeRecords`, and `disconnect` methods are available. The `observer` is automatically executed, and we expose the `disconnect` through `stop` to the user. We only need to expose the `takeRecords` to the user.

okxiaoliang4 (approved):

Issue #3481, opened 2023-10-18
useRafFn delta is always 0


slavanossar:
### Describe the bug

I tried using the new `fpsLimit` option for `useRafFn`, but noticed that when a value is set the callback function never runs.

This issue is that this line never runs

https://github.com/vueuse/vueuse/blob/aca6a79227f7f8b06760d756dfbb4cc5958b445d/packages/core/useRafFn/index.ts#L67

which causes `delta` to always be `0`. It needs to also be included in the conditional

https://github.com/vueuse/vueuse/blob/aca6a79227f7f8b06760d756dfbb4cc5958b445d/packages/core/useRafFn/index.ts#L60-L63

I will open a PR for this since it's a very simple fix.

### Reproduction

https://github.com/slavanossar/vueuse-useraffn-delta-bug

### System Info

```Shell
System:
    OS: macOS 12.6.9
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
    Memory: 35.24 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.17.0 - ~/.n/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 9.8.1 - /usr/local/bin/npm
    pnpm: 8.7.3 - /usr/local/bin/pnpm
  Browsers:
    Brave Browser: 117.1.58.137
    Safari: 16.6
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

bernhardberger:
[...]

PR #3482, opened 2023-10-18
fix(useRafFn): correctly set previousFrameTimestamp when fpsLimit option is used

slavanossar:

<details>
<br>

>

</details>

---

### Description

See #3481

### Additional context

---

copilot:summary

copilot:walkthrough

### DIFF
--- a/packages/core/useRafFn/index.ts
+++ b/packages/core/useRafFn/index.ts
@@ -58,6 +58,7 @@ export function useRafFn(fn: (args: UseRafFnCallbackArguments) => void, options:
     const delta = timestamp - (previousFrameTimestamp || timestamp)
 
     if (intervalLimit && delta < intervalLimit) {
+      previousFrameTimestamp = timestamp
       rafId = window.requestAnimationFrame(loop)
       return
     }


packages/core/useRafFn/index.ts
```diff
@@ -58,6 +58,7 @@ export function useRafFn(fn: (args: UseRafFnCallbackArguments) => void, options:
     const delta = timestamp - (previousFrameTimestamp || timestamp)
 
     if (intervalLimit && delta < intervalLimit) {
+      previousFrameTimestamp = timestamp
```
okxiaoliang4, line 61: I think we should move `previousFrameTimestamp = timestamp` to up line, and remove line 68.

slavanossar: True, that makes a lot more sense! I've updated my PR, thanks!

okxiaoliang4 (approved):

danwill:
I have come across the same issue noted in #3481 when using the `fpsLimit` option of `useRafFn`. However, the fix proposed in this pull request _also_ prevents the callback function from executing when certain values are used with the `fpsLimit` option. 

For example, in an environment with a 60hz refresh rate, using `fpsLimit: 60` may produce unexpected results: often the callback isn't executed at all, but will trigger when browser FPS dips temporarily. Setting an even lower value for `fpsLimit` prevents the callback from firing at all.

The below change in this PR updates `previousFrameTimestamp` on every loop. This results in the value of `delta` remaining roughly constant when idle (~16.6 @ 60hz). 

```
const delta = timestamp - (previousFrameTimestamp || timestamp)
previousFrameTimestamp = timestamp

if (intervalLimit && delta < intervalLimit) {
  rafId = window.requestAnimationFrame(loop)
  return
}
```
With this, setting a low `fpsLimit` to something like 12 will result in an `intervalLimit` value of 83.3, and will nearly always return early without triggering the callback function.

One potential solution to consider would be to revert the update in this PR and instead add a check to ensure `previousFrameTimestamp` has a non-zero value when considering the `intervalLimit`:

```
if (intervalLimit && delta < intervalLimit && previousFrameTimestamp) {
    rafId = window.requestAnimationFrame(loop);
    return;
}
```

One side-effect of the above is that the callback is always triggered the first time `loop()` is called. In my mind this is beneficial, as any fps throttling should apply to timing between callbacks. To this end, it may also be beneficial to reset `previousFrameTimestamp` to `0` when the `pause()` function is called.

Happy to discuss any further.

bernhardberger:
@danwill tbh, right now the officially documented feature is broken. it should either be fixed or removed for the time being.

daniandl:
I thought I was going crazy, this should really be fixed or removed from the documentation

PR #3549, opened 2023-11-14
feat(refBind): new function

serkodev:

---

### Description

Sometimes, when we want to reset the values of some `ref` variables after updating `props`, we need to use watch to monitor whether `props` has changed, and then reset the values of the `ref` variables. This approach can be cumbersome and not very intuitive. In such cases, we can use `refBind` to achieve this effect.

Original code:
```ts
const props = defineProps<{
  n: number
}>()

const start = ref(props.n)
watch(() => props.n, () => {
  start.value = props.n
})
```

Using `refBind`：

```ts
const props = defineProps<{
  n: number
}>()

const start = refBind(() => props.n)
```

In addition, I have also provided a custom watcher option. Based on the following example, when `props.n` is updated, `start` will be automatically reset to `0`:

```ts
const start = refBind(0, () => props.n)
```

Although VueUse currently provides features such as `syncRef`, I believe that this is addressing different problems and goals. Do you have any suggestions for better solutions? any thoughts or ideas are welcomed.

---

### <samp>🤖 Generated by Copilot at 1210c8b</samp>

This pull request adds a new utility function `refBind` to the `@vueuse/core` package, which allows creating one-way bindings between reactive sources and refs. It also includes a demo component, a documentation file, and a test file for the function.

### <samp>🤖 Generated by Copilot at 1210c8b</samp>

*  Export `refBind` function from `./refBind` module to enable one-way binding between refs ([link](https://github.com/vueuse/vueuse/pull/3549/files?diff=unified&w=0#diff-4ef755f3c945daaa416774a88556b30166725160cf2ce87e21ed80aca4ffd115R19))
*  Implement `refBind` function in `refBind/index.ts` using `vue-demi`, `watch`, and `toValue` ([link](https://github.com/vueuse/vueuse/pull/3549/files?diff=unified&w=0#diff-558fdca1bf7e2f754cf0f18cc100d1ff72a8aa43d065e548deada18e38aeab74R1-R31))
*  Document `refBind` function in `refBind/index.md` with usage, parameters, and examples ([link](https://github.com/vueuse/vueuse/pull/3549/files?diff=unified&w=0#diff-efc1612b077d39442a2bfe5cc309b84379427584d593c315ba600a4a9d98113dR1-R66))
*  Test `refBind` function in `refBind/index.test.ts` with three test cases using `vitest` ([link](https://github.com/vueuse/vueuse/pull/3549/files?diff=unified&w=0#diff-270ae73d40aa630105efd550b7c7b9e6fa0e98398aa5b847910b9a63b0e63559R1-R59))
*  Demonstrate `refBind` function in `refBind/demo.vue` with a Vue component that renders two input elements ([link](https://github.com/vueuse/vueuse/pull/3549/files?diff=unified&w=0#diff-60aa9fbfe5abb8f3b81e2c10c51b05a796346b4e06ee556841bcd354d7d21b3bR1-R12))


Doctor-wu:
Maybe you can simply use `computed`?

antfu:
I feel this can be covered by one-directional `syncRef` isn't it? It would be more like a shorthand/sugar for it. We are in general hesitant about adding new functions due to both maintainability and the learning curve. I personally feel the benefit of this rule might not be enough to make the exception.

PR #3571, opened 2023-11-19
feat(usePrint): add `printUrl` and `printSource`

stafyniaksacha:

<details>
<br>

>

</details>

---

### Description

This PR add a reactive [Printing](https://developer.mozilla.org/en-US/docs/Web/Guide/Printing) API. 

It adds methods to open the browser print dialog, in the current page, to print from an url or a html string using a sandboxed iframe

```vue
<script setup lang="ts">
import { usePrint } from '@vueuse/core'

const source = ref('<h1>Hello vueuse!</h1>\n<script'+'>alert("not executed")</'+'script>')
const { printSource, printUrl, pending } = usePrint(source)
</script>

<template>
  <textarea v-model="source" />
  <button :disabled="pending" @click="() => printSource()">
    Print
  </button>
  <button :disabled="pending" @click="() => printUrl('/path', { sandbox: false })">
    Print
  </button>
</template>
```


antfu:
Thanks for the PR - but I am honestly not very should if this should be in scope of VueUse - it seems you don't really need Vue reactivity in this case. I feel it can be achieved with a function call - it might be better if you have this as a standalone library.

PR #3618, opened 2023-12-05
fix: vitepress style

yikayiyo:
### Description
The previous commit modified some CSS variables, causing the button's background to become transparent on hover. This commit applies appropriate CSS variables.


antfu:
Thanks

PR #3623, opened 2023-12-06
feat(useAsyncState): add `immediateExecuteParams` option

phk422:

<details>
<br>

>

</details>

---

### Description

Add immediateExecuteParams option. When immediate is true, the immediateExecuteParams is used as a parameter for the first executing the promise

### DIFF
--- a/packages/core/useAsyncState/index.test.ts
+++ b/packages/core/useAsyncState/index.test.ts
@@ -19,6 +19,13 @@ describe('useAsyncState', () => {
       throw new Error('error')
     return id
   }
+  const p3 = (str: string, num: number) => {
+    return new Promise((resolve) => {
+      setTimeout(() => {
+        resolve(str + num)
+      }, 50)
+    })
+  }
 
   it('should work', async () => {
     const { execute, state } = useAsyncState(p1, 0)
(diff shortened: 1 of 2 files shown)


Alfred-Skyblue:
Do you think doing it this way would be better?

```ts
function fn(str = 'foo', num = 1) {
  return new Promise((resolve) => {
    setTimeout(() => {
      resolve(str + num)
    }, 50)
  })
}
useAsyncState(fn, 'bar')
```

phk422:
> This should be considered a rare case. We recommend addressing this issue using the following approach:
> 
> ```ts
> function fn(str: string, num: number) {
>   return new Promise((resolve) => {
>     setTimeout(() => {
>       resolve(str + num)
>     }, 50)
>   })
> }
> const { execute } = useAsyncState(fn, 'bar', { immediate: false })
> execute(0, 'foo', 1)
> ```
> 
> Alternatively, you can try defining the parameters externally:
> 
> ```ts
> let str = 'foo', num = 1
> function fn() {
>   return new Promise((resolve) => {
>     setTimeout(() => {
>       resolve(str + num)
>     }, 50)
>   })
> }
> useAsyncState(fn, 'bar')
> ```



> This should be considered a rare case. We recommend addressing this issue using the following approach:
> 
> ```ts
> function fn(str: string, num: number) {
>   return new Promise((resolve) => {
>     setTimeout(() => {
>       resolve(str + num)
>     }, 50)
>   })
> }
> const { execute } = useAsyncState(fn, 'bar', { immediate: false })
> execute(0, 'foo', 1)
> ```
> 
> Alternatively, you can try defining the parameters externally:
> 
> ```ts
> let str = 'foo', num = 1
> function fn() {
>   return new Promise((resolve) => {
>     setTimeout(() => {
>       resolve(str + num)
>     }, 50)
>   })
> }
> useAsyncState(fn, 'bar')
> ```

Thanks for your answer, you are right, I thought about it a little bit, adding a immediateExecuteParams as a parameter would be a little redundant

Alfred-Skyblue:
thank you for your understanding.

Issue #3589, opened 2023-11-26
Cookie is not a constructor


kissu:
### Describe the bug

The default example does not work with Nuxt3
https://vueuse.org/integrations/useCookies/#common-usage

With the following code
```html
<template>
  <div>
    <strong>locale</strong>: {{ cookies.get('locale') }}
    <hr>
    <pre>{{ cookies.getAll() }}</pre>
    <button @click="cookies.set('locale', 'ru-RU')">Russian</button>
    <button @click="cookies.set('locale', 'en-US')">English</button>
  </div>
</template>

<script setup>
import { useCookies } from '@vueuse/integrations/useCookies'

const cookies = useCookies(['locale'])
</script>
```

![image](https://github.com/vueuse/vueuse/assets/5133074/306d4687-6e54-48dc-b3aa-3d41d8af5917)

Btw, I do have a different error on Stackblitz: https://stackblitz.com/~/github.com/kissu/vueuse-cookie-nuxt3
`_universal_cookie.default is not a constructor`, but it's probably related

### Reproduction

[...]

PR #3629, opened 2023-12-08
doc(integrations): add dependencies version in docs

### LINKED ISSUE #3625: Vueuse/integration not working icw vueuse/nuxt in nuxt 3
### Describe the bug

I have a Nuxt 3 application with vueuse/nuxt module installed where I use some composables from VueUse. Now I wanted to use some composables from the integrations addon, but with no luck. 

I reproduced the issue here (with thanks to Muhammad Mahmoud):

https://stackblitz.com/edit/nuxt-starter-uffmnn?file=app.vue

### Reproduction

https://stackblitz.com/edit/nuxt-starter-uffmnn?file=app.vue

### System Info

```Shell
System:
    OS: Windows 10 10.0.19045
    Memory: 12.17 GB / 31.73 GB
  Binaries:
    Node: 18.18.2 - C:\Program Files\nodejs\node.EXE
    npm: 9.8.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.3636.0), Chromium (119.0.2151.97)
    Internet Explorer: 11.0.19041.3636
  npmPackages:
    @vueuse/integrations: ^10.7.0 => 10.7.0
    @vueuse/nuxt: ^10.7.0 => 10.7.0
```

### Used Package Manager

npm

### Validations

Doctor-wu:

<details>
<br>

>

</details>

---

### Description

Add `version` indicator after the library name to help users install suitable dependencies' version.

Recently, some users can't use `@vueuse/integrations` in `nuxt` like #3625 and #3589. After some investigations, it was found that users installed unsuitable versions of dependencies, this PR add `version` indicator in docs to help users install suitable version.

fix #3625


packages/integrations/useAsyncValidator/index.md
```diff
@@ -9,7 +9,7 @@ Wrapper for [`async-validator`](https://github.com/yiminghe/async-validator).
 ## Install
 
 ```bash
-npm i async-validator
+npm i async-validator@^4.2.5
```
antfu, line 12:
```suggestion
npm i async-validator@4
```

If that would be better?

Doctor-wu: It's hard to say that other libraries will not have a breaking change in a patch or minor release though they are not supposed to, so maybe more clearly version will be better

Doctor-wu: I have tried, if we execute `npm i async-validator@^4.2.5`, then npm will add a clear field `async-validator: "^4.2.5"` in `package.json`, that's maybe safer

antfu: > It's hard to say that other libraries will not have a breaking change in a patch or minor release though they are not supposed to, so maybe more clearly version will be better

Well, if that happens, it's their fault and we don't have control over it.

The main thing I am worried about is that we would only update them manually. For example, to say two years later, `async-validator` could go something like `4.99.0` with a lot of security fixes, but ppl come to our docs will install the old `4.2.5`.

Unless we will use a script to update this to the version we tested (from our dev deps) and update automatically when we bump the version, I'd prefer to not set them too strict to have ppl on the latest supported version.

Doctor-wu: Well, that's true. I'll commit some changes to all the docs in `integrations`

Doctor-wu: And I found we should use `npm i async-validator@^4`, or npm will add a specific version `sync-validator: "4"` in `package.json` and I don't think that's what we want

antfu (approved):

PR #3634, opened 2023-12-10
fix(useBattery): fake signal API support in some utilities

tachibana-shin:

<details>
<br>

>

</details>

---

### Description
when i tried to develop my app and use `useBattery` i soon realized some widgets that protect user privacy like `DuckDuckGo` cause spurious signal to Battery API which causes `'getBattery ' in navigator === true` even though it is actually `undefined`

image

### Additional context

This pr may not be necessary if the package targets standard browsers. But the more users the better so I created this pr


Jesse205:
Hi, `const isSupported = useSupported(() => navigator && 'getBattery' in navigator && typeof navigator.getBattery === 'function')` may pass the test.

packages/core/useBattery/index.ts
```diff
@@ -26,7 +26,7 @@ export function useBattery(options: ConfigurableNavigator = {}) {
   const { navigator = defaultNavigator } = options
   const events = ['chargingchange', 'chargingtimechange', 'dischargingtimechange', 'levelchange']
 
-  const isSupported = useSupported(() => navigator && 'getBattery' in navigator)
+  const isSupported = useSupported(() => navigator && typeof navigator.getBattery === 'function')
```
Jesse205, line 29:
```suggestion
  const isSupported = useSupported(() => navigator && 'getBattery' in navigator && typeof navigator.getBattery === 'function')
```

tachibana-shin: I fixed it even though I didn't like it. Can we approach it in ts style instead of js too? I mean it's redundant to check an unimportant condition
```ts

  const isSupported = useSupported(() => navigator && typeof (navigator as Navigator & {
    getBattery?: () => unknown
  }).getBattery === 'function')
```

Doctor-wu (approved):

PR #3650, opened 2023-12-14
fix(useCounter): allow `delta` to be negative

leewakin:
Sometimes I like this:

` `
import { useCounter } from '@vueuse/core'

const { count, inc } = useCounter()
const deltaValues = [-1, -2, -3, 1, 2, 3]

<button v-for="delta in deltaValues" :key="delta" @click="inc(delta)">
  {{delta}}
</button>
` `

So if delta could be negative, it would be useful.


Doctor-wu:
@leewakin Hey, thanks for contributing. However, I'm not sure if it's a good practice for `inc` to allow building a negative mutation when we already have a dedicated `dec` function for that purpose.

Doctor-wu:
Maybe you can consider adding a `mutate` function to implement mutate but not care about if the delta is negative or positive

leewakin:
> Maybe you can consider adding a `mutate` function to implement mutate but not care about if the delta is negative or positive

Thank you for your response. You're right, we should indeed implement a `mutate` function. However, I've encountered a situation as follows:
```
const { count, inc, dec } = useCounter(0, {max: 2, min: -2})
```
1. When delta is greater than or equal to 0, everything works as expected.
2. When delta is less than 0,
  - Calling `inc` makes `min` ineffective.
  - Calling `dec` makes `max` ineffective.

From an intuitive standpoint, we should either disallow negative delta values, or fix the bug where max and min become ineffective when delta is less than 0. However, I'm unsure if this change is necessary for the library.

Doctor-wu:
Also, I think this change should be approved by Anthony, requested him for review just now~

antfu (approved):

Issue #3328, opened 2023-08-20
Handle cancellation and multiple executions in `useAsyncState`, and allow to manually trigger refresh in `computedAsync`


segevfiner:
### Clear and concise description of the problem

`useAsyncState` and `computedAsync` are rather similar but each lack some functionality it would be nice to bring them more inline in terms of functionality with each other.

Two things I noticed is that `useAsyncState` doesn't handle cancellation or calling `execute` multiple times while the promise is still not resolved, and that `computedAsync` has no way to manually execute the query.

### Suggested solution

`useAsyncState` should handle multiple concurrent calls to execute and supply cancellation like `computedAsync`, `computedAsync` should allow to manually trigger execution.

### Alternative

Be frustrated if you need both missing behaviors in one composable.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

bodograumann:
Duplicate: https://github.com/vueuse/vueuse/issues/2958

segevfiner:
But that one is closed

bodograumann:
Indeed. Just wanted to point out that this has been brought up before and link the issues, if anybody stumbles upon one or the other ;-)

segevfiner:
Bump for stale bot (AKA Ignore all issues bot)

bodograumann:
"Closed as completed" doesn't sound right.
Also seems like the stale bot is deprecated and should be replaced with the stale action :thinking:

MuXiu1997:
A better option might be to use [`Vue Query`](https://vue-query.vercel.app/#/)

PR #3661, opened 2023-12-17
feat(useCancellableAsyncState): add `useCancellableAsyncState`

MuXiu1997:

<details>
<br>

>

</details>

---

### Description

Add a cancellable version of `useAsyncState` that does not directly modify `useAsyncState` and maintains the same style as `useAsyncState`

- ref: #3328
<br>
### Additional context

Known potential issue: When the `options.delay` is greater than the time interval between two consecutive calls to `execute`, the `onCancel` has not been invoked yet to register the callback

### DIFF
--- a/packages/core/useCancellableAsyncState/demo.vue
+++ b/packages/core/useCancellableAsyncState/demo.vue
@@ -1,0 +1,32 @@
+<script setup lang="ts">
+import axios from 'axios'
+import YAML from 'js-yaml'
+import { useCancellableAsyncState } from '@vueuse/core'
+
+const { isLoading, state, isReady, execute } = useCancellableAsyncState(
+  (onCancel, args: unknown) => {
+    const abortController = new AbortController()
+    onCancel(() => abortController.abort())
+    const id = args?.id || 1
+    return axios.get(
+      `https://jsonplaceholder.typicode.com/todos/${id}`,
+      { signal: abortController.signal },
+    ).then(t => t.data)
+  },
+  {},
+  {
+    resetOnExecute: false,
+  },
+)
+</script>
+
+<template>
+  <div>
+    <note>Ready: {{ isReady.toString() }}</note>
+    <note>Loading: {{ isLoading.toString() }}</note>
+    <pre lang="json" class="ml-2">{{ YAML.dump(state) }}</pre>
+    <button @click="execute(0, { id: 2 })">
+      Execute
+    </button>
+  </div>
+</template>
--- a/packages/core/useCancellableAsyncState/index.md
+++ b/packages/core/useCancellableAsyncState/index.md
@@ -1,0 +1,29 @@
+---
+category: State
+---
+
+# useCancellableAsyncState
+
+Cancellable version of `useAsyncState`
+
+## Usage
+
+```ts
+import axios from 'axios'
+import { useCancellableAsyncState } from '@vueuse/core'
+
+const { state, isReady, isLoading } = useCancellableAsyncState(
+  (onCancel) => {
+    const abortController = new AbortController()
+
+    onCancel(() => abortController.abort())
+
+    const id = args?.id || 1
+    return axios.get(
+      `https://jsonplaceholder.typicode.com/todos/${id}`,
+      { signal: abortController.signal },
+    )
+  },
+  {},
+)
+```
--- a/packages/core/useCancellableAsyncState/index.test.ts
+++ b/packages/core/useCancellableAsyncState/index.test.ts
@@ -1,0 +1,113 @@
+import { promiseTimeout } from '@vueuse/shared'
+import { describe, expect, it, vi } from 'vitest'
+import type { CancellableAsyncStateOnCancel } from '.'
+import { useCancellableAsyncState } from '.'
+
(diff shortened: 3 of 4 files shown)


Alfred-Skyblue:
If we want to add this feature, it seems like it should be added to `useAsyncState` instead of introducing a new function.

MuXiu1997:
> If we want to add this feature, it seems like it should be added to `useAsyncState` instead of introducing a new function.


I have also considered directly modifying the useAsyncState function, but it seems difficult to add this feature without breaking changes. That's why I chose to add a new function instead of modifying the existing one.

Alfred-Skyblue:
We can create a new Promise within the `execute` function so that the state of this Promise is controllable for us. This way, users don't have to create an AbortController externally to control the state of the external Promise.

MuXiu1997:
> We can create a new Promise within the `execute` function so that the state of this Promise is controllable for us. This way, users don't have to create an AbortController externally to control the state of the external Promise.

I believe that as a user, I don't just want the `execute` function to handle the previous `execute` promise internally, but I want to be able to decide what to do when the previous `execute` was interrupted. For example, I may want to interrupt a large data request or simply increment a counter

Alfred-Skyblue:
We can return an `abort` function for users to manually initiate an interrupt action. Specifically, I mean creating a new Promise within the `execute` function, but the timing of when to interrupt it is controlled by external calls.

MuXiu1997:
> We can return an `abort` function for users to manually initiate an interrupt action. Specifically, I mean creating a new Promise within the `execute` function, but the timing of when to interrupt it is controlled by external calls.

I don't know if I misunderstood, but what I want is not a function that can be manually called with `abort`, but rather something like `onCleanup` within `watchEffect`
```ts
function watchEffect(
  effect: (onCleanup: OnCleanup) => void,
  options?: WatchEffectOptions
): StopHandle
```

Alfred-Skyblue:
I'm sorry, I misunderstood your meaning, but I have reservations about adding this feature. We have many ways to implement it, such as:

```ts
let cleanup: () => void

useAsyncState(() => {
  cleanup?.()
  const abortController = new AbortController()
  cleanup = () => abortController.abort()
  return axios
    .get(`https://jsonplaceholder.typicode.com/todos/${id}`, {
      signal: abortController.signal
    })
    .then(t => t.data)
}, '')

```

Can you provide more reasons for adding this function? Thank you.

MuXiu1997:
> I'm sorry, I misunderstood your meaning, but I have reservations about adding this feature. We have many ways to implement it, such as:
> 
> ```ts
> let cleanup: () => void
> 
> useAsyncState(() => {
>   cleanup?.()
>   const abortController = new AbortController()
>   cleanup = () => abortController.abort()
>   return axios
>     .get(`https://jsonplaceholder.typicode.com/todos/${id}`, {
>       signal: abortController.signal
>     })
>     .then(t => t.data)
> }, '')
> ```
> 
> Can you provide more reasons for adding this function? Thank you.

This case has a timing error: 
`execute()` -> `isLoading = true` in `execute()` -> `cleanup()` -> `isLoading = false` in the previous `execute()`
 So when `execute()` is called rapidly multiple times, `isLoading` will be directly set to false, which does not meet the requirements

Alfred-Skyblue:
From a personal perspective, I don't believe that we should repeatedly execute the `execute` function multiple times because we can never guarantee when asynchronous operations will be completed. From the user's standpoint, this may also lead to unnecessary confusion. If we check for the existence of an ongoing asynchronous operation every time we modify `isLoading`, it could prevent users from knowing when the first asynchronous operation was completed. Personally, I feel that users should be prevented from triggering the `execute` function when `isLoading` is set to true.

Alfred-Skyblue:
I was just providing an example. We can internally return an `abort` function for cancellation, but other operations, I believe, can be handled externally. Can you describe any challenges you encountered when adding this functionality to `useAsyncState`?

MuXiu1997:
Can you describe why it has to be handled externally? I don't think users would consider calling an abort function as good design.

Clearly, when calling the execute function again, the internal analysis of the state can determine whether cleanup is needed or not. Why turn this into something users have to be concerned about?

In the existing philosophies of Vue and VueUse, the approach is to register a callback function that users want to perform when internal cancellation is required through a function. For example:

```typescript
function watchEffect(
  effect: (onCleanup: OnCleanup) => void,
  options?: WatchEffectOptions\n): StopHandle

type OnCleanup = (cleanupFn: () => void) => void
```

```typescript
function computedAsync<T>(
  evaluationCallback: (onCancel: AsyncComputedOnCancel) => T | Promise<T>,
  initialState?: T,
  optionsOrRef?: Ref<boolean> | AsyncComputedOptions,
): Ref<T>

type AsyncComputedOnCancel = (cancelCallback: Fn) => void
```

Alfred-Skyblue:
I'm just concerned about the cost of adding this function. I would prefer it to be implemented within the `useAsyncState`. As for why it needs to be handled externally, I'm just suggesting one possible solution.

MuXiu1997:
A better option might be to use [`Vue Query`](https://vue-query.vercel.app/#/)

PR #3676, opened 2023-12-26
🐞 fix(useWebSocket): useWebSocket bug

GarlandQian:
Fixed useWebSocket using computed or ref url and turning on the automatic reconnection function, resulting in multiple ws connection problems when the url is modified.

<details>
<br>

>

</details>

---

### Description

When using ref or computed url, immediate is set to false and reconnect is set to true, it will cause the url to be modified and the reconnection will not be closed and a new ws will be opened, which will cause the previous ws to be reconnected and unable to be closed.


antfu:
I am not sure. I don't think immediate should apply to all subsequent changes.

GarlandQian:
> I am not sure. I don't think immediate should apply to all subsequent changes.

Let me tell you what happened to me: with immediate: false and autoReconnect set to true, if you start ws, changing the path of the url (the url uses computedRef) will cause a new ws to be created, but the previous ws is not closed. This will cause the previous ws to keep reconnecting due to the automatic connection setup.

PR #3705, opened 2024-01-09
fix: 🧩 fixed reference issues for complex data types

17359898647:

<details>
<br>

>

</details>

---

### DIFF
--- a/packages/shared/refThrottled/index.ts
+++ b/packages/shared/refThrottled/index.ts
@@ -1,7 +1,11 @@
 import type { Ref } from 'vue-demi'
 import { ref, watch } from 'vue-demi'
+import { cloneFnJSON } from '../../core/useCloned'
 import { useThrottleFn } from '../useThrottleFn'
 
+function isReferenceType(value: any) {
+  return value !== null && (typeof value === 'object' || typeof value === 'function')
+}
 /**
  * Throttle execution of a function. Especially useful for rate limiting
  * execution of handlers on events like resize and scroll.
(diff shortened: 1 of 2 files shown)


17359898647:
@Doctor-wu I re-opened a pr because I didn't know how to clean up useless commit records, and the reason I was testing a lot of errors was because I imported cloneFnJSON from @vueuse/core.

packages/shared/refThrottled/index.ts
```diff
@@ -10,18 +14,30 @@ import { useThrottleFn } from '../useThrottleFn'
...
     return value
+  const throttled = ref() as Ref<T>
+  const isEqualityClone = cloneHandler === cloneFnJSON
+  const setThrottled = () => {
+    try {
+      throttled.value = isReferenceType(value.value) ? cloneHandler(value.value) : value.value
+    }
+    catch (error) {
+      if (isEqualityClone)
+        throttled.value = 'Error refThrottled' as any
```
Doctor-wu, line 30:
```suggestion
        throttled.value = value.value
        console.error('Some error occurred in cloneHandler, used uncloned value', error)
```

packages/shared/refThrottled/index.ts
```diff
@@ -10,18 +14,31 @@ import { useThrottleFn } from '../useThrottleFn'
...
   if (delay <= 0)
     return value
+  const throttled = ref() as Ref<T>
+  const isEqualityClone = cloneHandler === cloneFnJSON
+  const setThrottled = () => {
+    try {
+      throttled.value = isReferenceType(value.value) ? cloneHandler(value.value) : value.value
+    }
+    catch (error) {
+      if (isEqualityClone)
```
Doctor-wu: need a brackets here

17359898647: Do I need an effectScope here? Like this
```typescript
  const updater = useThrottleFn(setThrottled, delay, trailing, leading)
  const scope = effectScope(true)
  scope.run(() => {
    watch(value, updater, {
      deep: isReferenceType(value.value),
    })
  })

  tryOnScopeDispose(() => {
    scope.stop()
  })
```

Doctor-wu: No, if the user uses the hook in the setup function, it will be disposed automatically

packages/shared/refThrottled/index.ts
```diff
@@ -10,18 +14,41 @@ import { useThrottleFn } from '../useThrottleFn'
  * @param  delay  A zero-or-greater delay in milliseconds. For event callbacks, values around 100 or 250 (or even higher) are most useful.
  * @param [trailing] if true, update the value again after the delay time is up
  * @param [leading] if true, update the value on the leading edge of the ms timeout
+ * @param [deep] default true, Depth monitoring, if false, only monitor the first level
+ * @param [immediate] default true, Whether to execute immediately
+ * @param [cloneHandler] By default, it use `JSON.parse(JSON.stringify(value))` to clone
  */
-export function refThrottled<T>(value: Ref<T>, delay = 200, trailing = true, leading = true) {
+export function refThrottled<T>(value: Ref<T>, delay = 200, trailing = true, leading = true, deep = true, immediate = true, cloneHandler = cloneFnJSON) {
```
Alfred-Skyblue, line 21: I think that this should take an options object here instead of adding more parameters.

17359898647: I'm trying.

PR #3726, opened 2024-01-17
docs(syncRefs): add example of multiple targets

NozomuIkuta:

<details>
<br>

>

</details>

---

### Description

No related issues.

While `syncRefs` accepts both single ref and multiple refs as an array, the documentation only describes the former case.

When I saw documentation of `syncRefs` for the first time, I thought `syncRefs` is another version of `syncRef` that accepts multiple refs (although it's not if I'm correct). But I can't find how to sync multiple refs because the documentation only describes how to pass a single ref.

This PR is to add description so that users can understand all usecases without seeing the code.

Another point is that, with this PR, the sample code comes to be aligned with the demo UI.

### Additional context

This is my very first contribution to VueUse.
So, I might not be aware of the past decisions, although I read all guidelines.

### DIFF
--- a/packages/shared/syncRefs/index.md
+++ b/packages/shared/syncRefs/index.md
@@ -13,15 +13,24 @@ Keep target refs in sync with a source ref
 import { syncRefs } from '@vueuse/core'
 
 const source = ref('hello')
-const target = ref('target')
+const target1 = ref('target1')
+const target2 = ref('target2')
 
-const stop = syncRefs(source, target)
+const stop = syncRefs(source, [target1, target2])
 
-console.log(target.value) // hello
+console.log(target1.value) // hello
+console.log(target2.value) // hello
 
 source.value = 'foo'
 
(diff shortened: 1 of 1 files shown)


packages/shared/syncRefs/index.md
```diff
@@ -13,15 +13,24 @@ Keep target refs in sync with a source ref
 import { syncRefs } from '@vueuse/core'
 
 const source = ref('hello')
-const target = ref('target')
+const target1 = ref('target1')
+const target2 = ref('target2')
 
-const stop = syncRefs(source, target)
+const stop = syncRefs(source, [target1, target2])
```
antfu, line 19: I think we should keep the first example simple and easy to understand. We could have a sub title `### Sync with multiple targets for that

NozomuIkuta: Thank you for the feedback!
I added another section: 851a300

PR #3764, opened 2024-02-02
feat(computedRef): new function

### LINKED ISSUE #3485: computedRef (same as reactiveComputed but returns a ref)

As a developer using Vue for CRUD, and developing update forms in particular, I want an easy and DRY way to prepare a reactive object for field values. The problem is, the initial field values should be copied/constructed based on the data that is coming asynchronously from the API, and that may also update (such as after successful submit and pulling updated data from the server). When such object is first retrieved and/or updated, the set of fields (for the editable inputs) should be re-created. When the source data is not (yet) available, as when it's still loading or has never loaded, the object must be `undefined`.

The typical DRY code looks like:

```vue
<script setup lang="ts">
import { computedRef } from "@vueuse/core"

const { data } = useAsyncData() // fetch data asynchronously, the result is typed as Ref<Data | undefined>

const fields = computedRef(() => data.value && ({
  // only pick certain fields that are editable
  name: data.value.name,
  address1: data.value.address1,
  address2: data.value.address2,
  // possibly add form fields that are not present in the data, but represent client-side logic
  copyAddress1ToAddress2: data.value.address1 === data.value.address2,
}))

// TODO: define watcher for address sync, omitting this for brevity.

async function submit() {
  const res = await callSubmitApi(fields.value)
  if (res.data) {
    // On successful submit, `fields` should be re-created with fresh data coming from the server.
    data.value = res.data
  }
}
</script>

<template>
  <form v-if="fields" @submit.prevent="submit">
    <input v-model="fields.name" />
    <input v-model="fields.address1" />
    <input type="checkbox" v-model="fields.copyAddress1ToAddress2" />
    <input v-model="fields.address2" :disabled="fields.copyAddress1ToAddress2"/>
  </form>
</template>
```

### Suggested solution

The implementation is very simple, currently I am copy/pasting it between projects:

```ts
import type { UnwrapRef, WatchSource } from "vue"

/** @returns ref which resets when the source changes */
export function computedRef<T>(source: WatchSource<T>) {
  const value = ref(toValue(source))
  watch(source, (source) => {
    value.value = source as UnwrapRef<T>
  })
  return value
}
```

### Alternative

VueUse already has `reactiveComputed`, but it's not directly suitable, as it must always return an object. Therefore:

1. It can't be used when the source data could be is unavailable (`reactiveComputed` can't return `undefined`).

2. It can't be used when the ref is a non-object. For instance, consider a case when a list can be expanded and collapsed, which is a boolean ref. The default state (collapsed or expanded) could be dependent on the list length, such as it's expanded  by default when it's 10 or less items, and collapsed if there's more. Then, when a list is updated, it must recalculate its expanded status based on the new length.

Of course, we could always work around and return an object shaped as `{ value: <actual value, possibly undefined> }`, but why not have a function that uses a native Vue `Ref` for that case? It will also unwrap naturally in templates.

### Additional context

Naming is subject for discussion. For consistency with `reactiveComputed`, it could be `refComputed`. (To be honest, I would rather reverse both, `computedReactive` and `computedRef` read better in my thinking.)

### Validations

IlyaSemenov:

---

### Description

Adds `computedRef` function, which returns a ref that derives from a watch source.

Fixes #3485.

### Additional context

1. I am aware that there is a slowdown on new functions. I elaborated use-cases in #3485, and I personally use this approach very often on CRUD's. The function is fully generalized, it's short and doesn't bring along any new dependencies.
2. I didn't test this in Vue 2, and in particular I'm not sure if I correctly marked that ref.trigger() will only work in Vue 3. This is a copy/paste from `computedEager`.

### DIFF
--- a/packages/shared/computedRef/index.md
+++ b/packages/shared/computedRef/index.md
@@ -1,0 +1,43 @@
+---
+category: Reactivity
+---
+
+# computedRef
+
+Instead of returning a non-reactive ref that `computed` does, `computedRef` returns a fully reactive ref.
+
+It's similar to `reactiveComputed`, but returns a ref rather than reactive, and thus can be used for non-object values (including `undefined`).
+
+## Usage
+
+```ts
+import { computedRef } from '@vueuse/core'
+
+const data = ref('foo')
+
+const state = computedRef(() => data.value)
+
+console.log(state.value) // foo
+
+state.name = 'bar'
+
+console.log(state.value) // bar
+
+data.value = 'baz'
+
+console.log(state.name) // baz
+```
+
+### Manual Triggering
+
+You can also manually trigger the update by:
+
+```ts
+const ref = computedRef(() => { /* ... */ })
+
+ref.trigger()
+```
+
+::: warning
+Manual triggering only works for Vue 3
+:::
--- a/packages/shared/computedRef/index.test.ts
+++ b/packages/shared/computedRef/index.test.ts
@@ -1,0 +1,37 @@
+import { isVue3, ref } from 'vue-demi'
+import { describe, expect, it } from 'vitest'
+import { computedRef } from '.'
+
+describe('computedRef', () => {
+  it('should work', () => {
+    const data = ref('foo')
+
+    const state = computedRef(() => data.value.toUpperCase())
+
+    expect(state.value).toBe('FOO')
+
+    state.value = 'bar'
+
+    expect(state.value).toBe('bar')
+
+    data.value = 'baz'
+
+    expect(state.value).toBe('BAZ')
+  })
+
+  it.runIf(isVue3)('custom trigger', () => {
+    const data = ref('foo')
+
+    const state = computedRef(() => data.value.toUpperCase())
+
+    expect(state.value).toBe('FOO')
+
+    state.value = 'bar'
+
+    expect(state.value).toBe('bar')
+
+    state.trigger()
+
+    expect(state.value).toBe('FOO')
+  })
+})
--- a/packages/shared/computedRef/index.ts
+++ b/packages/shared/computedRef/index.ts
@@ -1,0 +1,33 @@
+import { ref, toValue, watch } from 'vue-demi'
+import type { Ref, WatchSource } from 'vue-demi'
+
+export interface RefWithControl<T> extends Ref<T> {
+  /**
+   * Force update the ref value.
+   */
+  trigger(): void
+}
+
+/**
+ * Ref that syncs to the watch source.
+ *
+ * It's similar to the built-in `computed`, but returns a fully reactive ref.
+ */
+export function computedRef<T>(source: WatchSource<T>) {
+  const result = ref() as RefWithControl<T>
+
+  watch(source, (value) => {
+    result.value = value
+  }, {
+    immediate: true,
+    flush: 'sync',
+  })
+
+  if (Object.isExtensible(result)) {
+    result.trigger = () => {
+      result.value = toValue(source)
+    }
+  }
+
+  return result
+}


antfu:
I am ok with having it, but I am concerned this can be potentially confusing. Can you help with:

- Providing an explanation in the docs to clear the difference between `comnputed` and other computed utils in VueUse?
- Considering the name isn't really self-explaining (`computed` returns a ref, `computedRef` doesn't seem to provide more context), how about `watchComputed`?

IlyaSemenov:
Please check the updated PR, I tried to come with better (still brief) explanation on what this does.

The naming is of course a tough question, but I'm hesitating about `watchComputed`. The fact that it uses watch in the implementation is not actually important (moreover, I suppose it could be implemented without watch with `customRef`). 
 
Basically (as I mentioned in the updated docs) this function follows the same contract as the built-in `computed`, the difference is that the result is itself reactive (and can be used as a `v-model` target). It's thus kind of a blend between `computed` and `ref`, hence `computedRef`.

It's also very similar to what `reactiveComputed` does, the difference is that `reactiveComputed` only works with objects. The documentation for `reactiveComputed` says: _"Computed reactive object"_ so I'm not sure why it's named `reactiveComputed` and not `computedReactive` in the first place. If it were called `computedReactive` (following its own description!) then again `computedRef` will be completely along the lines with it.

Thoughts?

IlyaSemenov:
Perhaps `refComputed` will work? This is both similar to `reactiveComputed` and to the existing family of `refDebounced`, `refWithControl`, etc.

antfu:
Oh now I see, you want a mutable computed ref, that updates when the source changes but that you can modify it freely from the outside right?

antfu:
Maybe instead a new function, it can be part of https://vueuse.org/shared/computedWithControl/#computedwithcontrol ?

IlyaSemenov:
@antfu That's correct, I want a mutable computed ref, that updates when the source changes but which I can modify freely from the outside. This is extremely useful for CRUD fields (that should be able to be modified with `<input v-model>`, but that are themselves dependent on external data which can be updated from the server).

> Maybe instead a new function, it can be part of `computedWithControl`?

Right, that's what I was actually thinking at first: we could extend `computedWithControl` to have **optional** second argument, and use the watch source as the value of `fn` by default.

However, `computedWithControl` returns a non-reactive computable. Consider a (now failing) test case:

```ts
it('is reactive', () => {
  const trigger = ref(0)
  const data = ref('foo')

  const fields = computedWithControl(trigger, () => ({ data: data.value }))
  const fieldsData = computed(() => fields.value.data) // For testing fields.value reactivity

  expect(fields.value.data).toBe('foo')
  expect(fieldsData.value).toBe('foo')

  fields.value.data = 'bar'

  expect(fields.value.data).toBe('bar')
  expect(fieldsData.value).toBe('bar') // <------- Here it fails, because fields.value is not reactive.

  trigger.value += 1

  expect(fields.value.data).toBe('foo')
  expect(fieldsData.value).toBe('foo')
})
```

IlyaSemenov:
I can add options parameter if that's what you think is better:

```ts
computedWithControl(() => data.value, () => ({ data: data.value }), { reactive: true })
```

or add a third overload:

```ts
computedWithControl(() => data.value, {
  get() {
    return { data: data.value }
  },
  reactive: true,
})
```

but personally I think it's kind of abusing it.

IlyaSemenov:
I just realized that the implementation is not behaving correctly, or better said, unexpectedly.

Consider the scenario:
1. Ref is calculated to A.
2. User changes it to B.
3. Ref dependencies update, the watch source runs but *recalculates* the result again to A.
4. Now what?
    a. We can reset the ref to A (this is what I would expect).
    b. We can keep the ref value as B (as technically the watch source didn't change and we advertise to only update when it updates).

The current PR goes with (b). Instead, I personally need (a) in most cases, and ended up with:

```ts
export function computedRef<T>(source: WatchSource<T>) {
  const result = ref() as Ref<T>
  watchEffect(() => {
    result.value = toValue(source)
  })
  return result
}
```

but this somewhat contradicts the naming.

OrbisK:
> I just realized that the implementation is not behaving correctly, or better said, in an expected manner.
> 
> Consider the scenario:
> 
> 1. Ref is calculated to A.
> 2. User changes it to B.
> 3. Ref dependencies update, the watch source runs but _recalculates_ the result again to A.
> 4. Now what?
>    a. We can reset the ref to A (this is what I would expect).
>    b. We can keep the ref value as B (as technically the watch source didn't change and we advertise to only update when it updates).
> 
> The current PR goes with (b). Instead, I personally need (a) in most cases, and ended up with:
> 
> ```ts
> export function computedRef<T>(source: WatchSource<T>) {
>   const result = ref() as Ref<T>
>   watchEffect(() => {
>     result.value = toValue(source)
>   })
>   return result
> }
> ```
> 
> but this somewhat contradicts the naming.

hey, is there anything I can do to support you? How should we proceed with the PR? :)

IlyaSemenov:
For starters, I am not sure what should be the correct behavior. When should the returned ref reset, when its dependencies change or when the computed value changes? Both approaches makes sense in different scenarios. Should it be two functions or one function with an option (required or having some default? which one then?). I don't have a clear vision myself hence marked this as draft.

antfu:

IlyaSemenov:
I see that you closed the PR due to "NOT accepting new functions" but didn't close the original issue that the PR was supposed to solve (#3485). The reasoning was included in the original issue, it explained two different and supposedly very common examples of use (CRUD forms and extending deeply nested props with reactive local state).

I read the contribution guide before submitting the issue and the PR and it wasn't mentioning that you're NOT accepting new functions (and it doesn't now). It only mentions "slowing down" which is reasonable. However, I believe the proposed `computedRef` is not just another one-of-zillion rare-edge-case function: it blends the two most popular concepts in Vue reactivity: a reactive ref and a computed value, allowing to create a computed value that is also reactive. I am not trying to object your decisions, I'm just saying I was under impression being the very generic function is a valid reason for the waiver for the rule.

This PR was stuck and not pushed by me because there were two different ways to implement it which behaved slightly differently. The PR included one implementation and I found myself actually using another one (which I included in the comments). I was hoping someone from the maintainers would make a decision on how to proceed.

That said, I agree that if there's no interested from maintainers then the function should not be included in the repo, so again, I'm not arguing, just wanted to express the state of things for possible future reference.

Issue #3203, opened 2023-07-02
useLongPress doesn't prevent mouse click


Andy-d-g:
### Describe the bug

Hello, 

Using UseLongPress with the mouse doesn't prevent the click event.
It works fine with the mobile press.

If I make a long press with my trackpad, both counter will be increase same if I add the `prevent` attribute which should avoid to increase the counter of the "basicCounter"

Thank you

### Reproduction

https://codesandbox.io/s/peaceful-rain-pqk7qt?file=/src/App.vue:72-92

### System Info

```Shell
System:
    OS: macOS 13.4.1
    CPU: (8) x64 Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
    Memory: 33.29 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.1 - ~/.nvm/versions/node/v18.16.1/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.1/bin/npm
  Browsers:
    Chrome: 107.0.5304.110
    Safari: 16.5.1
    Firefox: 114.0.2 (64 bits)
  npmPackages:
    @vueuse/components: ^10.2.1
    @vueuse/core": ^9.2.0
    vue: ^3.3.4
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

JiuRanYa:
[...]

PR #3791, opened 2024-02-15
feat(OnLongPress): `options.onMouseUp` callback

noook:

---

### Description

In issue #3203 it is briefly discussed that it's difficult to make `onLongPress` and `onClick` co-exist without having to hack it locally. This PR introduces the suggested logic into the `OnLongPress` renderless component.

### Additional context

- I wonder if we should add a way to use the click as an option to the composable
- I believe we can add this behaviour to the directive, but it relies on the composable and not the component, so finding a way to add this option to the composable could make sense
- This can be tested only if I add the click callback as an option to the composable
- I'll revert the commit for the demo, it is just here to show how it works
- I'm not sure whether I should update the documentation, as this is something that most users would do without expecting a different behaviour


antfu:
I am not very sure about this, ideally the `click` event should be a native event and a component should not emit it. Meanwhile, if the logic only exists in the component version, the behavior misalignment with the composable one might end up confusing.

noook:
I completely understand your point. I was trying to achieve what was described in the mentioned issue. Do you suggest doing it differently ? Like another event name so we keep the native click behavior, and possibly the `release` event ? all of that through the options ?

antfu:
Maybe we could have an extra option to pass a handler, like `onMouseUp(duration: number, distance: number, isLongPress: boolean)` so users could pass custom logic to handle if they expect to trigger a click event or so?

packages/core/onLongPress/index.ts
```diff
@@ -42,13 +50,41 @@ export function onLongPress(
...
+    startTimestamp = undefined
+    hasLongPressed = false
+  }
+
+  function onRelease(ev: PointerEvent) {
+    if (options?.modifiers?.self && ev.target !== elementRef.value)
+      return
+
+    if (!options?.onMouseUp)
+      return
```
noook, line 71: todo: early returns must also call `clear()`, or we could call clear early, and save a copy of variables that should be used in `options.onMouseUp()`

noook:
Sorry messed up the rebase, will probably reopen the PR

antfu (approved):

PR #3379, opened 2023-09-07
fix(useClipboard): use legacy way when without permission

Cat1007:
fixes #2494

### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

use legacy methods as a fallback when clipboard is supported but without permission

### Additional context

---

### <samp>🤖 Generated by Copilot at f04e403</samp>

This pull request enhances the `useClipboard` function with permission checks. It uses the `usePermission` function to query and request the clipboard permission before reading or writing to the clipboard.

### <samp>🤖 Generated by Copilot at f04e403</samp>

* Import `usePermission` function to check clipboard access ([link](https://github.com/vueuse/vueuse/pull/3379/files?diff=unified&w=0#diff-104a20ea8a35b01c36721b36a81d3138312e27a997280d66290ce7e8a9bc466aR11))
* Declare `permissionRead` and `permissionWrite` variables to store the result of `usePermission` ([link](https://github.com/vueuse/vueuse/pull/3379/files?diff=unified&w=0#diff-104a20ea8a35b01c36721b36a81d3138312e27a997280d66290ce7e8a9bc466aR66-R67))
[...]

PR #3812, opened 2024-02-22
fix(useClipboard): fix issue when permission is not defined

Mister-Hope:

  No, it's hard to simulate a browser like wechat browser

### Description

There are a lot of built-in browsers (Chinese app likes them), which does not implement permission api return value, for example on latest Wechat client  8.0.47, the demo of the usePermission page looks like this:

Screenshot_2024-02-22-14-21-02-41_e39d2c7de19156b

That is because when calling with `usePermission('xxx')`, these browsers return `undefined` (including the 'clipboard-write' and 'clipboard-read' used in `useClipboard`

When the value is undefined, we'd better treat it as `denied`, e.g. wechat browser just denied it with out any asking for security reasons:

Screenshot_2024-02-22-13-58-03-41_e39d2c7de19156b

The issue is included in https://github.com/vueuse/vueuse/pull/3379 as a feature for falling back to legacy if possible.

### DIFF
--- a/packages/core/useClipboard/index.ts
+++ b/packages/core/useClipboard/index.ts
@@ -71,7 +71,7 @@ export function useClipboard(options: UseClipboardOptions<MaybeRefOrGetter<strin
   const timeout = useTimeoutFn(() => copied.value = false, copiedDuring)
 
   function updateText() {
-    if (isClipboardApiSupported.value && permissionRead.value && permissionRead.value !== 'denied') {
+    if (isClipboardApiSupported.value && isAllowed(permissionRead.value)) {
       navigator!.clipboard.readText().then((value) => {
         text.value = value
       })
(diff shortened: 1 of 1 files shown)


posva (commented):
In this case, it's more about the feature being supported or not. So the fix should likely go in `isClipboardApiSupported`

posva:
If it always fail, what part of the API is exactly supported?

Mister-Hope:
The clipboard api is supported by wechat browsers, and a whitelist of websites can call `clipboard.write()` (don't be too amazing, this is how big company in china works). It's just normal website is not allowed "for security consideration".

Mister-Hope:
I made an update on the above comment, and if you inisit, I can do a full refactor, to only set `isSupported` to `true` if `clipboard.write()` permission is `granted` only.

posva:
I think the version is fine. I guess the `undefined` value is not meant to be valid? 😅 . I'm not 100% sure but it doesn't seem to be covered by the spec. I think it's worth adding a short comment about why not just !== denied for future maintainers. FYI I'm not making any calls here, just trying to help in the maintenance 😄

antfu (approved):

PR #3822, opened 2024-02-24
feat(useStorage): conditionally use event based on the used storage backend

ferferga:

<details>
<br>

>

</details>

---

### Description

Build upon the work of #2630. Adds a single event listener every time (instead of 2 in all cases), depending on the type of storage backend used.

### Additional context

Besides using strictly the minimum amount of event listeners needed, this might prevent race conditions in the future.

### DIFF
--- a/packages/core/useStorage/index.ts
+++ b/packages/core/useStorage/index.ts
@@ -180,8 +180,11 @@ export function useStorage<T extends(string | number | boolean | object | null)>
   if (window && listenToStorageChanges) {
     tryOnMounted(() => {
       // this should be fine since we are in a mounted hook
-      useEventListener(window, 'storage', update)
-      useEventListener(window, customStorageEventName, updateFromCustomEvent)
+      if (storage instanceof Storage)
+        useEventListener(window, 'storage', update)
+      else
+        useEventListener(window, customStorageEventName, updateFromCustomEvent)
+
       if (initOnMounted)
         update()
     })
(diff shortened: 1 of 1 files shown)


antfu (approved):

michaelcozzolino:
after this, not working on brave browser [Version 1.67.134 Chromium: 126.0.6478.186 (Official Build) (64-bit)](https://brave.com/latest/)

ferferga:
@michaelcozzolino Can you make a reproduction on play.vuejs.org please?

michaelcozzolino:
just go here with brave browser or probably chrome https://vueuse.org/core/useStorage/ and try the example 

https://github.com/user-attachments/assets/6fb5a014-5d4e-473c-b9d4-d2c6a1902a09

ferferga:
@michaelcozzolino Looks like the original behaviour before the custom event was introduced in [this commit](https://github.com/vueuse/vueuse/commit/c6185bf9fe2d6b7f2902ef459e731d7d2ee9e9f5) was modified. I had these changes locally for a while (the composable was fully tested informally before making this PR), but definitely something happened in the middle that slipped away in the final version (sorry!) and now the composable only works if it receives an event, which was not the original behaviour (and definitely less performant since the ref is overwritten everytime).

A revert of this PR is not a solution since the hard reliance of the composable on events also breaks the composable if ``listenToStoreChanges: false`` is passed as an option. I'm currently trying to bisect all the changes properly and will try to make a PR later today, thank you very much!

ferferga:
@michaelcozzolino Pushed a PR that fixes this https://github.com/vueuse/vueuse/pull/4152

Informally tested the composable but using the same storage in the same document is not an use case I use, sorry for the regression! Also added tests to avoid those kind of regressions in the future.

PR #3583, opened 2023-11-23
fix(useRouteParams,useRouteQuery): set route param/query to undefined when null or defaultValue

mtdvlpr:
### Before submitting the PR, please make sure you do the following

- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [X] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [X] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [X] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

This PR fixes the regression that was introduced in #3184. Before this PR and before `v10.3.0`, the route query/param was set to undefined when null or defaultValue. This makes sure that the query/param is removed from the url and not present with an empty value.

Currently, when the `page` query is set to null, the url looks like `website.com?another=something&page` instead of `website.com?another=something`

Fixes #3290

### Additional context

This PR is based on #3291. It's not a breaking change. Removing this behavior in `v10.3.0` might be considered a breaking change. This PR is just to fix a regression.

PR #3831, opened 2024-02-29
fix(useRouteParams,useRouteQuery): set route param/query to undefined when function-based defaultValue

henribru:

<details>
<br>

>

</details>

---

### Description

This PR follows up on the changes from https://github.com/vueuse/vueuse/pull/3583. That PR ensured that if you set a route param/query to the default value, it was removed from the query string. However, it doesn't handle function-based default values, so if you set the param/query to the value the default function evaluates to, it remains in the query string. This PR evaluates function-based default before the comparison to avoid that.

### DIFF
--- a/packages/router/useRouteParams/index.test.ts
+++ b/packages/router/useRouteParams/index.test.ts
@@ -243,4 +243,22 @@ describe('useRouteParams', () => {
     expect(page.value).toBe(2)
     expect(lang.value).toBe('en-US')
   })
+
+  it('should reset value when default function value', async () => {
+    let route = getRoute({
+      page: 2,
+    })
+    const router = { replace: (r: any) => route = r } as any
+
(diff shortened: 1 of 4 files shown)


henribru:
On second thought, maybe this isn't a good idea. When the default is a function, it could change. If you've explicitly set the query to a value, it might be unexpected for that query to change later if the value you set it to happened to be the default at that time.

henribru:
> On second thought, maybe this isn't a good idea. When the default is a function, it could change. If you've explicitly set the query to a value, it might be unexpected for that query to change later if the value you set it to happened to be the default at that time.

I suppose one way around this could be that when you set the query to the current default value, we still set the internal value, but keep it out of the URL params. We could then watch the default and if it changes to something else than the internal value, we could push the value to the URL params. But that's a lot of extra complexity just to keep the URL clean.

PR #3833, opened 2024-03-01
feat(ignoreDrag): adds ignoreDrag option to prevent onClickOutside handler to fire when a user is dragging outside target element

bobbybol:

- [ x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [ x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [ x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [ x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).

<details>
<br>

>

</details>

---

### DIFF
--- a/packages/core/onClickOutside/index.ts
+++ b/packages/core/onClickOutside/index.ts
@@ -12,6 +12,11 @@ export interface OnClickOutsideOptions extends ConfigurableWindow {
    */
   ignore?: (MaybeElementRef | string)[]
   /**
+   * Ignore drag so the handler does not fire if user drags instead of single clicks
+   * @default false
+   */
+  ignoreDrag?: boolean
+  /**
    * Use capturing phase for internal event listener.
    * @default true
    */
@@ -40,7 +45,7 @@ export function onClickOutside<T extends OnClickOutsideOptions>(
   handler: OnClickOutsideHandler<{ detectIframe: T['detectIframe'] }>,
   options: T = {} as T,
 ) {
-  const { window = defaultWindow, ignore = [], capture = true, detectIframe = false } = options
+  const { window = defaultWindow, ignore = [], ignoreDrag = false, capture = true, detectIframe = false } = options
 
   if (!window)
     return noop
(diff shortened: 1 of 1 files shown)


packages/core/onClickOutside/index.ts
```diff
@@ -68,7 +74,7 @@ export function onClickOutside<T extends OnClickOutsideOptions>(
       const dY = Math.abs(y - e.y)
       if (dX + dY > 4) {
         shouldListen = false
-        stopDragListener && stopDragListener()
+        stopDragListener?.()
```
bobbybol, line 82: ❤️

brokuka:
any news?

bobbybol:
@antfu What about this PR?

packages/core/onClickOutside/index.ts
```diff
@@ -55,6 +71,18 @@ export function onClickOutside<T extends OnClickOutsideOptions>(
   }
 
   let shouldListen = true
+  let stopDragListener: undefined | Function
+
+  const startDragListener = ({ x, y }: PointerEvent) => {
+    stopDragListener = useEventListener(window, 'pointermove', (e) => {
+      const dX = Math.abs(x - e.x)
```
OrbisK, line 78: not sure if we should call `useEventListener` inside a function. 

This might lead to unexpected behaviour.

bobbybol:
> @bobbybol can you please dont create the eventlistener inside the function. if we call this function multiple times, it will create a memory leak.

Hi @OrbisK I can go two ways with this:
A. Make absolutely sure the listener is removed before starting another one, while still having the listener inside the function. Something like:
```
if (stopDragListener) {
  stopDragListener()
  stopDragListener = undefined
}
```
B. Have a persistent listener (only attached if `ignoreDragging` isn't `false` of course) that compares mouse coordinates to the point where the original pointerdown occurred. This is a little more involved because now you need to store the coordinates of the original pointer down event in a variable outside of all the handler functions, but can be mutated by three different event handlers (click, pointerdown, and pointermove),

Let me know which route you prefer, and I will make it happen ☺️.

PR #3870, opened 2024-03-17
fix(useWebSocket): urlRef changes were not being tracked

ferferga:

<details>
<br>

>

</details>

---

### Description

URLRef changes were not being tracked inside useWebSocket

### DIFF
--- a/packages/core/useWebSocket/index.ts
+++ b/packages/core/useWebSocket/index.ts
@@ -1,5 +1,5 @@
 import type { Ref } from 'vue-demi'
-import { ref } from 'vue-demi'
+import { ref, watch } from 'vue-demi'
 import type { Fn, MaybeRefOrGetter } from '@vueuse/shared'
 import { isClient, isWorker, toRef, tryOnScopeDispose, useIntervalFn } from '@vueuse/shared'
 import { useEventListener } from '../useEventListener'
@@ -304,6 +304,8 @@ export function useWebSocket<Data = any>(
 
   if (immediate)
     open()
+
+  watch(urlRef, open)
 
   return {
     data,


GarlandQian:
If I haven't opened the WebSocket, I think it shouldn't be opened directly when the URL changes. It should maintain the current state.

GarlandQian:
If I haven't opened the WebSocket, I believe it shouldn't be opened directly when the URL changes. It should maintain the current state. Even when the immediate parameter is set to false, I think regardless of the current WebSocket state, it should be closed directly.

ferferga:
@GarlandQian In that case, what's the benefit of using useWebSocket instead of the "raw" WebSocket instance?

This is a regression, not a new feature.

DrColossos:
Sorry to jump in here: this just arose as a "bug" in our internal app. I was surprised with the behavior, especially since I use `immediate: false`. This is somehow unexpected since we have another server that retrieves the current websocket URL and passes it on to `useWebsocket` to be triggered at a later point in time.

ferferga:
@DrColossos First of all, this PR fixed a regression, so that behaviour worked in your application by "taking advantage" of that regression. But according to what's stated in the docs, what you say is the behaviour that it should have.

In either case, in my opinion the docs are incorrect and the implementation is right: in all other composables, `immediate` just controls if the composable starts at the beginning or not, instead of whether "reacting to effects" is enabled or not. Having the option to disable the "reactivity" makes no sense in this library, since the sole reason to exist is to make common APIs easy to use reactively.

In your case, I think your logic can easily be simplified: why not have that URL be computed (or `computedAsync` from this library if it's fetched from a server with reactive params) and pass that to this composable? If that computedref is turned nullish, the socket will disconnect, if the URL changes, the socket will point to the new one accordingly, in the same way you expect a component's DOM to update when it's reactive dependencies are updated.

DrColossos:
Thank you for you info. I was reading the `immediate` flag "Automatically open a connection" from the source of the types. Looking at the rest of the online docs, I see full the description of the flag states what you say. I was only looking at the types. 

Anyways, thanks for clarification!

PR #3316, opened 2023-08-15
feat(useWebWorkerFn): support local dependencies

dvh91:
### Description

fixes #2005 

`useWebWorkerFn` used to support remote dependencies for the function passed but not local functions.
the library which this function is a port of (`useWorker` by Alessio Koci) have the support of local dependencies.

The addition is also very much a port of `useWorker` implementation.

### Additional context

In order to support this feature, I've added a new dependency on core package called `isoworker`. as the core package is very light with dependencies, im unsure if thats fine or it's enough for us to split it into integration package of its own. open for thoughts 

---

### <samp>🤖 Generated by Copilot at 7e11dfb</samp>

This pull request adds a new feature to the `useWebWorkerFn` hook that allows users to pass local dependencies to the web worker. It uses the `isoworker` module to create a context for the dependencies and inject them into the worker blob URL. It also updates the documentation, the demo code, and the dependencies for the `@vueuse/core` package.

### <samp>🤖 Generated by Copilot at 7e11dfb</samp>

*  Add `isoworker` dependency to `@vueuse/core` package to support local dependencies in `useWebWorkerFn` hook ([link](https://github.com/vueuse/vueuse/pull/3316/files?diff=unified&w=0#diff-0b810c38f3c138a3d5e44854edefd5eb966617ca84e62f06511f60acc40546c7L43-R44), [link](https://github.com/vueuse/vueuse/pull/3316/files?diff=unified&w=0#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR256-R258), [link](https://github.com/vueuse/vueuse/pull/3316/files?diff=unified&w=0#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR8077-R8080))
[...]

PR #3899, opened 2024-03-31
feat(useWebWorkerFn): support local function dependencies

### LINKED ISSUE #2005: Extend useWebWorkerFn to also accept an array of local dependencies

`useWebWorkerFn` provides an option called `dependencies` in which you can give an array of strings that are paths to javascript files that the web worker will then import so that it can use any variables and functions that are globally defined in those javascript files.

However, often times you already have some functions in your local scope and it would be *much* more convenient to be able to just provide an array of references to those functions in an option called `localDependencies`.

So it might look something like this:

```js
import { useWebWorkerFn } from '@vueuse/core'
import { expensiveAdder } from './utils'

const { workerFn, workerStatus, workerTerminate } = useWebWorkerFn(
  (a, b) => expensiveAdder(a,b),
  {
    timeout: 50000,
    localDependencies: () => [expensiveAdder], // we pass the local function to the worker
  },
)
```

### Suggested solution

According to the [docs](https://vueuse.org/core/useWebWorkerFn/#credit) `useWebWorkerFn` is a port from [Alessio Koci's useWorker](https://github.com/alewin/useWorker), which according to [its docs](https://useworker.js.org/docs/api-useworker/#local-dependencies-example) allows the developer to provide an array of references to local functions so that they can be used inside the worker.

This is different from the `externalDependencies` option (which vueuse has renamed to `dependencies`) that already exists, because then you really need to create a separate file for that function and then provide it using the path to that file. The `localDependencies` option is just way more convenient in many cases.

And from what I understand, Alessio Koci also didn't create this feature himself, but that it ultimately comes from [Arjun Barrett's isoworker](https://github.com/101arrowz/isoworker#usage).

So basically my request is, can you please copy that feature into `useWebWorkerFn`? And maybe, if you're gonna add an option called `localDependencies`, then I would deprecate the `dependencies` and rename it to `externalDependencies`.

### Alternative

I can't think of an alternative other than just keeping everything as it is now... Like it isn't *impossible* to do certain things without this `localDependencies` option, it's just less convenient.

### Additional context

_No response_

### Validations

thefeymesaleng:
**Description**

This PR fixes #2005 and also addresses the issue with #3316 . The implementation provides a solution that deviates slightly from [alewin/useWorker](https://github.com/alewin/useWorker), due to a known issue with `isoworker`. Specifically, the localDependencies feature has been temporarily removed in their implementation. However, recognizing the high anticipation for this feature, I've introduced a simplified method inside`depsParser` to enable its functionality.

<details>
<br>

>

</details>

---

### DIFF
--- a/packages/core/useWebWorkerFn/index.md
+++ b/packages/core/useWebWorkerFn/index.md
@@ -34,6 +34,22 @@ const { workerFn, workerStatus, workerTerminate } = useWebWorkerFn(
 )
 ```
 
+### With local dependencies
+
+```ts {9-9}
+import { useWebWorkerFn } from '@vueuse/core'
+
+const pow = (a: number) => a * a
+
+const { workerFn, workerStatus, workerTerminate } = useWebWorkerFn(
+  numbers => pow(numbers),
+  {
+    timeout: 50000,
+    localDependencies: [pow]
+  },
+)
+```
+
 ## Web Worker
 
 Before you start using this function, we suggest you read the [Web Worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers) documentation.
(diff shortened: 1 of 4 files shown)


rkmuthuram:
Much needed

PR #3903, opened 2024-04-01
fix(useExtractedObservable,watchExtractedObservable): error when extractor is undefined

MstrVLT:

<details>
<br>

>

</details>

---

### Description

if useExtractedObservable or watchExtractedObservable extractor returns an undefined value. Errors appear in the console log

### DIFF
--- a/packages/core/onLongPress/index.md
+++ b/packages/core/onLongPress/index.md
@@ -118,7 +118,7 @@ function resetDirective() {
 
   <button
     v-on-long-press="[onLongPressCallbackDirective, { delay: 1000, modifiers: { stop: true } }]"
-    class="ml-2button small"
+    class="ml-2 button small"
   >
     Press long (with options)
   </button>
--- a/packages/core/useDevicePixelRatio/index.md
+++ b/packages/core/useDevicePixelRatio/index.md
@@ -4,7 +4,7 @@ category: Sensors
 
 # useDevicePixelRatio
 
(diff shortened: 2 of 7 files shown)


MstrVLT:
Tests were added but had to be removed I haven't found a good way to test this error

MstrVLT:
I also wrote a test, but it's too complicated
https://stackblitz.com/edit/vitest-dev-vitest-fgvmcd?file=test%2Fbasic.test.ts

Voltra:
Hi @MstrVLT !

That is by design as it's meant to always extract/map an observable from the vue reactive value.

IMHO, I think this is more of an XY problem: How to use `useExtractedObservable` and `watchExtractedObservable` if my function is partial and not total?

You can in fact return an empty observable to handle those cases, and in that case the value type should always be `T|undefined`.

These composables were made to behave like `useObservable` and `watch` but lifted to handle more complex cases based on vue reactive data.

The extractor is meant to be a 1-1 mapping (as per the typedef), and not a 1-0/1 mapping

Voltra:
For future reference: I don't mind being tagged in PR/issues regarding the composables I've written.

It'll probably help with intent and might result in adding laravel-style comments to the code (and addendum to the docs) to properly document these kinds of choices

antfu:
Closing due to inactivity. Please open a new issue referencing to this one if you can follow up with more information.

PR #3904, opened 2024-04-02
fix: side effect when element turn from null to element

Hengry:

- [v] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [v] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [v] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [v] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).

<details>
<br>

>

</details>

---

### Description

in `useMediaControls`, if the target element is assigned by a callback function, then the properties of the element will be set to the initial value of the refs.

For example:
```js
<script>
  const elRef = ref(null)
  const useMediaControls()
  const elRefRegister = (e) => {
    elRef.value = e
  }
</script>
<html>
  <video :ref="elRefRegister" muted />
</html>
```

In this case, elRef will be set to unmuted

### Additional context

I also considered to sync up internal refs with the target element if the element changes. While it changes the behavior, also the [previous commit](https://github.com/vueuse/vueuse/commit/b20aacf5f1f8da9bd5ebcb198480617526ce5ed1) shows it was designed to overwrite new the element when the element changing.

### DIFF
--- a/packages/core/useMediaControls/index.ts
+++ b/packages/core/useMediaControls/index.ts
@@ -294,7 +294,10 @@ export function useMediaControls(target: MaybeRef<HTMLMediaElement | null | unde
   /**
    * Apply composable state to the element, also when element is changed
    */
-  watch([target, volume], () => {
+  watch([target, volume], (_, [oldTarget]) => {
+    if (!oldTarget)
+      return
+
     const el = toValue(target)
     if (!el)
       return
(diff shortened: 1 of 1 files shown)


antfu:
Can you elaborate more? I don't understand what's the problem you are trying to solve here

Hengry:
For example:

```
<script>
  const elRef = ref(null)
  const controls = useMediaControls(elRef)
  const elRefRegister = (e) => {
    elRef.value = e
  }
</script>
<html>
  <video :ref="elRefRegister" muted />
</html>
```

I expect the `<video />` be muted. But actually, it would be unmuted.
This is due to the side effects of `useMediaControls`.

PR #3914, opened 2024-04-06
feat(useGamepad): add PS5 DualSense mapping

yue4u:

<details>
<br>

>

</details>

---

### Description

Add PS5 DualSense Controller mapping related to `useGamepad`

### Additional context

I guess developer needs will diff whether they want standard(ish) names (bumper / trigger) or more device specific names (L2 / R2). Current approach is trying to support both.

### DIFF
--- a/packages/core/useGamepad/index.ts
+++ b/packages/core/useGamepad/index.ts
@@ -12,43 +12,19 @@ export interface UseGamepadOptions extends ConfigurableWindow, ConfigurableNavig
 }
 
 /**
- * Maps a standard standard gamepad to an Xbox 360 Controller.
+ * Maps a [standard gamepad](https://w3c.github.io/gamepad/#remapping) to an Xbox 360 Controller.
+ * https://support.xbox.com/en-US/help/xbox-360/accessories/controllers
  */
 export function mapGamepadToXbox360Controller(gamepad: Ref<Gamepad | undefined>) {
   return computed(() => {
     if (gamepad.value) {
       return {
+        ...mapGamepadToStandardController(gamepad.value),
         buttons: {
           a: gamepad.value.buttons[0],
           b: gamepad.value.buttons[1],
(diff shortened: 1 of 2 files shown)


packages/core/useGamepad/index.ts
```diff
@@ -59,6 +35,69 @@ export function mapGamepadToXbox360Controller(gamepad: Ref<Gamepad | undefined>)
...
+          '○': gamepad.value.buttons[1],
+          '□': gamepad.value.buttons[2],
+          '△': gamepad.value.buttons[3],
+          'L1': gamepad.value.buttons[4],
+          'R1': gamepad.value.buttons[5],
+          'L2': gamepad.value.buttons[6],
+          'R2': gamepad.value.buttons[7],
+        },
+        create: gamepad.value.buttons[8],
+        options: gamepad.value.buttons[9],
```
yue4u, line 60: IMO `create`/`options` and (`back`/`start` for XBOX) should be inside `buttons` but that will be a breaking change and doing this only in `mapGamepadToPS5DualSenseController` will cause inconsistency.

packages/core/useGamepad/index.ts
```diff
@@ -59,6 +35,69 @@ export function mapGamepadToXbox360Controller(gamepad: Ref<Gamepad | undefined>)
...
+export function mapGamepadToPS5DualSenseController(gamepad: Ref<Gamepad | undefined>) {
+  return computed(() => {
+    if (gamepad.value) {
+      return {
+        ...mapGamepadToStandardController(gamepad.value),
+        buttons: {
+          '⨯': gamepad.value.buttons[0],
+          '○': gamepad.value.buttons[1],
+          '□': gamepad.value.buttons[2],
+          '△': gamepad.value.buttons[3],
```
yue4u, line 53: These are picked visually and may have better alternatives.

PR #3946, opened 2024-05-02
feat(watchOnce): allow watching until certain condition is met

BR0kEN-:
### Description

Add the optional `until` callback to the `watchOnce` options to keep watching until certain condition is met.


antfu:
I am not sure about this, having this would change the semantic of `watchOnce` - that might end up the handler being called multiple times. I guess it might be better for you to just call the stop handler returned by watch when you no longer need it.

Issue #2861, opened 2023-03-12
useWebSocket should allow separate ping/pong message


aphex:
### Clear and concise description of the problem

Added here https://github.com/vueuse/vueuse/pull/2170/files#diff-9b23df18f0f5236a301bc5926a49b4e0ea571aa84752304cc34a2f44047fa0f1R250

Current the heartbeat allows for a ping message to be customized and the same message is used to determine if it is a response. This seems odd as generally this message is `ping` with a response of `pong`. This gets stranger with something like AWS where we use the `{action: 'ping'}` syntax. Ideally we would be able to specify a `message` and a `response` in the heartbeat object. 

Ideally we wouldn't have to do any of this and we could just send a 0x9 opcode. (https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#pings_and_pongs_the_heartbeat_of_websockets) However I do not believe this is possible from the browser. If anyone can correct me though please do, AWS does not charge for 0x9 ping messages, whereas sending a custom ping will add up. It's frustrating.

This does exist in `ws` however I think this is only applicable for node (https://github.com/websockets/ws/blob/8a7016dc2fe4d9d63c428c67588d7c1f33a72e5c/lib/sender.js#L209)

this can also be tested with wscat (https://www.npmjs.com/package/wscat)
You need to enable slash codes like this `wscat -c wss://url --slash -P` you can then send a message `/ping` which should result in a pong response using ping/pong frames.

Again I don't think any of this is possible with the browsers websocket client, but would love to be shown I am wrong.

### Suggested solution

If opcode is not an option allowing for a custom heartbeat `message` and `response` will allow more common `ping` vs `pong` message as well as preventing the strange response from aws sockets

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #3950, opened 2024-05-03
feat(useWebSocket): allow different heartbeat response message

lavolpecheprogramma:
### Description

As mentioned in #2861 it can be useful to provide a custom response message.
For example, AWS does not charge for a ping/pong request, but with the current implementation the request must be ping/ping.
I made changes to be backwards compatible with the existing user implementation.

### Additional context

In my opinion, the default responseMessage should be "pong", but it will be a breaking change.


packages/core/useWebSocket/index.ts
```diff
@@ -251,8 +256,9 @@ export function useWebSocket<Data = any>(
         resetHeartbeat()
         const {
           message = DEFAULT_PING_MESSAGE,
+          responseMessage,
```
antfu, line 259:
```suggestion
          responseMessage = message,
```

packages/core/useWebSocket/index.ts
```diff
@@ -251,8 +256,9 @@ export function useWebSocket<Data = any>(
...
         } = resolveNestedOptions(options.heartbeat)
-        if (e.data === message)
+        if (e.data === (responseMessage === undefined ? message : responseMessage))
```
antfu, line 261:
```suggestion
        if (e.data === responseMessage)
```

antfu (approved):

Issue #3158, opened 2023-06-15
refDebounced doesn't work with objects


HugeLetters:
### Describe the bug

`debTest` gets updated immediately instead of after a 1000ms delay.
```typescript
const test = ref({ value: 'TEST' });
const debTest = refDebounced(test, 1000, { maxWait: 2000 });
```

Doing this works but is quite wordy with bigger objects.
```typesript
const test = ref({ value: 'TEST' });
const debTest = refDebounced(
	computed(() => ({ value: test.value.value })),
	1000,
	{ maxWait: 2000 }
);
```

### Reproduction

examples provided above

### System Info

```Shell
System:
    OS: Windows 10 10.0.22621
    CPU: (16) x64 AMD Ryzen 7 5700U with Radeon Graphics
    Memory: 2.19 GB / 15.33 GB
  Binaries:
    Node: 20.1.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.6.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.1702.0), Chromium (114.0.1823.43)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @vueuse/core: ^10.1.2 => 10.1.2
    @vueuse/nuxt: ^10.1.2 => 10.1.2
    vue: ^3.3.2 => 3.3.4
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

HugeLetters:
[...]

PR #3969, opened 2024-05-14
feat(refDebounced): add support for objects

XavierChevalier:

---

### Description

Objects did not work with `refDebounced`. An object need to be deeply clone to avoid JavaScript memory references.
Added a `deep` parameter to deeply watch object changes.

Closes: #3158

### DIFF
--- a/packages/shared/utils/index.ts
+++ b/packages/shared/utils/index.ts
@@ -119,3 +119,7 @@ export function objectEntries<T extends object>(obj: T) {
 export function getLifeCycleTarget(target?: any) {
   return target || getCurrentInstance()
 }
+
+export function deepClone<T>(obj: T): T {
+  return JSON.parse(JSON.stringify(obj))
+}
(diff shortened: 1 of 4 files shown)


XavierChevalier:
Any news on this?

packages/shared/utils/index.ts
```diff
@@ -119,3 +119,7 @@ export function objectEntries<T extends object>(obj: T) {
 export function getLifeCycleTarget(target?: any) {
   return target || getCurrentInstance()
 }
+
+export function deepClone<T>(obj: T): T {
+  return JSON.parse(JSON.stringify(obj))
```
antfu, line 124: This is not a safe clone, and the result might not be as expected.

Please follow the pattern like this: 
https://github.com/vueuse/vueuse/blob/5f584701a2247509d009d7e39264c7dc14fcfc97/packages/core/useRefHistory/index.ts#L32-L39

antfu (changes requested):

antfu:
Related to #3705 - I think it's better to design the API and implement on both `refDebounced` and `refThrottled` to keep the API consistent

OleksandrDzindziura:
Hello! Any updates? @XavierChevalier

PR #4019, opened 2024-06-05
fix(useWebNotification): prevent notifications when checking for support

### LINKED ISSUE #4017: useWebNotification emitting notification when mounted
### Describe the bug

When the user already has notifications enabled, it emits two notifications (without body) when the component is mounted.

_Including that the error apparently occurs in the documentation itself_

### Reproduction

https://vueuse.org/core/useWebNotification

### System Info

```Shell
System:
    OS: Windows 11 10.0.22621
    CPU: (8) x64 Intel(R) Core(TM) i5-10300H CPU @ 2.50GHz
    Memory: 1.94 GB / 15.84 GB
  Binaries:
    Node: 20.10.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.2.3 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (123.0.2420.97)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    @vueuse/nuxt: ^10.10.0 => 10.10.0
    vue: ^3.4.21 => 3.4.27
```

### Used Package Manager

npm

### Validations

9romise:

<details>
<br>

>

</details>

---

### Description
fixes #4017


Alfred-Skyblue:
In this PR, we are still triggering a notification, which I believe is unreasonable for the users. On [stackoverflow](https://stackoverflow.com/questions/29774836/failed-to-construct-notification-illegal-constructor/29895431), I found that it throws an error when used, and the error message is as follows:

```
Failed to construct 'Notification': Illegal constructor. Use ServiceWorkerRegistration.showNotification() instead. TypeError: Failed to construct 'Notification': Illegal constructor. Use ServiceWorkerRegistration.showNotification() instead.
```

I found that when I use `new Notification()` without passing any parameters, it does not issue a notification, and the error message is as follows:

```
Failed to construct 'Notification': 1 argument required, but only 0 present.
```

Can we determine support based on the error message? If the error message contains:

```
Failed to construct 'Notification': Illegal constructor
```

Can we then assume that `Notification` is not supported?

Of course, this is just my guess, as I do not have the corresponding device to test, so I am not sure about the priority of these two errors.

Alfred-Skyblue:
Additionally, as of now, it still triggers two notifications because `isSupported` is a computed property. We trigger the getter at:

https://github.com/vueuse/vueuse/blob/4d868f5e97d661dfab077108e1d650d011005584/packages/core/useWebNotification/index.ts#L126

and

https://github.com/vueuse/vueuse/blob/4d868f5e97d661dfab077108e1d650d011005584/packages/core/useWebNotification/index.ts#L131

This causes `new Notification` to be executed every time the getter is triggered. However, I believe it shouldn't be a computed property because whether the current environment supports `Notification` will never change. If `isSupported.value` is `false` when the page first loads, then it will always be `false` in subsequent checks. Therefore, should we cache the result upon the first check and return that in subsequent checks?

9romise:
Can you tell me what environment you still trigger two notifications?
I try to run the following code in the devtool after granting permission from this [demo](https://vueuse.org/core/useWebNotification/).
``` javascript
// Notifications will show
const notification = new Notification('')

// Notifications will not show
const notification = new Notification('')
notification.onshow = () => {
  notification.close()
}
```
I also verified the changes in the development environment. Btw, I'm using Chrome under Windows. And vitepress may have a cache and not update to the latest code immediately(I'm having this issue).

9romise:
> Additionally, as of now, it still triggers two notifications because `isSupported` is a computed property. We trigger the getter at:此外，截至目前，它仍然会触发两个通知，因为 `isSupported` 这是一个计算属性。我们在以下位置触发吸气器：
> 
> https://github.com/vueuse/vueuse/blob/4d868f5e97d661dfab077108e1d650d011005584/packages/core/useWebNotification/index.ts#L126
> 
> and 和
> 
> https://github.com/vueuse/vueuse/blob/4d868f5e97d661dfab077108e1d650d011005584/packages/core/useWebNotification/index.ts#L131
> 
> This causes `new Notification` to be executed every time the getter is triggered. However, I believe it shouldn't be a computed property because whether the current environment supports `Notification` will never change. If `isSupported.value` is `false` when the page first loads, then it will always be `false` in subsequent checks. Therefore, should we cache the result upon the first check and return that in subsequent checks?这会导致 `new Notification` 每次触发 getter 时执行。但是，我认为它不应该是一个计算属性，因为当前环境是否支持 `Notification` 永远不会改变。如果 `isSupported.value` 是 `false` 页面首次加载时，则它将始终 `false` 在后续检查中。因此，我们是否应该在第一次检查时缓存结果并在后续检查中返回该结果？

The fact that `computed` in `useSupported` already caches the results.
And `useSupported` has been used in other methods that need to check `isSupported`, I think this kind of change should be done together.

Alfred-Skyblue:
Sorry, I think I didn't describe it clearly just now. When I initialize, there are two notifications in my notification list. Here is my screen recording:

https://github.com/vueuse/vueuse/assets/55641773/af15fe7b-0703-4f56-a4c5-a3b964c192ef

The reason is that `isMounted` in `useSupported` triggered the setter:

https://github.com/vueuse/vueuse/blob/4d868f5e97d661dfab077108e1d650d011005584/packages/core/useSupported/index.ts#L4-L13

9romise:
Thanks for the reminder, but in this PR, I try to close the notification when it shows. And It works for me in Chrome under Windows. So this will not show the notification when `computed` was triggered. If it shows in your env, I need more info. 
``` javascript
const notification = new Notification('')
notification.onshow = () => {
  notification.close()
}
```

https://github.com/vueuse/vueuse/assets/38204901/59ce07ab-a59e-4662-8996-056349194b6f

Besides, I wonder if it's updated to the latest code, because I had this issue while debugging because of vitepress's caching.

Also, I'll try to use the error type to do this soon, and if it works, I'll submit a new commit in this PR.

Alfred-Skyblue:
I am on macOS and encountered this issue in the Arc browser. It seems that I did not reproduce it in Chrome. I am sure that I have updated to the latest code, as shown in the video:

https://github.com/vueuse/vueuse/assets/55641773/20fcbc25-3a00-449e-b4ec-c84cfefd235c

9romise:
I tested `new Notification()` without passing any parameters in Android Chrome. Unfortunately, this error has a higher priority.
```
Failed to construct 'Notification': 1 argument required, but only 0 present.
```

erkstruwe:
The two issues linked in `isSupported`'s source actually suggest a condition to prevent empty notifications from popping up (https://issues.chromium.org/issues/40415865#comment4):

```javascript
function isNewNotificationSupported() {
    if (!window.Notification || !Notification.requestPermission)
        return false;
    if (Notification.permission == 'granted')
        throw new Error('You must only call this *before* calling Notification.requestPermission(), otherwise this feature detect would bug the user with an actual notification!');
    try {
        new Notification('');
    } catch (e) {
        if ([e.name](http://e.name/) == 'TypeError')
            return false;
    }
    return true;
}
```

However, that code throws an error which we can't do in `isSupported`. @9romise Would it make sense to early-`return true` if the permission has been granted?

9romise:
> The two issues linked in `isSupported`'s source actually suggest a condition to prevent empty notifications from popping up (https://issues.chromium.org/issues/40415865#comment4):
> 
> ```js
> function isNewNotificationSupported() {
>     if (!window.Notification || !Notification.requestPermission)
>         return false;
>     if (Notification.permission == 'granted')
>         throw new Error('You must only call this *before* calling Notification.requestPermission(), otherwise this feature detect would bug the user with an actual notification!');
>     try {
>         new Notification('');
>     } catch (e) {
>         if ([e.name](http://e.name/) == 'TypeError')
>             return false;
>     }
>     return true;
> }
> ```
> 
> However, that code throws an error which we can't do in `isSupported`. @9romise Would it make sense to early-`return true` if the permission has been granted?

Sorry for late, I think this should be the best solution right now. Let's have a try.

erkstruwe:
Until this is merged, this is a workaround:

```typescript
    if (window.Notification) {
        window.Notification = class Notification extends window.Notification {
            constructor(title: string, options?: NotificationOptions) {
                super(title, options)

                if (!title) {
                    this.onshow = () => {
                        this.close()
                    }
                }
            }
        }
    }
```

9romise:
We should consider pushing this PR, as #4017 has blocked some users from upgrading to the latest version.

cc @Alfred-Skyblue @antfu

antfu (approved):

PR #4039, opened 2024-06-13
fix(until): cleanup at next tick to avoid memory leak

### LINKED ISSUE #4034: Watchers generated by `until` are not "one time" as said in the documentation
### Overview

When using [`until`](https://vueuse.org/shared/until/#until), the watchers defined are never stoped, meaning the expression will be evaluated again latter. This contradicts with the documentation, saying "Promised one-time watch for changes". Indeed, we have a watcher which we can use in line for a one-time use, but it can trigger again latter, while serving no use.

### With an example 
```vue
<script setup>
import { ref, onMounted } from 'vue';
import { until } from '@vueuse/core';

// we create a simple ref
const obj = ref({ hello: 'World' });

onMounted(async () => {
  // we expect the hello prop to be here when starting the app
  await until(() => obj.value.hello).toBeTruthy();
});
</script>

<template>
  <!--
    Now, we reset the obj, because we can do it latter in the app.
    No problem here, we have no reactive state with a direct dependency on the hello prop.
  -->
  <button @click="obj = null">Click me for the bug</button>
  <!--
    But in fact, it will trigger again the watcher defined for the `until` call, even though it serves no purposes.
    So you will have an error in the console, or given it's in vue's reactivy system, it will freeze the whole application.
    (onErrorCaptured may help here, I've yet to try it)
  -->
</template>
```

### Proposed fix
Either use the `once` `WatchOptions` from `vue 3.4` (but I'm not sure vueuse has access to this API), or simply use `watchOnce`. 

I've copy-pasted the code for `until`, but with replacing `watch` by `watchOnce`, and it solved my issue.

> I know my local version of vueuse is not the latest, but on my repro link it is. Also, FYI, the stackblitz link for doing the repro has out-of-date versions by default (`vueuse v7`, `vite v2`...)

### Reproduction

https://stackblitz.com/edit/vueuse-until-watchers?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: macOS 14.0
    CPU: (8) arm64 Apple M1
    Memory: 43.42 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.20.0 - ~/.local/share/volta/tools/image/node/18.20.0/bin/node
    Yarn: 1.22.19 - ~/.local/share/volta/tools/image/yarn/1.22.19/bin/yarn
    npm: 10.5.0 - ~/.local/share/volta/tools/image/node/18.20.0/bin/npm
    bun: 1.1.8 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 125.0.6422.142
    Safari: 17.0
  npmPackages:
    @vueuse/components: ^10.7.1 => 10.7.1
    @vueuse/core: ^10.7.1 => 10.7.1
    @vueuse/math: ^10.7.1 => 10.7.1
    @vueuse/router: ^10.7.1 => 10.7.1
    vue: ^3.4.9 => 3.4.18
```

### Used Package Manager

yarn

### Validations

Gianthard-cyh:

<details>
<br>

>

</details>

---

### Description

call stop in `until` in nextTick() instead of immediatly to fix #4034
close #4034. close  #3867

### DIFF
--- a/packages/shared/until/index.ts
+++ b/packages/shared/until/index.ts
@@ -1,5 +1,5 @@
 import type { WatchOptions, WatchSource } from 'vue-demi'
-import { isRef, watch } from 'vue-demi'
+import { isRef, nextTick, watch } from 'vue-demi'
 import { toValue } from '../toValue'
 import type { ElementOf, MaybeRefOrGetter, ShallowUnwrapRef } from '../utils'
 import { promiseTimeout } from '../utils'
@@ -76,7 +76,7 @@ function createUntil<T>(r: any, isNot = false) {
         r,
         (v) => {
           if (condition(v) !== isNot) {
-            stop?.()
+            nextTick(() => stop?.())
             resolve(v)
           }
         },
@@ -111,7 +111,7 @@ function createUntil<T>(r: any, isNot = false) {
         [r, value],
         ([v1, v2]) => {
           if (isNot !== (v1 === v2)) {
-            stop?.()
+            nextTick(() => stop?.())
             resolve(v1)
           }
         },
--- a/packages/shared/until/index.ts
+++ b/packages/shared/until/index.ts
@@ -76,7 +76,10 @@ function createUntil<T>(r: any, isNot = false) {
         r,
         (v) => {
           if (condition(v) !== isNot) {
-            nextTick(() => stop?.())
+            if (stop)
+              stop()
+            else
+              nextTick(() => stop?.())
             resolve(v)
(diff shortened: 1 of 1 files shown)


Gianthard-cyh:
This may duplicate with #3867. But for diffrent purposes. I wonder whether calling `nextTick()` or `setTimeout()` (in #3867) is a better solution.

antfu (dismissed):

PR #4070, opened 2024-06-27
feat(useStorage): make the second parameter of useStorage optional

Aaron-zon:

### Description

When only reading from Storage, there is no need to pass the second parameter. Before the modification, it could run normally with just one parameter, but since the second parameter is required, a warning would appear.

image

now make the second parameter of useStorage optional.

image


antfu:
It's required on purpose to infer the type of the value for a proper parser. Thanks for the PR

PR #4073, opened 2024-07-01
chore(core): update useUserMedia demo file camera selection reactivity

phojie:
### Description

🐞 The video stream does not react when changing the currentCamera selection.
✅ Added a watcher for the currentCamera selection to toggle the video streaming enable value.

### DIFF
--- a/packages/core/useUserMedia/demo.vue
+++ b/packages/core/useUserMedia/demo.vue
@@ -20,6 +20,13 @@ watchEffect(() => {
   if (video.value)
     video.value.srcObject = stream.value!
 })
+
+watch(() => currentCamera.value, () => {
+  enabled.value = false
+  setTimeout(() => {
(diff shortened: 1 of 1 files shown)


packages/core/useUserMedia/demo.vue
```diff
@@ -20,6 +20,13 @@ watchEffect(() => {
   if (video.value)
     video.value.srcObject = stream.value!
 })
+
+watch(() => currentCamera.value, () => {
+  enabled.value = false
+  setTimeout(() => {
+    enabled.value = true
+  }, 100)
+})
```
OrbisK, line 29: maybe you could briefly explain why this is necessary. Unfortunately, I don't have a second camera at the moment. 

And shouldn't this be controlled by the autoSwitch of useUserMedia? :)

PR #4086, opened 2024-07-09
fix(utils): objectEntries

heygsc:

<details>
<br>

>

</details>

---

### Description

I guess the original intention of this function, is to solve the problem about the key of the entries method is a string.

But in some cases, such as passing in number/string, Object.entries can run normally, while our objectEntries may have errors.Here is an example.

https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABHARgKwKbQKJigJxgwGcB5AGwBMAeAFUQwA8oMxLjl0soA+AClRoAXIloBKRAG8AUIkT4MUEPiSku0AHSsCRYgPQSAhhwCC+fIYCe1ANoBrDJbjBRAGlH3Hz0QF0fPaQBfaVBIWAROTBw8QhIAOQwAdzp+QRFxKVl5RWVVdSgtGN19NERjSO4jU3MrWwcnF1p3Wk8G338g6WkIBGI4cgwNcjgAcz41KILtWL0CEAwxMW7e-sHhscFuXB0SCko+OYWlnrA+gaHRkq2i+KSD-HnFoA

### DIFF
--- a/packages/shared/utils/index.ts
+++ b/packages/shared/utils/index.ts
@@ -112,8 +112,8 @@ export function objectOmit<O extends object, T extends keyof O>(obj: O, keys: T[
   })) as Omit<O, T>
 }
 
-export function objectEntries<T extends object>(obj: T) {
-  return Object.entries(obj) as Array<[keyof T, T[keyof T]]>
+export function objectEntries<T>(obj: T) {
+  return Object.entries(obj as object) as Array<[keyof T, T[keyof T]]>
 }
 
 export function getLifeCycleTarget(target?: any) {


antfu:
Can you explain why do you want to pass primitive values into it? I think the guard it why we have TypeScript - to catch up the mistake if you pass in wrong value

heygsc:
> Can you explain why do you want to pass primitive values into it? I think the guard it why we have TypeScript - to catch up the mistake if you pass in wrong value

In **Object.entries()**, if primitive is passed in, it will be converted to object.This PR is for this conversion.

A **low probability** event is when someone is accustomed to using **Object.entries()** to pass primitives and return an empty array to do something. They thought that **objectEntries** could do this, but after trying it out, they may find that it doesn't work.

Would it be more perfect to make **objectEntries** more consistent with **Object.entries()** when passing primitive?

antfu:
But what's the case where you want to pass a primitive? I would consider 95% when you do so it's a mistake - I would rather use `as` on the userland when I really want to pass a primitive but guard the rest 95% to be safe.

heygsc:
> But what's the case where you want to pass a primitive? I would consider 95% when you do so it's a mistake - I would rather use on the userland when I really want to pass a primitive but guard the rest 95% to be safe.`as`

Haha, yes, maybe in the actual use process, not passing primary is quite good. 
At first, this PR was more focused on maintaining consistency with Object.entries().

PR #4099, opened 2024-07-15
refactor(types): update `MaybeRef` types to match Vue3.5 and use local exports

jh-leong:
…ncy with Vue

<details>
<br>

>

</details>

---

### Description

I found that vueuse maintains its own definitions of `MaybeRef` and `MaybeRefOrGetter`, and there are mixed uses of locally defined types and types imported from `vue-demi`.

Since `vuejs/core` now has its own `MaybeRef` and `MaybeRefOrGetter` definitions, unifying the import from `vue-demi` will ensure better consistency.


antfu:
Could you explain more why the current approach would cause trouble?

The reason we can't import those types from `vue-demi` is that `vue-demi` does not shim those types for Vue 2

jh-leong:
Thanks for your comment.

Earlier, I tried to modify the `MaybeRef` type in [vuejs/core#11360](https://github.com/vuejs/core/pull/11360), which caused a compilation failure in vueuse within ecosystem-ci ([details of the failed jobs](https://github.com/vuejs/ecosystem-ci/actions/runs/9934713185/job/27439572949#step:7:906)).

Upon investigation, I found that the issue was due to the use of `MaybeRefOrGetter` exported from `vue-demi` in [useEventSource](https://github.com/vueuse/vueuse/blob/main/packages/core/useEventSource/index.ts#L107). In the subsequent call to [`toRef`](https://github.com/vueuse/vueuse/blob/efd507956638328bb74b584712ebb78ed3590112/packages/shared/toRef/index.ts#L17), which uses a locally defined type, the inconsistency in type definitions caused a compatibility issue, leading to the build failure.

> The reason we can't import those types from vue-demi is that vue-demi does not shim those types for Vue 2

As you mentioned, perhaps using the locally defined types for `MaybeRef` and `MaybeRefOrGetter` within vueuse might be a better solution?

packages/shared/utils/types.ts
```diff
@@ -1,4 +1,6 @@
-import type { ComputedRef, Ref, WatchOptions, WatchSource } from 'vue-demi'
+import type { ComputedRef, MaybeRef, Ref, WatchOptions, WatchSource } from 'vue-demi'
+
+export type { MaybeRef, MaybeRefOrGetter } from 'vue-demi'
```
yyx990803, line 3: I don't think this is gonna work for Vue 2 since Vue 2 doesn't expose `MaybeRef` and `MaybeRefOrGetter` types. Which is why `vueuse` is maintaining its own version of these types.

We should probably
1. Update the implementations here to match the type in 3.5
2. Change `useEventSource/index.ts` to import from `@vueuse/shared` instead of `vue-demi`

antfu (approved):

johuhype:
This now breaks the usage of unref (at least in 2.7)

For `dialogTitle: MaybeRef<string>`, when using
`const title = unref(dialogTitle);`
we are now getting 
image
instead of `title: string`

Does this mean unref has to get updated or should we use isRef (which is still working as expected for a MaybeRef type)?

PR #4100, opened 2024-07-15
feat(useElementVisibility): add `rootMargin` option

dritter:
### Description

`useElementVisibility` uses an IntersectionObserver under the hood, and this PR passes the `rootMargin` value to the IntersectionObserver. Note: I named it just `margin` (instead of `rootMargin`) here, because the fact that an IntersectionObserver is used might not known to the user. See [MDN IntersectionObserver/rootMargin](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/rootMargin).

### Additional Context

On our site, we want to lazy load content after it has become visible. With `useElementVisibility`, we can do that elegantlly. The only downside is that this happens very late (the element is about to enter the viewport). By setting a proper `rootMargin`, we want to trigger our callback sooner.


antfu:
Would be better:

- Rename it to `rootMargin` to align with the API
- Support MaybeRef

PR #4113, opened 2024-07-23
fix(useRouteHash, useRouteParams, useRouteQuery): fix effect triggering multiple times

### LINKED ISSUE #4112: useRouteQuery triggers reactivity twice
### Describe the bug

When updating a ref created with `useRouteQuery`, `trigger` is called twice:
- once during the `set` method (https://github.com/vueuse/vueuse/blob/main/packages/router/useRouteQuery/index.ts#L68)
- again during the `watch` callback on the route (https://github.com/vueuse/vueuse/blob/main/packages/router/useRouteQuery/index.ts#L94)

This doesn't appear to trigger watchers twice if watching the ref directly. This may be because vue isn't calling watchers if an equality check passes (old value === new value).

This **is** an issue if you create a `computed` with a non-scalar value using the value from the ref, like:

```js
const foo = useRouteQuery('foo');
const bar = computed(() => ({ bar })); // gets re-created twice whenever foo is updated
```

In this case watchers will be called twice because the object created by the computed will not pass an equality check, even if all the inner values are the same.

This is also presumably an issue with `useRouteParams` and `useRouteHash` with both take a similar approach.

I will try to create a PR to fix this by applying the same check inside the watcher that's used in the setter:

```js
  watch(
    () => route.query[name],
    (v) => {
        // add the same equality check to prevent unnecessary trigger
        if (query === v)
          return

      query = v

      _trigger()
    },
    { flush: 'sync' },
  )
```

### Reproduction

https://codesandbox.io/s/async-silence-q52wwm?file=/src/views/Home.vue

### System Info

```Shell
System:
    OS: Linux 6.8 Ubuntu 24.04 LTS 24.04 LTS (Noble Numbat)
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i7-13700H
    Memory: 9.01 GB / 31.02 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 20.13.1 - ~/.nvm/versions/node/v20.13.1/bin/node
    npm: 10.5.2 - ~/.nvm/versions/node/v20.13.1/bin/npm
  Browsers:
    Chrome: 126.0.6478.126
  npmPackages:
    @vueuse/core: ^10.11.0 => 10.11.0
    @vueuse/router: ^10.11.0 => 10.11.0
    vue: ^3.4.27 => 3.4.31
```

### Used Package Manager

npm

### Validations

aethr:
Resolves #4112.

<details>
<br>

>

</details>

---

### Description

When a ref is created with `useRouteQuery`, assigning a new value to the ref calls `trigger` twice for each change. It is called once immediately by the `set` operation, and the route is updated in a `nextTick`. When the route updates in the following tick, it triggers a watcher on the route query, which calls `trigger` again.

The extra call to `trigger` means that effects on the ref are triggered multiple times for the same value in different ticks.

More details are available in #4112.

### Additional context

This behavior was probably introduced in d52524441ae08c1ef7a9d3bc86c7fa32bbb0cd75.

Previously, the setter would only update the route, and we would rely on the route watcher to update the value and trigger the effect. This commit introduced a `trigger` call into the `set` method, but didn't add a guard against the `trigger` being called again in the subsequent tick.

### DIFF
--- a/packages/router/useRouteHash/index.test.ts
+++ b/packages/router/useRouteHash/index.test.ts
@@ -1,5 +1,5 @@
-import { nextTick, reactive, ref } from 'vue-demi'
-import { describe, expect, it } from 'vitest'
+import { computed, nextTick, reactive, ref, watch } from 'vue-demi'
+import { describe, expect, it, vi } from 'vitest'
 import { useRouteHash } from '.'
 
 describe('useRouteHash', () => {
@@ -72,6 +72,28 @@ describe('useRouteHash', () => {
     route.hash = 'foo'
 
     expect(hash.value).toBe('foo')
+  })
+
+  it('should trigger effects only once', async () => {
+    const route = getRoute()
+    const router = { replace: (r: any) => Object.assign(route, r) } as any
+    const onUpdate = vi.fn()
+
+    const hash = useRouteHash('baz', { route, router })
+    const hashObj = computed(() => ({
+      hash: hash.value,
+    }))
+
+    watch(hashObj, onUpdate)
+
+    hash.value = 'foo'
+
+    await nextTick()
+    await nextTick()
+
+    expect(hash.value).toBe('foo')
+    expect(route.hash).toBe('foo')
+    expect(onUpdate).toHaveBeenCalledTimes(1)
   })
 
   it('should allow ref or getter as default value', () => {
--- a/packages/router/useRouteHash/index.ts
+++ b/packages/router/useRouteHash/index.ts
@@ -51,6 +51,9 @@ export function useRouteHash(
   watch(
     () => route.hash,
     () => {
+      if (route.hash === _hash)
+        return
+
       _hash = route.hash
       _trigger()
     },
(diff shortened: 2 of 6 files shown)


packages/router/useRouteHash/index.test.ts
```diff
@@ -74,6 +74,28 @@ describe('useRouteHash', () => {
     expect(hash.value).toBe('foo')
   })
 
+  it('should trigger effects only once', async () => {
+    const route = getRoute()
+    const router = { replace: (r: any) => Object.assign(route, r) } as any
```
aethr, line 79: This mock router implementation is different from other tests in this file, which use:

```js
const router = { replace: (r: any) => route = r } as any
```

Using basic assignment to replace the value of `route` (ie `route = r`) doesn't trigger watchers on `route`, so this mock behaves closer to the real vue-router.

I don't think it would be a bad idea to update the other tests in this file to use the same method, or even create a more realistic mock router that the tests can all use. However, I didn't feel that change would be in the scope of this PR.

Bobakanoosh:
Would love for this to be merged. I have a chart that's data is derived from a query parameter, and this bug breaks the animation on it since the data re-calculates twice.

antfu (approved):

PR #4130, opened 2024-08-03
fix(useChangeCase): use the exported key for filter

thyngster:
**changeCaseTransforms** will return a null object of funcions becasue on the non-dev build the functions names comes minified/mangles, 
```
changeCase VALUES 
(14) [ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ, ƒ]
ƒ ir(t,e)
ƒ yt(t,e)
ƒ lr(t,e)
ƒ sr(t,e)
ƒ ur(t,e)
ƒ Oe(t,e)
ƒ or(t,e)
ƒ ar(t,e)
ƒ cr(t,e)
ƒ dr(t,e)
ƒ fr(t,e)
ƒ Et(t)
ƒ Kt(t)
ƒ hr(t,e)
```
making the v.name.endsWith('Case') never maching.

Not sure this is the best aproach to fix. changed Object.values to use Object.entries in other to be able to use theyh object key instead of the function name.

Issue: https://github.com/vueuse/vueuse/issues/4139
Duplicate PR: https://github.com/vueuse/vueuse/pull/4140

### DIFF
--- a/packages/integrations/useChangeCase/index.ts
+++ b/packages/integrations/useChangeCase/index.ts
@@ -12,7 +12,11 @@ type ChangeCaseKeys = FilterKeys<typeof changeCase>
 export type ChangeCaseType = ChangeCaseKeys[keyof ChangeCaseKeys]
 
 const changeCaseTransforms: any = /* @__PURE__ */ Object.entries(changeCase)
-  .filter(([name, fn]) => typeof fn === 'function' && name.endsWith('Case')).reduce((acc, [name, fn]) => ({ ...acc, [name]: fn }), {} as any)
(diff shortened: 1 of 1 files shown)


thyngster:
It seems I didnt linted before uploading ...

zyyv:
Sorry to see this PR, I didn't see it linked to the issue, so I made a duplicate version #4140. If you are willing to change the title and link it to the issue #4139, I will close my PR. Thank you.

Our solution is consistent, I am sure it actually solves the problem and I have verified it in the `production` environment. `function.name` will be compressed and cause the judgment to fail, we should use `key` to handle it.

antfu (approved):

Issue #4136, opened 2024-08-07
new type alias for `T & PromiseLike<T>`?


haoyun:
### Clear and concise description of the problem

https://github.com/vueuse/vueuse/blob/efd507956638328bb74b584712ebb78ed3590112/packages/core/useFetch/index.ts#L75-L88

The type `T & PromiseLike<T>` appears repeatedly.

### Suggested solution

 It might be good to have

```
type ToBeNamedType<T> = T & PromiseLike<T>

// packages/core/useFetch/index.ts
type AnotherToBeNamedType<T> = ToBeNamedType<useFetchReturn<T>>

// packages/core/useAsyncState/index.ts
type YetAnotherToBeNamedType = ToBeNamedType<UseAsyncValidatorReturn>

// packages/core/useAsyncState/index.ts
type ThirdToBeNamedType = ToBeNamedType<UseAsyncStateReturnBase<Data, Params, Shallow>>
```

That reduce the repeats in the source code and provides users a better interface than something like `<(userFetchReturn<boolean> & PromiseLike<useFetchReturn<boolean>>`.

However, I do not have a proper name for that type....

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
We actually have this: https://github.com/vueuse/vueuse/blob/efd507956638328bb74b584712ebb78ed3590112/packages/shared/utils/types.ts#L68

PR welcome if you want to help refactoring

haoyun:
> We actually have this:
> 
> https://github.com/vueuse/vueuse/blob/efd507956638328bb74b584712ebb78ed3590112/packages/shared/utils/types.ts#L68
> 
> PR welcome if you want to help refactoring

They are different types. `Promise<T> | T ` is a union while `T & PromiseLike<T>` is an intersetion. In fact, I think the former `Awaitable` union type could be better called `MaybeAwaitble` or `MaybePromise`. I named the latter `Thenable` in the PR.

PR #4137, opened 2024-08-07
new type alias `Thenable<T>` for `T & PromiseLike<T>`

haoyun:

### Description

#4136

### DIFF
--- a/packages/shared/utils/types.ts
+++ b/packages/shared/utils/types.ts
@@ -67,6 +67,8 @@ export type ShallowUnwrapRef<T> = T extends Ref<infer P> ? P : T
 
 export type Awaitable<T> = Promise<T> | T
 
+export type Thenable<T> = T & PromiseLike<T>
+
 export type ArgumentsType<T> = T extends (...args: infer U) => any ? U : never
 
 /**
--- a/packages/core/useAsyncState/index.ts
+++ b/packages/core/useAsyncState/index.ts
@@ -1,3 +1,4 @@
+import type { Thenable } from '@vueuse/shared'
 import { noop, promiseTimeout, until } from '@vueuse/shared'
 import type { Ref, UnwrapRef } from 'vue-demi'
 import { ref, shallowRef } from 'vue-demi'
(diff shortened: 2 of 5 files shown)


zyyv (approved):
Would you like to modify the `title` to mark the scope of this PR? You can refer to the history commits or other PRs.

packages/shared/utils/types.ts
```diff
@@ -67,6 +67,8 @@ export type ShallowUnwrapRef<T> = T extends Ref<infer P> ? P : T
 
 export type Awaitable<T> = Promise<T> | T
 
+export type Thenable<T> = T & PromiseLike<T>
```
antfu, line 70: Couldn't we just make `Awaitable<T> = PromiseLike<T> | T` instead of creating a new type? (their names do not indicates their difference either)

haoyun: union and intersection are different types. Anything of type `T` is of type `T | PromiseLike<T>` but might not be of type `T & PromiseLike<T>`.

antfu: Ah I see. The reason it's called `Awaitable` instead of `MaybeAwaitable` because non-Promise values also awaitable.

For the union, I don't honestly think `Thenable` is a good name here. I don't have a good name in my mind - if we didn't come up with a distinguishable name in the end, I'd suggest we keep things as-is to not having this refactor.

Maybe `WithPromiseLike` or `WithPromiseAPI`?

haoyun: Fair. The motivation for the refactor is that sometimes I found it tedious to explicitly type the return, though most of time unnecessary. A shorter alias will make it eariser  for that and also reduces repeats in the source code. 

From the definition, the union of `T | Promise(Like)<T>` **is** `MaybePromiseLike`. I suggested the name `MaybeAwaitable` because that's already called `Awaitable`. Meanwhile, the intersection, `T & Promise(Like)<T>` **is** `Awaitable`, but that's taken by the union.

The function `const fn = () => 1`, which is of type `() => number`, is of course of type `() => number | Promise<number>`. I would not say `fn() = 1` is awaitable, though `JS` wraps it into a promise and `await fn()` is allowed.

I think `WithPromiseLike` is a good choice.

OrbisK: @haoyun do you need any help/info to finish this up? :)

antfu: Yeah I think `WithPromiseLike` is a good name

Issue #2895, opened 2023-03-22
useStorageAsync does not read initial value from storage


aparajita:
### Describe the bug

First of all, thank you for such a fantastic lib! It has completely revolutionized the way I code. 🙏🏽 

There is a bug in `useStorageAsync` right here: https://github.com/vueuse/vueuse/blob/418585d87ab12fd5333c7bd66b806b9aae025153/packages/core/useStorageAsync/index.ts#L98

`read` is async, but the result is not being awaited. This results in a race condition that causes the value in storage to be overridden by `initialValue`. The reproduction illustrates this.

The simple solution is to make `useStorageAsync` itself async, because the call to `read` has to be awaited. This will unfortunately be a breaking change for existing code, but I don't see any way around this.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-rpc4fb?file=src/App.vue

### System Info

```Shell
System:
    OS: macOS 12.6.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 2.23 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 19.7.0 - ~/Library/pnpm/node
    npm: 9.5.0 - ~/Library/pnpm/npm
  Browsers:
    Brave Browser: 111.1.49.120
    Chrome: 111.0.5563.64
    Firefox: 107.0.1
    Safari: 16.3
  npmPackages:
    @vueuse/core: ^9.13.0 => 9.13.0 
    vue: ^3.2.47 => 3.2.47
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
[...]

PR #4158, opened 2024-08-17
feat(useStorageAsync): add `onReady` option and Promise return

asika32764:

<details>
<br>

>

</details>

---

### Description

This PR is aim to solve the issue https://github.com/vueuse/vueuse/issues/2895

Currently, `useStorageAsync` is impossible to wait the first value loaded, sometimes when app initializing, if we must use the storage, we may get the default initial value since the async storage not prepared.

This PR add an `onLoad` event to let developer can solve this problem by manually resolve promises in this callback.


ferferga (approved):
LGTM, but honestly, for the given use case, why not use the sync `useStorage`? I'm approving this though since other functions (like `useWebSocket`) supports similar callbacks as well.

asika32764:
For example, the capacitorStorage is asynchronous.

And sometimes people may write their own storage adapter since useStorageAsync and useStorage allows custom storage interface as an argument.

packages/core/useStorageAsync/index.ts
```diff
@@ -14,6 +14,11 @@ export interface UseStorageAsyncOptions<T> extends Omit<UseStorageOptions<T>, 's
    * Custom data serialization
    */
   serializer?: SerializerAsync<T>
+
+  /**
+   * On first value loaded hook.
+   */
+  onLoad?: (value: T) => void
```
OrbisK, line 21: If `router` and others use `onReady`, we might also think about `onReady` to be consistent. In addition, `onReady` implies to be called once.

packages/core/useStorageAsync/index.ts
```diff
@@ -94,7 +100,11 @@ export function useStorageAsync<T extends(string | number | boolean | object | n
     }
   }
 
-  read()
+  read().then(() => {
+    if (onLoad) {
+      onLoad(data.value)
+    }
```
OrbisK, line 106:
```suggestion
      onLoad?.(data.value)
```
I think this is a bit more consistent for example when someone accidentally passes a boolean. So `true` and `false` will both result in an error.

OrbisK (changes requested):

packages/core/useStorageAsync/index.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    })
+  }
+}
+
+describe('useStorageAsync', () => {
+  beforeEach(() => {
+    localStorage.clear()
+  })
+
+  it('onLoad', () => {
```
OrbisK, line 41:
```suggestion
  it('onReady', () => {
```

OrbisK (dismissed):

OrbisK (approved):

antfu:
I feel that exposing a `then` function to the ref might be more intuitive where users could also await for it.

It can be both:

```ts
const accessToken = await useStorageAsync('access.token', '', SomeAsyncStorage)
```

or 


```ts
const accessToken = useStorageAsync('access.token', '', SomeAsyncStorage)

accessToken.then(() => {
  // ...
})
```

asika32764:
Maybe

```ts
accessToken.ready.then()
```

Some libraries use this way.

asika32764:
So is there any update? Do we need to change to the new pattern?

asika32764:
@antfu @OrbisK Sorry to bother but I think this function is required for peoples since there are some issues discussion about this feature.

If wee need more changes, please leave some messages or a final decision so that I can modify the code.

Currently there has no any workaround to solve the async initialize issue without this PR.

ilyaliao:
> I feel that exposing a `then` function to the ref might be more intuitive where users could also await for it.
> 
> It can be both:
> 
> ```ts
> const accessToken = await useStorageAsync('access.token', '', SomeAsyncStorage)
> ```
> 
> or
> 
> ```ts
> const accessToken = useStorageAsync('access.token', '', SomeAsyncStorage)
> 
> accessToken.then(() => {
>   // ...
> })
> ```

It feels like using `useFetch`

```ts
const { isFetching, error, data } = await useFetch(url)
```

@asika32764 Sorry for the late reply. Would you be open to exploring how we could support something like this?

asika32764:
I implemented the Promise to the return value. But I still keep `onReady` options if someone needs it.

```ts
const accessToken = useStorageAsync('access.token', '', SomeAsyncStorage, {
  onReady(value) {
    // ...
  }
})

accessToken.value // Origin way, may be empty if storage not prepared

const resolvedToken = await accessToken

resolvedToken.value // Resolved, still needs a .value to get value
```

PR #4175, opened 2024-08-25
fix(useArrayFilter): match type to Array.prototype.filter

Jesse205:

<details>
<br>

>

</details>

---

### Description

Make the function signature of `useArrayFilter` match that of `Array.prototype.filter`.
```ts
/**
     * Returns the elements of an array that meet the condition specified in a callback function.
     * @param predicate A function that accepts up to three arguments. The filter method calls the predicate function one time for each element in the array.
     * @param thisArg An object to which the this keyword can refer in the predicate function. If thisArg is omitted, undefined is used as the this value.
     */
    filter<S extends T>(predicate: (value: T, index: number, array: T[]) => value is S, thisArg?: any): S[];
    /**
     * Returns the elements of an array that meet the condition specified in a callback function.
     * @param predicate A function that accepts up to three arguments. The filter method calls the predicate function one time for each element in the array.
     * @param thisArg An object to which the this keyword can refer in the predicate function. If thisArg is omitted, undefined is used as the this value.
     */
    filter(predicate: (value: T, index: number, array: T[]) => unknown, thisArg?: any): T[];
```

`Array.prototype.filter` 的函数签名具有以下特点：

The function signature of `Array.prototype.filter` has the following features:
 
1. When `fn` returns a type other than `boolean`, it can also work, just like using `if ()`.
2. The return type can filter the case of `undefined`.

```ts
// feature1
const list = [0, undefined, 2, undefined, 4, undefined]
const result = list.filter(i => i) // [2, 4]

// feature2
const list: (number|undefined)[] = [0, undefined, 2, undefined, 4, undefined]
const result: number[] = list.filter(i => i !== undefined)
```

### DIFF
--- a/packages/shared/useArrayFilter/index.test.ts
+++ b/packages/shared/useArrayFilter/index.test.ts
@@ -27,4 +27,9 @@ describe('useArrayFilter', () => {
     list.value.shift()
     expect(result.value).toStrictEqual([2, 4, 6, 8])
   })
+  it('should allow values other than boolean in fn.', () => {
+    const list = ref([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+    const result = useArrayFilter(list, i => i % 2)
+    expect(result.value).toStrictEqual([1, 3, 5, 7, 9])
+  })
 })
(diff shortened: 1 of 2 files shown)


Jesse205:
I can't use vitest, I tried to write type based tests (https://vitest.dev/guide/testing-types) but it doesn't work. So there is no type test within the final pr

antfu (approved):

PR #3848, opened 2024-03-06
restore watch urlRef

a982246809:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

  #3676应该已经被修复 , 所有不用去掉watch 
  #3676 should have been fixed , so there is no need to remove watch

PR #4204, opened 2024-09-05
feat(useEventSource): new `autoConnect` option to align with `useWebSocket`

ferferga:

---

### Description

As I was pointed [here](https://github.com/vueuse/vueuse/pull/3870#issuecomment-2324632526), there's a mismatch between the behaviour of the ``immediate`` setting in both ``useWebSocket`` and ``useEventSource``: in all the others composables, the ``immediate`` flag just specifies that the composable should "act" as soon as it's invoked, but in ``useWebSocket`` and ``useEventSource`` it also opts-out all the reactivity from the refs.

In my opinion, opting out from the reactivity doesn't make sense, since that's essentially the goal of VueUse (provide reactive wrappers around common browser APIs).

I believe the intended behaviour is to work like all the other composables ([1](https://github.com/vueuse/vueuse/pull/3676#issuecomment-1953854135)) ([2](https://github.com/vueuse/vueuse/pull/3848#issuecomment-2133363526)) (that PR was correct according to docs, but incorrect for my "concept" of immediate).

Feel free to close this PR in case the current behaviour is correct.

### Additional context

Not sure either if this should be considered a breaking change.


antfu:
https://github.com/vueuse/vueuse/pull/4417 improved the interoperation of `immediate` option, so I pivoted this PR to have `useEventSource` align with that design. Thanks!

antfu (approved):

PR #4223, opened 2024-09-14
feat(useInterval): support set the increase step for every interval

shuperry:
useInterval function can increase faster and decrease now by setting increase step.


packages/shared/useInterval/index.ts
```diff
@@ -42,11 +49,12 @@ export function useInterval(interval: MaybeRefOrGetter<number> = 1000, options:
   const {
     controls: exposeControls = false,
     immediate = true,
+    step,
```
antfu, line 52:
```suggestion
    step = 1,
```

packages/shared/useInterval/index.ts
```diff
@@ -42,11 +49,12 @@ export function useInterval(interval: MaybeRefOrGetter<number> = 1000, options:
...
     callback,
   } = options
 
   const counter = ref(0)
-  const update = () => counter.value += 1
+  const update = () => counter.value += (step || 1)
```
antfu, line 57:
```suggestion
  const update = () => counter.value += step
```

shuperry:
@antfu Thanks for reviewing this PR, your suggestion is done.

antfu:
I wonder when this would be useful? The `count` should means how many times the interval has been called. If you want count double, won't `const myCount = computed(() => count.value * 2)` serve your need already?

shuperry-antler:
> I wonder when this would be useful? The `count` should means how many times the interval has been called. If you want count double, won't `const myCount = computed(() => count.value * 2)` serve your need already?

Thanks for the reply, I joined a step param because I need to countdown, not always increase though.

antfu:
Oh you are mutating the counter? It shouldn't be designed usage tho

shuperry-antler:
It Seems that computed can also serve the need, this pr can close now.

Issue #4231, opened 2024-09-17
fix(useTimeoutPoll): fix incorrect default value of immediate property


MichaHuhn:
### Describe the bug

The default value of the `immediate` property is set to `false`:
https://github.com/vueuse/vueuse/blob/9943a7adc3220652c4034ed71ad5a9910f4640c0/packages/core/useTimeoutPoll/index.ts#L6

However, the default value in the docs is `true`:
https://github.com/vueuse/vueuse/blob/9943a7adc3220652c4034ed71ad5a9910f4640c0/packages/shared/useTimeoutFn/index.ts#L11

### Reproduction

not required

### System Info

```Shell
not required
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

phk422:
Hello @antfu, regarding this issue, I believe the first option is to let it execute immediately by setting immediate to true in timeoutPollOptions. However, this could affect the behavior of the feature. Another approach is to modify the type definition. I’m curious, if it were you, how would you modify it?

Borange:
It seems that the behaviour got change due to this. Previously it would wait until call and when you set it with immediate it would call directly. Now the behaviour is that is call as soon as you define it. Just wanted to mention as I got bug on this from my end, but the fix was easy to change.

version 12.3.0 and above
`const {resume, pause} = useTimeoutPoll(fetchData, 1000) // triggers function directly`

version 12.2.0 and less
`const {resume, pause} = useTimeoutPoll(fetchData, 1000) // waited until resume was called`

[...]

PR #4232, opened 2024-09-18
fix(useTimeoutPoll): `immediate` default value should be `true`

phk422:
reference #4231


OrbisK:
in the most composalbes do not handle default values in the function parameters. They use their own deconstruction. You should change this to avoid confusion?

```ts
  const {
    immediate = true,
  } = options
```

antfu (approved):

PR #4260, opened 2024-09-27
Add the defaultWindow variable in order to be able to send fetch requests in an SSR mode

proc07:
To be able to run a set of code on both the client and server side

### DIFF
--- a/packages/core/_configurable.ts
+++ b/packages/core/_configurable.ts
@@ -35,7 +35,7 @@ export interface ConfigurableLocation {
   location?: Location
 }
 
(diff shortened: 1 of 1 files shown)


proc07:
@antfu  in SSR mode, the component called useFetch, but the fetch function could not be called on the server side, resulting in the request failure. Is it possible to increase the line of code.
截屏2024-09-27 21 00 45

antfu:
Thanks, but `window` is designed to be undefined on the server side. Instead in `useFetch` we could try to detect from `globalThis.fetch`. I am closing this PR - feel free to bring up another approach in another PR.

PR #4278, opened 2024-10-15
feat(useMountedQueue): add utility to manage function queue on mount

bobby169:
### Description

This PR introduces a new composable function [`useMountedQueue`] that allows you to queue functions to be executed once the component is mounted. This is particularly useful for deferring operations that should only occur when the component is fully initialized.

### Additional context

The [`useMountedQueue`]function provides two main methods:
- [`addQueue(fn: () => void, priority = 0)`]: Adds a function to the queue with an optional priority. If the component is already mounted, the function is executed immediately.
- [`removeQueue(fn: () => void)`]:Removes a function from the queue if it exists.

The queue is processed in order of priority, with higher priority functions being executed first. This ensures that critical operations can be prioritized over less important ones.

### how to use?
```ts
const { addQueue } = useMountedQueue()

function onConnected(socket: Socket ) {
    socket?.on('gameReady', (data) => {
      addQueue(() => {
        tips.value = 'Ready Countdown'
        // the countdown component must onMounted
        readyCountDownRef.value?.startCountDown()    
      })
  }
```

### DIFF
--- a/packages/core/mountedQueue/index.md
+++ b/packages/core/mountedQueue/index.md
@@ -1,0 +1,64 @@
+---
+category: Utilities
+---
+
+# useMountedQueue
+
+Manage a queue of functions to be executed when a component is mounted and unmounted. Functions can be executed in order of priority.
+
+## Usage
+
+```ts
+import { useMountedQueue } from '@vueuse/core'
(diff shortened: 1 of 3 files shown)


antfu:

PR #4292, opened 2024-10-21
docs(useDraggable): example for `preventDefault: true`

### LINKED ISSUE #3755: bug with useDraggable
### Describe the bug

when i use useDraggable with a img element, or a div inner with a img tag, the element is not draggable; you need to set the img tag attribute 'draggable' to false to resolve this bug;

### Reproduction

https://stackblitz.com/edit/vitejs-vite-wdwgnn/?file=src%2FApp.vue

### System Info

```Shell
chrome
```

### Used Package Manager

npm

### Validations

mnenie:
resolves #3755

<details>
<br>

>

</details>

---

### Description

So, in this issue, we have a problem related to the handling of img and without  `preventDefault: true`  in `options`  we should not have a moving an img. When we set the  `preventDefault: true`  in `options` its work great, but without this, we get a bug that is reproduced in  #3755

This pr resolves this bug

### DIFF
--- a/packages/core/useDraggable/index.md
+++ b/packages/core/useDraggable/index.md
@@ -28,6 +28,17 @@ const { x, y, style } = useDraggable(el, {
 </template>
 ```
 
+Set `preventDefault: true` to override the default drag-and-drop behavior of certain elements in the browser.
+
+```ts
+const { x, y, style } = useDraggable(el, {
+  preventDefault: true,
+  // with `preventDefault: true`
+  // you can disable the native behavior (e.g., for img)
+  // and control the drag-and-drop, preventing the browser interference.
+})
+```
+
 ## Component Usage
 
 ```vue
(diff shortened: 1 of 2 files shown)


antfu:
I am not so sure about the special handling for IMG - it might happend to the other elements as well. I think the best cases would be to have users enable the preventDefaults if they expect that behavior. I would see this is more like a documentation issue - we could improve the docs and guide users to use that options.

antfu (changes requested):
As above

mnenie:
> I am not so sure about the special handling for IMG - it might happend to the other elements as well. I think the best cases would be to have users enable the preventDefaults if they expect that behavior. I would see this is more like a documentation issue - we could improve the docs and guide users to use that options.

yea maybe you're right, I'll do it 😉

mnenie:
> As above

@antfu check please, I've added edits to the documentation. I hope this is what we need 😉💯

PR #4308, opened 2024-10-28
Fixed missing types for fromEvent

### LINKED ISSUE #4307: fromEvent is not a wrapper around rxjs fromEvent
### Describe the bug

The [docs say](https://vueuse.org/rxjs/from/) from and fromEvent are "wrappers around RxJS's [from()](https://rxjs.dev/api/index/function/from) and [fromEvent()](https://rxjs.dev/api/index/function/fromEvent) to allow them to accept refs".

This sounds to me like the work exactly like rxjs from / fromEvent BUT also accept Vue refs. And if I am not mistaken this seems to be the case for from, but not for fromEvent:

```
import { from } from '@vueuse/rxjs';
import { io } from 'socket.io-client';
          
          const socket = io('//localhost:3100', {
            autoConnect: false,
          });
          
          fromEvent(socket, 'some-event');
```

Results in error:
```
Vue: Argument of type Socket<DefaultEventsMap, DefaultEventsMap> is not assignable to parameter of type MaybeRef<HTMLElement>
```

But if I import from rxjs `import { from } from 'rxjs';` it works.

This is very unhandy because if you have multiple fromEvent in one file, one needs to rename them:
```
import { fromEventRxjs } from 'rxjs';
import { fromEventVueUse } from '@vueuse/rxjs';
```

I think the cause is the typing [here](https://github.com/vueuse/vueuse/blob/main/packages/rxjs/from/index.ts#L14), it should be: `value: NodeStyleEventEmitter | ArrayLike<NodeStyleEventEmitter> | MaybeRef<T>`

### Reproduction

-

### System Info

```Shell
latest
```

### Used Package Manager

pnpm

### Validations

MickL:
Fixes #4307

### DIFF
--- a/packages/rxjs/from/index.ts
+++ b/packages/rxjs/from/index.ts
@@ -2,6 +2,7 @@ import type { MaybeRef } from '@vueuse/shared'
 import type { ObservableInput, Subscription } from 'rxjs'
 import type { Ref, WatchOptions } from 'vue-demi'
 import { fromEvent as fromEventRx, from as fromRxjs, Observable } from 'rxjs'
+import type { NodeStyleEventEmitter } from 'rxjs/internal/observable/fromEvent'
 import { isRef, watch } from 'vue-demi'
 
 export function from<T>(value: ObservableInput<T> | Ref<T>, watchOptions?: WatchOptions): Observable<T> {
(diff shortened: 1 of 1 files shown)


packages/rxjs/from/index.ts
```diff
@@ -2,6 +2,7 @@ import type { MaybeRef } from '@vueuse/shared'
 import type { ObservableInput, Subscription } from 'rxjs'
 import type { Ref, WatchOptions } from 'vue-demi'
 import { fromEvent as fromEventRx, from as fromRxjs, Observable } from 'rxjs'
+import type { NodeStyleEventEmitter } from 'rxjs/internal/observable/fromEvent'
```
antfu, line 5: I am not sure we we should use this as it's marked `internal`

PR #4317, opened 2024-10-31
feat(useSSRWidth): add optional support for SSR in useMediaQuery and useBreakpoints

### LINKED ISSUE #4102: useBreakpoint and useMediaQuery cause "Hydration text mismatch" on Nuxt
### Describe the bug

This bug can be related to #912

It seems like `useWindowSize`, we need to use the `tryOnMounted` function also for `useMediaQuery`, to prevent the mismatch error.

### Reproduction

https://stackblitz.com/edit/nuxt-starter-4j78pj

### System Info

```Shell
Doesn't matter.
```

### Used Package Manager

npm

### Validations

Tofandel:

---

### Description

Fixes #4102 

By adding an optional `ssrSize` option to `useBreakpoints` and `useMediaQuery` we can mock their behavior until hydration happens to avoid mismatch error

Tests which did not exist for them could now be added thanks to this feature

Notable fixes: 
 - the breakpoints are now sorted in ascending order in the `current()` method, this is to avoid the case of a user passing unordered breakpoints causing `active` to be wrong
 - the `current()` method now returns the list of active breakpoint in accordance with the `strategy` option it previously did not respect it
 - the types returned from useBreakpoints are now correct, previously `active` which was a function was not typed as one


packages/core/useBreakpoints/index.ts
```diff
@@ -89,39 +114,24 @@ export function useBreakpoints<K extends string>(
...
-  smaller: (k: MaybeRefOrGetter<K>) => ComputedRef<boolean>
-  smallerOrEqual: (k: MaybeRefOrGetter<K>) => ComputedRef<boolean>
-  between: (a: MaybeRefOrGetter<K>, b: MaybeRefOrGetter<K>) => ComputedRef<boolean>
-  isGreater: (k: MaybeRefOrGetter<K>) => boolean
-  isGreaterOrEqual: (k: MaybeRefOrGetter<K>) => boolean
-  isSmaller: (k: MaybeRefOrGetter<K>) => boolean
-  isSmallerOrEqual: (k: MaybeRefOrGetter<K>) => boolean
-  isInBetween: (a: MaybeRefOrGetter<K>, b: MaybeRefOrGetter<K>) => boolean
-  current: () => ComputedRef<string[]>
-  active: ComputedRef<string>
```
Tofandel, line 126: This type was wrong

packages/core/useBreakpoints/index.ts
```diff
@@ -89,39 +99,26 @@ export function useBreakpoints<K extends string>(
...
-} & Record<K, ComputedRef<boolean>>
+export type UseBreakpointsReturn<K extends string = string> = ReturnType<typeof useBreakpoints<K>>
```
Tofandel, line 124: This now automatically infer the type

```
declare function useBreakpoints<K extends string = string>(breakpoints: Breakpoints<K>, options?: UseBreakpointsOptions): Record<K, Ref<boolean, boolean>> & {
    greaterOrEqual: (k: MaybeRefOrGetter<K>) => Ref<boolean, boolean>;
    smallerOrEqual: (k: MaybeRefOrGetter<K>) => Ref<boolean, boolean>;
    greater(k: MaybeRefOrGetter<K>): Ref<boolean, boolean>;
    smaller(k: MaybeRefOrGetter<K>): Ref<boolean, boolean>;
    between(a: MaybeRefOrGetter<K>, b: MaybeRefOrGetter<K>): Ref<boolean, boolean>;
    isGreater(k: MaybeRefOrGetter<K>): boolean;
    isGreaterOrEqual(k: MaybeRefOrGetter<K>): boolean;
    isSmaller(k: MaybeRefOrGetter<K>): boolean;
    isSmallerOrEqual(k: MaybeRefOrGetter<K>): boolean;
    isInBetween(a: MaybeRefOrGetter<K>, b: MaybeRefOrGetter<K>): boolean;
    current: () => vue.ComputedRef<K[]>;
    active(): vue.ComputedRef<"" | K>;
};
type UseBreakpointsReturn<K extends string = string> = ReturnType<typeof useBreakpoints<K>>;
```

antfu:
Oh, that's brilliant! I was about to say that it's not possible to know the size of the client on the server side, but yeah, this feature at least provides a solution to that!

packages/core/useBreakpoints/index.md
```diff
@@ -44,6 +44,19 @@ const laptop = breakpoints.between('laptop', 'desktop')
...
+#### Server Side Rendering and Nuxt
+
+If you are using `useBreakpoints` with SSR enabled, then you need to specify which screen size you would like to render on the server and before hydration to avoid an hydration mismatch
+
+```js
+import { breakpointsTailwind, useBreakpoints } from '@vueuse/core'
+
+const breakpoints = useBreakpoints(
+  breakpointsTailwind,
+  { ssrSize: 768 } // Will enable SSR mode and render like if the screen was 768px wide
```
antfu, line 56: Maybe `ssrWidth` might be a better name?

Tofandel: Agreed, changed it

packages/core/useIntersectionObserver/index.ts
```diff
@@ -70,33 +70,33 @@ export function useIntersectionObserver(
...
-        targets.forEach(el => el && observer.observe(el))
-
-        cleanup = () => {
-          observer.disconnect()
-          cleanup = noop
-        }
-      },
-      { immediate, flush: 'post' },
-    )
+        () => [targets.value, unrefElement(root as MaybeComputedElementRef), isActive.value] as const,
```
Tofandel, line 73: Seems like there is a few lint issues that need fixing in the main repo, though unsure why it only started to popup now

userquin:
> I was about to say that it's not possible to know the size of the client on the server side, but yeah, this feature at least provides a solution to that!

There is a way using HTTP Client Hints via [Critical Client Hints](https://developer.mozilla.org/en-US/docs/Web/HTTP/Client_hints#critical_client_hints), but only on chromium based browsers:
- https://github.com/userquin/http-client-hints
- https://github.com/userquin/nuxt-http-client-hints (https://github.com/userquin/nuxt-http-client-hints/pull/10)

antfu:
Yeah, Critical Client Hints would be an interesting approach for this, but I agree it's more on the framework integration scope. I think we could have this feature first and discuss if we could provide better integration for Nuxt later.

packages/core/useSSRWidth/index.ts
```diff
@@ -0,0 +1,9 @@
+let SSRWidth: number | undefined
+
+export function useSSRWidth() {
+  return SSRWidth
+}
+
+export function setSSRWidth(width: typeof SSRWidth) {
+  SSRWidth = width
```
antfu, line 8: I am not so sure about having a global variable for this. It would be fine for an app that only has a single fixed width on SSR. But when the width could change per request, this would lead to context leak and undetermined SSR result. We could use provide/inject to store this value if we want to reuse it across multiple composables

antfu (changes requested):

packages/core/useSSRWidth/index.ts
Tofandel: Indeed, I guess it would also be subject to race conditions if the app renders multiple things async at the same time

Changed it to provide inject

packages/core/useSSRWidth/index.ts
```diff
@@ -0,0 +1,20 @@
+import type { App, InjectionKey } from 'vue'
+import { injectLocal, provideLocal } from '@vueuse/shared'
+import { hasInjectionContext } from 'vue'
+
+const ssrWidthSymbol = Symbol('SSR Width') as InjectionKey<number | null>
```
antfu, line 5:
```suggestion
const ssrWidthSymbol = Symbol('vueuse-ssr-width') as InjectionKey<number | null>
```

antfu (approved):

ashokgelal:
I'm working on getting a minimal repro example (my setup is fairly complex so trying to carve it out) but just for the time being - I upgraded to 12.1.0 and getting `node_modules/@vueuse/core/index.mjs?v=d9158410' does not provide an export named 'useSSRWidth'` error. This is a NuxtJS project with ssr set to false.

ashokgelal:
Spent too much time trying to repro but what problem hasn't been solved by just deleting `node_modules`? 😅 No longer an issue

xak2000:
Any hints on how to properly use this new feature?

I tried to use it like this:
```ts
const breakpoints = useBreakpoints(breakpointsTailwind, { ssrWidth: 768 })
```

But the problem is that as soon as I set `ssrWidth` the client width is not taken into account anymore. So, if I set `{ ssrWidth: 768 }`, then even on a large screen the result (on the client, after hydration) is `{ "sm": true, "md": true, "lg": false, "xl": false, "2xl": false }`. If I remove `{ ssrWidth: 768 }` parameter, then all works fine (`{ "sm": true, "md": true, "lg": true, "xl": false, "2xl": false }`), but obviously Hydration mistmatch problem appears.

I use `@vueuse/core@npm:12.5.0`, `@vueuse/nuxt@npm:12.5.0`.

Is it supposed to work this way? I'm just trying to understand what exactly does this feature do. The docs are scarce about this.

What exactly should happen when `ssrWidth` is set? I understand that it uses this width on the server rendering, but what should happen next? Should the values be "corrected" after hydration to values based on the real screen size (with jumpy layout effect)? Or `ssrWidth` will override the client's values even after hydration, like it does in my case?

A reproduce: https://stackblitz.com/edit/nuxt-starter-bt55593r?file=app.vue

Tofandel:
> Should the values be "corrected" after hydration to values based on the real screen size (with jumpy layout effect)?

Yes on mounted the values are corrected and yes if the values don't match you get a jump

This seems to be a reactivity issue with the object props as other methods don't have this issue

Tofandel:
Yes this issue appears because the properties are behind a getter and reevaluated on each pass, but it uses a composable under the hood, and the composable is not meant to be initialized all of the time, if you keep a reference to the shortcut, this does not happen

https://stackblitz.com/edit/nuxt-starter-cnydzchs?file=app.vue

PR #4325, opened 2024-11-04
fix(useDropZone): validate all file types individually when dropping multiple files

LouisMazel:

<details>
<br>

>

</details>

---

### Description

This PR fixes a bug in the `useDropzone` composable related to file type validation. Previously, the validation would pass if any of the dropped files matched any of the allowed types, potentially allowing invalid file types to be accepted.

**Bug details:**
- In the current implementation, the `some()` method checks if any file type matches the allowed types, making the validation pass even if some files don't match the criteria
- This creates a security issue as users could drop files with unauthorized types along with valid ones

**Fix:**
- Modified the `checkDataTypes` function to validate each file type individually
- Now ensures all dropped files match the allowed types before accepting them
- Maintains backward compatibility with both function and array validation approaches
- Preserves the existing API while fixing the validation logic

### Additional context

The fix changes the validation approach from an any-match to an all-match strategy, which better aligns with expected dropzone behavior. This is particularly important for security-conscious applications where file type validation is critical.

Example of the issue:
If allowed types are `['image/jpg', 'image/png']` and someone drops both a `.jpg` and `.exe` file, the previous implementation would incorrectly validate this as acceptable because at least one file (the `.jpg`) matched the criteria.

### DIFF
--- a/packages/core/useDropZone/demo.vue
+++ b/packages/core/useDropZone/demo.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { useDropZone, useEventListener } from '@vueuse/core'
+import { useDropZone } from '@vueuse/core'
 import { ref } from 'vue'
 
 const filesData = ref<{ name: string, size: number, type: string, lastModified: number }[]>([])
@@ -31,13 +31,8 @@ function onImageDrop(files: File[] | null) {
 
 const dropZoneRef = ref<HTMLElement>()
 const imageDropZoneRef = ref<HTMLElement>()
-const pngRef = ref()
 
 const { isOverDropZone } = useDropZone(dropZoneRef, onDrop)
-
-useEventListener(pngRef, 'dragstart', (event) => {
-  event.dataTransfer?.setData('image/png', '/vue.png')
-})
 
 const { isOverDropZone: isOverImageDropZone } = useDropZone(imageDropZoneRef, { dataTypes: ['image/png'], onDrop: onImageDrop })
 </script>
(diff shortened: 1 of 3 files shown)


ilyaliao:
Hey, can you try the [demo](https://github.com/vueuse/vueuse/blob/main/packages/core/useDropZone/demo.vue) result? It doesn't seem to work.


https://github.com/user-attachments/assets/8f1ffd7b-2cb5-44ab-bf58-244fc17c6941

LouisMazel:
> Hey, can you try the [demo](https://github.com/vueuse/vueuse/blob/main/packages/core/useDropZone/demo.vue) result? It doesn't seem to work.
> 
>  -.Compressed.with.FlexClip.mp4

This bug seems to come from the demo, if you try with real images and svgs everything works fine

Check out this video: https://jmp.sh/TDmH7Gzj

ilyaliao:
Sorry, can you test it in Safari?

After testing, it works well in Chrome, but in Safari, isOverDropZone is always true when dragging because types is empty (I guess this is related to Safari's security policy). However, the drop result is correct.

Please correct me if I'm wrong.

LouisMazel:
@ilyaliao 

Yes you are right, we must invalidate if no type is found in the DataTransferList.

I just updated the code : 

For Safari, the limitation is that we can check the validity of the files only during the drop event. For other events (drag*), the event.dataTransfer is not transmitted.
In my update, on Safari, isOverDropZone will always be true during the drag (I prefer this solution because the fact is that we are really "over" the drop zone)
We can explain that with Safari's limitation, we cannot validate the files during the drag.

Concerning the demo, the element drop is not a good use case. When using PNG and SVG, the dataTransfer has several items and therefore several types (e.g. 'text/uri-list', 'text/html', 'image/png' for PNG) because we drag an HTML element.
I think we should remove the use of these HTML elements in the demo to tell users to test with real files.

ilyaliao:
I agree, I'm looking forward to this PR being merged!

antfu (approved):

PR #4339, opened 2024-11-13
feat(useInputCompositionState): add function,the state change when a text composition system such as an input method editor starts

Nicholas003:
### Description

This PR introduces a new composable function [useInputCompositionState] input method editor display state

### Additional context

``` javascript
import { useInputComposition } from './index'

const state = useInputComposition()
```


antfu:

PR #4344, opened 2024-11-15
feat(useMediaControls): add 'media' option to define media query for resources

aaronLejeune:
Adding media option to sources. This specifies the media query for the resource's intended media.


43081j (commented):
makes sense. we should add a test though for when `media` is set, and for the default when it isn't

antfu (approved):

PR #4349, opened 2024-11-20
feat!: drop Vue 2 support, optimize bundles and clean up

antfu:
This will be a major breaking change which will be shipped as v12.0

Context: https://bsky.app/profile/vueuse.org/post/3lbfdavfy2s2p


README.md
```diff
@@ -66,23 +66,19 @@ Refer to [functions list](https://vueuse.org/functions) or [documentations](http
 
 ## 📦 Install
 
-> 🎩 From v4.0, it works for Vue 2 & 3 **within a single package** by the power of [vue-demi](https://github.com/vueuse/vue-demi)!
+> From v12.0, VueUse no longer supports Vue 2. Please use v10.x for Vue 2 support.
```
ibrahimBeladi, line 69: 10 or 11?

brolnickij: > 10 or 11?

11

eslint.config.js
```diff
@@ -14,7 +14,7 @@ const restricted = [
   resolve(dir, 'packages/core/index.ts'),
   resolve(dir, 'packages/shared/index.ts'),
   {
-    name: 'vue-demi',
+    name: 'vue',
     importNames: ['onMounted', 'onUnmounted', 'unref', 'toRef'],
```
Tofandel, line 18: @antfu Any idea why those imports needed to be restricted from the code base?

I can't think of a reason why

Tofandel: From what found in the gitlog it was to force using toValue from @vueuse/shared (which doesn't exist anymore) and tryOnMounted.

This just led to all the usage to be eslint ignored as internally as we can already be duplicating the checks those functions are doing in a way that cannot be externalised, so most of the time the "safe" function is not necessary.

PR #4353, opened 2024-11-22
fix(useStorageAsync): multi-tab deadlock

### LINKED ISSUE #4200: advice "useStorageAsync" "read" function change to  debounce
### Describe the bug

When there are two or more windows, quickly modifying values leads to an endless loop

video: https://github.com/afute/useStorageAsync-bug-example-p1/blob/main/example.mp4

```
// ...

const readDebounce: PromisifyFn<typeof read> = useDebounceFn(read, 50);
useEventListener(window, 'storage', e => readDebounce(e));
```

### Reproduction

https://github.com/afute/useStorageAsync-bug-example-p1

### System Info

```Shell
System:
    OS: Windows 10 10.0.19045
Binaries:
    Node: 20.17.0
    npm: 10.8.1
Browsers:
    Edge: Chromium (127.0.2651.105)
```

### Used Package Manager

npm

### Validations

afute:
fix #4200 

<details>
<br>

>

</details>

---

### Description

When "useStorageAsync" is used, when more than one tab is opened, the value of the two tabs changes to each other, causing a deadlock

[bug video](https://github.com/afute/useStorageAsync-bug-example-p1/blob/main/example.mp4)

### Reproduction

https://github.com/afute/useStorageAsync-bug-example-p1

### DIFF
--- a/packages/core/useCountdown/index.md
+++ b/packages/core/useCountdown/index.md
@@ -21,3 +21,27 @@ const { remaining, start, stop, pause, resume } = useCountdown(countdownSeconds,
   }
 })
 ```
+
+You can use a `ref` to change the initial countdown.
+`start()` and `resume()` also accept a new countdown value for the next countdown.
+
+```js
+import { ref } from 'vue'
+import { useCountdown } from '@vueuse/core'
+
+const countdown = ref(5)
+const { start, reset } = useCountdown(countdown, {
+})
+
+// change the countdown value
+countdown.value = 10
+
+// start a new countdown with 2 seconds
+start(2)
+
+// reset the countdown to 4, but do not start it
+reset(4)
+
+// start the countdown with the current value of `countdown`
+start()
+```
--- a/packages/core/useCountdown/index.test.ts
+++ b/packages/core/useCountdown/index.test.ts
@@ -1,7 +1,7 @@
 import type { Pausable } from '@vueuse/shared'
 import type { UseCountdownOptions } from '.'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
-import { effectScope } from 'vue'
+import { effectScope, ref } from 'vue'
 import { useCountdown } from '.'
 
 describe('useCountdown', () => {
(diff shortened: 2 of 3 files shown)


antfu:
Could you help to add some tests to prevent future regressions? Thanks

packages/core/useStorageAsync/queue.ts
```diff
@@ -0,0 +1,32 @@
...
+  private readonly queue: ((() => Promise<void>) | null)[]
+  private readonly emptyCallback: () => void
+
+  constructor(callback?: () => void) {
+    this.queue = [null, null]
+    this.emptyCallback = callback ?? (() => {})
+  }
+
+  enqueue(task: () => Promise<void>) {
+    this.queue[1] = task
```
Tofandel, line 11: Yikes!

Basically we just want to pause the watcher while the async read method is running, we have composables for that which could be used instead of this very shaky TaskQueue

Tofandel, line 11: Yikes!

Basically we just want to pause the watcher while the async read method is running, we have composables for that which could be used instead of this very shaky TaskQueue

packages/core/useStorageAsync/index.ts
```diff
@@ -95,31 +96,47 @@ export function useStorageAsync<T extends(string | number | boolean | object | n
...
+    })
+  })
+
+  if (window && listenToStorageChanges) {
+    useEventListener(window, 'storage', (e) => {
+      watchStop()
+      queue.enqueue(async () => {
+        await read(e)
+      })
+    })
```
Tofandel, line 139: From what I gathered this could just be simplified like this
```suggestion
    useEventListener(window, 'storage', async (e) => {
      watchStop()
      await read(e)
      watchStop = watchWithFilter(data, cb, {
        flush,
        deep,
        eventFilter,
      })
    })
```

Tofandel (changes requested):

packages/core/useStorageAsync/index.ts
afute: Sorry to reply to you now, you are correct, I was trying to reduce the side effects of "setItem" at the time.

packages/core/README.md
```diff
@@ -1 +1 @@
-../../README.md
\ No newline at end of file
+../../README.md
```
Tofandel, line 1: That's a weird changeset, what happened here?

PR #4362, opened 2024-11-27
fix issue#4357 (useAxios) : When executing the execute method, return a new promise

tinysimple:
### Description
Problem solved: The `execute` method is called multiple times, and the previous request is cancelled, resulting in the last actual request being successful but returning a Promise with a rejection status. 

And because the current return is the same shared promise object, even if the previous request has been canceled, multiple calls have to wait until all requests have been completed to obtain the same promise information, making it impossible to distinguish the data source of each request error.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/integrations/useAxios/index.ts
+++ b/packages/integrations/useAxios/index.ts
@@ -237,25 +237,30 @@ export function useAxios<T = any, R = AxiosResponse<T>, D = any>(...args: any[])
     const currentExecuteCounter = executeCounter
     isAborted.value = false
 
-    instance(_url, { ...defaultConfig, ...typeof executeUrl === 'object' ? executeUrl : config, signal: abortController.signal })
-      .then((r: any) => {
-        if (isAborted.value)
-          return
-        response.value = r
-        const result = r.data
-        data.value = result
-        onSuccess(result)
-      })
-      .catch((e: any) => {
-        error.value = e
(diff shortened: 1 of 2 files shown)


packages/integrations/useAxios/index.ts
```diff
@@ -237,25 +237,30 @@ export function useAxios<T = any, R = AxiosResponse<T>, D = any>(...args: any[])
...
+          error.value = e
+          reject(e)
+          onError(e)
+        })
+        .finally(() => {
+          options.onFinish?.()
+          if (currentExecuteCounter === executeCounter)
+            loading(false)
+        })
+    })
```
OrbisK, line 263:
```suggestion
    instance(_url, { ...defaultConfig, ...typeof executeUrl === 'object' ? executeUrl : config, signal: abortController.signal })
      .then((r: any) => {
        if (isAborted.value)
          return
        response.value = r
        const result = r.data
        data.value = result
        onSuccess(result)
        return result
      })
      .catch((e: any) => {
        error.value = e
        onError(e)
      })
      .finally(() => {
        options.onFinish?.()
        if (currentExecuteCounter === executeCounter)
          loading(false)
      })
    return promise
```

Should that (returning within than) not do the trick?

PR #4366, opened 2024-11-27
fix(useFetch): handle empty payload

ilyaliao:
fix: #4365 

<details>
<br>

>

</details>

---

### Description

This PR fix to prevent errors when payload is empty.

### Additional context

#4329

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -421,7 +421,7 @@ export function useFetch<T>(url: MaybeRefOrGetter<string>, ...args: any[]): UseF
       const payload = toValue(config.payload)
       // Set the payload to json type only if it's not provided and a literal object or array is provided and the object is not `formData`
       // The only case we can deduce the content type and `fetch` can't
-      const proto = Object.getPrototypeOf(payload)
+      const proto = Object.getPrototypeOf(payload || {})
       if (!config.payloadType && payload && (proto === Object.prototype || Array.isArray(proto)) && !(payload instanceof FormData))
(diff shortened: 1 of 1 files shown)


packages/core/useFetch/index.ts
```diff
@@ -421,7 +421,7 @@ export function useFetch<T>(url: MaybeRefOrGetter<string>, ...args: any[]): UseF
       const payload = toValue(config.payload)
       // Set the payload to json type only if it's not provided and a literal object or array is provided and the object is not `formData`
       // The only case we can deduce the content type and `fetch` can't
-      const proto = Object.getPrototypeOf(payload)
+      const proto = Object.getPrototypeOf(payload || {})
```
WingSMC, line 424: I think the use of the `??` operator would be more appropriate here.

ilyaliao: Thank you for your feedback! I've submitted a new commit based on your suggestions. Could you please check if this aligns with what you had in mind?

WingSMC (approved):
If `config.payload` is undefined or null then that is what `toValue` returns so this seems fine.

ilyaliao:
> If `config.payload` is undefined or null then that is what `toValue` returns so this seems fine. I don't have a real say in this though, I'm not a maintainer.

I know, but I like the suggestion you proposed. When the payload is empty, calling `toValue` will ultimately result in undefined, as you said. We should skip the scope of determining whether it is JSON.

Thank you again for your review.

antfu (approved):

PR #4370, opened 2024-11-28
docs(useClipboardItems): fix example

ShangChien:
#### Description:
Fixed an error in the usage of the Blob constructor when creating a ClipboardItem. 
The original code nested `{ type: mime }` in Array, which resulted in improperly generated Blob content.

#### Change:
modify
```javascript
new ClipboardItem({
    [mime]: new Blob(['\'<b>HTML content</b>\'', { type: mime }]),
})
```
to
```javascript
new ClipboardItem({
    [mime]: new Blob([`<b>HTML content</b>`], { type: mime }),
})
```

### DIFF
--- a/packages/core/useClipboardItems/index.md
+++ b/packages/core/useClipboardItems/index.md
@@ -20,7 +20,7 @@ import { useClipboardItems } from '@vueuse/core'
 const mime = 'text/html'
 const source = ref([
   new ClipboardItem({
-    [mime]: new Blob([`<b>HTML content</b>`], { type: mime }),
+    [mime]: new Blob(['<b>HTML content</b>'], { type: mime }),
   })
 ])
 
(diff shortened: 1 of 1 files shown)


packages/core/useClipboardItems/index.md
```diff
@@ -20,7 +20,7 @@ import { useClipboardItems } from '@vueuse/core'
 const mime = 'text/html'
 const source = ref([
   new ClipboardItem({
-    [mime]: new Blob(['\'<b>HTML content</b>\'', { type: mime }]),
+    [mime]: new Blob([`<b>HTML content</b>`], { type: mime }),
```
NaokiHaba, line 23: Thank you for submitting the code!

Since we're not using string interpolation, single quotes would be sufficient.
Please review.

```suggestion
    [mime]: new Blob(['<b>HTML content</b>'], { type: mime }),
```

NaokiHaba (approved):

NaokiHaba (approved):
Thank you for your support!

Issue #4381, opened 2024-12-04
BUG | `useWebSocket` | Reconnection occurs after explicitly calling the close method


ghimisradu:
### Describe the bug

Reconnection occurs after explicitly calling the `close` method when using `autoReconnect` options and a ref as websocket url.
The issue occurs when open it's called within the `autoReconnect.delay` interval.

In the stackblitz click login, wait for the `Unknown api key` message then logout and login again within 10 seconds. In the network tab there should be 2 active websocket connections using the same token.

The second connection it's triggered from [here](https://github.com/vueuse/vueuse/blob/fa7280728fab4705332a730be7f499006bba61d0/packages/core/useWebSocket/index.ts#L243)

### Reproduction

https://stackblitz.com/edit/vitejs-vite-pqu1gc?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: Windows 11 10.0.22000
    CPU: (8) x64 Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
    Memory: 7.28 GB / 31.92 GB
  Binaries:
    Node: 20.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Program Files\nodejs\yarn.CMD
    npm: 10.8.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.2600.0), Chromium (131.0.2903.51)
    Internet Explorer: 11.0.22000.2600
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #4383, opened 2024-12-05
fix(useWebSocket): clear retryTimer when connected

ilyaliao:
fix: #4381 

<details>
<br>

>

</details>

---

### Description

This PR fixes the ws retry timer during connection establishment to prevent duplicate connections.


packages/core/useWebSocket/index.ts
```diff
@@ -306,6 +312,10 @@ export function useWebSocket<Data = any>(
   const open = () => {
     if (!isClient && !isWorker)
       return
+
+    if (retryTimeout != null)
+      resetRetry()
```
43081j, line 317: i feel like we should be calling this in `close` instead of `open`, just the same as we call `resetHeartbeat` there

regardless of if we reopen later, closing should stop the timer

43081j (commented):
looks good

it looks like `useWebSocket` has no tests 😬 would be nice to add some but you'd have to play around with stubbing out `WebSocket` etc

packages/core/useWebSocket/index.ts
ilyaliao: I want to solve the issue where the auto-reconnect timeout might trigger unnecessarily when manually opening a new connection.

When a connection drops, our autoReconnect sets up a retryTimeout for delayed reconnection. If the user manually calls open() before this timeout triggers, we need to clear this timeout first - otherwise it would trigger another connection attempt even after we've already established a new connection manually.

Maybe there's a better way to handle this, but this is my current approach. Thanks for your review :>

43081j: `open` calls `close` before doing anything else, so it will also result in calling this additional logic if you put it there

ilyaliao: Sorry, I'm stuck. Could you help me test this [reproduction](https://stackblitz.com/edit/vitejs-vite-pqu1gc?file=src%2FApp.vue)? I tried to move it, but it doesn't seem to work 🥲

43081j: if i do the following in a test, it fails before the change and passes after:

1. open the socket (with `autoReconnect: true`)
2. close the socket (by causing the `WebSocket` to emit a `close` rather than by calling `close()` ourselves)
3. call `open()` (**at this point, the reconnect timer is queued already)

it seems to work

i just changed `close` to do this:

```ts
resetHeartbeat()
resetRetry()
heartbeatPause?.()
```

you can see in `open`, it calls `close` first thing:

```ts
const open = () => {
  if (!isClient && !isWorker)
    return
  close() // CALLED HERE
  // ...
}
```

so this change is doing exactly what your original code did (for this particular test case)

it'll fix another case if we do it this way too

antfu (approved):

PR #4384, opened 2024-12-06
fix(useElementHover): returns correctly hover state when element is removed

### LINKED ISSUE #4379: BUG | `useElementHover` | Value remains `true` after target is unmounted
### Describe the bug

The `useElementHover` value will stay `true` after element is unmounted.

Because internally `useElementHover` tracks hover state with a pair of `mouseenter/mouseleave` events, final `mouseleave` won't be triggered if target element is removed while mouse pointer still hovers the element. One could argue this is correct behavior as technically mouse never left the element. But it my opinion this is unexpected. 

This is likely a breaking change for those who actually rely on this quirk. 

I'm willing to submit a PR if this is confirmed to be a bug.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-4pfnez?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: macOS 15.1.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 110.88 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.14.0 - ~/.nvm/versions/node/v20.14.0/bin/node
    Yarn: 4.5.3 - ~/.nvm/versions/node/v20.14.0/bin/yarn
    npm: 10.7.0 - ~/.nvm/versions/node/v20.14.0/bin/npm
    bun: 1.1.31 - ~/.bun/bin/bun
  Browsers:
    Chrome: 131.0.6778.86
    Safari: 18.1.1
  npmPackages:
    @vueuse/core: ^12.0.0 => 12.0.0 
    vue: ^3.5.13 => 3.5.13
```

### Used Package Manager

yarn

### Validations

ben-lau:

<details>
<br>

>

</details>

---

### Description

closes #4379 , use mutationobserver to observe whether the element are removed.

### DIFF
--- a/packages/shared/watchWithFilter/index.md
+++ b/packages/shared/watchWithFilter/index.md
@@ -17,7 +17,7 @@ watchWithFilter(
   source,
   () => { console.log('changed!') }, // callback will be called in 500ms debounced manner
   {
-    eventFilter: debounceFilter(500), // throttledFilter, pausabledFilter or custom filters
+    eventFilter: debounceFilter(500), // throttledFilter, pausableFilter or custom filters
   },
 )
 ```


alexzhang1030:
How about making this an optional argument to avoid general performance overhead?

ben-lau:
> How about making this an optional argument to avoid general performance overhead?

Fair, I'll make this optional

PR #4403, opened 2024-12-18
fix(useMediaQuery): the return value should be computed

### LINKED ISSUE #4272: useBreakpoints doesn't adhere its own return type
### Describe the bug

`UseBreakpointsReturn` claims that methods such as `greater` return a computed ref but in fact they return a plain ref from the `useMediaQuery` composable so the actual return of `useBreakpoints` composable is not what `UseBreakpointsReturn` type claims it to be.

In my opinion the types should be in sync and the refs should not be mutable. In my project i solved it by running the refs through `readonly`.

### Reproduction

Not necessary

### System Info

```Shell
Not necessary
```

### Used Package Manager

npm

### Validations

alexzhang1030:

<details>
<br>

>

</details>

---

### Description

fixes #4272

The return value of useMediaQuery should be a getter(readonly computed value), for the general scenario. This will fix #4272 automatically.

### DIFF
--- a/packages/core/useColorMode/index.test.ts
+++ b/packages/core/useColorMode/index.test.ts
@@ -16,8 +16,10 @@ describe('useColorMode', () => {
     }
   })
 
+  const mockPreferredDark = usePreferredDark() as Ref<boolean>
+
   beforeEach(() => {
-    (usePreferredDark() as Ref<boolean>).value = false
+    mockPreferredDark.value = false
     localStorage.clear()
     htmlEl!.className = ''
   })
@@ -30,7 +32,7 @@ describe('useColorMode', () => {
   it('should translate auto mode when prefer dark', async () => {
     const mode = useColorMode()
     mode.value = 'auto'
-    ;(usePreferredDark() as Ref<boolean>).value = true
+    mockPreferredDark.value = true
     await nextTwoTick()
     expect(mode.value).toBe('dark')
     expect(localStorage.getItem(storageKey)).toBe('auto')
@@ -110,7 +112,7 @@ describe('useColorMode', () => {
 
   it('should only change html class when preferred dark changed', async () => {
     const mode = useColorMode({ emitAuto: true })
-    ;(usePreferredDark() as Ref<boolean>).value = true
+    mockPreferredDark.value = true
 
     await nextTwoTick()
     expect(mode.value).toBe('auto')
--- a/packages/core/useColorMode/index.test.ts
+++ b/packages/core/useColorMode/index.test.ts
@@ -1,3 +1,4 @@
+import type { Ref } from 'vue'
 import { afterAll, beforeEach, describe, expect, it, vi } from 'vitest'
 import { nextTick, ref } from 'vue'
 import { useColorMode } from '.'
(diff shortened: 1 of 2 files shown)


packages/core/useMediaQuery/index.ts
```diff
@@ -58,5 +58,5 @@ export function useMediaQuery(query: MaybeRefOrGetter<string>, options: Configur
     mediaQuery = undefined
   })
 
-  return matches
+  return computed(() => matches.value)
```
43081j, line 61: could it make more sense to wrap it in a readonly instead?

```suggestion
return readonly(matches)
```

presumably a `computed` is so you can compute something, whereas `readonly` will just wrap an existing ref as-is

(i could be wrong, tell me if i am!)

alexzhang1030: according to the vue docs, 

>  Takes an object (reactive or plain) or a ref and returns a readonly proxy to the original.

`readonly` is more proper for a `reactive` object.

and `computed` is lazy...

antfu: Yeah, I think both should work in this case. In terms of memory usage, `readonly` is definitely a bit more efficient. But the benefit of `computed` is that it deduces the value and guards unchanged effects to be triggered.

packages/core/useColorMode/index.test.ts
```diff
@@ -16,7 +17,7 @@ describe('useColorMode', () => {
   })
 
   beforeEach(() => {
-    usePreferredDark().value = false
+    (usePreferredDark() as Ref<boolean>).value = false
```
antfu, line 20: I am not sure why we tested like this before. But isn't that the change to `computed` make this statement invalid?

alexzhang1030: This is for type-checking.

The value is mocked as a ref. https://github.com/vueuse/vueuse/pull/4403/files#diff-7811d4dbacf020831d35d217e1b1d3036a472d6a79ede4e1172d06cf9b58937dR12-R16

alexzhang1030: image

alexzhang1030: I’m not an expert in mocking Vitest 🤔 Could you share a pseudocode example?

antfu (approved):

Tofandel:
Hmm, that's a bit redundant, the types of useBreakpoints where pretty much all incorrect and maybe it's better to just fix the types like in #4317 rather than introducing a computed of a ref which adds another effect scope for vue to track

packages/core/useMediaQuery/index.ts
Tofandel: But in this case the effects have already triggered because it's a wrapped ref, there is no deducing or effects to track, so readonly does make more sense

alexzhang1030:
> Hmm, that's a bit redundant, the types of useBreakpoints where pretty much all incorrect and maybe it's better to just fix the types like in #4317 rather than introducing a computed of a ref which adds another effect scope for vue to track

Nope, this is not a fix of #4272, instead, this is a fix of the behavior of `useMediaQuery`. It should not be a mutable ref here.

PR #4404, opened 2024-12-18
test: add useWebSocket tests

43081j:
Adds a few tests for `useWebSocket`.

Unfortunately, this creates a mock `WebSocket` for now. We could possibly use something like MSW in future.

I know we have a _lot_ of PRs to review right now, so this is ofc **a low priority PR**. feel free to ignore it for a while! I've also not contributed to this codebase in a long time, so its possible much of this is nonsense - let me know if it is


antfu (approved):

antfu:
Thanks! More tests are always welcome :)

PR #4414, opened 2024-12-20
chore: remove fs-extra

43081j:
Removes fs-extra and uses built-in `fs` functions.

This also introduces `jsonfile` as a dependency, the package `fs-extra` uses under the hood for read/write of JSON files. This way, we can continue using those utilities without the entirety of `fs-extra`.


antfu:
Why don't we just use JSON.stringify with fs, so we don't even need `jsonfile`?

43081j:
that'd be my preference too. using `JSON.parse` and `JSON.stringify`

i'll update

antfu (approved):

Issue #4193, opened 2024-09-02
useWebsocket watches url when immediate is false


zAlweNy26:
### Describe the bug

In my use case, I let the user change the parameters of the websocket url before starting the connection, but I saw that the connections starts every time the user changes the parameters even if `immediate` is set to `false`.
I found the PR #2367 that was correctly (imho) checking for the `immediate` property before adding the `watch` but a few months ago it was replaced by a commit removing that. Am I missing something? How can I fix my behaviour if this won't be changed on the package side ?

### Reproduction

https://stackblitz.com/edit/vitejs-vite-v4vfzt?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: Linux 6.5 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    CPU: (16) x64 AMD Ryzen 7 PRO 7730U with Radeon Graphics
    Memory: 41.76 GB / 61.60 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.10.0/bin/npm
    pnpm: 9.9.0 - ~/.nvm/versions/node/v20.10.0/bin/pnpm
    bun: 1.1.20 - ~/.bun/bin/bun
  npmPackages:
    @vueuse/components: ^11.0.3 => 11.0.3 
    @vueuse/core: ^11.0.3 => 11.0.3 
    @vueuse/nuxt: ^11.0.3 => 11.0.3 
    @vueuse/router: ^11.0.3 => 11.0.3 
    vue: 3.4.38 => 3.4.38
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
[...]

PR #4417, opened 2024-12-21
feat(useWebSocket): introduce `autoConnect` options to control auto connections on url changes

ilyaliao:
fix: #4193 

<details>
<br>

>

</details>

---

### Description

This PR fixes the WebSocket auto-reconnection behavior when the URL changes. According to the [documentation](https://vueuse.org/core/useWebSocket/#immediate), when `immediate` is set to `false`, the WebSocket should not automatically reconnect even when the URL changes. Users need to explicitly call `open()` to establish the connection.

---

I'm not entirely certain if this implementation is the optimal approach. Perhaps we should consider updating the documentation.

### DIFF
--- a/packages/core/useWebSocket/index.ts
+++ b/packages/core/useWebSocket/index.ts
@@ -76,11 +76,16 @@ export interface UseWebSocketOptions {
   }
 
   /**
-   * Automatically open a connection
+   * Immediately open the connection when calling this composable
    *
    * @default true
    */
   immediate?: boolean
+
+  /**
+   * Automatically connect to the websocket when URL changes
+   */
+  autoConnect?: boolean
 
   /**
    * Automatically close a connection
@@ -157,6 +162,7 @@ export function useWebSocket<Data = any>(
     onError,
     onMessage,
     immediate = true,
+    autoConnect = true,
     autoClose = true,
     protocols = [],
   } = options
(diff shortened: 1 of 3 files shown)


packages/core/useWebSocket/index.ts
```diff
@@ -315,7 +315,10 @@ export function useWebSocket<Data = any>(
   if (immediate)
     open()
 
-  watch(urlRef, open)
+  watch(urlRef, () => {
```
43081j, line 318: since the options can't be a ref, could we just avoid setting a watcher in the first place?

i.e.

```suggestion
if (immediate) {
  watch(urlRef, open)
}
```

afaict `immediate` can't change

ilyaliao: Actually, what you suggested matches exactly how it was in the [initial version](https://github.com/vueuse/vueuse/pull/2367). I'm not sure what happened in between that led to the current situation, but there appears to be some discrepancy between the current functionality and the documentation. 

If this proposal is approved, I plan for `immediate` to be ref/computed as the next step.

43081j: it looks like it was accidentally lost in 93b96ae0a

if your plan is to make `immediate` a ref anyway though, this code makes sense 👍

43081j (approved):

packages/core/useWebSocket/index.ts
antfu: Ah, indeed, `immediate` is in the concept of the watch's immediate behavior. Reactivity to the URL changes is a different concept. I changed this PR to add another option to control that.

antfu (approved):

PR #4426, opened 2024-12-23
feat(router): support partial transform

alexchexes:

---

### Description

Enhance the `transform` option in `useRouteQuery` and `useRouteParams` to support **partial transformations** by allowing **only `get`** or **only `set`** to be defined.  

This change improves flexibility, making it easier to handle cases where transformations are needed in one direction only.  

### Additional Context  
Previously, the `transform` option required defining both **`get` and `set`** methods when using the **object syntax**. This limitation made it cumbersome when only writing transformation was necessary.  

**Key changes:**  
1. **Logic:** Added fallbacks for missing `get` or `set` methods to maintain default behavior without requiring both functions.  
2. **Types:** Updated the type definition to enforce at least one of `get` or `set` must be defined when using the object syntax.  
3. **Tests:** Added tests for cases with only `get` and only `set` being provided.  

### Examples  
```js
// Only 'get' transformation
useRouteQuery('search', undefined, {
  transform: {
    get: (value: string) => value.toLowerCase(),
  },
})

// Only 'set' transformation
useRouteQuery('search', undefined, {
  transform: {
    set: (value: string) => value.trim(),
  },
})
```

### Why this is useful
- Supports scenario where only writing transformation is required.
- Avoids redundant boilerplate for defining both methods when only one is needed.
- Maintains backward compatibility with existing implementations.


packages/router/_types.ts
```diff
@@ -26,10 +26,16 @@ export interface ReactiveRouteOptions {
...
-  }
+  transform?:
+    | ((val: V) => R)
+    | ({
+      get?: (value: V) => R
+      set?: (value: R) => V
+    } & (
+        { get: (value: V) => R }
+        | { set: (value: R) => V }
+      ))
```
OrbisK, line 40: this is just this, right?
```suggestion
  transform?: ((val: V) => R) | Partial<{
    get: (value: V) => R
    set: (value: R) => V
  }>
```

packages/router/useRouteParams/index.ts
```diff
@@ -36,8 +36,18 @@ export function useRouteParams<
     transform,
   } = options
 
-  const transformGet = transform && 'get' in transform ? transform.get : transform ?? ((value: T) => value as any as K)
-  const transformSet = transform && 'set' in transform ? transform.set : (value: K) => value as any as T
+  const defaultGet = <T, K>(value: T) => value as any as K
```
43081j, line 39: Since these don't use anything from the current scope, we could move them out of the function to avoid re-creating them on every call

packages/router/_types.ts
alexchexes: Not exactly. The initial version is stricter and was added to prevent passing `transform: {}`, enforcing that either `get` or `set` must be provided. But you are right, the runtime code already handles this case safely with fallbacks, so even `transform: {}` wouldn't cause unexpected behavior. Strictly speaking, we could use `Partial` as well.

Eventually, I think it comes down to priorities:

- Do we prioritize cleaner types for readability and flexibility? Then we can go with `Partial`
- Or do we enforce API clarity for end users via stricter type checks, even though the runtime already handles nonsensical parameters gracefully?

I'm open to either approach and happy to go with what aligns best with the project's style

OrbisK: > Not exactly. The initial version is stricter and was added to prevent passing `transform: {}`, enforcing that either `get` or `set` must be provided

I don't think there's much point in having such a strict typecheck if it might just confuse users with (maybe) "complex" TS warnings. But thats just opionated 😄 i am fine with both 👍🏽

OrbisK: In the worst case, if someone uses `transform: {}` it'll give both the default behaviors.

packages/router/useRouteParams/index.ts
alexchexes: Thanks, done.

packages/router/useRouteParams/index.ts
```diff
@@ -36,8 +39,15 @@ export function useRouteParams<
     transform,
   } = options
 
-  const transformGet = transform && 'get' in transform ? transform.get : transform ?? ((value: T) => value as any as K)
-  const transformSet = transform && 'set' in transform ? transform.set : (value: K) => value as any as T
+  const transformGet = transform && typeof transform === 'object' && 'get' in transform
```
43081j, line 42: Just an opinion rather than a blocker: IMO we should move ternaries like this into `if` blocks to make the code more readable. Especially since we don't parenthesise any of it 

I suspect there's also a way to fall back to the default once rather than twice

43081j (approved):
Looks good to me, but would relax the type as suggested already (to a partial)

packages/router/_types.ts
rusproject: I ([alexchexes](https://github.com/alexchexes)) replied to this yesterday, but my reply got a "Pending" status, and nobody can see it. Anyone knows why, and how are we supposed to do open-source when simple real-time discussion is impossible here?

OrbisK: > I ([alexchexes](https://github.com/alexchexes)) replied to this yesterday, but my reply got a "Pending" status, and nobody can see it. Anyone knows why, and how are we supposed to do open-source when simple real-time discussion is impossible here?
> 
> P.S. Now all my replies from the initial account are blocked with "Pending" status as well. Wth?

I think you have started a rwiew instead of adding a comment. In top right should be a blue button. If you submit the Review all comments will be published.

alexchexes: Strange. I remember just adding a regular reply, just like I’m doing right now. Anyway, thanks, now it’s visible. Check it out

43081j: since the underlying logic does handle `{}`, and the error will be cleaner, I think we should just relax the type to a `Partial`

in reality, it should actually be `GetAndSet | Get | Set` since `Partial<GetAndSet> & (Get | Set)` means if you set both, the type will actually show as one of the properties being required

alexchexes: > but the error would be equally ugly

Just to clarify: the screenshots weren’t showing what happens with `transform: {}`. They were meant to compare how errors are reported when passing an invalid key like `foo` instead of `get` or `set`.  
With `Partial`, no error is raised when passing `{}` at all.  

So the trade-off is between:  

1. A cleaner type (`Partial`), which allows `{}` even though it makes no sense.  
2. A stricter type that looks more complex (including TS warnings) but prevents `{}`.  

I’m still open to whichever style is preferred here, just wanted to make sure we’re all clear on what the screenshots were illustrating

packages/router/useRouteParams/index.ts
43081j: I think option 1 seems fine if all works well

there's a small performance gain by having the defaults defined once, so it'd be nice to keep them outside

alexchexes: 👌done

packages/router/_types.ts
alexchexes: With the proposed example, it could look like this:
```typescript
interface GetFn<V, R> {
  get: (value: V) => R
}

interface SetFn<V, R> {
  set: (value: R) => V
}

type GetterOrSetter<V, R> = GetFn<V, R> | SetFn<V, R> | (GetFn<V, R> & SetFn<V, R>)

export interface ReactiveRouteOptionsWithTransform<V, R> extends ReactiveRouteOptions {
  /**
   * Function to transform data before returning, or an object with one or both functions:
   * `get` to transform data before returning, and `set` to transform data before setting
   */
  transform?: ((value: V) => R) | GetterOrSetter<V, R>
}
```
That's better than initial option, although I'm not sure if we should really complicate things like this or just stick with `Partial`. I'm not aiming to make it "perfect," so I'll leave this decision to you guys, whoever has the "last word" :)

antfu:
Yeah I think a simple partial would do,

```ts
{
  get?: (value: V) => R
  set?: (value: R) => V
}
```

PR #4427, opened 2024-12-24
perf(useAsyncState): check type of `promise`  once

TheLucifurry:

---

### Description
It seems, that its unnessesary to check type of recieved `promise` parameter on every execute


antfu:
I don't see a difference in your implementation. Meanwhile, we don't do pre-mature optimization without matrics. Thanks.

Issue #4306, opened 2024-10-27
[Docs]: linking to category filter not working


mtdvlpr:
### Describe the bug

When I want to share the functions list of a certain category, I'll send the filtered url, but when you open it, the list isn't filtered. When you press the filter manually it first removes the filter from the url and then you have to press again to get the desired filter working.

### Reproduction

https://vueuse.org/functions.html#category=Watch

### System Info

```Shell
Firefox - Android
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

CofCat456:
Developing locally doesn't seem to have this problem

babu-ch:
I tried serve and was able to reproduce it.
`nr docs:build && nr docs:serve`

If you get the query with onMounted, the discrepancy will be eliminated and it may be cured.

ilyaliao:
Duplicate of #4224, let's move the discussion there.

babu-ch:
@ilyaliao 
Oh, coincidentally I just created a PR 😲

ilyaliao:
> @ilyaliao Oh, coincidentally I just created a PR 😲

I'm sorry about that! Thank you for your contribution. Let’s see how we can resolve this issue.

Issue #4224, opened 2024-09-14
Doc: Filter parameters on the "Functions" page become misaligned with the search results after a page refresh.


heggria:
### Describe the bug

When I use filter parameters on the vueuse.org/functions.html page, the parameters are saved in the URL. However, when I refresh the page, the parameters and search results do not match.

![image](https://github.com/user-attachments/assets/2d691181-5102-4a17-9030-f0e58fc185fc)

### Reproduction

https://vueuse.org/functions.html#category=Animation

### System Info

```Shell
System:
    OS: macOS 14.4
    CPU: (8) arm64 Apple M3
    Memory: 70.25 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.20.4 - ~/.nvm/versions/node/v18.20.4/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v18.20.4/bin/yarn
    npm: 10.7.0 - ~/.nvm/versions/node/v18.20.4/bin/npm
    pnpm: 9.7.1 - ~/Documents/GitHub/heggria.site/node_modules/.bin/pnpm
  Browsers:
    Chrome: 128.0.6613.138
    Safari: 17.4
  npmPackages:
    @vueuse/core: ^11.0.0 => 11.0.0 
    vue: ^3.4.38 => 3.4.38
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

yuxi-ovo:
[...]

PR #4428, opened 2024-12-24
docs: fix linking to category filter

babu-ch:

<details>
<br>

>

</details>

---

### Description

fix 
https://github.com/vueuse/vueuse/issues/4306
https://github.com/vueuse/vueuse/issues/4224

To avoid hydration mismatches, I adjusted the code to use the query after onMounted.
If you know of a better approach, please let me know.
I also fixed the issue where the has directive filter was not cleared when pressing the clear button.

Execute the following command to read the filtered link and confirm that there is no hydration mismatch.
`nr docs:build && nr docs:serve`

### Additional context

🎄


packages/.vitepress/theme/components/FunctionsList.vue
```diff
@@ -15,12 +14,56 @@ useEventListener('click', (e) => {
...
+  category: string | null
+  component: boolean
+  directive: boolean
+  sort: 'category' | 'name' | 'updated' | null
+}>('hash-params', { removeFalsyValues: true })
+
+const search = computed<string | null>({
+  get() {
+    return isMounted.value ? query.search : null
+  },
```
EvgenyWas, line 30: I believe this syntax looks much clearer in this case
```suggestion
  get: () =>  isMounted.value ? query.search : null,
```

EvgenyWas:
It's also a solution, however, in this case, when this parameter actually affects what is rendered in HTML, in my opinion, it's better to use a query to have it on the server as well.

packages/.vitepress/theme/components/FunctionsList.vue
babu-ch: Thank you for your suggestion. I've edited it because it's definitely easier to understand.
https://github.com/vueuse/vueuse/pull/4428/commits/a2fd49fd49097db83f7aa663d37054609eaf27f0

babu-ch:
@EvgenyWas 
Thanks for the review😄
I believe the current configuration is for SSG, so I don't think it is possible to branch the output HTML by query!
We would like to support this if possible!

EvgenyWas:
> @EvgenyWas Thanks for the review😄 I believe the current configuration is for SSG, so I don't think it is possible to branch the output HTML by query! We would like to support this if possible!

Yes, you are right, I just mixed it up with SSR. Then your implementation is fine, in my opinion.

EvgenyWas (approved):

packages/.vitepress/theme/components/FunctionsList.vue
```diff
@@ -1,8 +1,7 @@
 <script setup lang="ts">
-import type { Ref } from 'vue'
-import { useEventListener, useUrlSearchParams } from '@vueuse/core'
+import { useEventListener, useMounted, useUrlSearchParams } from '@vueuse/core'
 import Fuse from 'fuse.js'
-import { computed, toRef } from 'vue'
+import { computed } from 'vue'
```
antfu, line 4: Actually I think we just need to use `toRef` from `@vueuse/core` instead of `vue`. In VueUse, we handle that additional later already.

babu-ch: Thanks for your comments antf!
I tried using vueuse toRef but I continue to get hydration mismatch
The way I tried it was to change toRef to vueuse in the original code

Issue #4092, opened 2024-07-11
`{combination: 'overwrite'}` option for `useFetch`


hequ9:
### Describe the bug

in #4091, i have to override the default `beforeFetch` when using it.
```typescript
// fetcher.ts
export default createFetch({
  options:{
    /**
     * Determine the inherit behavior for beforeFetch, afterFetch, onFetchError
     * @default 'chain'
     */
    combination: 'overwrite'
    beforeFetch(ctx){
      ctx = setToken(ctx)
      return ctx
    },
    afterFetch(ctx){
      console.log(ctx)
      return ctx
    }
  }
})
```
```typescript
// foo.vue
import fetcher from './fetcher'

const {data:fooData} = fethcer('/foo',{ 
/**
 * Actually i only wanted to overload `afterFetch`, 
 * but according to `combination`'s instructions i had to write a default `beforeFetch` here,
 * because the default `beforeFetch` would be invalidated
 */
  beforeFetch(ctx){
    ctx = setToken(ctx)
    return ctx
  },
  afterFetch(ctx){  
    console.log('remove default log.')
    return ctx
  }
}).post().json()
/**
 * in the following example, `{immediate:false}` affects the default `beforeFetch` & `afterFetch` options,
 * which i don't think matches the docs for `{combination: 'overwrite'}`.
 * i'm not sure if other options cause this (e.g. :`refetch`)
 * in practice i have to set {`immediate:false`} a lot to manually trigger requests
 */
const {data:barData,execute} = fethcer('/bar',{ 
    immediate:false
  }
}).post().json()
```
my opinion:
- in `fetcher('/foo')`, the default `beforeFetch` shouldn't be invalidated
- in `fetcher('/bar')`, `{immediate:false}` shouldn't affect the default `beforeFetch` & `afterFetch`, because according to the docs, i didn't override the relevant options
### Reproduction

https://stackblitz.com/edit/vitejs-vite-ntyswm?file=src%2FApp.vue

### System Info

```Shell
System:
  OS: Windows 11 10.0.22000
  CPU: (12) x64 Intel(R) Core(TM) i5-10600KF CPU @ 4.10GHz
  Memory: 2.54 GB / 15.91 GB
Binaries:
  Node: 21.6.2 - F:\nodejs\node.EXE
  Yarn: 1.22.22 - D:\pnpm\bin\yarn.CMD
  npm: 10.2.4 - F:\nodejs\npm.CMD
[...]

PR #4430, opened 2024-12-25
fix(useFetch): partial overwrite when `{combination: 'overwrite'}`

pkc918:
fix[#4092]

```js
  options = {
    ...options,
    ...args[0],
    beforeFetch: combineCallbacks(_combination, _options.beforeFetch, args[0].beforeFetch),
    afterFetch: combineCallbacks(_combination, _options.afterFetch, args[0].afterFetch),
    onFetchError: combineCallbacks(_combination, _options.onFetchError, args[0].onFetchError),
  }
```

When `args[0]` does not have a `beforeFetch`, `afterFetch`, or `onFetchError`,  the corresponding methods from `options` should be used instead.


pkc918:
> Hey @pkc918. Can you please include relevant tests that fail without this PR but pass with it?

@OrbisK Thanks for your review! I have added unit tests, but I have a question: should I add separate value for combination as a partial overwrite?

pkc918:
> > Hey @pkc918. Can you please include relevant tests that fail without this PR but pass with it?
> 
> @OrbisK Thanks for your review! I have added unit tests, but I have a question: should I add separate value for combination as a partial overwrite? Would that be more reasonable?

OrbisK (dismissed):

antfu (approved):

antfu (approved):

PR #4441, opened 2024-12-29
feat(useToNumber): `method` support custom function

btea:

<details>
<br>

>

</details>

---

### DIFF
--- a/packages/shared/useToNumber/index.ts
+++ b/packages/shared/useToNumber/index.ts
@@ -2,14 +2,15 @@ import type { ComputedRef } from 'vue'
 import type { MaybeRefOrGetter } from '../utils'
 import { computed, toValue } from 'vue'
 
-type Method = (value: string | number) => number
 export interface UseToNumberOptions {
   /**
    * Method to use to convert the value to a number.
    *
+   * Or a custom function for the conversion.
+   *
    * @default 'parseFloat'
    */
-  method?: 'parseFloat' | 'parseInt' | Method
+  method?: 'parseFloat' | 'parseInt' | ((value: string | number) => number)
 
   /**
    * The base in mathematical numeral systems passed to `parseInt`.
--- a/packages/shared/useToNumber/index.ts
+++ b/packages/shared/useToNumber/index.ts
@@ -41,6 +41,14 @@ export function useToNumber(
     let resolved = toValue(value)
     if (typeof resolved === 'string')
       resolved = Number[method](resolved, radix)
+    if (!Number.isNaN(resolved)) {
+      if (String(toValue(value)) > String(Number.MAX_SAFE_INTEGER)) {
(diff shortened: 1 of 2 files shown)


packages/shared/useToNumber/index.ts
```diff
@@ -41,6 +41,14 @@ export function useToNumber(
     let resolved = toValue(value)
     if (typeof resolved === 'string')
       resolved = Number[method](resolved, radix)
+    if (!Number.isNaN(resolved)) {
+      if (String(toValue(value)) > String(Number.MAX_SAFE_INTEGER)) {
+        console.warn('useToNumber: value is reaching the limit of Number.MAX_SAFE_INTEGER, resolved value may not be accurate')
+      }
+      else if (String(toValue(value)) < String(Number.MIN_SAFE_INTEGER)) {
+        console.warn('useToNumber: value is reaching the limit of Number.MIN_SAFE_INTEGER, resolved value may not be accurate')
+      }
```
huang-julien, line 50: How about using `Number.isSafeInteger()` ?

huang-julien (changes requested):

packages/shared/useToNumber/index.ts
btea: Looks good, I didn't even know about this API before. 

However, it seems that there is no way to distinguish whether it exceeds the maximum value or is less than the minimum value. Perhaps a comparison is still needed to give a corresponding warning prompt. What do you think?

huang-julien: Indeed there is not, I based my comment on the title of the PR.
Is there a reason you need to know if it's out of  max or min safe integer ?

btea: I have encountered similar scenarios, where I need to convert some string IDs into numbers. However, if the ID is out of range, the resulting number is a number with lost precision, which may not be what I expected. If a warning message is given, it can help developers find the problem more clearly and quickly.

huang-julien: I mean, is it enough if we just say that it's out of safe boudaries ?

btea: I can only give a warning. Do you have any good suggestions?

huang-julien: yes :+1: , I think a warning like `useToNumber: ${value} not a safe integer, resolved value may not be accurate`

btea: Looks good, updated.

huang-julien (approved):

EvgenyWas (approved):

antfu:
In VueUse, we don't usually emit warnings (bundle size, i18n, etc.). I'd prefer to allow users to provide custom function as `method` so they could handle the casting their own

antfu (approved):

PR #3191, opened 2023-06-28
feat: added `query` param to useFetch

jd-solanki:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.

---

### Description

Allow passing query param to useFetch like nuxt. Related issue: #3183

### Additional context

current implementation of useFetch is a bit complex so I've tried my best to implement this feature. Please let me know if I missed something.

@antfu why using `toRaw(reactive(toValue(options.query)))` on type `MaybeRefOrGetter<Record<string, any>>` doesn't give plain object as we discusssed here: #3190 - test is failing?

---

### <samp>🤖 Generated by Copilot at 15fe791</samp>

Added a query option to `useFetch` that allows dynamic URL search params. Updated the demo and tests to reflect the new feature.

### <samp>🤖 Generated by Copilot at 15fe791</samp>

* Add query option to useFetch to allow adding search params to the URL ([link](https://github.com/vueuse/vueuse/pull/3191/files?diff=unified&w=0#diff-0db62ed388d474bc80cd69e2ff1593cd6c9719b8794d8f88459a1704e8da25b5R136-R140), [link](https://github.com/vueuse/vueuse/pull/3191/files?diff=unified&w=0#diff-0db62ed388d474bc80cd69e2ff1593cd6c9719b8794d8f88459a1704e8da25b5L214-R219), [link](https://github.com/vueuse/vueuse/pull/3191/files?diff=unified&w=0#diff-0db62ed388d474bc80cd69e2ff1593cd6c9719b8794d8f88459a1704e8da25b5L406-R447))
  * Import reactive, toRaw, and watch from vue-demi to handle different types of query values ([link](https://github.com/vueuse/vueuse/pull/3191/files?diff=unified&w=0#diff-0db62ed388d474bc80cd69e2ff1593cd6c9719b8794d8f88459a1704e8da25b5L4-R4))
[...]

Issue #586, opened 2021-06-21
[Feature Request] useFetch transform payload into search params when using GET method


xuzaixian:
I think it's a quite usual case when sending request by GET method.

xuzaixian:
If there is anyone likes this feature, i'd like to create pr for this

wheatjs:
There has been discussion about this, but we are looking to slim down on `useFetch` and make it a minimal wrapper for the fetch api. Writing a `fetch` good library is pretty complicated so we would probably suggest using something like `useAxios` instead. We could also introduce something like `useKy`(https://github.com/sindresorhus/ky) if users wanted a more modern solution to `useAxios`

xuzaixian:
Ok, I will use `useAxios` instread.

Grawl:
I don't like Axios API and I don't like `useAxios` API, but I very liked `useFetch` API. I like it's error handling, before/after fetch hooks, and reactivity. Only two things that's missing is query builder for `.get()` request and mixed output type (some of my API methods returning different types: string in one scenario and JSON in another).

IamMille:
> There has been discussion about this, but we are looking to slim down on `useFetch` and make it a minimal wrapper for the fetch api. 

Maybe I'm misunderstanding something here, but there is essentially no logic needed to be added? 
Just use `URLSearchParamsParams` and allow a parameter to be passed to the .get() method here?

Grawl:
Also there is a good [`qs`](https://github.com/ljharb/qs) library to convert type safe object to query string, I use it like this:

```ts
export const getURL = (url: string, query?: Record<string, string | number>): string => {
	if (query !== undefined) {
		return `${url}?${qs.stringify(query)}`
	}
	return url
}
```

PR #4443, opened 2024-12-29
feat(useFetch): Add `query` option

bkeepers:

---

### Description

This adds a `query` option to `useFetch` that will append query parameters to the URL. This is particularly useful for `GET` requests, which don't currently have a way of including a dynamic payload like `POST`, `PUT`, etc.

> ### Query parameters
>
> `useFetch` accepts `query` option to generate URL query params automatically from an object. `query` object can also work with existing query parameters in the URL.
>
> ```ts
> useFetch('https://my-api.com/users', { query: { sortBy: 'username' } })
>
> // Object query will be merged with existing query in the URL
> useFetch('https://my-api.com/users?q=john', { query: { sortBy: 'username' } })
> ```

Replaces #3191, which is a little outdated and has merge conflicts. cc @jd-solanki

### DIFF
--- a/packages/core/useFetch/index.md
+++ b/packages/core/useFetch/index.md
@@ -25,6 +25,17 @@ import { useFetch } from '@vueuse/core'
 const { isFetching, error, data } = useFetch(url)
 ```
 
+### Query parameters
+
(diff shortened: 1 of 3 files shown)


ilyaliao:
Based on the discussion in #586, perhaps we should reconsider whether this feature is necessary.

PR #4450, opened 2025-01-01
fix(useDropZone): improve type matching logic for allowed data types

trandaison:

<details>
<br>

>

</details>

---

### Description

Support allow any format of a media file (e.g. `image/*`)

```ts
const { isOverDropZone: isOverImageDropZone } = useDropZone(
  imageDropZoneRef, 
  { 
    dataTypes: ['image/*'], 
    onDrop: onImageDrop 
  }
)
```

### DIFF
--- a/packages/core/useDropZone/index.ts
+++ b/packages/core/useDropZone/index.ts
@@ -64,7 +64,7 @@ export function useDropZone(
         return false
 
       return types.every(type =>
-        dataTypes.some(allowedType => type.includes(allowedType)),
+        dataTypes.some(allowedType => type.startsWith(allowedType.replace(/\*$/, ''))),
       )
     }
 


NaokiHaba (approved):

43081j:
i wonder if its safer to explicitly treat `/*` differently and continue using `includes` for everything else

for example, if i allow the type `video/3gpp` with this change, im also allowing `video/3gpp2` (a known type). and now there's no way for me to say "i specifically want exactly `video/3gpp`"

so it may be better to do something like:

```ts
if (type.endsWith('/*')) {
  // trim * off and use `startsWith`
} else {
  // existing includes logic
}
```

trandaison:
@43081j thanks for reviewing. I updated the PR as your suggestion.

However, the current logic is `type.includes(allowedType)`, this mean if the `allowedType = video/3gpp` and the `type = video/3gpp` will cause a bug.

I believe the correct logic should be an equal comparison.

```diff
  return types.every(type =>
-   dataTypes.some(allowedType => type.includes(allowedType)),
+   dataTypes.some(allowedType => type === allowedType),
  )
```

43081j:
ah you are correct, i missed that

we should double check with @antfu in that case. if we consider this an existing bug (the fact you can't exact match a type), it wouldn't be a breaking change to change this behaviour. but if we consider it purposely working that way, this'd be a breaking change and your original code is probably easier to ship than my suggestion

PR #4451, opened 2025-01-01
fix(useStorage): ensure `defaults` is only called once when it is a getter

### LINKED ISSUE #4449: BUG | `useStorage` | When `defaults` is a getter, it is invoked twice during initialization
### Describe the bug

When we provide a getter function as the `defaults` for `useStorage`, it gets executed twice during initialization. This can be observed by opening the devtools in the minimal reproduction.

截圖 2025-01-01 中午12 22 32

Perhaps we could first check if `defaults` is a getter function. If it is, we could retrieve the value upfront to avoid it being executed separately during the initialization of `data` and `rawInit`.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-p987jyfs?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M1
    Memory: 210.08 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    npm: 10.9.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
    pnpm: 9.15.1 - ~/Alex.Liu/__Library__/vueuse/node_modules/.bin/pnpm
  Browsers:
    Safari: 17.5
    Safari Technology Preview: 17.4
  npmPackages:
    @vueuse/core: workspace:* => 12.2.0 
    @vueuse/integrations: workspace:* => 12.2.0 
    @vueuse/math: workspace:* => 12.2.0 
    @vueuse/metadata: workspace:* => 12.2.0 
    @vueuse/nuxt: workspace:* => 12.2.0 
    @vueuse/rxjs: workspace:* => 12.2.0 
    @vueuse/shared: workspace:* => 12.2.0 
    vue: catalog: => 3.5.13 
```

### Used Package Manager

pnpm

### Validations

Mini-ghost:

<details>
<br>

>

</details>

---

### Description

fixes #4449 

Currently, during the initialization of `data` and `rawInit`, if `defaults` is provided as a getter function, it will be executed once for each initialization.

In this PR, a check is added to determine if `defaults` is a function. If it is, the returned value is retrieved upfront to ensure it is not executed repeatedly afterward. 

https://github.com/vueuse/vueuse/blob/eefc10f8afaa55efbdc59102fdb2046f06b31846/packages/core/useStorage/index.ts#L156-L171

### DIFF
--- a/packages/core/useStorage/index.ts
+++ b/packages/core/useStorage/index.ts
@@ -153,7 +153,8 @@ export function useStorage<T extends (string | number | boolean | object | null)
     initOnMounted,
   } = options
 
-  const data = (shallow ? shallowRef : ref)(typeof defaults === 'function' ? defaults() : defaults) as RemovableRef<T>
+  const _defaults = typeof defaults === 'function' ? defaults() : defaults
+  const data = (shallow ? shallowRef : ref)(_defaults) as RemovableRef<T>
 
   if (!storage) {
     try {
@@ -167,7 +168,7 @@ export function useStorage<T extends (string | number | boolean | object | null)
   if (!storage)
     return data
 
-  const rawInit: T = toValue(defaults)
+  const rawInit: T = toValue(_defaults)
   const type = guessSerializerType<T>(rawInit)
   const serializer = options.serializer ?? StorageSerializers[type]
 
--- a/packages/core/useStorage/index.test.ts
+++ b/packages/core/useStorage/index.test.ts
@@ -244,22 +244,6 @@ describe('useStorage', () => {
     expect(storage.removeItem).toBeCalledWith(KEY)
   })
 
-  it('getter', async () => {
-    expect(storage.getItem(KEY)).toEqual(undefined)
-
-    const defaults = vi.fn().mockReturnValue('1')
-    const store = useStorage(KEY, defaults, storage)
-
-    expect(defaults).toBeCalledTimes(1)
-    expect(storage.setItem).toBeCalledWith(KEY, '1')
-    expect(store.value).toBe('1')
-
-    store.value = '2'
-    await nextTwoTick()
-
-    expect(storage.setItem).toBeCalledWith(KEY, '2')
-  })
-
   it('pass ref as initialValue', async () => {
     expect(storage.getItem(KEY)).toEqual(undefined)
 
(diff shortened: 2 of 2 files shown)


packages/core/useStorage/index.ts
```diff
@@ -167,7 +168,7 @@ export function useStorage<T extends (string | number | boolean | object | null)
   if (!storage)
     return data
 
-  const rawInit: T = toValue(defaults)
+  const rawInit: T = toValue(_defaults)
```
babu-ch, line 171: Would it perhaps have worked if we just changed it to `toValue(data)` in the original code?

antfu: Yeah, since `defaults` is a `MaybeRefOrGetter`, I think we could use `toValue` every time we need the value. If you only want it to execute once, you can wrap it to `computed()` - so we support both use case for people that wants it to run once, and for people who wants to have it dynamic on every access

Mini-ghost: @babu-ch 
I looked into it, and this approach is simpler and works perfectly. A subtle difference is that with the approach below, the second parameter `defaults` in `mergeDefaults` will become a Reactive object.

```ts
useStorage('store', { data: 'test' }, window.localStorage, {
  mergeDefaults(storageValue, defaults) {
    // original (currently), defaults is { data: string }
    // after using the approach you provided, defaults is Reactive<{ data: string }>
  },
})
```

This is not an issue for me at all, but I’m not sure if it might affect a small portion of users.

Mini-ghost: @antfu Do you mean that the current behavior is expected? If so, I will close this PR and the related issue. Thank you for your review.

antfu: Yeah kinda off. In VueUse we basically assume all getters are pure and cheap to execute.  If you want to use side-effects or heavy computation inside, a `computed` or more controlled ref might give you more freedom to do so. We could improve our docs about that.

Mini-ghost: I also expect getter functions to be pure and strictly followed. However, since they are pure functions, repeated execution should not produce different results. It’s also unlikely that developers would intentionally make them behave differently. Wouldn’t it be better to ensure the getter function is executed only once, especially when it’s only used to retrieve default values?

I might have overlooked something. Thank you for your feedback!

antfu: We accept a Ref for that, meaning that it could be changed. I don't know how useful it is to change the default in this specific case, but that's the pattern throughout the VueUse functions. To me, with or without this PR makes very little difference and probably the majority of the users won't even notice, but what I am concerned about is that with this change, it will limit the capability of doing dynamic getters. Also, if we have this, it may require us to revisit all the functions that doing so. Otherwise, it would create inconsistent expectations across functions.

Mini-ghost: After understanding the reasoning, I have decided to close this PR and the corresponding issue. Thank you for your patience and thoughtful response 💚.

PR #4456, opened 2025-01-02
feat(useElementBounding): immediateWithDelay

stephenjason89:
## Overview

This PR introduces an enhancement to the `useElementBounding` composable by adding an `immediateWithDelay` option. This option allows developers to specify a delay (in milliseconds) before the `update()` function is invoked after the component is mounted.

---

### Changes

- **Added `immediateWithDelay` Option:**  
   The `immediateWithDelay` option accepts a numeric value, defining the delay before the bounding box is recalculated upon mount.

---

### Why This Is Needed

The **`immediateWithDelay`** option enhances the existing `immediate` behavior by allowing a configurable delay for the initial bounding box calculation. While it is true that users can manually call `update()` on `onMounted` with their own delay, the same argument could be made against the existing `immediate` option. 

The inclusion of `immediate` was chosen for **convenience and consistency**, saving users from needing to explicitly call `update()` themselves. Similarly, **`immediateWithDelay`** extends this convenience to scenarios where a delay is required, providing a more integrated and declarative solution.

#### Key Points:
1. **Consistency**: The composable already provides the `immediate` option for convenience. Adding `immediateWithDelay` builds on this pattern to address a common use case.
   
2. **Convenience**: By offering a built-in delay mechanism, users can handle bounding box updates declaratively, avoiding boilerplate code or the need to manage `setTimeout` manually.

3. **Declarative API**: Exposing `immediateWithDelay` aligns with the composable's purpose of simplifying reactive bounding box calculations, keeping the implementation centralized and user-friendly.

---

### Implementation

The `immediateWithDelay` option complements the existing `immediate` property by allowing bounded delays. When both are enabled:
- `immediate` triggers an immediate update.
- `immediateWithDelay` adds a configurable delay for cases requiring stabilization.

---

### Example

```typescript
useElementBounding(target, {
  immediateWithDelay: 200, // Wait 200ms after mount before updating bounding box
});
```

This ensures that bounding box calculations occur after the specified delay, allowing animations or transitions to complete.

---

### Additional Context

This enhancement cannot be implemented externally because the `update()` function is not exposed, making it impossible to manually trigger bounding box recalculations in userland code. By including this option, the composable gains greater flexibility without requiring users to modify its internal behavior.

---

### Checklist

- [x] Checked that there isn't already a PR solving the problem.
- [x] Relevant tests included.

---

### Final Notes

This change prioritizes usability while maintaining compatibility with existing functionality. By keeping the default value for `immediateWithDelay` as `0`, the new feature introduces no breaking changes to the current API.

### DIFF
--- a/packages/core/useElementBounding/index.ts
+++ b/packages/core/useElementBounding/index.ts
@@ -35,6 +35,19 @@ export interface UseElementBoundingOptions {
   immediate?: boolean
 
   /**
+   * Delay (in milliseconds) before calling the `update()` function after the component is mounted.
+   *
+   * This ensures the initial bounding box is computed correctly, especially when the target element
+   * has a transition effect. Without a delay, `update()` might be called prematurely, leading to
+   * inaccurate calculations if the element's layout is still in flux.
+   *
+   * Use this option to allow animations or transitions to complete before the bounding box is measured.
+   *
+   * @default 0
+   */
+  immediateWithDelay?: number
+
+  /**
    * Timing to recalculate the bounding box
    *
    * Setting to `next-frame` can be useful when using this together with something like {@link useBreakpoints}
@@ -60,6 +73,7 @@ export function useElementBounding(
     windowResize = true,
     windowScroll = true,
     immediate = true,
+    immediateWithDelay = 0,
     updateTiming = 'sync',
   } = options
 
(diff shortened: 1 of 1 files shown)


stephenjason89:
Hey @OrbisK, whenever you have a moment, could you also take a look at this PR? It's related to the other one. Thanks a lot! 😊

PR #4459, opened 2025-01-04
fix(useElementBounding): modify the `handler` parameter

pkc918:
The parameter of the handler shouldn't be a ref.

fix: #4609


pkc918:
> Could you elaborate on how your PR solves the problem?你能详细说明一下你的 PR 是如何解决这个问题的吗？

Hi @ilyaliao. Each property that `handler` receives was a Ref, but now those properties have been changed to regular values.

ilyaliao:
Ah, I see. But it seems like the "Maximum recursive updates exceeded" issue is still occurring.

pkc918:
> Ah, I see. But it seems like the `Maximum recursive updates exceeded` issue is still occurring.


Ok！Thx for your reply.  I see I made a mistake. Aside from making the parameters readonly, I can't think of any other solutions. I will close this PR.

Issue #1767, opened 2022-07-06
useBluetooth : isConnected never change from true


KaliaJS:
### Describe the bug

The computed isConnected never changes when I try to stop the device connexion. This when I manually deactivate the sensor (function stop) or if I do it physically by removing the battery from the Polar H7.
  
  ```javascript
  function stop() {
    if (device.value) {
      console.log(server.value.connected) // return true
      console.log(isConnected.value) // return true
      device.value.gatt?.disconnect()

      setTimeout(() => {
        console.log(server.value.connected) // return false
        console.log(isConnected.value) // return true
      }, 3000)
    }
  }
  ```

### Reproduction
https://vueuse.org/core/useBluetooth/#demo

### System Info

```Shell
System:
    OS: macOS 12.3.1
    CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
    Memory: 2.45 GB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.15.0 - /usr/local/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.5.5 - /usr/local/bin/npm
  Browsers:
    Chrome: 103.0.5060.114
    Chrome Canary: 105.0.5163.0
    Firefox: 90.0.2
    Safari: 15.4
  npmPackages:
    @vueuse/core: ^8.8.1 => 8.8.1 
    vue: ^3.2.25 => 3.2.37
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

[...]

PR #4460, opened 2025-01-04
fix(useBluetooth): isConnected state not changed when disconnected

ilyaliao:
fix: #1767 

<details>
<br>

>

</details>

---

### Description

This PR fix `isConnected` state never changes when the device is disconnected.

The solution is to manually change the `isConnected` state by listening to the `gattserverdisconnected` event.


TiiFuchs (approved):
looks good! :)

OrbisK (approved):

Mini-ghost:
Would it be better if `isConnected` were a readonly value?

ilyaliao:
> Would it be better if `isConnected` were a readonly value?

sounds great, but I'm not sure if I should change it in this PR.

Mini-ghost:
> > Would it be better if `isConnected` were a readonly value?
> 
> sounds great, but I'm not sure if I should change it in this PR.


The `isConnected` here is originally a `ComputedRef`, which means it is already readonly for users. For implementation, we changed to using `ref`, but wouldn't it be better to keep it read-only for users?

If there are any factors I may have overlooked, please let me know. Thank you!

ilyaliao:
> > > Would it be better if `isConnected` were a readonly value?
> > 
> > 
> > sounds great, but I'm not sure if I should change it in this PR.
> 
> The `isConnected` here is originally a `ComputedRef`, which means it is already readonly for users. For implementation, we changed to using `ref`, but wouldn't it be better to keep it readonly for users?
> 
> If there are any factors I may have overlooked, please let me know. Thank you!

Oh, I didn't notice that, you're right! I'll add it right away. Thanks for catching that!

packages/core/useBluetooth/index.ts
```diff
@@ -87,23 +87,27 @@ export function useBluetooth(options?: UseBluetoothOptions): UseBluetoothReturn
...
 
   async function connectToBluetoothGATTServer() {
     // Reset any errors we currently have:
     error.value = null
 
     if (device.value && device.value.gatt) {
-      // Add callback to gattserverdisconnected event:
-      device.value.addEventListener('gattserverdisconnected', () => {})
+      // Add reset fn to gattserverdisconnected event:
+      device.value.addEventListener('gattserverdisconnected', reset)
```
ferferga, line 105: This should be replaced with ``useEventListener`` so it's disposed accordingly.

packages/core/useBluetooth/index.ts
```diff
@@ -87,23 +87,27 @@ export function useBluetooth(options?: UseBluetoothOptions): UseBluetoothReturn
     }
   }
 
-  const server = ref<undefined | BluetoothRemoteGATTServer>()
+  const server = shallowRef<undefined | BluetoothRemoteGATTServer>()
+  const isConnected = ref(false)
```
ferferga, line 91:
```suggestion
  const isConnected = shallowRef(false)
```

ferferga (changes requested):

packages/core/useBluetooth/index.ts
antfu: Nice spotting! We could do that in another PR

antfu (approved):

antfu (approved):

PR #4467, opened 2025-01-06
docs(useMediaControls): use useTemplateRef instead of ref

andrewvasilchuk:
`useTemplateRef` is the preferred approach for direct DOM access, as recommended in the [Vue.js documentation](https://vuejs.org/guide/essentials/template-refs.html)

### DIFF
--- a/packages/core/useMediaControls/index.md
+++ b/packages/core/useMediaControls/index.md
@@ -12,9 +12,12 @@ Reactive media controls for both `audio` and `video` elements
 
 ```vue
 <script setup lang="ts">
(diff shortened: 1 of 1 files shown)


OrbisK:
Hey, thank you for this PR.

I am not sure if this should be merged (now). `useTemplateRef` is not supported for `vue<3.5`.

huang-julien (dismissed):
👍

antfu:
I think we could just assume people are on the latest Vue version and use them directly with the commnets

antfu (approved):

Issue #4466, opened 2025-01-06
REQUEST | `useDropZone` | isOverDropZoneWithInvalidFile


alexcroox:
### Clear and concise description of the problem

Currently useDropZone exports a single ref: isOverDropZone

I show a dashed area when the user drags a file over the drop zone.

This doesn't appear if they drag the wrong file type, I'd like to still show the drop zone but with warnings about invalid file type.

Currently I cannot do this with just `isOverDropZone` which will always be false in any state with an invalid file type.

### Suggested solution

Export new ref such as isOverDropZoneWithInvalidFile (feel free to suggest something better), which will always be true if dragging over the drop zone, no matter the file

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #4469, opened 2025-01-07
feat(useDropZone): add `isOverDropZoneWithInvalidFile` to handle invalid file types

phk422:
reference: #4466 

This commit introduces a new ref `isOverDropZoneWithInvalidFile` to the `useDropZone` composable. 
This ref remains true when dragging over the drop zone, regardless of file type validity, 
allowing developers to show UI feedback (e.g., warnings) for invalid file types


packages/core/useDropZone/index.ts
```diff
@@ -94,6 +96,15 @@ export function useDropZone(
...
         }
+        switch (eventType) {
+          case 'enter':
+            isOverDropZoneWithInvalidFile.value = true
+            break
+          case 'leave':
+          case 'drop':
+            isOverDropZoneWithInvalidFile.value = false
+            break
+        }
```
OrbisK, line 107: if we do this we dhould think about `onEnter` and `onOver`

OrbisK: @phk422 I actually meant that we should think about it. If we call the callbacks there, it will be breaking for everyone who relies on it to be valid.

phk422: @OrbisK Thank you for the reminder! However, in the original code, when the file is invalid, onEnter and onOver do not execute, so it seems that this won’t affect users who rely on this behavior, right?

Issue #4393, opened 2024-12-10
REQUEST | `useCloned` | Track changes in original and cloned values


maximepvrt:
### Clear and concise description of the problem

As a developer using VueUse, I want to easily detect if the original or cloned values in `useCloned` have changed so that I can simplify state management, validation logic, and ensure efficient tracking of data modifications.

Currently, `useCloned` requires manual setup with watchers or external comparison libraries to detect changes. Adding this feature would improve usability and reduce boilerplate.

### Suggested solution

In the `useCloned` module, we could add two reactive attributes to the returned object:

1. **`isOriginalModified`**: Tracks if the original value has been modified after cloning.
2. **`isClonedModified`**: Tracks if the cloned value has been modified.

Example:

```javascript
import { ref } from 'vue';
import { useCloned } from '@vueuse/core';

const original = ref({ name: 'Alice', age: 25 });
const { cloned, isOriginalModified, isClonedModified } = useCloned(original);

console.log(isOriginalModified.value); // false
console.log(isClonedModified.value);  // false

// Modify the original value
original.value.name = 'Bob';
console.log(isOriginalModified.value); // true

// Modify the cloned value
cloned.value.age = 30;
console.log(isClonedModified.value); // true
```

This can be implemented by:
- Watching both `original` and `cloned` deeply.
- Using a comparison (e.g., `lodash.isEqual` or `JSON.stringify`) to check for changes.
- Updating the reactive `isOriginalModified` and `isClonedModified` values.

### Alternative

Developers can currently set up their own watchers manually for `original` and `cloned`, but this introduces unnecessary complexity and repetition. 

Another option could be to provide a utility function like `useHasChanged` to check for modifications in any reactive objects, but adding this directly into `useCloned` would be more seamless for this specific use case.

### Additional context

This feature would be particularly useful for use cases like:
- **Form Validation**: Detect if a cloned form state diverges from the original.
- **State Management**: Easily track modifications for undo/redo or syncing logic.
- **Real-Time Applications**: Monitor and handle changes without redundant boilerplate.

Thank you for considering this request! I’d be happy to assist in further detailing the use cases or even work on a pull request if needed.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
[...]

PR #4470, opened 2025-01-07
feat(useCloned): return `isModified`

pkc918:
reference: #4393 

Add a new `isModifed` feature to `useCloned`. This reactive attribute tracks whether the cloned value has been modified.

### DIFF
--- a/packages/core/useCloned/index.ts
+++ b/packages/core/useCloned/index.ts
@@ -24,7 +24,7 @@ export interface UseClonedReturn<T> {
    */
   cloned: Ref<T>
   /**
-   * IsModified ref
+   * Ref indicates whether the cloned data is modified
    */
   isModified: Ref<boolean>
   /**
@@ -45,7 +45,8 @@ export function useCloned<T>(
 ): UseClonedReturn<T> {
   const cloned = ref({} as T) as Ref<T>
   const isModified = ref<boolean>(false)
-  let _isSync = false
+  let _lastSync = false
+
   const {
     manual,
     clone = cloneFnJSON,
@@ -55,8 +56,8 @@ export function useCloned<T>(
   } = options
 
   watch(cloned, () => {
-    if (_isSync) {
-      _isSync = false
+    if (_lastSync) {
+      _lastSync = false
       return
     }
     isModified.value = true
@@ -66,7 +67,7 @@ export function useCloned<T>(
   })
 
   function sync() {
-    _isSync = true
+    _lastSync = true
     isModified.value = false
 
     cloned.value = clone(toValue(source))
--- a/packages/core/useCloned/index.ts
+++ b/packages/core/useCloned/index.ts
@@ -56,7 +56,7 @@ export function useCloned<T>(
 
   watch(cloned, () => {
     if (_isSync) {
-    	_isSync = false
+      _isSync = false
       return
     }
     isModified.value = true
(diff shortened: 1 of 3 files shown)


packages/core/useCloned/index.ts
```diff
@@ -63,5 +68,13 @@ export function useCloned<T>(
...
   }
 
-  return { cloned, sync }
+  const { stop } = watch(cloned, () => {
+    isClonedModified.value = true
+    stop()
+  }, {
+    deep: true,
+    flush: 'sync',
+  })
```
OrbisK, line 77: I think if we sync we should watch again.

OrbisK (changes requested):

packages/core/useCloned/index.ts
pkc918: Thank you for your review. Does `isClonedModified` need to be reset to false inside the `sync` function?

packages/core/useCloned/index.ts
```diff
@@ -23,6 +23,10 @@ export interface UseClonedReturn<T> {
    * Cloned ref
    */
   cloned: Ref<T>
+  /**
+   * IsCloneModified ref
+   */
+  isClonedModified: Ref<boolean>
```
antfu, line 29:
```suggestion
  isModified: Ref<boolean>
```

I think this would do already

packages/core/useCloned/index.ts
antfu: Yeah I think that would make more sense

pkc918: OK! I will fix it as soon as possible. Thx!

OrbisK:
looks good so far. If you have the time, please consider adding some tests. 👍🏽

pkc918:
> looks good so far. If you have the time, please consider adding some tests. 👍🏽

Thx! Should the syntax of Vue3.5 be avoided? like `pause & resume`.

pkc918:
Hi，@OrbisK ！The unit tests have been supplemented. If you have time, I hope you can review then again. Thx a lot!

OrbisK:
> > looks good so far. If you have the time, please consider adding some tests. 👍🏽
> 
> Thx! Should the syntax of Vue3.5 be avoided? like `pause & resume`.

We have recently added the +vue@3.3 syntax (`toValue`). I am not sure if we will drop vue@3.4 yet. But I think this use of pause and resume is pretty clean. I think Anthony will have an opinion on that 😄

OrbisK (approved):

pkc918:
> > > looks good so far. If you have the time, please consider adding some tests. 👍🏽
> > 
> > 
> > Thx! Should the syntax of Vue3.5 be avoided? like `pause & resume`.
> 
> We have recently added the +vue@3.3 syntax (`toValue`). I am not sure if we will drop vue@3.4 yet. But I think this use of pause and resume is pretty clean. I think Anthony will have an opinion on that 😄我们最近添加了+vue@3.3 语法（ `toValue` ）。我不确定我们是否还会放弃 vue@3.4。但我觉得这种暂停和恢复的使用相当干净。我认为安东尼会有看法😄

OK! The syntax for Vue3.5 has been canceled. If I need to use it after future version updates, I will modify it then.

packages/core/useCloned/index.ts
```diff
@@ -48,7 +54,19 @@ export function useCloned<T>(
     immediate = true,
   } = options
 
+  watch(cloned, () => {
+    if (_isSync)
+      return _isSync = false
```
antfu, line 59:
```suggestion
    if (_isSync) {
    	_isSync = false
      return
    }
```

antfu (approved):

PR #4471, opened 2025-01-07
feat(useRefHistory): add `shouldCommit`

### LINKED ISSUE #4468: REQUEST | `useDebouncedRefHistory ` |  Can you provide a hook function to provide prev and next data to block whether to continue committing
### Describe the bug

When directly assigning an array ref, commit will definitely trigger. I hope that a prefix function can be provided that can return me the previous and current values to compare whether commit is really needed

### Reproduction

https://stackblitz.com/edit/vitejs-vite-xo6udi6k

### System Info

```Shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i5-10500 CPU @ 3.10GHz
    Memory: 4.45 GB / 15.75 GB
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i5-10500 CPU @ 3.10GHz
    Memory: 4.45 GB / 15.75 GB
    CPU: (12) x64 Intel(R) Core(TM) i5-10500 CPU @ 3.10GHz
    Memory: 4.45 GB / 15.75 GB
    Memory: 4.45 GB / 15.75 GB
  Binaries:
    Node: 20.16.0 - E:\nvm\nodejs\node.EXE
    Yarn: 1.22.22 - E:\nvm\nodejs\yarn.CMD
    npm: 10.8.1 - E:\nvm\nodejs\npm.CMD
    pnpm: 9.12.1 - E:\nvm\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (129.0.2792.79)
    Internet Explorer: 11.0.19041.4355
  npmPackages:
    @vueuse/components: ^11.1.0 => 11.1.0
    @vueuse/core: ^11.1.0 => 11.1.0
    vue: ^3.5.11 => 3.5.11
```

### Used Package Manager

npm

### Validations

JonathanSchndr:
### Description

This PR adds a `beforeCommit` option to `useRefHistory` that allows users to control whether a commit should proceed based on comparing the previous and new values. This is particularly useful when you want to prevent unnecessary history entries when values haven't meaningfully changed.

The `beforeCommit` function receives two parameters:
- `oldValue`: The previous value from history
- `newValue`: The current value being committed

It returns a boolean indicating whether the commit should proceed.

Example usage:
```ts
const { history } = useRefHistory(myRef, {
  beforeCommit: (oldValue, newValue) => {
    return JSON.stringify(oldValue) !== JSON.stringify(newValue)
  }
})
```

### Additional context

- The implementation preserves all existing functionality
- The `beforeCommit` check is performed just before the actual commit
- All original comments and documentation have been maintained
- The change is fully typed with TypeScript
- This feature is particularly useful for array/object references where you want to compare actual content rather than just references

Fixes #4468 

[x] Read the Contributing Guidelines
[x] Read the Pull Request Guidelines
[x] Checked for duplicate PRs
[x] Provided description of the solution
[x] Added relevant tests (to be added if required by maintainers)


43081j:
someone who understands filters more should confirm im not talking nonsense

but i wonder if its better this be implemented as some kind of event filter, such that it never reaches `commit` (similar to how you can probably do that now via the `eventFilter` option)

if it turns out that's a silly idea - this PR does look good meanwhile, although i'd be tempted to call it something like `shouldCommit` since its specifically to check if it should commit, rather than being a plain hook

tetap:
> someone who understands filters more should confirm im not talking nonsense
> 
> but i wonder if its better this be implemented as some kind of event filter, such that it never reaches `commit` (similar to how you can probably do that now via the `eventFilter` option)
> 
> if it turns out that's a silly idea - this PR does look good meanwhile, although i'd be tempted to call it something like `shouldCommit` since its specifically to check if it should commit, rather than being a plain hook
> 
> edit:
> 
> basically a filter like this:
> 
> ```ts
> function equalityFilter<T>(val: Ref<T>, isEqual: (oldValue: T, newValue: T) => boolean): EventFilter {
>      const lastValue = ref(val.value)
> 
>      return (invoke) => {
>        if (!isEqual(lastValue.value, val.value))
>          lastValue.value = val.value
>          invoke()
>        }
>     }
>    }
> ```
> 
> but yeah i could just be overcomplicating things 🤷

ShouldCommit is a more appropriate name, beforeCommit is more like a hook.
eventFilter currently does not support multiple scenarios. If equalityFilter is used, debounceFilter will fail again. This should be a broader scenario and a heavier workload. I think this design is the best. Just consider whether you really need to add multiple to eventFilter.

OrbisK:
Looks good so far. Maybe you can add tests to prevent future regressions. 👍🏽

JonathanSchndr:
> Looks good so far. Maybe you can add tests to prevent future regressions. 👍🏽

Added :)

OrbisK (dismissed):

packages/core/useRefHistory/index.ts
```diff
@@ -45,6 +45,13 @@ export interface UseRefHistoryOptions<Raw, Serialized = Raw> extends Configurabl
    * Deserialize data from the history
    */
   parse?: (v: Serialized) => Raw
+  /**
+   * Function to determine if the commit should proceed
+   * @param oldValue Previous value
+   * @param newValue New value
+   * @returns boolean indicating if commit should proceed
+   */
+  shouldCommit?: (oldValue: Serialized | undefined, newValue: Raw) => boolean
```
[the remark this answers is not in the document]
JonathanSchndr: you mean as `shouldCommit?: (oldValue: Raw | undefined, newValue: Raw) => boolean` ?
then they would be the same, but is it really better? what do you think speaks in favor of this?

ilyaliao (approved):

packages/core/useRefHistory/index.ts
```diff
@@ -138,6 +150,12 @@ export function useRefHistory<Raw, Serialized = Raw>(
     // so we do not trigger an extra commit in the async watcher
     ignorePrevAsyncUpdates()
 
+    if (shouldCommit) {
+      if (!shouldCommit(lastRawValue, source.value))
+        return
+    }
```
OrbisK: lets give `shouldCommit` the default value of `()=>true`, that way, we dont need this type cast

packages/core/useRefHistory/index.ts
```diff
@@ -93,6 +100,7 @@ export function useRefHistory<Raw, Serialized = Raw>(
     deep = false,
     flush = 'pre',
     eventFilter,
+    shouldCommit,
```
OrbisK:
```suggestion
    shouldCommit = () => true,
```

packages/core/useRefHistory/index.ts
```diff
@@ -138,6 +150,12 @@ export function useRefHistory<Raw, Serialized = Raw>(
     // so we do not trigger an extra commit in the async watcher
     ignorePrevAsyncUpdates()
 
+    if (shouldCommit) {
+      if (!shouldCommit(lastRawValue, source.value))
+        return
+    }
+
+    lastRawValue = source.value
```
OrbisK:
```suggestion
    if (!shouldCommit(lastRawValue, source.value))
      return

    lastRawValue = source.value
```

OrbisK (approved):
We don't need this check anymore. Besides that, it LGTM.

OrbisK (approved):

Issue #4035, opened 2024-06-10
add automatic scroll while using useDraggable


3aluw:
### Clear and concise description of the problem

While useDraggable offers a convenient way to make elements draggable within their parent container, it encounters limitations when dealing with scrollable parents. In such scenarios, users might need to perform multiple drags to reach the desired position within the scrollable area, leading to a clunky and inefficient user experience

You can see this in action [https://stackblitz.com/edit/vitejs-vite-mmp6rw?file=src%2FApp.vue](here)

### Suggested solution

Add an auto scroll functionality to useDraggable.

### Alternative

Add a new module built on top `useMouse `and `usescroll `that scrolls automatically and can be used alongside useDraggable 

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Gazoon007:
I think this issue has been solved. @ilyaliao

PR #4440, opened 2024-12-28
test: add browser tests with `*.browser.test.ts` files

OrbisK:
related to #3792

### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

This pr adds an optional vitest browser test to the `*.browser.test.ts` files. This allows us to test demo pages and functions that require a browser (such as useAnimate).

### Additional context

I've no clue how to add deps to the catalog instead the root package.

PR #4472, opened 2025-01-07
feat(useDraggable): auto-scroll with restricted dragging within the container

Gazoon007:

<details>
<br>

>

</details>

---

### Description

This PR resolves this issue https://github.com/vueuse/vueuse/issues/4035

I created the auto-scroll functionality and added 4 new options:  

- **autoScroll**  
  Enable auto-scroll when dragging near the edges.  
- **scrollSpeed**  
  Control the speed of auto-scroll.
- **scrollMargin**  
  Set the margin from the edge that triggers auto-scroll.  
- **scrollDirection**  
  Determine the direction of auto-scroll. 

https://github.com/user-attachments/assets/975604d1-83c0-491f-baa6-20a9eb7f22da

### Additional context

Before that, I made another adjustment by adding another option and setting the style: 

- **restrictInContainer** (please tell me if someone has a better name for this)
  Restrict dragging within the container. If the pointer drags the element beyond the container area, the element will stay within the edge of the current dragging axis.  

  This adjustment is necessary (imo) because I want the auto-scroll to behave similarly (kinda) to [PlainDraggable](https://anseki.github.io/plain-draggable/#options-autoscroll). I am unsure whether this option should be the default behavior for `useDraggable` or not.

  If `autoScroll` is enabled, this option is automatically activated because (imo) it creates a poor user experience if an element is dragged out of view but auto-scroll continues.

  Additionally, `containerElement` must be defined for this to work.

- **text-wrap: nowrap**  
  Prevents text from wrapping when auto-scroll is initialized.

### DIFF
--- a/packages/core/useDraggable/index.md
+++ b/packages/core/useDraggable/index.md
@@ -39,6 +39,17 @@ const { x, y, style } = useDraggable(el, {
 })
 ```
 
+Set `autoScroll: true` to enable auto-scroll when dragging near the edges.
+
+```ts
+const { x, y, style } = useDraggable(el, {
+  autoScroll: true,
+  scrollSpeed: 2, // Control the speed of auto-scroll.
+  scrollMargin: 30, // Set the margin from the edge that triggers auto-scroll.
+  scrollDirection: 'both' // Determine the direction of auto-scroll.
+})
+```
+
 ## Component Usage
 
 ```vue
--- a/packages/core/useDraggable/index.ts
+++ b/packages/core/useDraggable/index.ts
@@ -148,37 +148,6 @@ export interface UseDraggableOptions {
   scrollDirection?: 'x' | 'y' | 'both'
 }
 
-function handleAutoScroll(
-  container: HTMLElement | SVGElement,
-  targetRect: DOMRect,
-  position: Position,
-  scrollSpeed: number,
-  scrollMargin: number,
-  direction: 'x' | 'y' | 'both',
-) {
-  const { clientWidth, clientHeight } = container
-  let deltaX = 0
-  let deltaY = 0
-
-  if ((direction === 'x' || direction === 'both')) {
-    if (position.x <= scrollMargin)
-      deltaX = -scrollSpeed
-    else if ((position.x + targetRect.width >= clientWidth - scrollMargin))
-      deltaX = scrollSpeed
-  }
-
-  if ((direction === 'y' || direction === 'both')) {
-    if (position.y <= scrollMargin)
-      deltaY = -scrollSpeed
-    else if ((position.y + targetRect.height >= clientHeight - scrollMargin))
-      deltaY = scrollSpeed
-  }
-
-  if (deltaX || deltaY) {
-    container.scrollBy({ left: deltaX, top: deltaY, behavior: 'auto' })
-  }
-}
-
 /**
  * Make elements draggable.
  *
@@ -228,6 +197,37 @@ export function useDraggable(
       e.preventDefault()
     if (toValue(stopPropagation))
       e.stopPropagation()
+  }
+
+  const handleAutoScroll = (
+    container: HTMLElement | SVGElement,
+    targetRect: DOMRect,
+    position: Position,
+    scrollSpeed: number,
+    scrollMargin: number,
+    direction: 'x' | 'y' | 'both',
+  ) => {
+    const { clientWidth, clientHeight } = container
+    let deltaX = 0
+    let deltaY = 0
+
+    if ((direction === 'x' || direction === 'both')) {
+      if (position.x <= scrollMargin)
+        deltaX = -scrollSpeed
+      else if ((position.x + targetRect.width >= clientWidth - scrollMargin))
+        deltaX = scrollSpeed
+    }
+
+    if ((direction === 'y' || direction === 'both')) {
+      if (position.y <= scrollMargin)
+        deltaY = -scrollSpeed
+      else if ((position.y + targetRect.height >= clientHeight - scrollMargin))
+        deltaY = scrollSpeed
+    }
+
+    if (deltaX || deltaY) {
+      container.scrollBy({ left: deltaX, top: deltaY, behavior: 'auto' })
+    }
   }
 
   const intervalId = ref<NodeJS.Timeout | null>(null)
--- a/packages/core/useDraggable/index.ts
+++ b/packages/core/useDraggable/index.ts
@@ -1,7 +1,7 @@
 import type { MaybeRefOrGetter } from '@vueuse/shared'
 import type { PointerType, Position } from '../types'
 import { isClient, toRefs } from '@vueuse/shared'
-import { computed, ref, toValue } from 'vue'
+import { computed, ref, toValue, watch } from 'vue'
 import { defaultWindow } from '../_configurable'
 import { useEventListener } from '../useEventListener'
 
@@ -111,6 +111,72 @@ export interface UseDraggableOptions {
    * @default [0]
    */
   buttons?: MaybeRefOrGetter<number[]>
+
+  /**
+   * Whether to restrict dragging within the container.
+   *
+   * @default false
+   */
+  restrictInContainer?: MaybeRefOrGetter<boolean>
+
+  /**
+   * Whether to enable auto-scroll when dragging near the edges.
+   *
+   * @default false
+   */
+  autoScroll?: MaybeRefOrGetter<boolean>
+
+  /**
+   * Speed of auto-scroll.
+   *
+   * @default 1
+   */
+  scrollSpeed?: MaybeRefOrGetter<number>
+
+  /**
+   * Margin from the edge to trigger auto-scroll.
+   *
+   * @default 30
+   */
+  scrollMargin?: MaybeRefOrGetter<number>
+
+  /**
+   * Direction of auto-scroll.
+   *
+   * @default 'both'
+   */
+  scrollDirection?: 'x' | 'y' | 'both'
+}
+
+function handleAutoScroll(
+  container: HTMLElement | SVGElement,
+  targetRect: DOMRect,
+  position: Position,
+  scrollSpeed: number,
+  scrollMargin: number,
+  direction: 'x' | 'y' | 'both',
+) {
+  const { clientWidth, clientHeight } = container
+  let deltaX = 0
+  let deltaY = 0
+
+  if ((direction === 'x' || direction === 'both')) {
+    if (position.x <= scrollMargin)
+      deltaX = -scrollSpeed
+    else if ((position.x + targetRect.width >= clientWidth - scrollMargin))
+      deltaX = scrollSpeed
+  }
+
+  if ((direction === 'y' || direction === 'both')) {
+    if (position.y <= scrollMargin)
+      deltaY = -scrollSpeed
+    else if ((position.y + targetRect.height >= clientHeight - scrollMargin))
+      deltaY = scrollSpeed
+  }
+
+  if (deltaX || deltaY) {
+    container.scrollBy({ left: deltaX, top: deltaY, behavior: 'auto' })
+  }
 }
 
 /**
@@ -138,6 +204,11 @@ export function useDraggable(
     containerElement,
     handle: draggingHandle = target,
     buttons = [0],
+    restrictInContainer,
+    autoScroll = false,
+    scrollSpeed = 2,
+    scrollMargin = 30,
+    scrollDirection = 'both',
   } = options
 
   const position = ref<Position>(
@@ -159,6 +230,64 @@ export function useDraggable(
       e.stopPropagation()
   }
 
+  const intervalId = ref<NodeJS.Timeout | null>(null)
+  const startAutoScroll = () => {
+    const container = toValue(containerElement)
+    if (container && !intervalId.value) {
+      intervalId.value = setInterval(() => {
+        const targetRect = toValue(target)!.getBoundingClientRect()
+        const { x, y } = position.value
+        let adjustedX = x - container.scrollLeft
+        let adjustedY = y - container.scrollTop
+        if (adjustedX >= 0 && adjustedY >= 0) {
+          handleAutoScroll(
+            container,
+            targetRect,
+            { x: adjustedX, y: adjustedY },
+            toValue(scrollSpeed),
+            toValue(scrollMargin),
+            toValue(scrollDirection),
+          )
+          adjustedX += container.scrollLeft
+          adjustedY += container.scrollTop
+          position.value = {
+            x: adjustedX,
+            y: adjustedY,
+          }
+        }
+      }, 1000 / 60)
+    }
+  }
+  const stopAutoScroll = () => {
+    if (intervalId.value) {
+      clearInterval(intervalId.value)
+      intervalId.value = null
+    }
+  }
+  const checkAutoScroll = () => {
+    if (toValue(options.disabled) || !pressedDelta.value)
+      return
+    const container = toValue(containerElement)
+    if (!container)
+      return
+
+    const isInScrollableArea
+      = (container.scrollLeft > 0
+        && container.scrollLeft + 1
+        < container.scrollWidth - container.clientWidth)
+      || (container.scrollTop > 0
+        && container.scrollTop + 1
+        < container.scrollHeight - container.clientHeight)
+
+    if (isInScrollableArea)
+      startAutoScroll()
+    else stopAutoScroll()
+  }
+
+  if (toValue(autoScroll)) {
+    watch(position, checkAutoScroll)
+  }
+
   const start = (e: PointerEvent) => {
     if (!toValue(buttons).includes(e.button))
       return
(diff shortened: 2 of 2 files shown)


packages/core/useDraggable/index.ts
```diff
@@ -111,6 +111,41 @@ export interface UseDraggableOptions {
    * @default [0]
    */
   buttons?: MaybeRefOrGetter<number[]>
+
+  /**
+   * Whether to restrict dragging within the container.
+   *
+   * @default false
+   */
+  restrictInContainer?: MaybeRefOrGetter<boolean>
```
OrbisK, line 120: i think this might be misleading. Maybe this should be e.g. `restrictInView`

Gazoon007: @OrbisK Could you tell me why `restrictInContainer` is misleading?
Personally, I think it makes it more ambiguous.

The reason "container" is included in the name is that the function is closely associated with `elementContainer`, which is why I named it that way.

OrbisK: Well, basically the container is the entire scrollable element. The drag element never leaves the container. It's more a question of whether it should remain in the visible area.

Gazoon007: @OrbisK Okay, that makes sense I guess. I'll adjust to it since I'm not great at naming things.

packages/core/useDraggable/index.ts
```diff
@@ -159,6 +202,102 @@ export function useDraggable(
...
+    container: HTMLElement | SVGElement,
+    targetRect: DOMRect,
+    position: Position,
+    scrollSpeed: number | Position,
+    scrollMargin: number | Position,
+    direction: 'x' | 'y' | 'both',
+  ) => {
+    const { clientWidth, clientHeight } = container
+
+    const getAxisValues = (value: number | Position): [number, number] =>
```
43081j, line 215: nit: you could move this outside of the `useDraggable` function (within this file still) to avoid creating the function on every call (since its a pure function, doesn't need any state)

packages/core/useDraggable/index.ts
```diff
@@ -159,6 +202,102 @@ export function useDraggable(
...
+      else if (position.y + targetRect.height >= clientHeight - marginY)
+        deltaY = speedY
+    }
+
+    if (deltaX || deltaY) {
+      container.scrollBy({ left: deltaX, top: deltaY, behavior: 'auto' })
+    }
+  }
+
+  const intervalId = ref<NodeJS.Timeout | null>(null)
```
43081j, line 243: iirc elsewhere we used `ReturnType<typeof setTimeout>` because in browsers it won't be a `NodeJS.Timeout` (node types override window's types from what i remember)

packages/core/useDraggable/index.ts
Gazoon007: @43081j I'm not sure of the idea of bringing the local function that can only be used inside handleAutoScroll to be accessible globally. How about I still keep `getAxisValues` inside the handleAutoScroll by implementing closure?

```
const handleAutoScroll = (() => {
  const getAxisValues = (value: number | Position): [number, number] =>
    typeof value === 'number' ? [value, value] : [value.x, value.y]

  return (
    // ...
  ) => {
    // ...

    const [marginX, marginY] = getAxisValues(scrollMargin)
    const [speedX, speedY] = getAxisValues(scrollSpeed)

    // ...
  }
})()
```

Afaik, this will avoid creating the function on every call

packages/core/useDraggable/index.ts
Gazoon007: @43081j Yes you're right, roger that sir

packages/core/useDraggable/index.ts
43081j: It isn't accessible globally, it's inside a module (so will be local to the module)

You can put whatever utils and constants you want within the file as they're local/private to it

Gazoon007: @43081j Oh yes, sorry. What I meant by 'global' is still within the `useDraggable` module's domain, not literally global. In the future, any new functionality besides auto-scroll can access it within the same module. Is that still okay?

43081j: I think it's fine. If you wanted to make it clear it's for scrolling, you could name it more explicitly (like `getScrollAxisValues`)

This was just a minor comment anyway so we shouldn't think too much about it. If you do it, you're saving some memory by having one copy of the function.

It isn't a blocker though

43081j: i think using an iife will make the syntax more confusing than it needs to be

i would just take it out into the module scope, but if you disagree with that then feel free to leave it where it is unless another reviewer says otherwise. i don't want to block this as its a tiny comment anyway 👍 

don't worry about it

OrbisK:
would it make sense to combine these options under one?:

```ts
{
  autoScroll: boolean
  scrollSpeed: number
  scrollMargin: number
  scrollDirection: '...'
}
```

to:

```ts
{
  autoScroll: boolean | {
    // enabled: boolean
    speed: number
    margin: number
    direction: '...'  
  }
}
```
all scroll options only make sense if autoscroll is set. 

@Gazoon007 what do you think?

Gazoon007:
> would it make sense to combine these options under one?:
> @Gazoon007 what do you think?

Nice, I prefer this one too, I'll commit the changes soon

packages/core/useDraggable/index.md
```diff
@@ -39,6 +39,17 @@ const { x, y, style } = useDraggable(el, {
...
 
+Set `autoScroll: true` to enable auto-scroll when dragging near the edges.
+
+```ts
+const { x, y, style } = useDraggable(el, {
+  autoScroll: true,
+  scrollSpeed: 2, // Control the speed of auto-scroll.
+  scrollMargin: 30, // Set the margin from the edge that triggers auto-scroll.
+  scrollDirection: 'both' // Determine the direction of auto-scroll.
+})
```
OrbisK, line 50:
```suggestion
const { x, y, style } = useDraggable(el, {
  autoScroll: {
    speed: 2, // Control the speed of auto-scroll.
    margin: 30, // Set the margin from the edge that triggers auto-scroll.
    direction: 'both' // Determine the direction of auto-scroll.
  },
})
```

OrbisK (changes requested):

OrbisK:
@Gazoon007 I will try to write some tests in the next days, so we can merge this sooner 👍🏽

3aluw:
When will this PR get merged ?

packages/core/useDraggable/index.ts
```diff
@@ -159,6 +199,106 @@ export function useDraggable(
       e.stopPropagation()
   }
 
+  const scrollConfig = toValue(autoScroll)
+  const defaultScrollConfig = { speed: 2, margin: 30, direction: 'both' }
```
43081j, line 203: you could move this outside the function to avoid creating it every time (since it is constant)

Gazoon007:
> When will this PR get merged ?

Wait for @OrbisK to finish some tests. If he's tied up with something else, I'll do it myself.

43081j:
Can you also catch the branch? Main has changed since so there will be some conflicts I think

OrbisK:
> > When will this PR get merged ?
> 
> Wait for @OrbisK to finish some tests. If he's tied up with something else, I'll do it myself.

Ohhh. Not sure why it was removed from my todo list 😅

Gazoon007:
> > > When will this PR get merged ?
> > 
> > 
> > Wait for @OrbisK to finish some tests. If he's tied up with something else, I'll do it myself.
> 
> Ohhh. Not sure why it was removed from my todo list 😅

Because it takes too long, I'll do it myself guys, lemme cook. 

I'll create tests for existing behaviour and also the auto-scroll functionality

OrbisK:
@Gazoon007 Sorry, I was pretty busy. I had started, but never finished it 😅

packages/core/useDraggable/index.browser.test.ts
```diff
@@ -0,0 +1,214 @@
...
+  }: {
+    axis?: 'x' | 'y'
+    containerRect: DOMRect
+    el: HTMLElement
+    margin?: number
+    offset: number
+  }) {
+    if (axis === 'x') {
+      const scrollTriggerX = containerRect.right - margin
+      const leftEdge = scrollTriggerX - el.offsetWidth
```
OrbisK, line 137: we should use `useTemplateRef` here

Gazoon007: To confirm, did you mean to suggest using `useTemplateRef` for the refs in the setup function?  If so, then, yes, my bad. I was relying on VTU’s behavior, not Vue’s standard convention for getting DOM elements.

packages/core/useDraggable/index.browser.test.ts
```diff
@@ -0,0 +1,214 @@
...
+        <div style="width: 1000px; height: 1000px; position: relative;">
+          <div ref="el" style="width: 100px; height: 100px; position: absolute; top: 100px; left: 100px; background: darkslategray;" />
+        </div>
+      </div>
+    `
+    return mount(defineComponent({
+      template,
+      setup() {
+        const el = shallowRef<HTMLElement | null>(null)
+        const container = shallowRef<HTMLElement | null>(null)
```
OrbisK: https://github.com/vueuse/vueuse/pull/4472/files/af58439785d8623262c43d2b3ca18c4f19f89c28#r2202271351

not sure how my comment ended up there 😅 but i mean using `useTemplateRef` instead of shallowRef.

43081j:
sorry just now getting around to thoroughly reviewing this

we're computing the position in an interval to know if to scroll the container or not. this feels like its an inefficient way of doing it

is it not possible to observe the drag events to track the mouse position, and scroll when it changes?

drag events contain the mouse coords, so i wonder if we can just listen for that?

on another note, we should generally keep the tests in one or the other - browser or not. rather than having two suites. in this case, we should probably just have all the tests in the browser suite

Gazoon007:
> we're computing the position in an interval to know if to scroll the container or not. this feels like its an inefficient way of doing it
> 
> is it not possible to observe the drag events to track the mouse position, and scroll when it changes?
> 
> drag events contain the mouse coords, so i wonder if we can just listen for that?
> 

@43081j Relying solely on pointer events is tricky. If the mouse stays near the edge, scrolling stops even though it should continue. To keep the dragged element under the pointer and maintain a smooth experience, we need a loop that keeps scrolling as long as the pointer is near the edge, especially if it’s not moving and remains at the exact same screen coordinate.

Tbh, my math ain't mathing to solve this without some kind of loop, like `setInterval` or `requestAnimationFrame`. If you’ve figured out a way to handle it without using one, I’m genuinely curious and would love to learn your approach.

Even [PlainDraggable ](https://github.com/anseki/plain-draggable/blob/18be3a2d4998d4623579a440fa6277adb0f3d7d5/src/plain-draggable.js#L77) uses a loop with `requestAnimationFrame` for this.

Using RAF is actually more performant and efficient than `setInterval`.

But when using RAF, I notice that scroll values may vary slightly between frames (maybe) due to browser timing and system load. This makes strict `.toBe()` assertions unreliable in tests.

If you're okay with expecting results as a range when using RAF, instead of exact values like with `setInterval`, I'm totally fine with that too.

Gazoon007:
> on another note, we should generally keep the tests in one or the other - browser or not. rather than having two suites. in this case, we should probably just have all the tests in the browser suite

Actually, I'm following what useScroll tests by providing 2 tests that require a browser, and not.

43081j:
> Actually, I'm following what useScroll tests by providing 2 tests that require a browser, and not

that too should provide one suite rather than two

all vue functions will run in a browser ultimately, so if we do add tests which require the browser, we may as well just move the whole suite.

> Relying solely on pointer events is tricky. If the mouse stays near the edge, scrolling stops even though it should continue

I understand now 👍  thanks for clarifying. ill have a think

Gazoon007:
> > Actually, I'm following what useScroll tests by providing 2 tests that require a browser, and not
> 
> that too should provide one suite rather than two
> 
> all vue functions will run in a browser ultimately, so if we do add tests which require the browser, we may as well just move the whole suite.

@43081j But if I look into some modules, a lot of them are following this pattern—like `useAnimate`, `useClipboard`, `useScroll`, `useWebSocket`, `useElementBounding`, and many others. Should we still proceed to keep everything in one browser-only suite? I personally prefer keeping the separation since it gives a nice separation of concern.

If moving everything into a single browser suite is the intended pattern from now on, maybe it’s worth refactoring the existing ones and documenting it in `CONTRIBUTING.md` to make it clearer?

43081j:
>  I personally prefer keeping the separation since it gives a nice separation of concern.

where do you draw the line? what is the separation of concern?

i.e. what makes a test a "browser test" vs not? how do you decide which file the test belongs in, given all tests could be browser tests

sorry for all the questions - this will be good feedback for what we do in future. forget about it for now in this PR and ill catch up with the other maintainers around it

Gazoon007:
> where do you draw the line? what is the separation of concern?
> 
> i.e. what makes a test a "browser test" vs not? how do you decide which file the test belongs in, given all tests could be browser tests

Intuitively `*.browser.test.ts` means it requires a browser to test, the reason is because I read this pr description https://github.com/vueuse/vueuse/pull/4440,

And also I think something that import from `@vitest/browser/context` is considered to be a browser test, and for my case although it doesn't import from there but, it specifically needs a browser (with headless set to false) to test, some test cases which are auto scroll functionality needs to know scrollbar height and width, if it runs without browser or with headless set to true, the scrollbar height and width are always 2px, and it makes calculation not reliable.

I hope there is a clearer explanation about it in the contribution guide doc.

43081j:
makes sense 👍 

really, all tests we have should be browser tests (and the same goes for any other project shipping browser components/libraries).

historically, people didn't do this because the infrastructure didn't exist or browser tests were slow. the downside, though, is that JSDOM isn't a real browser, and isn't representative of where our code will run in production. so we're testing against a made up platform nobody has in the real world.

these days, playwright (and chrome headless) are pretty fast. i suspect, if we invented vueuse today, all our tests would be browser tests because of this. 

i'll catch up with the rest of the team, but i think if a function needs even one test to be a browser test, we should probably just move the whole suite.

Gazoon007:
You make a good point about why all tests should be browser tests in the current infrastructure.

Alright then, I’ll commit the adjustment shortly.

I also hope that in the future, all tests will be refactored into a single suite to avoid confusion for new contributors like myself, and/or that this guideline will be documented in the `CONTRIBUTING.md` file.

Thank you for telling me 😀

43081j:
i've started the discussion with the other maintainers 👍  so we will likely update the contribution docs and such in the near future

thanks for helping us notice this :D

Gazoon007:
Hi! Just a friendly follow-up on this PR, it’s been open for exactly a year.
Happy to help if anything is needed from my side.

packages/core/useDraggable/index.browser.test.ts
```diff
@@ -0,0 +1,459 @@
...
+    const containerRect = container.getBoundingClientRect()
+    return { wrapper, el, container, elRect, containerRect }
+  }
+
+  async function simulateAutoScrollDrag({
+    el,
+    container,
+    pointerdown,
+    pointermove,
+    duration = 1000,
```
43081j: this seems like a long time to be delaying tests. we should find out how long the browser really needs, it might even be instant (i.e. we wouldn't need the delay, or just use `0`)

anything to explain the duration basically, rather than it being arbitrary

packages/core/useDraggable/index.ts
```diff
@@ -159,6 +201,118 @@ export function useDraggable(
...
+        speed: toValue(scrollConfig.speed) ?? defaultScrollConfig.speed,
+        margin: toValue(scrollConfig.margin) ?? defaultScrollConfig.margin,
+        direction: scrollConfig.direction ?? defaultScrollConfig.direction,
+      }
+    : defaultScrollConfig
+
+  const getScrollAxisValues = (value: number | Position): [number, number] =>
+    typeof value === 'number' ? [value, value] : [value.x, value.y]
+
+  function clampContainerScroll(container: HTMLElement) {
```
43081j: this is a pure function so can be moved outside of the vue function (top level in this file). otherwise it'll get created for every call

packages/core/useDraggable/index.ts
```diff
@@ -171,21 +325,25 @@ export function useDraggable(
...
+
   const move = (e: PointerEvent) => {
     if (toValue(options.disabled) || !filterEvent(e))
       return
     if (!pressedDelta.value)
       return
 
     const container = toValue(containerElement)
+    if (container)
+      clampContainerScroll(container as HTMLElement)
```
43081j: the types declare we can pass an SVG into this function, so this cast is making an incorrect assumption.

either the `clampContainerScroll` should also accept `SVGElement`, or we should somehow check here that it is _actually_ a `HTMLElement`

packages/core/useDraggable/index.browser.test.ts
Gazoon007: I had adjusted the duration (and also the interval, because it was intended to be 1000 / 60 to match the function)

I investigated the minimum timing the browser actually needs. Auto-scroll in useDraggable advances every ~16.7 ms (the hook’s setInterval(..., 1000 / 60)), so our test now dispatches pointer moves at the same cadence and only for 300 ms. That short window still yields about 12–13 ticks at the default 2 px per tick (~24 px total), so we tightened the assertions to require at least 20 px of scroll on each axis rather than zero.

Anything less than ~300 ms becomes unreliable because fewer ticks means <20 px of motion, which no longer proves the auto-scroll actually advanced. Setting the duration to 0 would skip all pointer moves entirely.

OrbisK (approved):

PR #4479, opened 2025-01-10
feat: use `useEventListener` where it was not being used

ferferga:

<details>
<br>

>

</details>

---

### Description

`useEventListener` was not being used in all the contexts. With this PR, we're sure the events are correctly disposed always.

### Additional context

* Some composables have been refactored a little, given `useEventListener` simplifies the cleanup logic
* All references to `àddEventListener` in docs have been updated as well to give the best practices rightaway to users.
* Only 1 usage of ``addEventListener`` has been left untouched because it's been removed in #4481

### DIFF
--- a/packages/core/useBluetooth/index.ts
+++ b/packages/core/useBluetooth/index.ts
@@ -4,6 +4,7 @@ import { tryOnMounted, tryOnScopeDispose } from '@vueuse/shared'
 import { readonly, shallowRef, watch } from 'vue'
 
 import { defaultNavigator } from '../_configurable'
+import { useEventListener } from '../useEventListener'
 import { useSupported } from '../useSupported'
 
 export interface UseBluetoothRequestDeviceOptions {
@@ -102,7 +103,7 @@ export function useBluetooth(options?: UseBluetoothOptions): UseBluetoothReturn 
 
     if (device.value && device.value.gatt) {
       // Add reset fn to gattserverdisconnected event:
-      device.value.addEventListener('gattserverdisconnected', reset)
+      useEventListener(device, 'gattserverdisconnected', reset, { passive: true })
 
       try {
         // Connect to the device:
--- a/packages/core/useBroadcastChannel/index.ts
+++ b/packages/core/useBroadcastChannel/index.ts
@@ -3,6 +3,7 @@ import type { ConfigurableWindow } from '../_configurable'
 import { tryOnMounted, tryOnScopeDispose } from '@vueuse/shared'
 import { ref, shallowRef } from 'vue'
 import { defaultWindow } from '../_configurable'
+import { useEventListener } from '../useEventListener'
 import { useSupported } from '../useSupported'
 
 export interface UseBroadcastChannelOptions extends ConfigurableWindow {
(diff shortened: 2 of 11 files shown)


packages/core/useBroadcastChannel/index.ts
```diff
@@ -49,17 +50,21 @@ export function useBroadcastChannel<D, P>(options: UseBroadcastChannelOptions):
...
+      useEventListener(channel, 'messageerror', (e: MessageEvent) => {
         error.value = e
-      }, { passive: true })
+      }, listenerOptions)
 
-      channel.value.addEventListener('close', () => {
+      useEventListener(channel, 'close', () => {
         isClosed.value = true
-      })
+      }, listenerOptions)
```
ilyaliao, line 67: Unsure if these parameters have any effect.

截圖 2025-01-10 晚上9 39 06

```suggestion
      const listenerOptions = {
        passive: true,
      }

      useEventListener(channel, 'message', (e: MessageEvent) => {
        data.value = e.data
      }, listenerOptions)

      useEventListener(channel, 'messageerror', (e: MessageEvent) => {
        error.value = e
      }, listenerOptions)

      useEventListener(channel, 'close', () => {
        isClosed.value = true
      }, listenerOptions)
```

packages/core/useBluetooth/index.ts
```diff
@@ -102,7 +103,7 @@ export function useBluetooth(options?: UseBluetoothOptions): UseBluetoothReturn
 
     if (device.value && device.value.gatt) {
       // Add reset fn to gattserverdisconnected event:
-      device.value.addEventListener('gattserverdisconnected', reset)
+      useEventListener(device, 'gattserverdisconnected', reset, { passive: true })
```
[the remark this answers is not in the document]
ilyaliao: > FYI `BlueToothDevice` is an `EventTarget` so does have the underlying, generic `addEventListener(target, name, options)`. so this should work
> 
> but we didn't set `passive` before i guess so it is a change worth noting

Thanks for the explanation!

packages/core/useBroadcastChannel/index.ts
ferferga: @ilyaliao Yes, for the same reason listed above (BroadcastChannel is part of the standard DOM API and the listened events are also standard).

I purposefully skipped custom events in my other PR yesterday (#4477). An example:

https://github.com/vueuse/vueuse/pull/4477/files#diff-015d17be17c2a02d87f6b40bc47f526a5e82a01ee334f2883d271c6d98543d21L193

ilyaliao (approved):

antfu (approved):
I consider this a feature that worth noting in the release note. I am not 100% if this is complete side-effect free when we use this inside a callback. I updated the title.

ferferga:
@antfu Yes it is (in my app I do use watchers inside `Promise.all` and other "strange" places and they're garbaged correctly).

This is the only case I found where `effectScope` is not disabled properly: https://github.com/vuejs/core/issues/12681
In fact, I was thinking at opening a PR at a later time with a new util like `tryDispose` that checks if there's an effect scope, if not then an appScope, so we toggle between `onScopeDispose` and `onUnmount` depending on the situation. But honestly, I prefer to have this fixed in Vue core.

PR #4481, opened 2025-01-10
refactor(useDevicePixelRatio): use `useMediaQuery`

ferferga:

<details>
<br>

>

</details>

---

### Description

Use the `useMediaQuery` composable inside `useDevicePixelRatio` to greatly simplify the logic.

### Additional context

* The returned object from the composable has been changed. Does this qualify as a breaking change? Given a new property was added (not removed/modified) I believe it's good to go in a minor release.
* Related to https://github.com/vueuse/vueuse/pull/4479

### DIFF
--- a/packages/core/useDevicePixelRatio/index.ts
+++ b/packages/core/useDevicePixelRatio/index.ts
@@ -1,6 +1,6 @@
 import type { ConfigurableWindow } from '../_configurable'
 import { noop, watchImmediate } from '@vueuse/shared'
-import { ref } from 'vue'
+import { readonly, ref } from 'vue'
 import { defaultWindow } from '../_configurable'
 import { useMediaQuery } from '../useMediaQuery'
 
@@ -22,7 +22,10 @@ export function useDevicePixelRatio(options: ConfigurableWindow = {}) {
     stop = watchImmediate(query, () => pixelRatio.value = window!.devicePixelRatio)
   }
 
-  return { pixelRatio, stop }
+  return {
+    pixelRatio: readonly(pixelRatio),
+    stop,
+  }
 }
 
 export type UseDevicePixelRatioReturn = ReturnType<typeof useDevicePixelRatio>
--- a/packages/core/useDevicePixelRatio/index.ts
+++ b/packages/core/useDevicePixelRatio/index.ts
@@ -1,7 +1,8 @@
 import type { ConfigurableWindow } from '../_configurable'
-import { tryOnScopeDispose } from '@vueuse/shared'
+import { noop, watchImmediate } from '@vueuse/shared'
 import { ref } from 'vue'
 import { defaultWindow } from '../_configurable'
+import { useMediaQuery } from '../useMediaQuery'
 
 /**
  * Reactively track `window.devicePixelRatio`.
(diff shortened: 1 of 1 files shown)


ilyaliao (dismissed):
I tested it in Chrome devTools, and it looks good to me.

packages/core/useDevicePixelRatio/index.ts
```diff
@@ -14,24 +15,14 @@ export function useDevicePixelRatio(options: ConfigurableWindow = {}) {
...
-      media?.removeEventListener('change', observe)
-    }
-
-    observe()
-    tryOnScopeDispose(cleanup)
+    stop = watchImmediate(query, () => pixelRatio.value = window!.devicePixelRatio)
   }
 
-  return { pixelRatio }
+  return { pixelRatio, stop }
```
43081j, line 25: instead of introducing a new `stop` value, should we not just stop on dispose like `useMediaQuery` itself does?

```ts
tryOnScopeDispose(() => {
  stop();
});
```

ferferga: That's unnecessary since the watcher will be disposed by the parent effectScope directly, it's redundant (in fact, I've seen many places where this is done unnecessarily, but that's for another PR)

This is why I added it: https://vueuse.org/guide/best-practice.html#side-effect-clean-up

43081j: that makes sense

the doc you linked to does say not all functions return a `stop` handler. what do you think means we should return one in this case?

OrbisK, line 25:
```suggestion
  return { readonly(pixelRatio), stop }
```

maybe we should readonly this too

ferferga: From the source:

``Some functions return a stop handler just like the watch function. Not all functions return a stop handler so a more general solution is to use the effectScope API from Vue``

In reality, the stop function is never needed in most cases (as I said before, everything effect-based from Vue will be picked automatically by the parent component and recursively if nested some levels deep). However, if there is no parent component/scope (detected by the getCurrentInstance or getCurrentScope), which happens when used in JS/TS normal modules, they will stand forever.

From now on, forget about normal components, this is all about the usage Vue's context.

As suggested in the docs, you can create an effectScope and don't need to have the composable returning anything (and in my opinion that's the best thing, because everything is picked automatically, even if we forgot to add it to the stop callback). Exporting the callback and calling it is another way of doing things.

At the end of the day it's a matter of preference. Before effectScope existed in Vue, the only way to do it was calling a handler and, if you look to other composables that use watchers internally, they export both the watcher options and the stop callback. It's just another way of doing the same thing and, for consistency, I added it here to match the behaviour of other composables.

43081j: Makes sense to me 👍 thanks for the explanation. Will help when I'm reviewing future PRs for sure

antfu:
Hey @ferferga! Thanks for your great contributions, I just send you an invitation to the org without granting you the triage right, if you are interested in helping out. Also come to say hi on Discord if you use it, we have a channel for collaborating as well.

ferferga:
Hello @antfu! Thank you very much for the invitation! I'm very glad to be helpful when I can. I just joined your Discord server as well 🙌!

PR #4483, opened 2025-01-10
test(useDevicePixelRatio): add basic browser tests with zoom

OrbisK:

<details>
<br>

>

</details>

---

### Description

I was preparing the same changes as #4481. 

Got stuck testing for a while. Zooming in playwright/vitest is pretty hard 😄

Added 2 vitest commands to manipulate `document.body.style.zoom`. This has no effect when executed in a test iframe, so it needs to be evaluated on the top level document via `page.evaluate` (command).

If there is a cleaner or better solution please hit me up 😅

### DIFF
--- a/packages/.test/commands.ts
+++ b/packages/.test/commands.ts
@@ -1,0 +1,38 @@
+// Hilfstyp, um den ersten Parameter aus einem Tuple zu entfernen
+import type { BrowserCommand } from 'vitest/node'
+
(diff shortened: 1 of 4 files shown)


43081j:
you could also test that firing a `change` event of the media query you're observing runs the expected logic etc

it'd be less e2e than this, though. so if we'd rather do it this way, it seems fine to me

PR #4497, opened 2025-01-16
fix(debounceFilter): invoke on maxWait with the latest invoker

### LINKED ISSUE #3665: Debounce filter: defers the last debounced event until a new key is pressed
### Bug:

image

(See reproduction link)

As you can see, the last few words weren't there after the debounce completes. It shows up when I press a new character into the input, which is unexpected.

### How it should work:

I have created a minimal re-implementation of the debounce filter to verify the behavior. I'm not sure how `lastRejectOnCancel` works. But it is likely the cause I can tell.

https://jsfiddle.net/4tgk2n5r/1/

image

### Reproduction

https://play.vueuse.org/#N4IgDghgxg1hDmBTAziAXAbVAOwgW0XRADcBXRAWgBNE8BLEAGhGQHtSAnKQtEU7MDHhMQNZFA50wAFzqtsRAGrkABABFadFQAoAFhAA2AMxV1sKgGIdE2KLoCUp5CogqaxRAdZgz8FaVkDOmkATxcDLwB3ZxD2FWlWFUjJaUQVAFVsOg8OZEMVZTSAGToAIw4ISRQVI1YOAtUAJhUAMhUAZhFOAyJdaWkwZDQAemH+QXgAOihWPGGyShp6YaDS4bMaAA9JvAArVABfRhx8HhAAAQXSZERh5H1rKhE2Tm4icaERMQkpWXkiADKD0QVAaiHSN38gWCdBQky6HB6vD6AyGow+Uxmc0u5Gut3ulRB5wAjAAGSbEyak9bYLY7fYgI4nAhEHGIPHDGbWZ7sLhnDFfFA-GRyBS8ADCrAiiCgf3MrBMKBu2Fk+UKKkleDArGQMPkGWhshQCKRIBRgxGYwEQmms2GbI5XMQJPJlOpG0Q2z2hwAuswjHQDMa0KAAIJgMCTBboUCpLUGCCpIgAHjMYACKmIFDwrBoBgAvAAdECpTbSUqsTbFlTnNMBIssRDSUhgACiCcGIOrwwAfIXsP3k2Ae8BgPFPeXKyoDgdk8Nh4Ph+2IJ3QaOVJ4VzdQTO5wvsEOR2OaKVp7P532FMxxJIZEQ6Fq6tIVGPrCYDjUOLMVAByBY--sH21DhnzHSJEzsDQK34bgd0-b8fwdG5OTqRAAIHDCZmwZBn1LSdNhUfMVDfbQfx-ex+ywnCNw7bdCOIxAjG0UkKOwINn1KRB4DMABuSj5Gok96JIsjWP7cDpEgxBoNsEFtDwitNkYHRiEcfMexffsVDcaSo0MVQiOIPjsCOF8dJk7g0BUMlSVJZS8AgTYAHUIGCKz2ls0lpzE7AjBguUVBuZs21ouTHGALTTBMbQAEJOO47Bwsi7SqKlRBJi8eBSIgcjkpUeKzHotRE3S7BWEibR7GM7TtKCgBJFVEA4YhDG0SrCI0iLzBqlKBLSjLWCyn9Sly7qes3Vc9IMAydG0YrUkmMqKscCh8q4sxHGGazPPsSYEgsOhNjkli8tMmzWO0g5+yuhQZwOIA

### System Info

```Shell
Tried all online, on Firefox 121 and macOS Sonoma.
```

### Used Package Manager

npm

### Validations

43081j:
Fixes #3665
Closes #4473 

A filter works by binding a watch callback and its args every time the watcher fires. This bound function is then held back until your "filter" calls its `invoke` parameter.

In `debounceFilter`, we setup a `maxWait` timer to fire after a certain amount of time. Once this fires, it calls `invoke`.

The problem here is that the `invoke` will be the paired cb/args from **the point in time where we started the timer**.

For example:

```ts
const cb = vi.fn();
const someRef = ref(0);

watchWithFilter(
  someRef,
  cb,
  {
    eventFilter: debounceFilter(100, {maxWait: 500})
  }
);

someRef.value = 1; // cb(1, 0) is held back, `maxWait` timer is kicked off _with `invoke = () => cb(1, 0)`_
someRef.value = 2; // cb(2, 0)
// somehow 500ms passes
// maxWait timer now calls `cb(1, 0)` rather than cb(2, 0)
```

To fix this, i've added `lastInvoker`, which is the last invoke function we created for this filter. Once the max timer fires, we will clear the regular timer, and call `lastInvoker` (which a regular timer originally set up).


packages/shared/watchDebounced/index.test.ts
```diff
@@ -42,6 +42,33 @@ describe('watchDebounced', () => {
     num.value = 5
     await vi.advanceTimersByTimeAsync(75)
     expect(cb).toHaveBeenCalledTimes(2)
-    expect(cb).toHaveBeenCalledWith(4, 2, expect.anything())
+    expect(cb).toHaveBeenCalledWith(5, 4, expect.anything())
```
43081j, line 45: as far as i can tell, this test was already showing the bug but we just rolled with it

second pair of eyes on this test would be helpful

packages/shared/watchDebounced/index.test.ts
```diff
@@ -42,6 +42,33 @@ describe('watchDebounced', () => {
...
+    await constantUpdateOverTime(1)
+    expect(cb).toHaveBeenCalledTimes(1)
+    await constantUpdateOverTime(50)
+    expect(cb).toHaveBeenCalledTimes(2)
+    await constantUpdateOverTime(50)
+
+    expect(cb).toHaveBeenCalledTimes(3)
+    expect(cb.mock.calls[0][0]).toBe(50) // or 51
+    expect(cb.mock.calls[1][0]).toBe(100) // or 101
+    expect(cb.mock.calls[2][0]).toBe(150) // or 151
```
OrbisK, line 72:
```suggestion
    expect(cb.mock.calls[0][0]).toBe(50)
    expect(cb.mock.calls[1][0]).toBe(100)
    expect(cb.mock.calls[2][0]).toBe(150)
```

i was not sure which tick it should trigger :)

OrbisK (changes requested):

OrbisK (approved):

pavi2410:
Great work @43081j James!

PR #4500, opened 2025-01-16
fix(useCssVar): new Behavior

### LINKED ISSUE #4493: BUG | `useCssVar` | `initialValue` is not working
### Describe the bug

在复现场景中，用例1和用例2为什么不一样？从源码的角度来看，这两个用例应该是差不多的。

Why are Case 1 and Case 2 different in reproducing scenarios? From the perspective of source code, these two use cases should be similar.

### Reproduction

https://play.vueuse.org/#N4IgDghgxg1hDmBTAziAXAbVAOwgW0XRADcBXRAWgBNE8BLEAGhGQHtSAnKQtEU7MDHhMQNZFA50wAFzqtsRAGrkABABFadFQAoAFhAA2AMxV1sKgGIdE2KLoCUp5CogqaxRAdZgz8FaVkDOmkATxcDLwB3ZxD2FWlWFUjJaUQVAFVsOg8OZEMVZTSAGToAIw4ISRQVI1YOAtUAJhUAMhUAZhFOAyJdaWkwZDQAemH+QXgAOihWPGGyShp6YaDS4bMaAA9JvAArVABfRhx8HhAAAQXSZERh5H1rKhE2Tm4icaERMQkpWXkiADKD0QVAaiHSN38gWCdBQky6HB6vD6AyGow+Uxmc0u5Gut3ulRB5wAjAAGSbEyak9bYLY7fYgI4nAhEHGIPHDGbWZ7sLhnDFfFA-GRyBS8ADCrAiiCgf3MrBMKBu2Fk+UKKkleDArGQMPkGWhshQCKRIBRgxGYwEQmms2GbI5XMQJPJlOpG0Q2z2hwAuswjHQDMa0KAAIJgMCTBboUCpLUGCCpIgAHio2RU1iMAF4ADogTx5gB8wGAKhmXnqBwOyeGaeIhZzHBz2FT2WLpbwyD8VZrdcLz2F0iIdC1dWkKlLePFyGQikqKgONQ4sxUAHIHTdOXVEKvmyPtRxx6XM4wkom7Kf5ABZdgqkELpcr1cLXfYZujMsQSHE5szbDIcdPBULMM0QIxtHsX95AAsspTqYD-BuadZ0qbRVwoChyzqVdT08SC32wSJz10bQsI4U9tGIQxyEcLNCwnZsVFg-8pUQSYvHgbQAAM7AgbAkDQFQABJgCogxyAOLj8IOfDm2vW9UiobQIOA+jgEY2CKyjai0hA1dHlfGTm3fYZP0hRooP-cdOz8EDMzQ3RPC8JI6gMKhV1kwjiO0GyKLEmjVIY8xmLYIN2NYTieP0fjEEEkT-MQSTpM8+T+EU5TaLUjSbO08TdLXRyIlYQz7EZKsgA

### System Info

```Shell
无

None
```

### Used Package Manager

pnpm

### Validations

ilyaliao:
resolves #4493 

<details>
<br>

>

</details>

---

### Description

This PR enhances several behaviors:

- Automatically adds `variable` to `window.document.documentElement` or the `target` on first use, and also when the `target element` changes.
- Allows users to change the color again in `onMounted` when setting `initialValue`.
- Adds some new `tests` to ensure that both the `state` and the `element` are indeed changed.

### DIFF
--- a/packages/core/useCssVar/index.ts
+++ b/packages/core/useCssVar/index.ts
@@ -69,6 +69,7 @@ export function useCssVar(
           elRef.value.style.setProperty(raw_prop, val)
       }
     },
+    { immediate: true },
   )
 
   return variable
--- a/packages/core/useCssVar/index.ts
+++ b/packages/core/useCssVar/index.ts
@@ -55,7 +55,7 @@ export function useCssVar(
         old[0].style.removeProperty(old[1])
       updateCssVar()
     },
-    { immediate: true },
+    { flush: 'post', immediate: false },
   )
 
   watch(
--- a/packages/core/useCssVar/index.ts
+++ b/packages/core/useCssVar/index.ts
@@ -37,7 +37,7 @@ export function useCssVar(
     const el = toValue(elRef)
     if (el && window && key) {
       const value = window.getComputedStyle(el).getPropertyValue(key)?.trim()
-      variable.value = value || initialValue
+      variable.value = value || variable.value || initialValue
     }
   }
 
@@ -55,7 +55,6 @@ export function useCssVar(
         old[0].style.removeProperty(old[1])
       updateCssVar()
     },
-    { flush: 'post' },
   )
 
   watch(
--- a/packages/core/useCssVar/index.ts
+++ b/packages/core/useCssVar/index.ts
@@ -59,8 +59,8 @@ export function useCssVar(
   )
 
   watch(
-    variable,
-    (val) => {
+    [variable, elRef],
+    ([val]) => {
       const raw_prop = toValue(prop)
       if (elRef.value?.style && raw_prop) {
         if (val == null)
--- a/packages/core/useCssVar/index.ts
+++ b/packages/core/useCssVar/index.ts
@@ -59,13 +59,13 @@ export function useCssVar(
 
   watch(
     [variable, elRef],
-    ([val]) => {
+    ([val, el]) => {
       const raw_prop = toValue(prop)
-      if (elRef.value?.style && raw_prop) {
+      if (el?.style && raw_prop) {
         if (val == null)
-          elRef.value.style.removeProperty(raw_prop)
+          el.style.removeProperty(raw_prop)
         else
-          elRef.value.style.setProperty(raw_prop, val)
+          el.style.setProperty(raw_prop, val)
       }
     },
     { immediate: true },
--- a/packages/core/useCssVar/index.ts
+++ b/packages/core/useCssVar/index.ts
@@ -55,7 +55,7 @@ export function useCssVar(
         old[0].style.removeProperty(old[1])
       updateCssVar()
     },
-    { flush: 'post', immediate: false },
+    { flush: 'post' },
   )
 
   watch(
--- a/packages/core/useCssVar/index.test.ts
+++ b/packages/core/useCssVar/index.test.ts
@@ -77,4 +77,28 @@ describe('useCssVar', () => {
     expect(vm.variable).toBe('blue')
     expect(vm.el.style.getPropertyValue('--color')).toBe('blue')
   })
+
+  it('should have existing value', async () => {
+    const vm = mount(defineComponent({
+      setup() {
+        const el = ref()
+
+        const color = '--color'
+        const variable = useCssVar(color, el)
+
+        return {
+          el,
+          variable,
+        }
+      },
+      template: `
+        <div ref="el" style="--color: red"></div>
+      `,
+    }))
+
+    await nextTick()
+    expect(vm.variable).toBe('red')
+    expect(window.getComputedStyle(vm.el).getPropertyValue('--color')).toBe('red')
+    expect(vm.el.style.getPropertyValue('--color')).toBe('red')
+  })
 })
--- a/packages/core/useCssVar/index.test.ts
+++ b/packages/core/useCssVar/index.test.ts
@@ -1,6 +1,6 @@
 import { defaultWindow } from '@vueuse/core'
 import { describe, expect, it } from 'vitest'
-import { defineComponent, h, nextTick, onMounted, useTemplateRef } from 'vue'
+import { defineComponent, h, nextTick, onMounted, ref, shallowRef, useTemplateRef } from 'vue'
 import { useCssVar } from '.'
 import { mount } from '../../.test'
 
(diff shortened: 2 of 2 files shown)


ilyaliao:
Should this work?

```vue
<script setup lang="ts">
import { useCssVar } from '@vueuse/core'
import { ref } from 'vue'

const el = ref(null)
const color = useCssVar('--color')

function switchColor() {
  if (color.value === '#df8543')
    color.value = '#7fa998'
  else
    color.value = '#df8543'
}
</script>

<template>
  <div ref="el" style="color: var(--color)">
    Sample text, {{ color }}
  </div>
  <button @click="switchColor">
    Change Color
  </button>
</template>

```

OrbisK:
> Should this work? If so, then this is a new bug.

Do you mean the color switch?

YiMo1:
但是依旧没有解决`initialValue`的问题。

However, the issue of `initialValue` has not been resolved yet.

https://github.com/vueuse/vueuse/issues/4493#issuecomment-2594319783

ilyaliao:
> > Should this work? If so, then this is a new bug.
> 
> Do you mean the color switch?

When I load the page, the color state changes to `#df8543`, but the actual element color doesn't change until the next update. This happens because `--color: #7fa998` isn't added to `window.document.documentElement` tag on the initial load.

If this isn't the expected behavior, we can fix it in this PR.

https://github.com/user-attachments/assets/1412f3d2-a7e2-4d74-81a2-d85a08031100

ilyaliao:
> 但是依旧没有解决`initialValue`的问题。
> 
> However, the issue of `initialValue` has not been resolved yet.
> 
> [#4493 (comment)](https://github.com/vueuse/vueuse/issues/4493#issuecomment-2594319783)

對，這個 PR 主要是修復關於在 `onMounted` 中修改 `color` 狀態時未改變的問題，另一個問題如同上面我提到的，如果他不是預期行為，我們可以考慮在這個 PR 中一起修復。

Yes, this PR mainly fixes the issue where the `color` state doesn't change when modified in `onMounted`. As for the other issue, if it's not the expected behavior, we can fix it in this PR as well.

ilyaliao:
All done.

EvgenyWas (approved):

packages/core/useCssVar/index.ts
```diff
@@ -55,12 +55,12 @@ export function useCssVar(
         old[0].style.removeProperty(old[1])
       updateCssVar()
     },
-    { immediate: true },
+    { flush: 'post' },
```
OrbisK, line 58: dont we need both (post and immediate)? And if not why? 😄

ilyaliao: I think we don't need `immediate` because lines `54-55` are used to clean up the `prop` from the previous element. 

https://github.com/vueuse/vueuse/blob/0ec82c0b1737d008c6971c0b83f6320bbb76ad88/packages/core/useCssVar/index.ts#L54-L55

Initially, both old[0] and old[1] will be undefined, so `immediate: true` is unnecessary.

The reason we need `flush: 'post'` is because at line 39 `window.getComputedStyle(el).getPropertyValue(key)`

https://github.com/vueuse/vueuse/blob/0ec82c0b1737d008c6971c0b83f6320bbb76ad88/packages/core/useCssVar/index.ts#L39

if you change the `color` value at the `onMounded`, since the `color` on the `element` hasn't **really changed yet**, we won't get the **correct value** here. Therefore, we need to wait for `Vue` to update before we get the state from the element.

OrbisK: > both old[0] and old[1] will be empty, so immediate: true is unnecessary.

what about `updateCssVar()` to get the initial value.

ilyaliao: Sorry, I missed this.

I believe it's also not needed because `variable` will always equal `initialValue` at the initial. It only needs to call `updateCssVar` when subsequent changes occur.

Perhaps I'm wrong.

OrbisK: > Sorry, I missed this.
> 
> I believe it's also not needed because `variable` will always equal `initialValue` at the initial. It only needs to call `updateCssVar` when `variable` changes occur.
> 
> Perhaps I'm wrong.

I think if we dont do this immediate we might miss that one, no?

https://github.com/vueuse/vueuse/blob/0ec82c0b1737d008c6971c0b83f6320bbb76ad88/packages/core/useCssVar/index.ts#L38-L41

ilyaliao: > > Sorry, I missed this.
> > I believe it's also not needed because `variable` will always equal `initialValue` at the initial. It only needs to call `updateCssVar` when `variable` changes occur.
> > Perhaps I'm wrong.
> 
> I think if we dont do this immediate we might miss that one, no?
> 
> https://github.com/vueuse/vueuse/blob/0ec82c0b1737d008c6971c0b83f6320bbb76ad88/packages/core/useCssVar/index.ts#L38-L41

However, we have already set variable to initialValue at:

https://github.com/vueuse/vueuse/blob/0ec82c0b1737d008c6971c0b83f6320bbb76ad88/packages/core/useCssVar/index.ts#L32

In fact, in my testing, whether `immediate` is `true` or `false` doesn't affect the actual result. ofc, we can add `immediate` back first in case I missed something.

OrbisK: I think this should work:

```ts
  it('should have existing value', async () => {
    const vm = mount(defineComponent({
      setup() {
        const el = ref()

        const color = '--color'
        const variable = useCssVar(color, el)

        return {
          el,
          variable,
        }
      },
      template: `
        <div ref="el" style="--color: red"></div>
      `
    }))

    await nextTick()
    expect(vm.variable).toBe('red')
    expect(window.getComputedStyle(vm.el).getPropertyValue("--color")).toBe('red')
    expect(vm.el.style.getPropertyValue('--color')).toBe('red')
  })
  ```
  
  But it does'nt. `immediate` does not change the test result here. 
  
  this one does remove the `--color` from the element.
  
  https://github.com/vueuse/vueuse/blob/0ec82c0b1737d008c6971c0b83f6320bbb76ad88/packages/core/useCssVar/index.ts#L61-L72

Comment (ilyaliao) [reply packages/core/useCssVar/index.ts:58]: Oh...It seems quite tricky, thank you for noticing.

I need some time to study it.

**Comment by ilyaliao** [comment]
I believe I have completed the task. I made some minor modifications, which are smaller than before but still resolved the issue. Additionally, I added more tests to ensure the original behavior remains intact. 

Please lmk if I missed anything :)

File: packages/core/useCssVar/index.ts [hunk 2]
```diff
@@ -55,12 +55,11 @@ export function useCssVar(
...
       updateCssVar()
     },
-    { immediate: true },
   )
 
   watch(
-    variable,
-    (val) => {
+    [variable, elRef],
+    ([val]) => {
```
OrbisK, line 62: Maybe not such a big difference, but I think we should add the new element here as well?
```suggestion
    [variable, elRef],
    ([val, el]) => {
```

I can't suggest the full function because it's not considered a change. So the function should use el too. 😄 what do you think?

ilyaliao: Sure, no problem, I think it's great.

OrbisK (approved):

OrbisK:
@ilyaliao isnt this a fix of the initial value instead of new behavior?

ilyaliao:
> @ilyaliao isnt this a fix of the initial value instead of new behavior?

At first it was like this, but later I also `added`/`modified` some other features.

Issue #4089, opened 2024-07-10
useMagicKeys needs the keys to be pressed in order (sometimes)


joelearn:
### Describe the bug

When setting up a keyboard shortcut like Ctrl + Shift + Period, the ref returned by useMagicKeys only responds the first time when the keys are pressed in that exact sequence.

In the reproduction link, try holding down the keys in this order: Shift + Ctrl + Period. The ref remains `false`. However, if you do that exact order again it turns `true`, and from then on, any order of the keys can be held to make the ref `true`.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-z1ss3e?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: Linux 5.10 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i5-7440HQ CPU @ 2.80GHz
    Memory: 8.86 GB / 12.38 GB
    Container: Yes
    Shell: 3.1.0 - /usr/bin/fish
  Binaries:
    Node: 20.15.0 - ~/.nvm/versions/node/v20.15.0/bin/node
    npm: 10.7.0 - ~/.nvm/versions/node/v20.15.0/bin/npm
    pnpm: 9.5.0 - ~/.local/share/pnpm/pnpm
  npmPackages:
    @vueuse/core: ^10.11.0 => 10.11.0
    vue: ^3.4.31 => 3.4.31
```

### Used Package Manager

pnpm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

weicheng2138:
[...]

PR #4505, opened 2025-01-17
fix(useMagicKeys): fix key order issue on first use

babu-ch:

<details>
<br>

>

</details>

---

### Description

fix https://github.com/vueuse/vueuse/issues/4089

### DIFF
--- a/packages/core/useMagicKeys/index.ts
+++ b/packages/core/useMagicKeys/index.ts
@@ -165,7 +165,7 @@ export function useMagicKeys(options: UseMagicKeysOptions<boolean> = {}): any {
         if (!(prop in refs)) {
           if (/[+_-]/.test(prop)) {
             const keys = prop.split(/[+_-]/g).map(i => i.trim())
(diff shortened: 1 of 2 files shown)


packages/core/useMagicKeys/index.ts
```diff
@@ -165,7 +165,7 @@ export function useMagicKeys(options: UseMagicKeysOptions<boolean> = {}): any {
         if (!(prop in refs)) {
           if (/[+_-]/.test(prop)) {
             const keys = prop.split(/[+_-]/g).map(i => i.trim())
-            refs[prop] = computed(() => keys.every(key => toValue(proxy[key])))
+            refs[prop] = computed(() => keys.map(key => toValue(proxy[key])).every(Boolean))
```
babu-ch, line 168: The first time, shift is not registered in proxy[key], so pressing the key from shift will return false.
If ctrl, shift, period is in the order, every will be tracked in order and become true.
By tracking everything with map, the order is ignored even for the first time.

OrbisK (approved):

EvgenyWas (approved):

PR #4511, opened 2025-01-19
docs(useTitle): update tip for auto import in Nuxt 3

### LINKED ISSUE #4487: BUG | `useTitle` | Nuxt 3 does not provide a composable named `useTitle()`
### Describe the bug

The documentation of [useTitle()](https://vueuse.org/core/useTitle/#usetitle) state that:
> When using with Nuxt 3, this function will NOT be auto imported in favor of Nuxt's built-in useTitle(). Use explicit import if you want to use the function from VueUse.

Nuxt 3 doesn't not provide a composable named useTitle(), neither does unjs/unhead which Nuxt 3 used under the hood to manage the head tag

### Reproduction

.

### System Info

```Shell
System:
    OS: macOS 15.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 478.20 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.18.0 - ~/.nvm/versions/node/v20.18.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.18.0/bin/npm
  Browsers:
    Chrome: 131.0.6778.265
    Chrome Canary: 134.0.6953.0
    Edge: 118.0.2088.61
    Safari: 18.1
    Safari Technology Preview: 18.2
  npmPackages:
    @vueuse/components: ^12.3.0 => 12.3.0
    @vueuse/core: ^12.0.0 => 12.3.0
    @vueuse/integrations: ^12.3.0 => 12.3.0
    @vueuse/nuxt: ^12.3.0 => 12.3.0
    vue: latest => 3.5.13
```

### Used Package Manager

npm

### Validations

EvgenyWas:

<details>
<br>

>

</details>

---

### Description

Resolves #4487

### DIFF
--- a/packages/core/useTitle/index.md
+++ b/packages/core/useTitle/index.md
@@ -7,7 +7,7 @@ category: Browser
 Reactive document title.
 
 ::: tip
-When using with Nuxt 3, this function will **NOT** be auto imported in favor of Nuxt's built-in `useTitle()`.
+This composable isn't compatible with SSR. When using with Nuxt 3, this function will **NOT** be auto imported in favor of Nuxt's built-in composables such as `useHead()` and `useSeoMeta()`.
 Use explicit import if you want to use the function from VueUse.
 :::
 
--- a/packages/core/useTitle/index.md
+++ b/packages/core/useTitle/index.md
@@ -6,9 +6,8 @@ category: Browser
 
 Reactive document title.
 
-::: tip
-This composable isn't compatible with SSR. When using with Nuxt 3, this function will **NOT** be auto imported in favor of Nuxt's built-in composables such as `useHead()` and `useSeoMeta()`.
-Use explicit import if you want to use the function from VueUse.
+::: warning
+This composable isn't compatible with SSR.
 :::
 
 ## Usage
--- a/packages/core/useTitle/index.ts
+++ b/packages/core/useTitle/index.ts
@@ -42,6 +42,7 @@ export type UseTitleOptions = ConfigurableDocument & UseTitleOptionsBase
  * @see https://vueuse.org/useTitle
  * @param newTitle
  * @param options
+ * @description It's not SSR compatible. Your value will be applied only on client-side.
  */
 export function useTitle(
   newTitle: ReadonlyRefOrGetter<string | null | undefined>,
--- a/packages/nuxt/index.ts
+++ b/packages/nuxt/index.ts
@@ -17,7 +17,6 @@ const disabledFunctions = [
   'useFetch',
   'useCookie',
   'useHead',
-  'useTitle',
   'useStorage',
   'useImage',
 ]


OrbisK (dismissed):

packages/core/useTitle/index.md
```diff
@@ -7,7 +7,7 @@ category: Browser
 Reactive document title.
 
 ::: tip
```
ilyaliao, line 9: I'm just curious, would it be better if we use `Warning` or `Note`?

ilyaliao, line 9: I'm just curious, would it be better if we use `Warning` or `Note`?

ilyaliao, line 9: I'm just curious, would it be better if we use `Warning` or `Note`?

EvgenyWas: For me, it looks more like a tip except for this sentence - `This composable isn't compatible with SSR.`. If you think it's better I can move just that to a warning.

ilyaliao: Sorry, I deleted it by mistake.

I think it was originally "tip", so it should be fine :)

ilyaliao:
Sorry, I deleted it by mistake.

I think it was originally "tip", so it should be fine :)

ilyaliao (approved):

packages/core/useTitle/index.md
```diff
@@ -7,7 +7,7 @@ category: Browser
...
-When using with Nuxt 3, this function will **NOT** be auto imported in favor of Nuxt's built-in `useTitle()`.
+This composable isn't compatible with SSR. When using with Nuxt 3, this function will **NOT** be auto imported in favor of Nuxt's built-in composables such as `useHead()` and `useSeoMeta()`.
```
huang-julien, line 10: >  When using with Nuxt 3, this function will **NOT** be auto imported in favor of Nuxt's built-in composables such as `useHead()` and `useSeoMeta()`.

I think we hsould remove this. `useTitle` allows to retrieve the current title in any component.
`useHead` and `useSeoMeta` does not allow it. While it returns the `ActiveHead` instance, it does not return to the final and resolved value of the title in the head.

huang-julien (changes requested):

packages/core/useTitle/index.md
EvgenyWas: Do you suggest to add auto import for this composable in Nuxt?
Can you please explain what you mean it can be used in any component, but useHead not?

huang-julien: We can add it back. But we need to tell users that it is not SSR compatible in the JSDoc.

If you call `useHead({ title: 'hello' })`, it will return an `ActiveHeadEntry`. This entry is not injected into child components. 
Even if you provide this entry, this modification doesn't not reflect the final state your `Head` instance because other components can also call `useHead` -> this entry only contains the modification of the `useHead` call.

You could call `getActiveHead` from `unhead` but that also wouldn't work because `unhead` resolves asynchronously all tags and you would only get the state of the `Head` when the component is being mounted, not when everything has been rendered in SSR.

EvgenyWas: Thank you for the wide explanation, @huang-julien! I got the difference :)
I updated the documentation, added a description to JSDoc, and removed the disabled auto import in Nuxt for `useTitle`.

huang-julien (approved):

OrbisK (approved):

ilyaliao (approved):

PR #4534, opened 2025-01-24
fix(useBreakpoints,useMediaQueries)!: don't use the SSR width if app hydration has already been done when calling the composable

Tofandel:
### Description

An issue regarding the reactivity of the shortcut breakpoints was mentionned in https://github.com/vueuse/vueuse/pull/4317#issuecomment-2610854689

Repro: https://stackblitz.com/edit/nuxt-starter-cnydzchs?file=app.vue

### Additional context

This led me to see that SSR width behavior is still used everytime a component mounts (v-if/v-else) even if the app has already been hydrated. In those cases we should skip SSR support entirely and go straight to normal behavior


packages/core/useBreakpoints/index.ts
```diff
@@ -46,27 +48,35 @@ export function useBreakpoints<K extends string>(
 
   const { window = defaultWindow, strategy = 'min-width', ssrWidth = useSSRWidth() } = options
 
-  const ssrSupport = typeof ssrWidth === 'number'
-  const mounted = ssrSupport ? ref(false) : { value: true }
-  if (ssrSupport) {
-    tryOnMounted(() => mounted.value = !!window)
+  const ssrSupport = ref(getCurrentInstance()?.root.isMounted ? typeof ssrWidth === 'number' : false)
```
xak2000, line 51: Probably I don't fully understand the intention but shouldn't this be inverted?

```
const ssrSupport = ref(getCurrentInstance()?.root.isMounted ? false : typeof ssrWidth === 'number')
```

`getCurrentInstance()?.root.isMounted` means that hydration is finished, right? In this case `ssrSupport` should be disabled, shoudn't it?

packages/core/useBreakpoints/index.ts
```diff
@@ -46,27 +48,35 @@ export function useBreakpoints<K extends string>(
...
+  if (ssrSupport.value) {
+    tryOnMounted(() => ssrSupport.value = !!window)
```
xak2000, line 53: Is the intention to disable `ssrSupport` after mount is completed? Shouldn't this be `ssrSupport.value = !window` then?

Tofandel: Indeed, good catch

packages/core/useBreakpoints/index.ts
Tofandel: Yep

Daverine:
Since `window.matchMedia` is a client-only feature, it should only be executed during the mounted lifecycle of a component. Instead of using the `ssrWidth` option, I recommend introducing a `matchOnSSR` option, which should be a Boolean and defaults to false. This `matchOnSSR` option will determine the return value of `useMediaQuery` when operating in server-side rendering (SSR) mode. Once the mounted lifecycle is triggered, the return value will rely on `window.matchMedia` as intended.

Tofandel:
> Since window.matchMedia is a client-only feature

A *rather simple* mock is implemented server side with ssrWidth (that should handle 99% of cases for now)

matchOnSSR would just introduce confusion as a media query can represent a lot of ranges and can be negated, just saying "disable everything" means that you could possibly not be rendering anything or rendering too many things if you had overlapping queries

While the ssrWidth provides a way to say "Ok render everything on the server as it should be on mobile" or "render as desktop"

It can even be plugged with experimental headers to detect the client width and put that dynamically into ssrWidth to avoid layout shifts

We of course just need this PR to be merged to get the issue fixed

xak2000:
> It can even be plugged with experimental headers to detect the client width and put that dynamically into ssrWidth to avoid layout shifts

Oh yeah, I'm really waiting for this to happen. With current implementation whatever default width I choose (mobile or desktop), other half of the users will be unhappy with elements jumping on screen. :)

I hope these experimental headers will be implented by most browsers "soon" (c) and we all be happier. 😸

Tofandel:
Drafting because it turns out that suspense is messing up root.isMounted

I'll see to add a new tested composable to check both of those
```
getCurrentInstance().root.isMounted
getCurrentInstance().root.subTree.suspense.isHydrating
```

fredericDelaporte:
This trouble causes `useBreakpoints` to be tricky to use with Nuxt. What can actually be used on it and how? 

 - ✔️Destructured refs (`const { sm, lg } = useBreakpoint(breakpointsList, { ssrWidth: 1280 })`)
 - ❓Methods as `current` and the like which return a ref: should they be called only in setup for storing the ref, for them to work properly? (It seems to me a good practice anyway.)
 - ❓Methods as `isGreaterOrEqual` and the like which does not return a ref: should they be destructured too for working properly?
 - ❌ Ref properties (non-destructured) (`breakpoints.lg` out of setup code).

It also forces to put comments at each usage reminding why it must stayed destructured, to avoid any later change to cause bugs. It is non-obvious that its properties must be destructured in the setup for them to work properly.

Tofandel:
I just checked again and I was somewhat blocked by vue's internal API not exposing any flag on the appContext if you use createApp or createSSRApp, so I can fix the issue at hand but there is no clean way to distinguish an app that is server side rendering and then hydrating vs an app with no SSR support where there would be no hydration and so no need to run any checks.

We'd probably need to post an SFC in vue to have a `isSSR` flag on the `appContext`

eslint.config.js
```diff
@@ -15,10 +15,6 @@ const restricted = [
   '../..',
   resolve(dir, 'packages/core/index.ts'),
   resolve(dir, 'packages/shared/index.ts'),
-  {
```
Tofandel, line 18: I could not see a good reason why those imports were restricted so I removed them. From what I see it used to be restricted to force usage of the helpers from @vueuse/core, but unref doesn't exist anymore and tryOnMounted and onMounted don't have the same semantics so onMounted is a valid usage in a lot of cases

Tofandel:
I found a reliable way to detect if SSR using `onServerPrefetch()` and if hydrating using `instance.vnode.el && !instance.isMounted`

The isSSR flag would have allowed to avoid the onServerPrefetch which currently means 2 render steps on the server (first the render function then a last render after onServerPrefetch)

PR #4535, opened 2025-01-24
feat(createReusableTemplate): explicit props

antfu:
Before this PR, `createReusableTemplate` always treated all props as attributes and made the props appear on the DOM. In some cases, it might not be desired. This PR allows explicit provide the props to prevent those props to be leaking on the DOM.


OrbisK (approved):

ilyaliao (approved):

benjamincanac:
Thanks @antfu! For context, I use this composable a lot in Nuxt UI v3 but get all props passed to the template in the DOM: https://github.com/nuxt/ui/blob/v3/test/components/__snapshots__/NavigationMenu.spec.ts.snap#L7

PR #4104, opened 2024-07-18
feat(core): adding a custom event name to `onClickOutside`

HydenLiu:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description
I adding a custom event name to `onClickOutside`

### Additional context
fixes #3911

Issue #3911, opened 2024-04-05
onClickOutside: Option to use mousedown event


anjoseb121:
### Clear and concise description of the problem

As a developer using `onClickOutside` function I want trigger the listener earlier. Common actions such as clicking outside to close, are being prevented from happening when you press down outside to perform another action, such as resizing or drag and drop. 
I'm willing to submit a PR for the issue.

### Suggested solution

In `OnClickOutsideOptions` we could have an option to call the listener on the `click` or `mousedown` event.

### Alternative

Having an `onMouseDownOutside` function.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

fabkho:
bump

bendemartin97:
🚀🚀

kevinluo201:
Hello, I'd like to ask, is using modifier to add mousedown event handler a good idea? (Directive Usage), e.g.
```javascript
v-on-click-outside.mousedown="closeSomething"
```
So the users know this is only for mousedown event

PR #4537, opened 2025-01-25
feat(onClickOutside): add `controls`

sibbng:

<details>
<br>

>

</details>

---

### Description

Gives end users more flexibility over triggering the listener. Technically unlocks fixing issues like #4104 and #3911 on end-user side. IMO after merging this PR we should close this issues and let end-users to decide what to do in their specific situation. I shared my thoughts on these PRs why they are not mergeable and clash with the goals of onClickOutside. This is like a never ending chase game. It's better end-users to handle situation base on their needs.

### DIFF
--- a/packages/core/onClickOutside/index.md
+++ b/packages/core/onClickOutside/index.md
@@ -29,7 +29,7 @@ onClickOutside(target, event => console.log(event))
 If you need more control over triggering the handler, you can use the `controls` option.
 
 ```ts
-const { cancel, fire } = onClickOutside(
+const { cancel, trigger } = onClickOutside(
   modalRef,
   (event) => {
     modal.value = false
@@ -40,7 +40,7 @@ const { cancel, fire } = onClickOutside(
 useEventListener('pointermove', (e) => {
   cancel()
   // or
-  fire(e)
+  trigger(e)
 })
 ```
 
--- a/packages/core/onClickOutside/index.md
+++ b/packages/core/onClickOutside/index.md
@@ -26,21 +26,21 @@ onClickOutside(target, event => console.log(event))
 </template>
 ```
 
-If you need more control over triggering the handler, you can use the `controller` option.
+If you need more control over triggering the handler, you can use the `controls` option.
 
 ```ts
-const controller = onClickOutside(
+const { cancel, fire } = onClickOutside(
   modalRef,
   (event) => {
     modal.value = false
   },
-  { controller: true },
+  { controls: true },
 )
 
 useEventListener('pointermove', (e) => {
-  controller.cancel()
+  cancel()
   // or
-  controller.fire(e)
+  fire(e)
 })
 ```
 
--- a/packages/core/onClickOutside/index.md
+++ b/packages/core/onClickOutside/index.md
@@ -32,7 +32,6 @@ If you need more control over triggering the handler, you can use the `controlle
 const controller = onClickOutside(
   modalRef,
   (event) => {
-    console.log(event)
     modal.value = false
   },
   { controller: true },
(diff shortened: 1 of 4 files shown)


packages/core/onClickOutside/index.ts
```diff
@@ -146,5 +176,16 @@ export function onClickOutside<T extends OnClickOutsideOptions>(
...
+      stop,
+      cancel: () => { shouldListen = false },
+      fire: (event: Event) => {
+        shouldListen = true
+        listener(event)
+      },
+    }
+  }
+
   return stop
```
OrbisK, line 190: IMO the long-term return of the controller should be the default here.

Maybe we can do a backwards-comparable solution for now. Not sure if this is considered hacky or not:

What if we return something like this:

```js
const useXY = ()=>{
  const controller /* is actually stop function */  = ()=>{
    // stop logic
  }
  controller.cancel = ()=>{
    shouldListen = false 
  }
  controller.stop = ()=>{
    // stop logic
  }
  controller.fire = (event)=>{
    shouldListen = true
    listener(event)
  }
  
  return controller
}
```

This should allow
```ts
const stop = useXY()
stop()

// AND
const {stop, fire, cancel} = useXY()
```

This means that we do not need an extra option.

OrbisK: > it feels like anti-pattern and look ugly

I agree with that. But it would ensure that a breaking change can also be a little softer.

> If we decide to pursue controller approach, in next major, we can drop overloads, default to Event (or even `Event | undefined`) and significantly simplify types.

When we do, we might consider making sure that all composables return an object. That way we are consistent and future proof.

OrbisK (dismissed):

packages/core/onClickOutside/index.ts
OrbisK: > I considered this option but it feels like anti-pattern and look ugly. And more importantly this is a breaking change for types. Without controller, `handler` function's first argument receives `PointerEvent` type. When controller is enabled it becomes Event. Similar to when `detectIframe` enabled, it's `FocusEvent | PointerEvent`.

Just for reference, the vue 3.5 watcher does exactly this pattern: https://github.com/vuejs/core/blob/7ecd2a22c799d85b0c9857e48d97de5755197e87/packages/reactivity/src/watch.ts#L324-L328. Not to say it is not an anti-pattern.

Partially related: https://github.com/vueuse/vueuse/discussions/4548#discussioncomment-11985061

OrbisK (dismissed):

packages/core/onClickOutside/index.md
```diff
@@ -26,6 +26,24 @@ onClickOutside(target, event => console.log(event))
...
+    modal.value = false
+  },
+  { controller: true },
+)
+
+useEventListener('pointermove', (e) => {
+  controller.cancel()
+  // or
+  controller.fire(e)
+})
```
antfu, line 44:
```suggestion
const { cancel, fire } = onClickOutside(
  modalRef,
  (event) => {
    modal.value = false
  },
  { controls: true },
)

useEventListener('pointermove', (e) => {
  cancel()
  // or
  fire(e)
})
```

packages/core/onClickOutside/index.md
```diff
@@ -26,6 +26,24 @@ onClickOutside(target, event => console.log(event))
 </template>
 ```
 
+If you need more control over triggering the handler, you can use the `controller` option.
```
antfu, line 29:
```suggestion
If you need more control over triggering the handler, you can use the `controls` option.
```

antfu (approved):

antfu (approved):

PR #4554, opened 2025-01-30
fix: useCountdown `start()` can provide a custom initial value

ralacerda:

<details>
<br>

>

</details>

---

### Description

`useCountdown` return type includes the `start()` function with the following signature:
```ts
start: (initialCountdown?: MaybeRefOrGetter<number>) => void
```

But the implementation before this PR does not use the value.
This PR makes so `start()` will actually set a custom value for `initialCountdown` (only that time).

### Additional context

I'm using `let` but not `ref`. I assume this would consume less memory and be faster, but I'm not sure if there is a drawback.

Current for behaviour for:
```ts
const { start, reset, resume } = useCountdown(5)

start(2)
reset()
```
is going back to the original 5 seconds countdown, but I'm not sure if this is the expected behaviour.

I included a simple test that check if everything is working as expected (setting a custom inicial value with `start(X)` and then checking if it goes back to the original value with `start()`)

There might be other edge cases with `stop()` or `resume()`, so maybe more tests are required.

I also included a test to check if the `interval` used in the composable can be a ref, and if changing the ref also changes the initial countdown when starting it again. (This passes even without the changes in this PR).

### DIFF
--- a/packages/core/useCountdown/index.test.ts
+++ b/packages/core/useCountdown/index.test.ts
@@ -123,7 +123,7 @@ describe('useCountdown', () => {
   })
 
   it('start can provide a custom interval', async () => {
-    const { start } = useCountdown(countdown, options)
+    const { start, reset } = useCountdown(countdown, options)
     vi.advanceTimersByTime(countdown * interval + 10)
     expect(completeCallback).toHaveBeenCalledTimes(1)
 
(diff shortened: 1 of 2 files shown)


ralacerda:
Yeah, that would be even better.
I'll also update the documentation to make it clear that `reset()` and `start()` will always reset back to the original value/ref.

ralacerda:
Okay, let me know what you think.

OrbisK (approved):

ilyaliao (approved):

43081j (approved):

OrbisK (approved):

PR #4555, opened 2025-01-31
fix(createEventHook): type check for multiple arguments

isimehmeti:

<details>
<br>

>

</details>

---

### Description

This PR improves the typing of `createEventHook`  to correctly support multiple arguments when used with tuple types. Previously, tuple types were treated as a single argument instead of separate parameters.

**Before (Issue)**
```js
const eventHook = createEventHook<[string, number]>();
eventHook.on((foo, bar) => {  // ❌ Incorrect: foo is [string, number], bar is unknown
  console.log(foo, bar);
});
```

- The foo parameter was inferred as [string, number] instead of string, and bar was inferred as unknown.

**After (Fix)**
```js
const eventHook = createEventHook<[string, number]>();
eventHook.on((foo, bar) => {  // ✅ Correct: foo is string, bar is number
  console.log(foo, bar);
});
```

- Now, multiple parameters are properly inferred when passing a tuple type.

### Additional context

-The fix introduces an extra condition:

```js
[T] extends [any[]] ? (...param: T) => void 
```
This ensures that tuples are spread as separate arguments instead of being treated as a single parameter.

- The change is backward-compatible and does not affect single-parameter usage.

### DIFF
--- a/packages/core/useFileDialog/index.ts
+++ b/packages/core/useFileDialog/index.ts
@@ -49,7 +49,7 @@ export interface UseFileDialogReturn {
   files: Ref<FileList | null>
   open: (localOptions?: Partial<UseFileDialogOptions>) => void
   reset: () => void
-  onChange: EventHookOn<[FileList | null]>
+  onChange: EventHookOn<FileList | null>
   onCancel: EventHookOn
 }
 
--- a/packages/core/useFileDialog/index.ts
+++ b/packages/core/useFileDialog/index.ts
@@ -49,7 +49,7 @@ export interface UseFileDialogReturn {
   files: Ref<FileList | null>
   open: (localOptions?: Partial<UseFileDialogOptions>) => void
   reset: () => void
-  onChange: EventHookOn<FileList | null>
+  onChange: EventHookOn<[FileList | null]>
   onCancel: EventHookOn
 }
 
--- a/packages/core/useCountdown/index.md
+++ b/packages/core/useCountdown/index.md
@@ -21,3 +21,27 @@ const { remaining, start, stop, pause, resume } = useCountdown(countdownSeconds,
   }
 })
 ```
+
+You can use a `ref` to change the initial countdown.
+`start()` and `resume()` also accept a new countdown value for the next countdown.
+
+```js
+import { ref } from 'vue'
+import { useCountdown } from '@vueuse/core'
+
+const countdown = ref(5)
+const { start, reset } = useCountdown(countdown, {
+})
+
+// change the countdown value
+countdown.value = 10
+
+// start a new countdown with 2 seconds
+start(2)
+
+// reset the countdown to 4, but do not start it
+reset(4)
+
+// start the countdown with the current value of `countdown`
+start()
+```
--- a/packages/core/useCountdown/index.test.ts
+++ b/packages/core/useCountdown/index.test.ts
@@ -1,7 +1,7 @@
 import type { Pausable } from '@vueuse/shared'
 import type { UseCountdownOptions } from '.'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
-import { effectScope } from 'vue'
+import { effectScope, ref } from 'vue'
 import { useCountdown } from '.'
 
 describe('useCountdown', () => {
@@ -109,5 +109,37 @@ describe('useCountdown', () => {
     expect(isActive.value).toBeFalsy()
     vi.advanceTimersByTime(60)
     expect(tickCallback).toHaveBeenCalledTimes(0)
+  })
+
+  it('initial interval can be changed', async () => {
+    const countdown = ref(3)
+
+    const { start } = useCountdown(countdown, { ...options, immediate: false })
+
+    countdown.value = 2
+    start()
+    vi.advanceTimersByTime(210)
+    expect(completeCallback).toHaveBeenCalledTimes(1)
+  })
+
+  it('start can provide a custom interval', async () => {
+    const { start, reset } = useCountdown(countdown, options)
+    vi.advanceTimersByTime(countdown * interval + 10)
+    expect(completeCallback).toHaveBeenCalledTimes(1)
+
+    start(1)
+    vi.advanceTimersByTime(110)
+    expect(completeCallback).toHaveBeenCalledTimes(2)
+
+    start()
+    vi.advanceTimersByTime(110)
+    expect(completeCallback).toHaveBeenCalledTimes(2)
+    vi.advanceTimersByTime(countdown * interval + 10)
+    expect(completeCallback).toHaveBeenCalledTimes(3)
+
+    start(1)
+    reset()
+    vi.advanceTimersByTime(110)
+    expect(completeCallback).toHaveBeenCalledTimes(3)
   })
 })
--- a/packages/core/useCountdown/index.ts
+++ b/packages/core/useCountdown/index.ts
@@ -32,7 +32,7 @@ export interface UseCountdownReturn extends Pausable {
   /**
    * Resets the countdown and repeatsLeft to their initial values.
    */
-  reset: () => void
+  reset: (countdown?: MaybeRefOrGetter<number>) => void
   /**
    * Stops the countdown and resets its state.
    */
@@ -40,7 +40,7 @@ export interface UseCountdownReturn extends Pausable {
   /**
    * Reset the countdown and start it again.
    */
-  start: (initialCountdown?: MaybeRefOrGetter<number>) => void
+  start: (countdown?: MaybeRefOrGetter<number>) => void
 }
 
 /**
@@ -64,8 +64,8 @@ export function useCountdown(initialCountdown: MaybeRefOrGetter<number>, options
     }
   }, options?.interval ?? 1000, { immediate: options?.immediate ?? false })
 
-  const reset = () => {
-    remaining.value = toValue(initialCountdown)
+  const reset = (countdown?: MaybeRefOrGetter<number>) => {
+    remaining.value = toValue(countdown) ?? toValue(initialCountdown)
   }
 
   const stop = () => {
@@ -81,8 +81,8 @@ export function useCountdown(initialCountdown: MaybeRefOrGetter<number>, options
     }
   }
 
-  const start = () => {
-    reset()
+  const start = (countdown?: MaybeRefOrGetter<number>) => {
+    reset(countdown)
     intervalController.resume()
   }
 
--- a/packages/integrations/useDrauu/index.ts
+++ b/packages/integrations/useDrauu/index.ts
@@ -45,7 +45,7 @@ export function useDrauu(
 
   const onChangedHook = createEventHook<void>()
   const onCanceledHook = createEventHook<void>()
-  const onCommittedHook = createEventHook<[SVGElement | undefined]>()
+  const onCommittedHook = createEventHook<SVGElement | undefined>()
   const onStartHook = createEventHook<void>()
   const onEndHook = createEventHook<void>()
   const canUndo = ref(false)
--- a/packages/integrations/useDrauu/index.ts
+++ b/packages/integrations/useDrauu/index.ts
@@ -45,7 +45,7 @@ export function useDrauu(
 
   const onChangedHook = createEventHook<void>()
   const onCanceledHook = createEventHook<void>()
-  const onCommittedHook = createEventHook<SVGElement | undefined>()
+  const onCommittedHook = createEventHook<[SVGElement | undefined]>()
   const onStartHook = createEventHook<void>()
   const onEndHook = createEventHook<void>()
   const canUndo = ref(false)
--- a/packages/shared/createEventHook/index.test.ts
+++ b/packages/shared/createEventHook/index.test.ts
@@ -139,7 +139,7 @@ describe('createEventHook', () => {
   it('should pass union type', () => {
     let count = 0
 
-    const { on: onResult, trigger } = createEventHook<number | string>()
+    const { on: onResult, trigger } = createEventHook<[number | string]>()
 
     // union type should be inferred
     onResult(value => count = 2)
--- a/packages/shared/createEventHook/index.test.ts
+++ b/packages/shared/createEventHook/index.test.ts
@@ -139,7 +139,7 @@ describe('createEventHook', () => {
   it('should pass union type', () => {
     let count = 0
 
-    const { on: onResult, trigger } = createEventHook<[number | string]>()
+    const { on: onResult, trigger } = createEventHook<number | string>()
 
     // union type should be inferred
     onResult(value => count = 2)
--- a/packages/shared/createEventHook/index.ts
+++ b/packages/shared/createEventHook/index.ts
@@ -13,8 +13,8 @@ type Callback<T> = IsAny<T> extends true
       [T] extends [void]
         ? (...param: unknown[]) => void
         : [T] extends [any[]]
-          ? (...param: T) => void
-          : (...param: [T, ...unknown[]]) => void
+            ? (...param: T) => void
+            : (...param: [T, ...unknown[]]) => void
     )
 
 export type EventHookOn<T = any> = (fn: Callback<T>) => { off: () => void }
--- a/packages/shared/createEventHook/index.ts
+++ b/packages/shared/createEventHook/index.ts
@@ -8,13 +8,13 @@ import { tryOnScopeDispose } from '../tryOnScopeDispose'
 // any extends void = true
 // so we need to check if T is any first
 type Callback<T> = IsAny<T> extends true
-  ? (...param: any) => void 
+  ? (...param: any) => void
   : (
       [T] extends [void]
         ? (...param: unknown[]) => void
-        : T extends any[] 
-        ? (...param: T) => void 
-        : (...param: [T, ...unknown[]]) => void
+        : T extends any[]
+          ? (...param: T) => void
+          : (...param: [T, ...unknown[]]) => void
     )
 
 export type EventHookOn<T = any> = (fn: Callback<T>) => { off: () => void }
--- a/packages/shared/createEventHook/index.ts
+++ b/packages/shared/createEventHook/index.ts
@@ -12,7 +12,7 @@ type Callback<T> = IsAny<T> extends true
   : (
       [T] extends [void]
         ? (...param: unknown[]) => void
-        : T extends any[]
+        : [T] extends [any[]]
           ? (...param: T) => void
           : (...param: [T, ...unknown[]]) => void
     )
--- a/packages/shared/createEventHook/index.ts
+++ b/packages/shared/createEventHook/index.ts
@@ -8,16 +8,18 @@ import { tryOnScopeDispose } from '../tryOnScopeDispose'
 // any extends void = true
 // so we need to check if T is any first
 type Callback<T> = IsAny<T> extends true
-  ? (...param: any) => void
+  ? (...param: any) => void 
   : (
       [T] extends [void]
         ? (...param: unknown[]) => void
+        : T extends any[] 
+        ? (...param: T) => void 
         : (...param: [T, ...unknown[]]) => void
     )
 
 export type EventHookOn<T = any> = (fn: Callback<T>) => { off: () => void }
 export type EventHookOff<T = any> = (fn: Callback<T>) => void
-export type EventHookTrigger<T = any> = (...param: IsAny<T> extends true ? unknown[] : [T, ...unknown[]]) => Promise<unknown[]>
+export type EventHookTrigger<T = any> = (...param: Parameters<Callback<T>>) => Promise<unknown[]>
 
 export interface EventHook<T = any> {
   on: EventHookOn<T>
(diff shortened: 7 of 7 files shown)


OrbisK (approved):

EvgenyWas (approved):

ilyaliao (dismissed):

OrbisK (dismissed):

ilyaliao (approved):

ilyaliao:
Should we care about the `workflows`? @OrbisK

OrbisK:
> Should we care about the `workflows`? Or can we merge directly? @OrbisK

image

looks like an typecheck issue (https://app.netlify.com/sites/vueuse/deploys/67a08230f106910008a97622). So we should care.

isimehmeti:
@OrbisK 

the test that fails is this one 

```js
  it('should pass union type', () => {
    let count = 0

    const { on: onResult, trigger } = createEventHook<number | string>()

    // union type should be inferred
    onResult(value => count = 2)
    trigger(1)
    trigger(2)

    expect(count).toBe(2)
  })
  ```
  Adding
  `T extends any[] ? (...param: T) => void `
  
  makes the type more complex so in order to interfere union types it should be declared 
  `createEventHook<[number | string]>()` 
  
  we should find a common ground for this case

**Review by ilyaliao — APPROVED** [review]

File: packages/core/useFileDialog/index.ts [hunk 1]
```diff
@@ -49,7 +49,7 @@ export interface UseFileDialogReturn {
   files: Ref<FileList | null>
   open: (localOptions?: Partial<UseFileDialogOptions>) => void
   reset: () => void
-  onChange: EventHookOn<FileList | null>
+  onChange: EventHookOn<[FileList | null]>
```
ilyaliao, line 52: Here isn't there only one `argument`? Why need to use `[]` to define the `type`?

isimehmeti: because it's a union type and here

```js
type Callback<T> = IsAny<T> extends true
  ? (...param: any) => void
  : (
      [T] extends [void]
        ? (...param: unknown[]) => void
        : T extends any[]
          ? (...param: T) => void
          : (...param: [T, ...unknown[]]) => void
    )
    ```
   adding `T extends any[] ? (...param: T) => void `  makes the type complex so when is interfered it gives type `any`

Comment (OrbisK) [reply packages/core/useFileDialog/index.ts:52]: https://www.typescriptlang.org/play/?#code/C4TwDgpgBAwghgGwQIzgYwNYB4AqA+KAXihyggA9gIA7AEwGco5qQBtAXSgH4oAKAOkFg4AJzgBbegC4SASiIEAbgHsAlrSgzewseJk55hJWtoBuAFCXQkKACEisRCnTZ6wEauoBzPOevQYB3gkVEwsVjcPbwAaKGoAV3FkCBFY5GVlBAhmdgIgA

this should work no?

Comment (isimehmeti) [reply packages/core/useFileDialog/index.ts:52]: for me this is better and simpler. I just wanted to keep the backward capability for like the case `[T] extends [void]
        ? (...param: unknown[]) => void` case

**Comment by isimehmeti** [comment]
@OrbisK  @ilyaliao 

changed the case from `T extends any[] ? (...param: T) => void` to `T extends any[] ? (...param: T) => void`

this keeps the backward capability also does not require `[]` for union types

**Comment by 43081j** [comment]
i wonder if we should simplify slightly and reduce our use of `any`

something like [this playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAkgZgQQHYgDwBUA0UCa2ByAfFALxQAMUEAHsBEgCYDOUAFAIxQBkU6AlFAD8uKAC4o+AFCTQkWE2Rp0xMvEUZswAE4BXCNjgBDADZMIhaQHpLEgKIB1XjgAKtgMrTZ0AMInjAI0MAYwBrDBV5dWUqWnpmKG09SSghNgA6DLBDLUMAW3EdJBCkAHsAdyQAbQBdARJiADcSgEsGZLE2dpSUyvRqmLpGFkqm1uqu7pThVjrGlrbJyfF0AbiWQuLyqvHF3dTWDLSsnPzeWahRhb2ltmO85fPLib5PcGhbBvpgAAkSkpCAPJIDCkKCGFARVhwJDiXzGALBMLKc4AbygJTgcHEM1Ic1aUAAvjI3lAPl9fv8AZiQWRwSBIdDYX5AqFwo95sS5GSkD8-iF0FpmgBzIUQLQ0sEQ0EHTLZe5QZxy3IQOhaJioOEI1nKQjnZxaEq5ZpmVAbUoVGoWSTNHlioxBd6fHkUpGgunEFHtEow0lO3mU4HKL2Y8Tc-2A6lBlLaYWirShv0ugWxsXhdpBYwQbLY9mtSREyTWKAAgAyABEnK4PJyfMzEQDjAxwqCYAoUM2aIN4okIO1poc7qc6bmrtiJj0+qshlARvMdnt+7KTgUiubtiPx91sQOleJethDmatjVariLhzJnwANzSGu+8l85MisUNpsrWlSsgyo67yLt6KdmsCS6NAwhHha-R7lgUCHqux7VKe9QKgaRomuB2wWF497OnyQKvhK7rSoyUCaiygKNmyZ5ohiWJsCOhJ3mGLpUnA+HvpK9JET6pH1hRyJnk81q2lo9qOg+lJ8W6Uqeik3oJuJgJIGxFiySG2HhixyntDGz7xupSaCrpWkpBmWZ6TiSFPAWRboO46DVgwEAZtk0BBN6TDAFAAAWfLyTh-yoJUHmCkgQrYEgOi5P4YrVIQN6Oc5Wiue5nk+RJDB+RpFFBTGoXhZF0VaLFN6SGlIRpN6rBQn82CBFoqLtHAfw3ik1gAHqCNIKR1S1UDtZ1BLXqVuGNhVSBVU1JS1dkDUpJNvX9V1UA9e0i2DSVZVpDpcasAA5MFNpCrt2AAMzkCdQ1la+W2GTt+25Udp3nUNkgJcYLlQG5SAeVATCHZmCBaEKLqZS6qAHaFcWvU571JZ9KW-f9ECA8DI0ZfpaPgw9UOSH9oUA0DLpjVVwWzb92gLZYHX5kNeNCgTqPpcTrCk1R7TBZT1PrbjSMo0T21intENHbTvOE2jN0ploQsPbtL1vR9X0-XSIMYwF7rxTDisIyraOg5jGvSLr-zM5N031Wzc3NatVOde0K2tbbNM3sb5EMKbNXLTNltQPNNvU-b2ScwNQ2u5Lul7WUXmGHQnxaMdUAABzkInocoMxo0C9Lu1RzHEBxwnyepyVCtw0rnmXKrTF8qglxQ6XyXfRX8wZ+j1fpbX8w45XfIeyUDXcz3TOVdV-ds4PLe91nkfR7HYqFynQ1D274d3bnc-x9gRfXkAA)

```ts
type Callback<T> = IsAny<T> extends true
  ? (...param: unknown[]) => void
  : (
      [T] extends [void]
        ? () => void
        : T extends unknown[]
            ? (...param: T) => void
            : (param: T) => void
    )
```

its a weird type at the minute in `main` and in this PR because:

- `Callback<void>` is `(...params: unknown[]) => void` which seems odd
  - who is using this and why?
  - `trigger` is nonsense in main, because the first param is then `void` (i.e. `(param0: void) => void`)
- `Callback<any>` has `...params: any` but should really at least be `unknown[]` IMO
  - though it could cause confusion since the type param is still `any`. people just shouldn't be passing this really
- `Callback<string>` results in `(param0: string, ...params: unknown[])`
  - is this because people couldn't specify multiple params before?
  - seems like this should be `(param0: string)` without the made up rest params, and the caller is responsible for defining any extra params

isimehmeti:
> i wonder if we should simplify slightly and reduce our use of `any`
> 
> something like [this playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAkgZgQQHYgDwBUA0UCa2ByAfFALxQAMUEAHsBEgCYDOUAFAIxQBkU6AlFAD8uKAC4o+AFCTQkWE2Rp0xMvEUZswAE4BXCNjgBDADZMIhaQHpLEgKIB1XjgAKtgMrTZ0AMInjAI0MAYwBrDBV5dWUqWnpmKG09SSghNgA6DLBDLUMAW3EdJBCkAHsAdyQAbQBdARJiADcSgEsGZLE2dpSUyvRqmLpGFkqm1uqu7pThVjrGlrbJyfF0AbiWQuLyqvHF3dTWDLSsnPzeWahRhb2ltmO85fPLib5PcGhbBvpgAAkSkpCAPJIDCkKCGFARVhwJDiXzGALBMLKc4AbygJTgcHEM1Ic1aUAAvjI3lAPl9fv8AZiQWRwSBIdDYX5AqFwo95sS5GSkD8-iF0FpmgBzIUQLQ0sEQ0EHTLZe5QZxy3IQOhaJioOEI1nKQjnZxaEq5ZpmVAbUoVGoWSTNHlioxBd6fHkUpGgunEFHtEow0lO3mU4HKL2Y8Tc-2A6lBlLaYWirShv0ugWxsXhdpBYwQbLY9mtSREyTWKAAgAyABEnK4PJyfMzEQDjAxwqCYAoUM2aIN4okIO1poc7qc6bmrtiJj0+qshlARvMdnt+7KTgUiubtiPx91sQOleJethDmatjVariLhzJnwANzSGu+8l85MisUNpsrWlSsgyo67yLt6KdmsCS6NAwhHha-R7lgUCHqux7VKe9QKgaRomuB2wWF497OnyQKvhK7rSoyUCaiygKNmyZ5ohiWJsCOhJ3mGLpUnA+HvpK9JET6pH1hRyJnk81q2lo9qOg+lJ8W6Uqeik3oJuJgJIGxFiySG2HhixyntDGz7xupSaCrpWkpBmWZ6TiSFPAWRboO46DVgwEAZtk0BBN6TDAFAAAWfLyTh-yoJUHmCkgQrYEgOi5P4YrVIQN6Oc5Wiue5nk+RJDB+RpFFBTGoXhZF0VaLFN6SGlIRpN6rBQn82CBFoqLtHAfw3ik1gAHqCNIKR1S1UDtZ1BLXqVuGNhVSBVU1JS1dkDUpJNvX9V1UA9e0i2DSVZVpDpcasAA5MFNpCrt2AAMzkCdQ1la+W2GTt+25Udp3nUNkgJcYLlQG5SAeVATCHZmCBaEKLqZS6qAHaFcWvU571JZ9KW-f9ECA8DI0ZfpaPgw9UOSH9oUA0DLpjVVwWzb92gLZYHX5kNeNCgTqPpcTrCk1R7TBZT1PrbjSMo0T21intENHbTvOE2jN0ploQsPbtL1vR9X0-XSIMYwF7rxTDisIyraOg5jGvSLr-zM5N031Wzc3NatVOde0K2tbbNM3sb5EMKbNXLTNltQPNNvU-b2ScwNQ2u5Lul7WUXmGHQnxaMdUAABzkInocoMxo0C9Lu1RzHEBxwnyepyVCtw0rnmXKrTF8qglxQ6XyXfRX8wZ+j1fpbX8w45XfIeyUDXcz3TOVdV-ds4PLe91nkfR7HYqFynQ1D274d3bnc-x9gRfXkAA)
> 
> ```ts
> type Callback<T> = IsAny<T> extends true
>   ? (...param: unknown[]) => void
>   : (
>       [T] extends [void]
>         ? () => void
>         : T extends unknown[]
>             ? (...param: T) => void
>             : (param: T) => void
>     )
> ```
> 
> its a weird type at the minute in `main` and in this PR because:
> 
> * `Callback<void>` is `(...params: unknown[]) => void` which seems odd
>   
>   * who is using this and why?
>   * `trigger` is nonsense in main, because the first param is then `void` (i.e. `(param0: void) => void`)
> * `Callback<any>` has `...params: any` but should really at least be `unknown[]` IMO
>   
>   * though it could cause confusion since the type param is still `any`. people just shouldn't be passing this really
> * `Callback<string>` results in `(param0: string, ...params: unknown[])`
>   
>   * is this because people couldn't specify multiple params before?
>   * seems like this should be `(param0: string)` without the made up rest params, and the caller is responsible for defining any extra params
> 
> you can merge this PR as it currently stands, don't count me as a blocker. but you should probably at least change `[T] extends [unknown[]]` to `T extends unknown[]`.
> 
> does anyone know why we test against `[void]` and `[unknown[]]`? what usage fails if we change it to `void` and `unknown[]`?

For me, I would go with the simplest and cleanest solution:
 `type Callback<T> = [T] extends [any[]] ? (...params: T) => void : (param: T) => void;`
 
 with this one typing arguments is easy and clean you get what you expect and you are not limited 
 
 example: [TsPlayground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAwghgGwQIzgYwNYB4AqA+KAXigG0cBdKCAD2AgDsATAZ1LnpBPMoH4oAKAHTCwcAE5wAtswBcUHAEoiBAG4B7AJaMoc-qImS5i5VHVaA3AChroSFABCRWIhTpszYGI30A5nku20DBO8EiomFgeXr5QAD5Q9ACuksgQYv6BUAAiIS7h2CRR3j4ANAnJqWLkGeDQAKK5YW5YifQY9GoA7vRcNXYAYo2uEUW+vQG1UADiQ-lYhZ7FZUkpaWXCgq3tXT3c-gD0+1CdABYMCWpQ4j7JDMCs4tA0YAgQaHSME3YAErPNvUA)

43081j:
yes but the conditional types are there to allow `Callback<any>` and `Callback<void>`

it'd be a breaking change to the types if we drop those

```ts
Callback<any>;
// actually represents (...params: unknown[]) => void
// DOES NOT represent (param: any) => void

Callback<void>;
// actually represents (...params: unknown[]) => void
// DOES NOT represent (param: void) => void
```

for whatever reason, we have these two special cases. and that's why the type looks so wonky

isimehmeti:
that's why I added the change as an extra condition not to have breaking changes

but if I change `[T] extends [any[]]` to `T extends unknown[]`.  It will not interfere with union types like this test case on

`packages/shared/createEventHook/index.test.ts`

```js
  it('should pass union type', () => {
    let count = 0

    const { on: onResult, trigger } = createEventHook<number | string>()

    // union type should be inferred
    onResult(value => count = 2)
    trigger(1)
    trigger(2)

    expect(count).toBe(2)
  })
  ```

**Comment by hndregjoni** [comment]
> i wonder if we should simplify slightly and reduce our use of `any`
> 
> something like [this playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAkgZgQQHYgDwBUA0UCa2ByAfFALxQAMUEAHsBEgCYDOUAFAIxQBkU6AlFAD8uKAC4o+AFCTQkWE2Rp0xMvEUZswAE4BXCNjgBDADZMIhaQHpLEgKIB1XjgAKtgMrTZ0AMInjAI0MAYwBrDBV5dWUqWnpmKG09SSghNgA6DLBDLUMAW3EdJBCkAHsAdyQAbQBdARJiADcSgEsGZLE2dpSUyvRqmLpGFkqm1uqu7pThVjrGlrbJyfF0AbiWQuLyqvHF3dTWDLSsnPzeWahRhb2ltmO85fPLib5PcGhbBvpgAAkSkpCAPJIDCkKCGFARVhwJDiXzGALBMLKc4AbygJTgcHEM1Ic1aUAAvjI3lAPl9fv8AZiQWRwSBIdDYX5AqFwo95sS5GSkD8-iF0FpmgBzIUQLQ0sEQ0EHTLZe5QZxy3IQOhaJioOEI1nKQjnZxaEq5ZpmVAbUoVGoWSTNHlioxBd6fHkUpGgunEFHtEow0lO3mU4HKL2Y8Tc-2A6lBlLaYWirShv0ugWxsXhdpBYwQbLY9mtSREyTWKAAgAyABEnK4PJyfMzEQDjAxwqCYAoUM2aIN4okIO1poc7qc6bmrtiJj0+qshlARvMdnt+7KTgUiubtiPx91sQOleJethDmatjVariLhzJnwANzSGu+8l85MisUNpsrWlSsgyo67yLt6KdmsCS6NAwhHha-R7lgUCHqux7VKe9QKgaRomuB2wWF497OnyQKvhK7rSoyUCaiygKNmyZ5ohiWJsCOhJ3mGLpUnA+HvpK9JET6pH1hRyJnk81q2lo9qOg+lJ8W6Uqeik3oJuJgJIGxFiySG2HhixyntDGz7xupSaCrpWkpBmWZ6TiSFPAWRboO46DVgwEAZtk0BBN6TDAFAAAWfLyTh-yoJUHmCkgQrYEgOi5P4YrVIQN6Oc5Wiue5nk+RJDB+RpFFBTGoXhZF0VaLFN6SGlIRpN6rBQn82CBFoqLtHAfw3ik1gAHqCNIKR1S1UDtZ1BLXqVuGNhVSBVU1JS1dkDUpJNvX9V1UA9e0i2DSVZVpDpcasAA5MFNpCrt2AAMzkCdQ1la+W2GTt+25Udp3nUNkgJcYLlQG5SAeVATCHZmCBaEKLqZS6qAHaFcWvU571JZ9KW-f9ECA8DI0ZfpaPgw9UOSH9oUA0DLpjVVwWzb92gLZYHX5kNeNCgTqPpcTrCk1R7TBZT1PrbjSMo0T21intENHbTvOE2jN0ploQsPbtL1vR9X0-XSIMYwF7rxTDisIyraOg5jGvSLr-zM5N031Wzc3NatVOde0K2tbbNM3sb5EMKbNXLTNltQPNNvU-b2ScwNQ2u5Lul7WUXmGHQnxaMdUAABzkInocoMxo0C9Lu1RzHEBxwnyepyVCtw0rnmXKrTF8qglxQ6XyXfRX8wZ+j1fpbX8w45XfIeyUDXcz3TOVdV-ds4PLe91nkfR7HYqFynQ1D274d3bnc-x9gRfXkAA)
> 
> ```ts
> type Callback<T> = IsAny<T> extends true
>   ? (...param: unknown[]) => void
>   : (
>       [T] extends [void]
>         ? () => void
>         : T extends unknown[]
>             ? (...param: T) => void
>             : (param: T) => void(...param: T) => void
>     )
> ```
> 
> its a weird type at the minute in `main` and in this PR because:
> 
> * `Callback<void>` is `(...params: unknown[]) => void` which seems odd
>   
>   * who is using this and why?
>   * `trigger` is nonsense in main, because the first param is then `void` (i.e. `(param0: void) => void`)
> * `Callback<any>` has `...params: any` but should really at least be `unknown[]` IMO
>   
>   * though it could cause confusion since the type param is still `any`. people just shouldn't be passing this really
> * `Callback<string>` results in `(param0: string, ...params: unknown[])`
>   
>   * is this because people couldn't specify multiple params before?
>   * seems like this should be `(param0: string)` without the made up rest params, and the caller is responsible for defining any extra params
> 
> you can merge this PR as it currently stands, don't count me as a blocker. but you should probably at least change `[T] extends [unknown[]]` to `T extends unknown[]`.
> 
> does anyone know why we test against `[void]` and `[unknown[]]`? what usage fails if we change it to `void` and `unknown[]`?

That is a trick to tackle the way the conditional type expressions work. `T extends void ? M : N` would, when faced with `Callback<never>` resolve to `never`, regardless of `M` or `N`. When we check against `[void]`, `Callback<never>` works a bit more intuitively, and it resolves to `(param: unknown) => void`. Similarly, for union types (`never` is just a case of this, since it is a union of no types), the conditional distributes over the unioned types. So, `Callback<string | number>` resolves to `((param_0: string, ...param: unknown[]) => void) | (param_0: number, ...param: unknown[]) => void)` which, I think, is not as intuitive (and neither intended) compared to `(param_0: number | string, ...param: unknown[]) => void`.

Other than that, I do agree that adding the additional `...param: unknown[]` seems a bit superfluous.

**Comment by 43081j** [comment]
> but if I change `[T] extends [any[]]` to `T extends unknown[]`. It will not infer with union types like this test case on

aha yeah, ok makes sense!

otherwise it just applies to each member of the union, so would be something like `(p0: string) => void | (p0: number) => void`

> Callback<never> works a bit more intuitively, and it resolves to (param: unknown) => void

to be honest i'd love to see these three weird cases vanish in the next major version

```ts
Callback<never>; // (...params: unknown[]) => void
Callback<any>; // (...params: any) => void
Callback<void>; // (..params: unknown[]) => void
```

i actually don't think it is intuitive that any of these resolve to `(param: unknown) => void`. i think they're all very strange and are the sole reason `Callback` isn't a much simpler type

i would've expected this:

```ts
Callback<never>; // (param: never) => void
Callback<any>; // (param: any) => void
Callback<void>; // (param: void) => void
```

then i'd ask the consumer what they're trying to do, as the `never` and `void` examples here seem like nonsense

the `void` one seems like a hack to produce `Callback<[]>` (i.e. no args), and `never`, who knows.

i've taken this on a bit of a tangent so maybe we should discuss outside of this PR. but i do feel like this type is a bit of a mess right now

OrbisK:
> but i do feel like this type is a bit of a mess right now

maybe we can fine tune this, to solve the problem with v12 and do a follow up with the breaking changes for v13?

43081j:
i think so

i think this PR is all good and we can discuss other changes for the next major

EvgenyWas (approved):

ilyaliao (approved):

Issue #4556, opened 2025-01-31
BUG | `useTextareaAutosize` | "Uncaught ResizeObserver loop completed with undelivered notifications"


alexchexes:
### Describe the bug

When using `<textarea>` of the `useTextareaAutosize` with non-static width (e.g., width: 100%), an error occurs if the textarea contains more than one line of text and the window is resized in a way that affects the textarea's width:

`Uncaught ResizeObserver loop completed with undelivered notifications`

Reproducible in Chrome both locally and on play.vueuse.org, with the following code:

```vue
<script setup>
import { useTextareaAutosize } from '@vueuse/core'
const { textarea, input } = useTextareaAutosize()
input.value = 'long text'.repeat(30)
</script>

<template>
  <textarea
    ref="textarea"
    v-model="input"
    class="w-full resize-none text-black"
  />
</template>
```

### Reproduction

[...]

PR #4557, opened 2025-02-02
fix(useTextareaAutosize): improve resize handling with requestAnimationFrame

ilyaliao:
fix: #4556 

<details>
<br>

>

</details>

---

### Description

This PR fixes `ResizeObserver` loop errors by wrapping resize triggers in `requestAnimationFrame`.


packages/core/useTextareaAutosize/index.ts
```diff
@@ -52,8 +52,11 @@ export function useTextareaAutosize(options?: UseTextareaAutosizeOptions) {
   useResizeObserver(textarea, ([{ contentRect }]) => {
     if (textareaOldWidth.value === contentRect.width)
       return
-    textareaOldWidth.value = contentRect.width
-    triggerResize()
+
+    requestAnimationFrame(() => {
```
OrbisK, line 56: do we need `window.requestAnimationFrame()`? If yes we also need `options.window` and can pass it to `useResizeObserver` 😄

OrbisK: why don't we implement this in the resizeobserver? dont we need this in most cases?

ilyaliao: > why don't we implement this in the resizeobserver? dont we need this in most cases?

Are you mean useResizeObserver? Or what are you referring to?

OrbisK: > Are you mean useResizeObserver? Or what are you referring to?

should we wrap every callback from `useResizeObserver` into `requestAnimationFrame` to prevent this.

OrbisK: here: https://github.com/vueuse/vueuse/blob/main/packages/core/useResizeObserver/index.ts#L78

OrbisK (approved):

antfu (approved):

packages/core/useTextareaAutosize/index.ts
```diff
@@ -52,8 +54,11 @@ export function useTextareaAutosize(options?: UseTextareaAutosizeOptions) {
   useResizeObserver(textarea, ([{ contentRect }]) => {
     if (textareaOldWidth.value === contentRect.width)
       return
-    textareaOldWidth.value = contentRect.width
-    triggerResize()
+
+    window?.requestAnimationFrame(() => {
```
antfu, line 58: I wonder if we should fallback to execute this immediately when `window` does not exist?

OrbisK: what about `tryRequestAnimationFrame` 😄

ilyaliao: Do you mean ?

```ts
function tryRequestAnimationFrame(fn: Fn) {
  if (typeof requestAnimationFrame === 'function') {
    requestAnimationFrame(fn)
  }
  else {
    fn()
  }
}
```

ilyaliao: Maybe we could add this to #4548 and replace all instances with this, if needed.

antfu: Do we have other usage like this? If not I think we could just implement this inline for now and refactor later if needed

ilyaliao: Maybe not, but I think we can implement this inline directly, and then make changes to the rest in another PR.

OrbisK (approved):

antfu (approved):

PR #4560, opened 2025-02-04
docs: apply twoslash

serkodev:

---

### Description

Related: https://github.com/vueuse/vueuse/discussions/4548#discussioncomment-11997908

This PR applies [Twoslash](https://github.com/shikijs/twoslash) to document of VueUse and still in progress.

- Makes all `ts` and `vue` code blocks in the `index.md` document automatically apply Twoslash.
- Automatically imports `vue` in all code blocks (silently).
- Some sample code includes `vue` imports while others do not. Instead of enforcing uniform importing or non-importing, I believe that keeping shorter code snippets import-free can enhance readability. Therefore, I aim to preserve the original code snippets as much as possible.
- I believe some sample code is more like pseudo-code. Although certain null checks or type checks may not pass, they improve readability. As a result, I have enabled Twoslash’s `noErrors` option by default.
- Updates `vitepress > vite` to `npm:rolldown-vite@latest` to reduce the memory usage and build time on Netlify.

#### Preview

screenshot

##### Core

| Category   | Status |
| ---------- | ------ |
| State      | ✅      |
| Elements   | ✅      |
| Browser    | ✅      |
| Sensors    | ✅      |
| Network    | ✅      |
| Animation  | ✅      |
| Component  | ✅      |
| Watch      | ✅      |
| Reactivity | ✅      |
| Array      | ✅      |
| Time       | ✅      |
| Utilities  | ✅      |

 ##### Add-ons

| Category     | Status |
| ------------ | ------ |
| Electron     | ✅      |
| Firebase     | ✅      |
| Head         | 🔗      |
| Integrations | ✅      |
| Math         | ✅      |
| Motion       | 🔗      |
| Router       | ✅      |
| RxJS         | ⚠️      |
| SchemaOrg    | 🔗      |
| Sound        | 🔗      |


userquin:
@serkodev You need to change the node `max_old_space_size` here https://github.com/vueuse/vueuse/blob/main/netlify.toml#L6

serkodev:
@userquin Oh! Thanks for the reminder. I’m currently researching whether there’s a way to reduce build memory usage (it does require more than 16GB at the moment). So, it’s still uncertain whether this flag will be changed. If I decide to modify it later, I’ll add it to Netlify’s config. 👍

userquin:
@serkodev don't waste your time, looks like it is a problem with Vue version, check:
- https://github.com/vuejs/vitepress/issues/4242
- https://github.com/shikijs/shiki/issues/796

If we can run the build at Netlify with 16GB it should be fine /cc @antfu 

In the meantime you can also try using `buildConcurrency: 10` (or even small number, beware, will be slow) at VP config file (by default it is 64) here https://github.com/vueuse/vueuse/blob/main/packages/.vitepress/config.ts#L67:
- https://github.com/vuejs/vitepress/issues/4227#issuecomment-2469892828

userquin:
@serkodev Maybe you can try with VP 2.0.1.alpha (using Vite 6) in a new local branch from this one (or even using a new clean clone of the branch, no idea if updating to Vite 6 will have some side effect on the build): https://github.com/vuejs/vitepress/releases

(If you don't have time, I can give it a try next Saturday)

userquin:
we can also try updating the dependency to latest version 2.3.2: https://github.com/vueuse/vueuse/blob/main/pnpm-workspace.yaml#L12

userquin:
Thx @brc-dd :
> You can try enabling the [fs cache thing](https://shiki.style/packages/vitepress#file-system-cache) that twoslash added (it might not work though 🫠). Twoslash is rebuilding whole ts graph for each file which is both slow + goes OOM.

serkodev:
@userquin Thanks for providing this information and saving me a lot of time! 💚 I’ll take a look and see if it fixes the issue.

serkodev:
I tried the following:
- ✅ Using twoslash’s `createFileSystemTypesCache` successfully reduced memory usage, bringing it down from 24GB to 12GB.
- ❌ Reducing `buildConcurrency` didn’t seem to improve the issue.
- ❌ Updating the following dependencies also didn’t help with memory usage:
```
'@shikijs/vitepress-twoslash': ^2.3.2  
vitepress: ^2.0.0-alpha.2  
```

serkodev:
I’m not sure if it’s Netlify’s memory limit or build timeout, but after applying twoslash to all code blocks, Netlify encounters an error when running `build:docs`. 🧐

```
Failed during stage 'building site': Build script returned non-zero exit code: 2 (https://ntl.fyi/exit-code-2)
ELIFECYCLE  Command failed with exit code 129. (https://ntl.fyi/exit-code-129)
```
https://app.netlify.com/sites/vueuse/deploys/67a5e6ad2313310008ea2696

userquin:
can you try updating shikijs and shikijs vitepress to latest versions?

serkodev:
Updated but seems Netlify still can't build.

OrbisK:
@serkodev can we support you somehow?

serkodev:
> @serkodev can we support you somehow?

I’m currently still stuck on the issue of not being able to build successfully on Netlify. Since the build works after reducing the number of twoslash code blocks, I believe the issue is caused by twoslash. Recently, I’ve been investigating twoslash to see if there might be any memory leak issues. If possible, I would really appreciate any help in finding other solutions. Thank you very much!

43081j:
@serkodev can you catch up from main and also set this in the vitepress config:

```ts
  vite: {
    build: {
      sourcemap: false,
    },
  },
```

this should solve the problem with slow/out of memory builds

43081j:
looks like it still fails 😞 

i wonder if its worth trying to put the config in `.vitepress/config.ts`. it shouldn't make a difference but who knows

will see if i can dig into it a bit more with the e18e folks again some time (they're the ones that found the sourcemaps problem)

serkodev:
Inside `.vitepress/config.ts`, the `vite: viteConfig` is imported from `vite.config.ts` so it should be the same?

43081j:
ah i hadn't noticed that. you're right in that case 👍 

so we probably solved the memory issue but somehow still not the build... 🤔

serkodev:
On my device, if there is a memory leak issue during the build process, an error message like below will be displayed.
```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

However, since Netlify only shows the error below:
```
ELIFECYCLE  Command failed with exit code 129
```
I am unable to confirm whether the issue is same or related to a memory leak.

43081j:
yep i suspect it is two problems

however, you have solved a huge memory issue by disabling sourcemaps. i've had a few people check out this branch and fail to build it thanks to OOM exceptions. with sourcemaps disabled, it builds successfully

but we still have whatever problem CI is having

serkodev:
But my situation is the opposite. With `max-old-space-size=12288`, setting `sourcemap: false` results in a `JavaScript heap out of memory` error. However, if I don’t set it, the build succeeds. Anyway, I can make the build succeed by increasing `max-old-space-size` tho.

On my device, the build has always been successful, but on Netlify it fails unless I reduce the number of Twoslash code blocks.

43081j:
aha and that is absurd, since of course having sourcemaps results in _more_ memory (we're computing more of something, in general)

no idea why it would be the reverse for you! very strange

there's almost certainly a memory issue here either way. there's no way it should need you to specify memory limits at all. if it doesn't succeed under defaults, we have a problem

userquin:
can you try with VP ^1.6.3 instead 2 alpha?

serkodev:
Seems still cannot build on Netlify, seems like the memory usage issue is related to twoslash.
I’m thinking about whether it’s possible to build it using GitHub Actions and then upload it to Netlify?

ilyaliao:
Looks like it works 👀

serkodev:
Yep, after plenty of trial-and-error, I finally got it building successfully on Netlify! 😭

packages/.vitepress/plugins/markdownTransform.ts
```diff
@@ -67,10 +67,14 @@ ${snippet}
...
             // remove twoslash notations
-            snippetForCompare = twoslasher(snippet, 'ts', {
+            let r = twoslasher(snippet, 'ts', {
               handbookOptions: { noErrors: true },
               customTags: ['include'],
-            }).code
+            })
+            snippetForCompare = r.code
+
+            // @ts-expect-error set null
```
43081j: can we not just type `r` as whatever `twoslasher` returns ` | null`?

serkodev: Seems your code is outdated. the `r` variable is for debug purpose.

the latest code is just
```ts
snippetForCompare = twoslasher(snippet, 'ts').code
```

serkodev:
Currently I'm working on passing the `test:browser` test, seems there are some bug related to rolldown.
I'm trying to only apply `rolldown-vite` to `vitepress`

---

Update: test units passed

packages/.vitepress/plugins/markdownTransform.ts
43081j: sorry it was github freaking out! showing me an old diff

all good

OrbisK (approved):

antfu (approved):

PR #4561, opened 2025-02-05
feat(useDebounceFn): Add cancel option

### LINKED ISSUE #3375: useDebounceFn: Stop debouncing

I'd like to have the possibility of stopping the execution of a debounced function that is already planned to be executed in the future.

Example use case: I have an input control where typing would open a panel for autocompletions/suggestions. The typing keyboard events are debounced so that the panel shows after some delay (may be due to avoid the panel doing unecessary network fetches for example). I would also need the escape key to close the panel straight away and/or cancel the fetch, without waiting. But in that case, if they were some keys pressed just before the escape key, the debounce effect is already running and will ultimately re-opens the closed panel. Ideally, I'd like to cancel the planned debounce effect execution before it occurs.

### Suggested solution

This would require breaking changes in the api.
```
// Before
const debounced = useDebounceFn(() => ..., 1000);

// After
const { debounced, stop } = useDebounceFn(() => ..., 1000);

// Call the debounced function (same as before)
debounced().then(...) // or await

// Stop the debounce if the execution of the function is planned but not yet started
stop() // This would eventually throw in the debounced() call depending on the already existing rejectOnCancel parameter.
```

### Alternative

Current alternative is some custom code to track events in the timeline.

### Additional context

_No response_

### Validations

RaulTrombin:
### Description

Implements cancel option for debounced functions (fixes #3375)
Show:
https://github.com/user-attachments/assets/885f326b-9c86-44e2-ab71-8c6b177f8d74

This implementation:
- Adds a `cancel()` method to debounced functions
- Extends `EventFilter` type to support cancellation

### Documentation
- Basic cancellation usage
- `rejectOnCancel` option usage

### DIFF
--- a/packages/shared/useDebounceFn/index.ts
+++ b/packages/shared/useDebounceFn/index.ts
@@ -15,9 +15,9 @@ export function useDebounceFn<T extends FunctionArgs>(
   fn: T,
   ms: MaybeRefOrGetter<number> = 200,
   options: DebounceFilterOptions = {},
-): PromisifyFn<T> {
-  return createFilterWrapper(
-    debounceFilter(ms, options),
-    fn,
-  )
+): PromisifyFn<T> & { cancel: () => void } {
+  const filter = debounceFilter(ms, options)
+  const wrapped = createFilterWrapper(filter, fn) as PromisifyFn<T> & { cancel: () => void }
+  wrapped.cancel = () => filter.cancel?.()
+  return wrapped
 }
--- a/packages/shared/useDebounceFn/demo.vue
+++ b/packages/shared/useDebounceFn/demo.vue
@@ -4,13 +4,23 @@ import { ref } from 'vue'
 
 const updated = ref(0)
 const clicked = ref(0)
+const pending = ref(false)
 const debouncedFn = useDebounceFn(() => {
   updated.value += 1
-}, 1000, { maxWait: 5000 })
+  pending.value = false
+}, 1000, { maxWait: 5000, rejectOnCancel: true })
 
 function clickedFn() {
   clicked.value += 1
+  pending.value = true
   debouncedFn()
+}
+
+function cancelPending() {
+  if (pending.value) {
+    debouncedFn.cancel()
+    pending.value = false
+  }
 }
 </script>
 
@@ -18,8 +28,19 @@ function clickedFn() {
   <button @click="clickedFn">
     Smash me!
   </button>
+  <button
+    :disabled="!pending"
+    class="warn opacity-50 disabled:opacity-25"
+    @click="cancelPending"
+  >
+    Cancel pending
+  </button>
+
   <note>Delay is set to 1000ms and maxWait is set to 5000ms for this demo.</note>
 
   <p>Button clicked: {{ clicked }}</p>
   <p>Event handler called: {{ updated }}</p>
+  <p v-if="pending">
+    Pending update...
+  </p>
 </template>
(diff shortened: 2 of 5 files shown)


RaulTrombin:
> Thanks for your PR! Would you mind adding tests for this feature?

Hey, tests available.

before:

 ```
Test Files  171 passed (171)
      Tests  1204 passed | 1 skipped (1205)
   Start at  00:27:15
   Duration  57.95s (transform 1.77s, setup 635ms, collect 22.12s, tests 14.83s, environment 93.82s, prepare 15.77s)
```

now:
```
 Test Files  172 passed (172)
      Tests  1212 passed | 1 skipped (1213)
   Start at  02:55:49
   Duration  59.25s (transform 2.13s, setup 644ms, collect 22.76s, tests 15.06s, environment 95.35s, prepare 15.97s)
```

CI results available here:
https://github.com/RaulTrombin/vueuse/actions/runs/13170958187

File: packages/shared/useDebounceFn/index.ts [hunk 1]
```diff
@@ -15,9 +15,9 @@ export function useDebounceFn<T extends FunctionArgs>(
...
   options: DebounceFilterOptions = {},
-): PromisifyFn<T> {
-  return createFilterWrapper(
-    debounceFilter(ms, options),
-    fn,
-  )
+): PromisifyFn<T> & { cancel: () => void } {
+  const filter = debounceFilter(ms, options)
+  const wrapped = createFilterWrapper(filter, fn) as PromisifyFn<T> & { cancel: () => void }
+  wrapped.cancel = () => filter.cancel?.()
```
43081j, line 21: this is unfortunate 😞 because we return a function, we have no backwards-compat way of doing this other than sticking an arbitrary property on the end like you've done here

I wonder if we should change the return type based on the options. but maybe that's just as sketchy

some of this was/is being discussed in #4548

ilyaliao: Maybe for the end user, there won't be much difference.

Is your point about the `return type`?

43081j: It's more that we are tagging a property onto a function rather than having a structured object

Probably ok for now but this is a good example of one that really should return an object I imagine

43081j: I think some discussion is already happening in #4548, but yes it would be good to discuss at some point 👍 

this looks good for now though

OrbisK: Maybe we should use an AbortSignal here. If we pass an AbortSignal as an option, we can still abort without messing with the return value 🤔

(sorry @RaulTrombin sorry for the back and forth)

VividLemon:
This could be a useful addition if finished

pcoterecollective:
So? What's up with this PR? I can't believe it's not already merged! Need help with anything?? @RaulTrombin

OrbisK:
@43081j Lets get this over the line! I think we could make a breaking change. Maybe I will find some time next week to adjust this.

43081j:
i think it has quite a few conflicts, if someone can catch it up i agree we should probably get this one merged when we can

pcoterecollective:
@RaulTrombin @43081j Created [this PR](https://github.com/vueuse/vueuse/pull/5258) with a slightly different implementation (No breaking changes, just additive changes)

And [this one](https://github.com/vueuse/vueuse/pull/5259) is a Composition API flavour variant

Issue #4567, opened 2025-02-06
`useDevicePlatform` | Detect user device platform


arashsheyda:
### Clear and concise description of the problem

I want to detect the user's device platform (e.g., macOS, Windows, Linux, iOS, Android) so that I can make platform-specific UI adjustments (like shortcuts, icons, or ...)

### Suggested solution

`useDevicePlatform`: that detects the platform from `navigator.platform` or `navigator.userAgentData.platform` and returns a reactive reference with values like 'macOS', 'Windows', 'Linux', 'Android', 'iOS', or 'Unknown'.

### Additional context

there is an issue about [useDeviceDetection](https://github.com/vueuse/vueuse/issues/4463) but this would be different than that as `useDevicePlatform` only shows `macOs` | `windows` | `linux` | `android` | `ios` 

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vueuse.org/guide).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

OrbisK:
I think [this](https://github.com/vueuse/vueuse/issues/4492) is also related.

PR #4570, opened 2025-02-06
feat: add useDevicePlatform

aryjm:

### Description

Detects the platform of the user agent and returns a reactive reference with values like 'macOS', 'Windows', 'Linux', 'Android', 'iOS', or 'null'.

Coauthored by: @jonathanschndr and @Intevel

### Additional context

#4567


userquin:
can you take a look at this repo https://github.com/userquin/detect-browser-es ? you can also use it with SSR

JonathanSchndr:
> can you take a look at this repo https://github.com/userquin/detect-browser-es ? you can also use it with SSR

Thanks to mention it. So we inspired by your code. I would like to use your Package as dependency , but I think the idea of vueuse is not to include dependents for parts

userquin:
> Thanks to mention it. So we inspired by your code. I would like to use your Package as dependency , but I think the idea of vueuse is not to include dependents for parts

uhmm, check integrations, detect-browser-es is about ~15KB and std-env ~6KB

JonathanSchndr:
> > Thanks to mention it. So we inspired by your code. I would like to use your Package as dependency , but I think the idea of vueuse is not to include dependents for parts
> 
> 
> 
> uhmm, check integrations, detect-browser-es is about ~15KB and std-env ~6KB

Yes 💯 that's right, but it's not the size, it's the amount. Vueuse use less dep and we need only a part of the lib. So the decision is up to the maintainer.

userquin:
Just a note: you can remove Windows 11 detection, you can only detect it using [High Entropy Values](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues) or update the logic using https://github.com/userquin/detect-browser-es/blob/main/src/index.ts#L514-L523 

(there is no way to get Windows 11 from user agent: https://learn.microsoft.com/en-us/microsoft-edge/web-platform/how-to-detect-win11 and https://learn.microsoft.com/en-us/microsoft-edge/web-platform/how-to-detect-win11#sample-code-for-detecting-windows-11)

JonathanSchndr:
> Just a note: you can remove Windows 11 detection, you can only detect it using [High Entropy Values](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues) or update the logic using https://github.com/userquin/detect-browser-es/blob/main/src/index.ts#L514-L523 
> 
> 
> 
> (there is no way to detect Windows 11 from user agent: https://learn.microsoft.com/en-us/microsoft-edge/web-platform/how-to-detect-win11 and https://learn.microsoft.com/en-us/microsoft-edge/web-platform/how-to-detect-win11#sample-code-for-detecting-windows-11)
> 
> 
> 
> 

Hmm UserAgentData ist currently experimental :( so that was the point to go with the old one .. thanks for your note, we remove this check!

packages/core/useDevicePlatform/index.ts
```diff
@@ -0,0 +1,97 @@
...
+      return os
+  }
+
+  return null
+}
+
+export function useDevicePlatform(options: ConfigurableWindow = {}) {
+  const { window = defaultWindow } = options
+
+  const os = ref<string | null>(null)
```
userquin, line 80: cannot use here `OperatingSystem` instead string?

if so, we'll need to also export `OperatingSystem` (to avoid problems with Typescript >= 5.6 about required anotation)

packages/core/useDevicePlatform/index.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export function useDevicePlatform(options: ConfigurableWindow = {}) {
+  const { window = defaultWindow } = options
+
+  const os = ref<OperatingSystem | null>(null)
+
+  if (window) {
+    watch(
+      () => window.navigator.userAgent,
+      () => {
```
OrbisK, line 84: can `window.navigator.userAgent` be reactive?

Intevel: We were actually not sure about this, the problem is that it is not “reactive” either way. The user agent cannot change just like that.

However, the watcher through the watch immediate ensures that the value can be read directly at the beginning

antfu:
Thanks for your PR, but I think this does not belong to VueUse, but more like a library that mentioned above. Since this information does not need reactiveness and does not need to be coupled with Vue.

packages/core/useDevicePlatform/index.ts
mrleblanc101: Doesn't changing the emulated device in the DevTools update the User-Agent ?
Screenshot 2025-03-31 at 11 47 27 AM

OrbisK: `window.navigator.userAgent` isnt a signal/ref (or is it?), so the watcher cant detect the changes.

OrbisK: image

https://play.vuejs.org/#eNp9UctOwzAQ/BXLF1IpckGFS9VGKqgHOACCHi2hKNm6Lolt+ZFUivLvrB21cEC92TOzuzO7A90Yw7oAdElXrrLSeOLAB1NwJVujrScD6UtfHXJS6dYEDzUZyd7qltxg2Q1XXCU+y2bropeq1j1TZSdF6bVlwYHdCFA+J5mCPj1nZF2QgSuCHZXTDbBGi4zT1IZ4K4UAy2lOLgVcjTkZZNtCLUsPSxQFGBHmajWfXKNf/HhoTYOKYgfOr+aXL82pdzhtLwU7Oq0wbTLAaQwlG7Bvxkt0w+lysha5sml0/5KwODA/49UBqu9/8KM7RYzTdwuYuwMMceZ8aQX4id5+vsIJ3xey1XVoUH2F/ADcVIgeJ9ljUDXa/qNLbp/TzaQSO7c9eVDuHCoajcox6TnF2z1dif5rd8HuUx3eALf41YGNPXGBC/bA7m7p+AMEpMIk

Even tho the userAgent changed (see console), the watch will never detect it. Unless we can get a browser event to detect it 😅

PR #4574, opened 2025-02-09
feat: add useOnceInAllTabs (Web Locks API)

d--j:

---

### Description

Uses [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) to elect one of possible many tabs to do work. This can be used to e.g. refresh a session only in one tab.


packages/core/useOnceInAllTabs/index.ts
```diff
@@ -0,0 +1,89 @@
...
+   *
+   * @return `true` when the workload function was executed `false` otherwise.
+   */
+  doOnce: (workload: Workload) => boolean
+}
+
+/**
+ * Ensure that code gets executed only in one of many tabs.
+ *
+ * @see https://vueuse.org/core/useOnceInAllTabs/
```
OrbisK, line 42:
```suggestion
 * @see https://vueuse.org/useOnceInAllTabs/
```

packages/core/useOnceInAllTabs/index.ts
```diff
@@ -0,0 +1,89 @@
...
+ * @param [options]
+ */
+export function useOnceInAllTabs<Workload extends AnyFn = AnyFn, Keys extends string = string>(options: UseOnceInAllTabsOptions<Keys>): UseOnceInAllTabsReturn<Workload> {
+  const {
+    name,
+    navigator = defaultNavigator,
+  } = options
+  const isSupported = useSupported(() => navigator?.locks)
```
OrbisK, line 50: https://github.com/vueuse/vueuse/pull/4514

We should include the existence of AbortController to supported.

packages/core/useOnceInAllTabs/index.md
```diff
@@ -0,0 +1,21 @@
...
+Uses the [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) to elect one tab to do work.
+
+Use this composable to e.g. refresh a session only once even when multiple tabs of your application are open.
+
+## Usage
+
+```ts
+import { useOnceInAllTabs } from '@vueuse/core'
+
+const { doOnce, isElected, isSupported } = useOnceInAllTabs({ name: 'lock-name' })
```
OrbisK, line 16: maybe we should go for: `useWebLocks()` instead of `useOnceInAllTabs()` 🤔

OrbisK (changes requested):

packages/core/useOnceInAllTabs/index.ts
d--j: OK, did that. Anything that currently supports the [Web Locks API](https://caniuse.com/mdn-api_navigator_locks) does support the [AbortController](https://caniuse.com/abortcontroller), too.

OrbisK: > OK, did that. Anything that currently supports the [Web Locks API](https://caniuse.com/mdn-api_navigator_locks) does support the [AbortController](https://caniuse.com/abortcontroller), too.

This is true. But I think this will help with future regressions if for some reason the AbortController is not available (bug, security settings, new runtime/browser ...).

packages/core/useOnceInAllTabs/index.md
d--j: You are totally right, the name `useOnceInAllTabs` is not the best name. The Web Locks API has multiple use cases. Leader election is only one of them. IMHO, when we name the composable `useWebLocks` it should also handle the other use cases and features of the Web Locks API (shared locks, lock stealing, non-blocking lock requests). I did not figure out a good way to do that, tho.

What about the name `useWebLocksLeader` (and renaming `doOnce` to `asLeader` or `asLeaderDo`)?

OrbisK: I think I need to take a closer look at Web Locks. I have never used them before.

As described in: https://github.com/vueuse/vueuse/discussions/4548#discussioncomment-12115352

IMO we should add `useWebLocks`. `useOnceInAllTabs` (or whatever the name is) should wrap `useWebLocks`. So if we add other use cases we will have a common ground at `useWebLocks` and increase visibility with related composables.

d--j: @OrbisK I think I have found a reasonably flexible way for a `useWebLocks` composable. Please take a look at it.

packages/core/useWebLocks/index.ts
```diff
@@ -0,0 +1,113 @@
...
+          releaseLocks.delete(name)
+          abortControllers.delete(abortController)
+          reject(onrejected)
+        })
+      })
+    })
+  }
+  tryOnScopeDispose(() => {
+    abortControllers.forEach(abortController => abortController.abort())
+    abortControllers.clear()
```
OrbisK, line 107: Question: Is there a scenario where we only want to abort one controller? otherwise it seems we can use one abort controller for all lock requests. So we do not need to loop.
 
EDIT: Ok, maybe if we have several different reqeusts. But I still think we can use something like
 `AbortSignal.any` here to have an internal signal for all that we abort on dispose and we can create one per request to abort only one when needed.
 
 Ref: https://kettanaito.com/blog/dont-sleep-on-abort-controller

packages/core/useWebLocks/index.test.ts
```diff
@@ -0,0 +1,8 @@
+import { describe, expect, it } from 'vitest'
+import { useWebLocks } from '.'
+
+describe('useWebLocks', () => {
+  it('should be defined', () => {
+    expect(useWebLocks).toBeDefined()
+  })
+})
```
OrbisK, line 8: Maybe we can test some things here.

We can also use browser tests to use a real headless browser.

packages/core/useWebLocks/index.ts
```diff
@@ -0,0 +1,113 @@
...
+ * @param [options]
+ */
+export function useWebLocks<Keys extends string = string>(options: UseWebLocksOptions = {}): UseWebLocksReturn<Keys> {
+  const { navigator = defaultNavigator } = options
+  const isSupported = useSupported(() => navigator?.locks && typeof AbortController === 'function')
+  const releaseLocks = new Map<Keys, (value: unknown) => void>()
+  const abortControllers = new Set<AbortController>()
+  function request(name: Keys, callback: (signal: AbortSignal) => Promise<unknown> | unknown): Promise<unknown>
+  function request(name: Keys, options: UseWebLocksRequestOptions, callback: (signal: AbortSignal) => Promise<unknown> | unknown): Promise<unknown>
+  function request(name: Keys, optionsOrFn: UseWebLocksRequestOptions | ((signal: AbortSignal) => Promise<unknown> | unknown), callback?: (signal: AbortSignal) => Promise<unknown> | unknown): Promise<unknown> {
```
OrbisK, line 68: i think we can have better types for `unknown` here

OrbisK (changes requested):
i like it.

packages/core/useWebLocks/index.ts
d--j: How about this solution? library users can opt to use better types:
```ts
type ReturnTypeOne = void
type ReturnTypeTwo = { api: string, result: any }
const { request } = useWebLocks<{ lockName1: ReturnTypeOne, lockName2: ReturnTypeTwo }>()

request('lockName1', () => {})
request('lockName2', () => {
  // TypeScript error because api: number != api: string
  return { api: 1, result: 2}
})

```

packages/core/useWebLocks/index.test.ts
d--j: Added tests. But unfortunately, my Vitest knowledge is insufficient – maybe you can tell me how to catch the [Unhandled Rejection](https://github.com/vueuse/vueuse/blob/78aa3c47127877b9b5dbf6fde73942d99fa5bbdc/packages/core/useWebLocks/index.browser.test.ts#L57-L58) error?

packages/core/useWebLocks/index.ts
d--j: I switched to one shared `AbortController` – and added documentation/demo that makes clear that the `request` callback is responsible to clean up event handlers to the `abort` event (when it adds one).

packages/core/useWebLocks/index.test.ts
OrbisK: not that easy. I think we dont need to test the `tryOnScopeDispose`.

packages/core/useWebLocks/index.ts
OrbisK: We just passing the signal. Not the whole controller. okay. so this should be fine.

d--j: [AbortSignal.any](https://developer.mozilla.org/de/docs/Web/API/AbortSignal/any_static) is kinda new (baseline 2024). Should we require it in the `isSupported` check or use a [ponyfill](https://gist.github.com/CNSeniorious000/9fc1a72e45358dd7c9e2f16e5d26df5c)?

OrbisK: > [AbortSignal.any](https://developer.mozilla.org/de/docs/Web/API/AbortSignal/any_static) is kinda new (baseline 2024). Should we require it in the `isSupported` check or use a [ponyfill](https://gist.github.com/CNSeniorious000/9fc1a72e45358dd7c9e2f16e5d26df5c)?

I think we dont need this, because we will currently dont have an external signal. I have removed my prev comments :)

packages/core/useWebLocks/index.test.ts
d--j: I found [a way to test/catch it](https://github.com/vitest-dev/vitest/pull/6016)

d--j:
I just pushed a new version but also converted this PR to a draft since I am not finished polishing it (`useWebLocks`) and coding the demo (`useLeaderElection`). Unfortunately, my day job demands some attention and I need to step away for a bit.

Thank you, @OrbisK, for very valuable pointers and code reviews so far!

packages/core/useLeaderElection/index.md
```diff
@@ -0,0 +1,68 @@
...
+Uses the [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) to elect one tab as leader.
+
+Use this composable to e.g. refresh a session only once even when multiple tabs of your application are open or multiple worker are running.
+
+## Usage
+
+```ts
+import { useLeaderElection } from '@vueuse/core'
+
+const { asLeader, isElected, isSupported } = useLeaderElection({ name: 'lock-name' })
```
Glandos, line 16:
```suggestion
const { asLeader, isLeader, isSupported } = useLeaderElection({ name: 'lock-name' })
```

PR #4581, opened 2025-02-14
feat!: drop CJS build, now it's ESM-only

antfu:
Consider the majority of Vue users are using bundlers, dropping CJS should only affect CJS-SSR usage, which is considered rather nitch.

We plan to ship this as v13 on 2025/3/10 (Monday), with this as the sole breaking change.

If you have any feedback, please leave comments below. Thanks.


packages/core/package.json
```diff
@@ -22,23 +22,16 @@
   ],
   "sideEffects": false,
   "exports": {
-    ".": {
-      "import": "./index.mjs",
-      "require": "./index.cjs"
-    },
+    ".": "./index.mjs",
```
43081j, line 25: in a later release, would these eventually become `.js` (i.e. `index.js`)

seeing as we already have `"type": "module"`, they're already loaded as ES modules

43081j (dismissed):

ilyaliao:
Will this be released in v13?

OrbisK (approved):

packages/core/package.json
antfu: I prefer the `.mjs` for explicitness (in Node it might also slightly improve the performance as it skip one check to find package.json)

OrbisK (approved):

userquin:
Why are we still generating CJS stuff? Why are we including CJS stuff in the tgz (cannot remove `d.cts` and `.cjs` from files array)?

packages/components/package.json
```diff
@@ -21,19 +21,15 @@
...
-  "main": "./index.cjs",
+  "main": "./index.mjs",
   "module": "./index.mjs",
   "unpkg": "./index.iife.min.js",
   "jsdelivr": "./index.iife.min.js",
-  "types": "./index.d.ts",
+  "types": "./index.d.mts",
   "files": [
-    "*.cjs",
     "*.d.cts",
```
OrbisK, line 33: @userquin this?

userquin: yes, also .d.cts and remove it from the rollup.config.ts file

userquin: L122:

imagen

userquin: We should remove `.d.cts` from all files entries in all package.json files

userquin:
Just a note about node10 module resolution (I guess @antfu wants also to drop support for it, shouldn't be a problem): it is not resolving correctly the types but not the impplementation. I'm testing on my local without generating CJS stuff...

43081j (approved):

packages/core/package.json
```diff
@@ -22,26 +22,17 @@
...
     "./*": "./*",
-    "./metadata": {
-      "import": "./metadata.mjs",
-      "require": "./metadata.cjs"
-    }
+    "./metadata": "./metadata.mjs"
```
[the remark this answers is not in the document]
antfu: I personally don't like to ask users to explicitly add `.mjs` to the exports. Bundler should take creat of that I think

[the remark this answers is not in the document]
43081j: It's not to ask users, it's to support the ones that do 

Anyone who builds to esm can't use this because they need the extension, but typescript will tell them no (since it isn't in the export map) 

Generally we should have both if people are to continue using it without an extension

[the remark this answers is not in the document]
antfu: >  note again that we're inconsistent 

Do you mean the ecosystem, or VueUse itself is inconsistent?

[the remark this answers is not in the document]
43081j: Vueuse itself exports `./*` in some places, which means the exports have extensions

Meanwhile, we export individual things without extensions

I think we should do both so people following "pure" esm can import the _actual file_. Now that these are not aliases anymore, they're just acting as a constraint

Before, they were aliases for the sake of conditionally importing the right module. But now that there's only one, it's a one to one mapping of the file on disk... But without the extension

In other projects, we either have real paths (i.e extensions), or we have both (two entries for each file)

9romise:
aha, typo: ~~2024~~ 2025

userquin:
Uhmm, rigth now `@vueuse/core` types are broken (latest v12.8.1 and beta v13.0.0-beta.1), cannot resolve `@vueuse/core/_configurable`:

https://arethetypeswrong.github.io/?p=%40vueuse%2Fcore%4012.8.1

https://arethetypeswrong.github.io/?p=%40vueuse%2Fcore%4013.0.0-beta.1

userquin:
I guess we can map `./*` to `./dist/*.mjs` or use `imports` entry (`#`) to map internal/private modules (otherwise we'll need to map individual entries, maybe related to @43081j comment)

43081j:
It doesn't seem to exist regardless of our export map 😬

userquin:
> It doesn't seem to exist regardless of our export map 😬

Yeah, missing in the tgz file 😬😬

43081j:
I guess it's introduced by #4591

Since we didn't import types from configurable before then maybe? 

So the rollup bundle has never had to deal with it

antfu:
That was a mistake, fixed in https://github.com/vueuse/vueuse/commit/3f5da12be86a9e6193f9a708078066a3d2a321f4

Issue #4575, opened 2025-02-10
Tracking inconsistencies: replace many `ref` with `shallowRef`


OrbisK:
Replace most primitive `ref` with `shallowRef`

_Originally posted by @OrbisK in https://github.com/vueuse/vueuse/discussions/4548#discussioncomment-11985416_

ilyaliao:
We can make a list to trace the progress of refactoring

- [ ] useBluetooth
- [ ] useBreakpoints
- [ ] useBroadcastChannel
- [ ] useBrowserLocation
- [ ] useClipboard
- [ ] useColorMode
- [ ] useCssVar
- [ ] useDark
- [ ] useEventListener
- [ ] useEyeDropper
- [ ] useFavicon
- [ ] useFileDialog
- [ ] useFileSystemAccess
- [ ] useFullscreen
- [ ] useGamepad
- [ ] useImage
- [ ] useMediaControls
- [ ] useMediaQuery
- [ ] useMemory
- [ ] useObjectUrl
- [ ] usePermission
- [ ] usePreferredColorScheme
- [ ] usePreferredDark
- [ ] usePreferredLanguages
- [ ] useScreenOrientation
- [ ] useScreenSafeArea
- [ ] useScriptTag
- [ ] useShare
- [ ] useStyleTag
- [ ] useTextareaAutosize
- [ ] useTitle
- [ ] useUrlSearchParams
- [ ] useVibrate
- [ ] useWakeLock
- [ ] useWebNotification
- [ ] onClickOutside
- [ ] onKeyStroke
- [ ] onLongPress
- [ ] onStartTyping
- [ ] useBattery
- [ ] useDeviceMotion
- [ ] useDeviceOrientation
- [ ] useDevicePixelRatio
- [ ] useDevicesList
- [ ] useDisplayMedia
- [ ] useElementByPoint
- [ ] useElementHover
- [ ] useFocus
- [ ] useFocusWithin
- [ ] useFps
- [ ] useGeolocation
- [ ] useIdle
- [ ] useInfiniteScroll
- [ ] useKeyModifier
- [ ] useMagicKeys
- [ ] useMouse
- [ ] useMousePressed
- [ ] useNavigatorLanguage
- [ ] useNetwork
- [ ] useOnline
- [ ] usePageLeave
- [ ] useParallax
- [ ] usePointer
- [ ] usePointerSwipe
- [ ] useScroll
- [ ] useScrollLock
- [ ] useSpeechRecognition
[...]

PR #4582, opened 2025-02-14
refactor: change primitive to `shallowRef`, rename `ref` usage to `deepRef`

antfu:
Part of #4575

This PR is a refactor, should not introduce any behavioral change

- change primitive refs like `ref(0)` to `shallowRef(0)`
- rename other `ref()` usage to `deepRef()` (alias from `ref()`) for better explicitness
- local ESLint rule to restrict `ref()` usage

### DIFF
--- a/packages/core/onLongPress/component.ts
+++ b/packages/core/onLongPress/component.ts
@@ -1,6 +1,6 @@
 import type { RenderableComponent } from '../types'
 import type { OnLongPressOptions } from './index'
-import { defineComponent, h, ref } from 'vue'
+import { ref as deepRef, defineComponent, h } from 'vue'
 import { onLongPress } from './index'
 
 export interface OnLongPressProps extends RenderableComponent {
@@ -12,7 +12,7 @@ export const OnLongPress = /* #__PURE__ */ defineComponent<OnLongPressProps>({
   props: ['as', 'options'] as unknown as undefined,
   emits: ['trigger'],
   setup(props, { slots, emit }) {
-    const target = ref()
+    const target = deepRef()
     onLongPress(
       target,
       (e) => {
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -129,6 +129,11 @@ export default antfu(
         from: 'vue',
         name: 'shallowRef',
       },
+      {
+        from: 'vue',
+        name: 'ref',
+        as: 'deepRef',
+      },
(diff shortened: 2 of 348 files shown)


antfu:
2. With the ESLint rule, contributors should get the error with hint to choose between `deep` and `shallow` - this brings the awareness of what ref you are using.
3. This PR itself does not improve or change the performance (there should be follow-up PRs that changes some deepRef to shallowRef, which would actually improve the performance, but that would be breaking and we need to handle it carefully). But more like a move that we are trying to make the shallow default. You can argue that most of the time performance is imperceptible even on nested objects/arrays (as long as they are not crazily big). Micro or not, I consider this not an "optimization", but more like a default performant practice that _should_ be the default, where optimization is not even relevant.

TheAlexLichter:
Thanks for the insights @antfu!

Do you think the move to `shallowRef` by default (similar to how other frameworks do it) would be one that should be adopted by the whole ecosystem, so also for "end users"? If yes, why that? 🤔

antfu:
I don't honestly sure what would be the best practice for this. Theoretically, yes, I think `ref` should be shallow by default and ideally the ecosystem should move on that, but the status quote has a strong inertia, where benefits might not always justify the cost of migration. I think we still need time to try and feel about it, there is no final answer yet. But I think VueUse could a good place to start and explore.

packages/core/onLongPress/component.ts
```diff
@@ -12,7 +12,7 @@ export const OnLongPress = /* #__PURE__ */ defineComponent<OnLongPressProps>({
   props: ['as', 'options'] as unknown as undefined,
   emits: ['trigger'],
   setup(props, { slots, emit }) {
-    const target = ref()
+    const target = deepRef()
```
43081j, line 15: should this be a `shallowRef` too? since its a ref to an element afaict

43081j (approved):
looks good to me

@antfu there's various places where we use a `deepRef` but possibly shouldn't. i'll ping you on discord and, if im right, will open separate PRs 👍

packages/core/onLongPress/component.ts
antfu: Yes. I was using a simple regexp to replace those simple ones to shallow, I didn't go through these details (requires context) in this pr. Definitely something for the future.

PR #4583, opened 2025-02-14
feat(useRafFn): add `once` option

btea:

<details>
<br>

>

</details>

---

### Description
Sometimes users may only want to execute `fn` once, so I think adding this option is more flexible.

### DIFF
--- a/packages/core/useRafFn/index.ts
+++ b/packages/core/useRafFn/index.ts
@@ -1,8 +1,8 @@
 import type { Pausable } from '@vueuse/shared'
 import type { MaybeRefOrGetter } from 'vue'
-import type { ConfigurableWindow } from '../_configurable'
 import { tryOnScopeDispose } from '@vueuse/shared'
 import { computed, readonly, ref, toValue } from 'vue'
+import type { ConfigurableWindow } from '../_configurable'
 import { defaultWindow } from '../_configurable'
 
 export interface UseRafFnCallbackArguments {
@@ -60,6 +60,7 @@ export function useRafFn(fn: (args: UseRafFnCallbackArguments) => void, options:
   })
   let previousFrameTimestamp = 0
   let rafId: null | number = null
+  let hasCalledOnce = false
 
   function loop(timestamp: DOMHighResTimeStamp) {
     if (!isActive.value || !window)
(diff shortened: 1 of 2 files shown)


packages/core/useRafFn/index.ts
```diff
@@ -70,10 +78,19 @@ export function useRafFn(fn: (args: UseRafFnCallbackArguments) => void, options:
...
+      isActive.value = false
+      rafId = null
+      return
+    }
     rafId = window.requestAnimationFrame(loop)
   }
 
   function resume() {
+    if (hasCalledOnce)
+      return
```
NaokiHaba, line 92: Currently, the implementation prevents `resume()` from restarting the loop if `once: true` has been set and the function has already executed once. This makes it impossible for users to explicitly call `resume()` to re-run the function, even when they intend to do so.

btea: Maybe I should just remove the logic around the `hasCalledOnce` variable?

ilyaliao: Depending on your goal, do you want to **stop** after executing once? Or just automatically **pause** and then be able to **resume** to run again?

NaokiHaba: If we want to allow resume() to restart the function even when once: true is set, we could introduce an option like resume(force: true).

This way, the function would still execute only once automatically, but users could explicitly call resume(force: true) to manually restart it when needed.

Would this approach align with the expected behavior?

btea: I think both are fine. Which way do you think is better?

NaokiHaba: In my opinion, to achieve the same effect without increasing code complexity, we should consider removing the existing conditional branching.

btea: OK, now it is adjusted like this.

OrbisK:
Can you please add some basic tests to cover this: https://github.com/vueuse/vueuse/blob/main/packages/core/useRafFn/index.test.ts

btea:
> Can you please add some basic tests to cover this: [`main`/packages/core/useRafFn/index.test.ts](https://github.com/vueuse/vueuse/blob/main/packages/core/useRafFn/index.test.ts)

Done.

antfu (approved):

PR #4591, opened 2025-02-15
feat(useCached): add `options.deepRefs`

OrbisK:

<details>
<br>

>

</details>

---

### Description

This does add some utils to handle the `options.deepRefs`.

Transform `useCached` to support `options.deepRefs`.

### Additional context

Because watcher has also `options.deepRefs` we might want to have a closer look there.


packages/shared/createRef/index.ts
```diff
@@ -0,0 +1,22 @@
+import type { Ref, ShallowRef } from 'vue'
+import { ref as deepRef, shallowRef } from 'vue'
+
+export type MaybeDeepRef<T = any, D extends boolean = false> = D extends true ? Ref<T> : ShallowRef<T>
+
+export function createRef<T = any, D extends boolean = false>(value: T, deep?: D): MaybeDeepRef<T, D> {
```
OrbisK, line 6: maybe better naming here 😅

packages/core/_configurable.ts
```diff
@@ -39,3 +39,10 @@ export const defaultWindow = /* #__PURE__ */ isClient ? window : undefined
 export const defaultDocument = /* #__PURE__ */ isClient ? window.document : undefined
 export const defaultNavigator = /* #__PURE__ */ isClient ? window.navigator : undefined
 export const defaultLocation = /* #__PURE__ */ isClient ? window.location : undefined
+
+export interface ConfigurableDeepRefs<D extends boolean> {
+  /**
+   * Return deep refs instead of shallow refs.
+   */
```
antfu, line 46:
```suggestion
  /**
   * Return deep refs instead of shallow refs.
   *
   * @default true - will be changed to `false` by default in the next major
   */
```

packages/core/useCached/index.ts
```diff
@@ -1,17 +1,27 @@
...
+  const cachedValue = createRef(refValue.value, deep)
 
   watch(() => refValue.value, (value) => {
     if (!comparator(value, cachedValue.value))
       cachedValue.value = value
-  }, watchOptions)
+  }, options)
 
-  return cachedValue
+  return cachedValue satisfies UseCachedReturn<T, D>
```
antfu, line 24: better put it at the end of the function with `:`

packages/shared/createRef/index.ts
antfu: The name sounds good to me. We should add some jsdoc

packages/shared/createRef/index.ts
```diff
@@ -0,0 +1,22 @@
+import type { Ref, ShallowRef } from 'vue'
+import { ref as deepRef, shallowRef } from 'vue'
+
+export type MaybeDeepRef<T = any, D extends boolean = false> = D extends true ? Ref<T> : ShallowRef<T>
```
antfu, line 4:
```suggestion
export type ShallowOrDeepRef<T = any, D extends boolean = false> = D extends true ? Ref<T> : ShallowRef<T>
```

Look for better names. But `Maybe` prefix mostly indicates "both are fine", while here it's either, so I suggest we better avoid using `Maybe`

OrbisK: My one semester of Haskell is trying to get: `EitherShallowOrDeep` (`EitherShallowOrDeepRef`). But I think this is too long.

packages/core/_configurable.ts
```diff
@@ -39,3 +39,12 @@ export const defaultWindow = /* #__PURE__ */ isClient ? window : undefined
...
 export const defaultNavigator = /* #__PURE__ */ isClient ? window.navigator : undefined
 export const defaultLocation = /* #__PURE__ */ isClient ? window.location : undefined
+
+export interface ConfigurableDeepRefs<D extends boolean> {
+  /**
+   * Return deep refs instead of shallow refs.
+   *
+   * @default true - will be changed to `false` by default in the next major
+   */
+  deep?: D
```
antfu, line 49: Oh I remember why I picked `shallow` before, as `deep` actually conflicts with watch's deep. Maybe name it explicitly as `deepRef`?

OrbisK: Actually this should be fine, no? If we have a `deepRef` and we watch it, we might want to end up with `deep: true` anyway.

For most composables, we do not really extend `WatchOptions` (except for watch composables). And even if we do, in most cases we want to do it in a controlled way.

antfu: Yeah, while it might make sense to do both, I am worried that having one flag controlling two things could result in footgun in future design. Consider what we have for `immediate`. I think it would better to keep on the safe path and avoid making premature assumptions of the usage cases.

> I think if we are going to expose watchOptions strait, we should do options.watchOptions?

that would be a bit too late to change that, as most of composable already designed it in a flat way

antfu (approved):

PR #4592, opened 2025-02-15
refactor: move to `shallowRef` in shared utilities

43081j:
This switches various tests and `useLastChanged` to use a `shallowRef` rather than a `deepRef` on primitive values (such as numbers).

Achieved through `npx eslint --fix packages/shared`, using the `no-primitive-shallow` lint rule (currently on a branch).

### DIFF
--- a/packages/shared/useLastChanged/index.ts
+++ b/packages/shared/useLastChanged/index.ts
@@ -1,5 +1,5 @@
 import type { Ref, WatchOptions, WatchSource } from 'vue'
-import { ref as deepRef, watch } from 'vue'
+import { shallowRef, watch } from 'vue'
 import { timestamp } from '../utils'
 
 export interface UseLastChangedOptions<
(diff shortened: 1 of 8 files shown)


packages/shared/useLastChanged/index.ts
```diff
@@ -17,7 +17,7 @@ export interface UseLastChangedOptions<
 export function useLastChanged(source: WatchSource, options?: UseLastChangedOptions<false>): Ref<number | null>
 export function useLastChanged(source: WatchSource, options: UseLastChangedOptions<true> | UseLastChangedOptions<boolean, number>): Ref<number>
 export function useLastChanged(source: WatchSource, options: UseLastChangedOptions<boolean, any> = {}): Ref<number | null> | Ref<number> {
-  const ms = deepRef<number | null>(options.initialValue ?? null)
+  const ms = shallowRef<number | null>(options.initialValue ?? null)
```
43081j, line 20: if this is considered breaking, we can revert it for now so this PR only touches tests

antfu: This should be safe to do

antfu (approved):

PR #4598, opened 2025-02-15
refactor(useEventSource): use `shallowRef` for event & allow any data

43081j:
Uses `shallowRef` for the event names as they are always strings.

Uses `deepRef` for the event data as it could be anything.

This also raised that the `data` is currently typed as `string | null`, but event source events can actually pass data of any type. This compiled successfully in the past because a `MessageEvent` defaults to `MessageEvent<any>` (i.e. `{data: any}`).

For now, i have introduced a `Data` type parameter which defaults to `any` to keep backwards compat.

In future, we should default this to `unknown`.

### DIFF
--- a/packages/core/useEventSource/index.ts
+++ b/packages/core/useEventSource/index.ts
@@ -1,7 +1,7 @@
 import type { Fn, MaybeRefOrGetter } from '@vueuse/shared'
 import type { Ref, ShallowRef } from 'vue'
 import { isClient, toRef, tryOnScopeDispose } from '@vueuse/shared'
-import { ref as deepRef, shallowRef, watch } from 'vue'
+import { ref as deepRef, shallowReadonly, shallowRef, watch } from 'vue'
 import { useEventListener } from '../useEventListener'
 
 export type EventSourceStatus = 'CONNECTING' | 'OPEN' | 'CLOSED'
@@ -55,23 +55,23 @@ export interface UseEventSourceReturn<Events extends string[], Data = any> {
    * Reference to the latest data received via the EventSource,
    * can be watched to respond to incoming messages
    */
-  readonly data: ShallowRef<Data>
+  readonly data: Readonly<ShallowRef<Data>>
 
   /**
    * The current state of the connection, can be only one of:
    * 'CONNECTING', 'OPEN' 'CLOSED'
    */
-  readonly status: ShallowRef<EventSourceStatus>
+  readonly status: Readonly<ShallowRef<EventSourceStatus>>
 
   /**
    * The latest named event
    */
-  readonly event: ShallowRef<Events[number] | null>
+  readonly event: Readonly<ShallowRef<Events[number] | null>>
 
   /**
    * The current error
    */
-  readonly error: ShallowRef<Event | null>
+  readonly error: Readonly<ShallowRef<Event | null>>
 
   /**
    * Closes the EventSource connection gracefully.
@@ -92,7 +92,7 @@ export interface UseEventSourceReturn<Events extends string[], Data = any> {
    * The last event ID string, for server-sent events.
    * @see https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/lastEventId
    */
-  readonly lastEventId: ShallowRef<string | null>
+  readonly lastEventId: Readonly<ShallowRef<string | null>>
 }
 
 function resolveNestedOptions<T>(options: T | true): T {
@@ -214,12 +214,12 @@ export function useEventSource<Events extends string[], Data = any>(
 
   return {
     eventSource,
-    event,
-    data,
-    status,
-    error,
+    event: shallowReadonly(event),
+    data: shallowReadonly(data),
+    status: shallowReadonly(status),
+    error: shallowReadonly(error),
     open,
     close,
-    lastEventId,
+    lastEventId: shallowReadonly(lastEventId),
   }
 }
--- a/packages/core/useEventSource/index.ts
+++ b/packages/core/useEventSource/index.ts
@@ -55,23 +55,23 @@ export interface UseEventSourceReturn<Events extends string[], Data = any> {
    * Reference to the latest data received via the EventSource,
    * can be watched to respond to incoming messages
    */
-  data: Ref<Data>
+  readonly data: ShallowRef<Data>
 
   /**
    * The current state of the connection, can be only one of:
    * 'CONNECTING', 'OPEN' 'CLOSED'
    */
-  status: Ref<EventSourceStatus>
+  readonly status: ShallowRef<EventSourceStatus>
 
   /**
    * The latest named event
    */
-  event: Ref<Events[number] | null>
+  readonly event: ShallowRef<Events[number] | null>
 
   /**
    * The current error
    */
-  error: Ref<Event | null>
+  readonly error: ShallowRef<Event | null>
 
   /**
    * Closes the EventSource connection gracefully.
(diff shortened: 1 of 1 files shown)


43081j:
for reference:
https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/data

> The data sent by the message emitter; this can be any data type, depending on what originated this event

meanwhile, a `MessageEvent` in typescript is defined as:

```ts
interface MessageEvent<T = any> extends Event {
  readonly data: T;
  // ...
}
```

so our `data` should indeed at least be `Ref<any>`, and preferably `Ref<unknown>` one day

im guessing this is a fairly low usage helper in the wild so nobody ran into it yet

packages/core/useEventSource/index.ts
```diff
@@ -115,8 +115,8 @@ export function useEventSource<Events extends string[]>(
   events: Events = [] as unknown as Events,
   options: UseEventSourceOptions = {},
 ): UseEventSourceReturn<Events> {
-  const event: Ref<string | null> = deepRef(null)
-  const data: Ref<string | null> = deepRef(null)
+  const event: Ref<string | null> = shallowRef(null)
```
OrbisK, line 118:
```suggestion
  const event: ShallowRef<string | null> = shallowRef(null)
```
no?

packages/core/useEventSource/index.ts
```diff
@@ -115,8 +115,8 @@ export function useEventSource<Events extends string[]>(
...
+  const data: Ref<unknown> = deepRef(null)
```
OrbisK, line 119: generally **I prefere** `const data = deepRef<unknown>(null)`  i think we can even drop the `null` >  `const data = deepRef<unknown>()`

packages/core/useEventSource/index.ts
```diff
@@ -115,8 +115,8 @@ export function useEventSource<Events extends string[]>(
...
   const status = shallowRef('CONNECTING') as Ref<EventSourceStatus>
   const eventSource = deepRef(null) as Ref<EventSource | null>
   const error = shallowRef(null) as Ref<Event | null>
```
OrbisK, line 122:
```suggestion
  const status = shallowRef<EventSourceStatus>('CONNECTING')
  const eventSource = deepRef<EventSource | null>(null)
  const error = shallowRef<Event | null>(null)
```

packages/core/useEventSource/index.ts
43081j: this and your other comment are just because its a lint fix up. i wouldn't have written it this way myself

ill update

packages/core/useEventSource/index.ts
43081j: same as the other comments, it was just already this way

ill update though

packages/core/useEventSource/index.ts
OrbisK: I know. 😄 

Just wanted to mention if we touch this anyway 👍🏽

43081j:
@OrbisK you're right

ive updated to add a type parameter and published the PR 👍

OrbisK (dismissed):

43081j:
im happy to merge this whenever but a second review would be helpful, particularly because we're slightly changing behaviour

```ts

const {data} = useEventSource(url);

// ...

// Before this PR
data.trim(); // typescript thinks it is a string and strongly types the existence of `trim`, etc

// After this PR
data.trim(); // typescript allows it because it is `any`, but has no idea what `trim` is
data.trim('nonsensical', 'args', 'trim', 'doesnt', 'take'); // works because it is `any`
```

the joys of `any`

we could alternatively default `Data = string` so it behaves exactly as before, however it'll be wrong because an event source can emit non-string data 🤔

packages/core/useEventSource/index.ts
```diff
@@ -110,16 +110,16 @@ function resolveNestedOptions<T>(options: T | true): T {
...
-  const event: Ref<string | null> = deepRef(null)
-  const data: Ref<string | null> = deepRef(null)
-  const status = shallowRef('CONNECTING') as Ref<EventSourceStatus>
-  const eventSource = deepRef(null) as Ref<EventSource | null>
-  const error = shallowRef(null) as Ref<Event | null>
+  const event: ShallowRef<string | null> = shallowRef(null)
+  const data: ShallowRef<Data | null> = shallowRef(null)
+  const status = shallowRef<EventSourceStatus>('CONNECTING')
+  const eventSource = deepRef<EventSource | null>(null)
+  const error = shallowRef<Event | null>(null)
```
OrbisK, line 122: I think this can be done like this: https://github.com/vueuse/vueuse/pull/4591

So we do not actually break deep refs on data.

43081j: I'd personally not bother as it clearly should be shallow. We need to decide when to break these to make them that way

Otherwise we will end up with this deep flag workaround everywhere, and it'll be even more difficult to get rid of in future

Now is our chance to fix this and it'll be less breaking than removing the deep flag one day

OrbisK: From discord discussion with @43081j:

TL;DR:

- returned immutable data should be `readonly` and `shallow`
- complex state that should not be `readonly` should default to shallow and have an opt-out (`deepRefs`) (rare)
- exceptions always exist

43081j:
@OrbisK this now should have read-only return types

I think it should be good to go. we will be weakening the types but its the only correct thing to do since its currently wrong 😬

43081j:
ok @OrbisK this is ready to go again. it now has your suggested readonly wrapper around each ref

we should get this merged soon since its been sat around a while now

packages/core/useEventSource/index.ts
```diff
@@ -50,28 +50,28 @@ export interface UseEventSourceOptions extends EventSourceInit {
...
 }
 
-export interface UseEventSourceReturn<Events extends string[]> {
+export interface UseEventSourceReturn<Events extends string[], Data = any> {
   /**
    * Reference to the latest data received via the EventSource,
    * can be watched to respond to incoming messages
    */
-  data: Ref<string | null>
+  readonly data: Readonly<ShallowRef<Data>>
```
OrbisK, line 58: ~~Actually, it looks like there is a difference between `readonly data: ...` and `data: Readonly<...>'. 

https://www.typescriptlang.org/play/?#code/C4TwDgpgBAglC8UDeAoKUBOECGATA9gHYA2IUARgFzJrpQDG1AzsBgJaEDmtAvinylCQoAIQQ10VKACUcBEiAA8qOg2asO3dDwB8-FCgBmAV0L1gbIlGAQWMABTZqMAJQSo2AHTlP9cQCIbFn8ofSNTc0tCa1tgEXspETcVCm9fAKDgEP4gA

The return type of `shallowReadonly` is `Readonly<T>`, not sure if this matters.~~

looks like you already did. but do we need still the readonly prefix?

43081j: there is a difference. it is why I wrote both

43081j: `readonly` - you can't re-assign the property in the returned object
`Readonly` - you can't mutate the shallow ref

OrbisK: so we always want to use both? :)

OrbisK (approved):

packages/core/useEventSource/index.ts
43081j: pretty much and afaict, `shallowReadonly` has special treatment for if you give it a ref, so it makes the ref read-only and throws a warning if you change it

and `readonly` will return a `Readonly<Ref<T>>` rather than a `Readonly<ShallowRef<T>>` (doesn't matter much but means we lose the shallow type)

PR #4613, opened 2025-02-19
fix(useScroll): handle negative scroll values

ilyaliao:

<details>
<br>

>

</details>

---


EvgenyWas (approved):

ilyaliao:
> 1. we should add an example to the docs for how to use direction top

Agree, maybe we can add it in this PR.

> 2. we should evaluate if we can/should resprect:

Humm, I think a CSS solution would be simpler. At least we can more easily explain to users "how to use it."

packages/core/useInfiniteScroll/index.md
```diff
@@ -50,6 +50,17 @@ function resetList() {
...
+## Direction
+
+Different scroll directions require different CSS style settings:
+
+| Direction          | Required CSS                                          |
+| ------------------ | ----------------------------------------------------- |
+| `bottom` (default) | No special settings required                          |
+| `top`              | `display: flex;`<br>`flex-direction: column-reverse;` |
+| `left`             | `display: flex;`<br>`flex-direction: row-reverse;`    |
+| `right`            | `display: flex;`                                      |
```
OrbisK, line 62: if bottom does not need settings, why does right? 😄

ilyaliao: because the elements must first be arranged horizontally.

Galllarate:
Hi, sorry I don't know how all this process in open source work, but how long can it take? I made an issue two weeks ago and as I saw here, the fix has already been implemented but not merged.

OrbisK (approved):

Azurox:
Unsure if I'm doing something wrong but it seems that the value `distance` is not respected when using the top direction

**Repro**
[https://codesandbox.io/p/devbox/staging-currying-zyqr7s](https://codesandbox.io/p/devbox/nostalgic-browser-qpwnh4?workspaceId=ws_HgCzJZgijHVi48nTgLc2XT)

Issue #4614, opened 2025-02-19
BUG | `onStartTyping` | Incorrect accepted valid characters


Bernard-Borg:
### Describe the bug

When using `onStartTyping`, multiple valid keys are not accepted. For example, 0 is accepted, but Numpad0 isn't (and then, Numpad1 is).

The issue seems to stem from the following function.

```js
function isTypedCharValid({
  keyCode,
  metaKey,
  ctrlKey,
  altKey,
}: KeyboardEvent) {
  if (metaKey || ctrlKey || altKey)
    return false

  // 0...9
  if (keyCode >= 48 && keyCode <= 57)
    return true

  // A...Z
  if (keyCode >= 65 && keyCode <= 90)
    return true

  // a...z
  if (keyCode >= 97 && keyCode <= 122)
    return true

  // All other keys.
  return false
}
```

JavaScript keyCode 97 - 122 is not a...z, but rather Numpad1 to F11. 
Please note that I linked the documentation as a reproduction as this behaviour can be observed on the onStartTyping DEMO.

### Reproduction

https://vueuse.org/core/onStartTyping/

### System Info

```Shell
System:
    OS: Windows 10 10.0.20348
    CPU: (6) x64 Intel(R) Xeon(R) Gold 6230 CPU @ 2.10GHz
    Memory: 4.07 GB / 24.00 GB
  Binaries:
    Node: 23.7.0 - C:\nvm4w\nodejs\node.EXE
    Yarn: 4.6.0 - C:\nvm4w\nodejs\yarn.CMD
    npm: 10.9.2 - C:\nvm4w\nodejs\npm.CMD
  Browsers:
    Internet Explorer: 11.0.20348.2849
  npmPackages:
    @vueuse/core: ^12.7.0 => 12.7.0
    vue: 3.5.13 => 3.5.13
```

### Used Package Manager

yarn

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vueuse.org/guide).
- [x] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #2932, opened 2023-04-04
fix(onStartTyping): fix char ascii

meenie-net:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

`0-9` annotation did not match it's owner statement. And should consider add `if` of `A-Z`.

### Additional context

Nothing

---

### <samp>🤖 Generated by Copilot at 9a80dc6</samp>

Improved the `onStartTyping` hook to accept lowercase letters as valid inputs. Modified the `isTypedCharValid` function in `packages/core/onStartTyping/index.ts` to check for both uppercase and lowercase letters.

### <samp>🤖 Generated by Copilot at 9a80dc6</samp>

* Allow lowercase letters as valid typed characters in `isTypedCharValid` function ([link](https://github.com/vueuse/vueuse/pull/2932/files?diff=unified&w=0#diff-6f8b89b356b64f48bdacac50d856f35df86cb2671b234e83f3c39846038cb3d0L37-R47))

PR #4616, opened 2025-02-19
fix(onStartTyping): Incorrect accepted valid characters

Bernard-Borg:

---

### Description

This PR fixes the characters which are accepted when using onStartTyping. Specifically, it adds back the ability for users to start typing with Numpad0 to focus on the connected element (#4614, related: #2932).

### Additional context

Should more characters be accepted or are just the basic letters and numbers enough?

### DIFF
--- a/packages/core/onStartTyping/index.test.ts
+++ b/packages/core/onStartTyping/index.test.ts
@@ -1,0 +1,56 @@
+import type { Ref } from 'vue'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { shallowRef } from 'vue'
+import { onStartTyping } from './index'
+
+describe('onStartTyping', () => {
+  let element: Ref<HTMLInputElement>
+  let callBackFn: any
+
+  beforeEach(() => {
+    element = shallowRef(document.createElement('input'))
+    element.value.tabIndex = 1
+    callBackFn = vi.fn()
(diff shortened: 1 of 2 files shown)


packages/core/onStartTyping/index.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  let element: Ref<HTMLInputElement>
+  let callBackFn: any
+
+  beforeEach(() => {
+    element = shallowRef(document.createElement('input'))
+    element.value.tabIndex = 1
+    callBackFn = vi.fn()
+  })
+
+  function createKeyDownEvent(keyCode: number) {
```
43081j, line 16: just a tiny nit, can you call this `dispatchKeyDownEvent` to make it clear it not only creates it, but also dispatches it

43081j (changes requested):
can you add one more test 👀  a negative

e.g. if you type arrow keys or some such thing, we _do not_ trigger the callback

packages/core/onStartTyping/index.test.ts
Bernard-Borg: Done

43081j (approved):

OrbisK (approved):

ilyaliao (approved):

PR #4617, opened 2025-02-20
feat(useObjectVModel): Implement new useObjectVModel composable to allow update of an object passed through v-model in child components

michaelcozzolino:

<details>
<br>

>

</details>

---

### Description

At the moment vue does not provide any way to update the property of an object passed through props by using v-model without breaking the parent -> child data flow. even by using defineModel
in the child component as in the following:

```
const obj = defineModel<{propertyA: string; propertyB: string}>({ required: true })
</script>

<template>
    <div>
        <input v-model="obj.propertyB">
    </div>
</template>
```

it would break the data flow. (https://vuejs.org/guide/components/props.html#mutating-object-array-props)

In order to solve this problem i created a composable `useObjectVModel` that takes a `modelRef` as parameter (obtainable thanks to the `defineModel()` macro) and leverages the js proxy to update the whole object with the same data as before plus the changed one for each property by triggering an update on the  `modelRef ` that will automatically trigger an emit to update the data in the parent component.

The idea has been taken from https://skirtles-code.github.io/vue-examples/patterns/computed-v-model
I just made a composable out of it.

Probably the same outcome can be achieved with the `useVModel` but i think it would be more expensive when used with nested properties in big objects as there should be a deep watcher involved.

### Additional context

If the feature gets approved, I will add docs and tests.

### DIFF
--- a/packages/core/useObjectVModel/index.ts
+++ b/packages/core/useObjectVModel/index.ts
@@ -1,0 +1,23 @@
+import type { ModelRef, WritableComputedRef } from 'vue'
+import { computed } from 'vue'
+
+export function useObjectVModel<T extends object, M extends PropertyKey = string, G = T, S = T>(
+  obj: ModelRef<T, M, G, S>,
+): WritableComputedRef<T> {
+  const o = computed({
+    get: (): T => {
+      return new Proxy(obj.value as unknown as T, {
+        set(target, propertyName, newValue): boolean {
+          o.value = { ...target, [propertyName]: newValue }
+
+          return true
+        },
+      })
+    },
+    set: (newValue: T): void => {
+      obj.value = newValue as unknown as S
+    },
+  })
+
+  return o
+}
--- a/packages/core/useObjectVModel/index.ts
+++ b/packages/core/useObjectVModel/index.ts
@@ -4,14 +4,39 @@ import { computed } from 'vue'
 export function useObjectVModel<T extends object, M extends PropertyKey = string, G = T, S = T>(
   obj: ModelRef<T, M, G, S>,
 ): WritableComputedRef<T> {
+  function createDeepProxy(target: T, update: (newValue: any) => void): T {
+    return new Proxy(target, {
+      get(target: T, p, receiver) {
+        const value = Reflect.get(target, p, receiver)
+
+        if (typeof value === 'object') {
+          return createDeepProxy(value as T, (newChildValue: any) => {
+            const targetCopy = Array.isArray(target) ? [...target] : { ...target }
+
+            targetCopy[p] = newChildValue
+            update(targetCopy)
+
+            return newChildValue
+          })
+        }
+
+        return value
+      },
+      set(target, p, newValue, receiver) {
+        const targetCopy = Array.isArray(target) ? [...target] : { ...target }
+
+        targetCopy[p] = newValue
+        update(targetCopy)
+
+        return Reflect.set(target, p, newValue, receiver)
+      },
+    })
+  }
+
   const o = computed({
     get: (): T => {
-      return new Proxy(obj.value as unknown as T, {
-        set(target, propertyName, newValue): boolean {
-          o.value = { ...target, [propertyName]: newValue }
-
-          return true
-        },
+      return createDeepProxy(obj.value as unknown as T, (newValue: T) => {
+        o.value = newValue
       })
     },
     set: (newValue: T): void => {


OrbisK:
I like the idea, but lets talk about some thoughts:

1. what happens if we use `useObjectVModel` in the child and access it from the parent like: `<Child :model-value="obj" />`. Will this still break the dataflow?
2. this should work for array too, right?
3. I do not like that we cannot know if `v-model` on `<Child>` can now have two behaviours. Maybe (if possible) we should do this as a directive only, like 
Parent:
```html
<Child v-deep-model="obj" />
<!-- If possible: -->
<Child v-deep-model:my-model-value="obj" />
```
This will always indicate that we can have a "deep emit".

michaelcozzolino:
> I like the idea, but lets talk about some thoughts:
> 
> 1. what happens if we use `useObjectVModel` in the child and access it from the parent like: `<Child :model-value="obj" />`. Will this still break the dataflow?
> 2. this should work for array too, right?
> 3. I do not like that we cannot know if `v-model` on `<Child>` can now have two behaviours. Maybe (if possible) we should do this as a directive only, like
>    Parent:
> 
> ```
> <Child v-deep-model="obj" />
> 
> <Child v-deep-model:my-model-value="obj" />
> ```
> 
> This will always indicate that we can have a "deep emit".
> 
> Let me know what you think and if I got anything wrong. 😄

1. That is allowed by vuejs itself as stated here https://github.com/vuejs/rfcs/discussions/503#discussioncomment-7828483

> If the parent provides the necessary v-model prop, the child ref value will always be in sync with parent prop value.
Otherwise, the child ref will behave like a normal local ref.

2. right, indeed i noticed that my implementation was not working with arrays, so i updated it, now the proxy gets created deeply

3. As already stated in 1 it is a vuejs feature, but i took into account your proposal and here is my new proposal

We implement both the composable and the directive where

The Composable
1. is called `useVDeepModel`
2. it's basically the one i already implemented
3. can be used like:

```
// types.ts

export interface Person {
    name:     string;
    surname:  string;
    children: Person[];
}
```

```
<!-- Parent.vue -->

<template>
<Child v-model="person"  />
</template>

<script setup lang="ts">
import type { Person }   from './types';
import { ref } from 'vue';

const person = ref<Person>({
    name:     'michael',
    surname:  'cozzolino',
    children: [
        {
            name:     'first',
            surname:  'child',
            children: [],
        },
        {
            name:     'second',
            surname:  'child',
            children: [],
        },
    ],
});
```

```
<!-- Child.vue -->

<template>
    <div>
        <input v-model="editablePerson.name">
        <input v-model="editablePerson.surname">
        <input v-model="editablePerson.children[0].name">
    </div>
</template>

<script setup lang="ts">
import type { Person }     from './types';
import { useVDeepModel } from './useVDeepModel';

const person = defineModel<Person>({ required: true });
const editablePerson = useVDeepModel(person);
</script>
```


The directive
1. is called `v-deep-model`
2. does exactly the same job as the composable, but instead of using that the user will only have to provide the v-deep-model on the child component and a normal v-model in the parent with the following api:

- [x] `v-model object syntax`: `v-deep-model:property-name="object"` to modify the property of an object 
- [ ] `v-model array syntax`: `v-deep-model:property-name="[object]"` to modify the property of an object ( for consistency with the next one)
- [x] `v-model array syntax with indexable object`: `v-deep-model:array-property-name:property-name="[object, index]"` to modify the property of an object being an element of an array belonging to the outer object
- [ ]  `shorthand v-model array syntax with indexable object`:  `v-deep-model:property-name="[object.arrayPropertyName, index]"` to modify the property of an object being an element of an array belonging to the outer object, where the array object is passed directly as the first value of the v-model array instead of being retrieved through the v-model args

The checked syntaxes have already been implemented in the following proof of concept, of course by being a POC i just wanted it to work with the most minimal setup, therefore i only implemented one (buggy) hook of the directive and nothing more, just to see if it would work, and indeed it seems it does, Moreover i cannot say that i can fully achieve the full api as i'm not so much expert with vuejs directives but I can try as much as i can.

```
export const vDeepModel = {
    created(el: Element, { value, arg }, vnode) {
        if (Array.isArray(value)) {
            const index                = value[1];
            const [arrayProp, objProp] = arg.split(':');
            el.value = value[0][arrayProp][index][objProp];
        }
        else {
            el.value = value[arg];
        }

        el.addEventListener('input', () => {
            console.log('something changed', arg, el, value, vnode, el.value);
            const newValue = el.value;

            if (Array.isArray(value)) {
                const index                = value[1];
                const [arrayProp, objProp] = arg.split(':');
                const newArr = value[0][arrayProp];
                newArr[index][objProp] = newValue;

                vnode.ctx.emit('update:modelValue', { ...value[0], [arrayProp]: newArr });
            }
            else {
                vnode.ctx.emit('update:modelValue', { ...value, [arg]: newValue });
            }
        });
    },

    updated(el: Element, { value }, vnode) {
        console.log('updated: ', value);
    },
};
```

3. can be used like:

```
// types.ts

export interface Person {
    name:     string;
    surname:  string;
    children: Person[];
}
```

```
<!-- Parent.vue -->

<template>
<Child v-model="person"  />
</template>

<script setup lang="ts">
import type { Person }   from './types';
import { ref } from 'vue';

const person = ref<Person>({
    name:     'michael',
    surname:  'cozzolino',
    children: [
        {
            name:     'first',
            surname:  'child',
            children: [],
        },
        {
            name:     'second',
            surname:  'child',
            children: [],
        },
    ],
});
```

```
<!-- Child.vue -->

<template>
    <div>
        <!-- v-model object syntax -->
        <input v-deep-model:name="editablePerson">

        <!-- v-model array syntax -->
        <input v-deep-model:name="editablePerson">

        <!-- v-model array syntax with indexable object -->
        <input v-deep-model:children:name="[editablePerson, index]">

        <!-- shorthand  v-model array syntax with indexable object -->
        <input v-deep-model:name="[editablePerson.children, index]">
    </div>
</template>

<script setup lang="ts">
import type { Person }     from './types';

const person = defineModel<Person>({ required: true });
</script>
```

michaelcozzolino:
before starting i actually proposed an RFC for Vue, https://github.com/vuejs/rfcs/discussions/751 who knows it might get approved.

michaelcozzolino:
After many attempts in trying to make a draft for the directive I have to say that, at least to me and my knowledge, it's almost impossible to achieve it, as working with references(objects) will always end up in mutating the original object directly through the directive, even destructuring recursively and assigning the new value ended up in side effects when working with more than 1 input field. Same as for the composable, it works good if used in a non recursive way, meaning that you have to pass into it only the root object, if you want to edit a sub object you will have to create another component and pass through it the sub object, because passing the sub object directly in the composable and working with that might still mutate the property of a prop without realizing it, and it's very difficult if not impossible to ensure that it won't happen. So, to conclude, if you need to mutate deeply nested object properties through props, the best option is to go with a pinia/vuex store and there will be neither problems with reactivity nor with vue props mutation violations.

PR #4619, opened 2025-02-22
fix(useAnimate): Correct condition for updating keyframes

Mrlilili:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

When keyframes is a ref value, after updating keyframes, the keyframes held by useAnimate is not updated.

related: https://github.com/vueuse/vueuse/issues/4612

PR #4622, opened 2025-02-24
test(useAnimate): add test for keyframe reactivity

43081j:
Adds a test for asserting that keyframes can be passed as a ref and will trigger an update when they change.

Depends on #4619 

### DIFF
--- a/packages/core/useAnimate/index.browser.test.ts
+++ b/packages/core/useAnimate/index.browser.test.ts
@@ -37,4 +37,42 @@ describe('useAnimate', () => {
     })
     wrapper.unmount()
   })
+
+  it('should support keyframes refs', async () => {
+    const keyframes = shallowRef<PropertyIndexedKeyframes>({
+      transform: 'rotate(360deg)',
+    })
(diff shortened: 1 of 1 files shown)


packages/core/useAnimate/index.browser.test.ts
```diff
@@ -37,4 +37,42 @@ describe('useAnimate', () => {
     })
     wrapper.unmount()
   })
+
+  it('should support keyframes refs', async () => {
```
OrbisK, line 41:
```suggestion
  it.fails('should support keyframes refs', async () => {
```
If we commit it as failing we can merge this first into main and than into https://github.com/vueuse/vueuse/pull/4619 to see if this is working :)

43081j: it is. i had the changes from the other PR locally to prove it out

but i don't mind either way. we can wait, or i can add `fails` and we do a follow-up PR. it may be worth just waiting tbh given we know the fix in the other PR does fix it

OrbisK (approved):

Issue #4564, opened 2025-02-05
`refManualReset` | reset ref to inital value


isimehmeti:
### Clear and concise description of the problem

The VueUse library currently provides refAutoReset, which resets a ref’s value to its initial state after a specified timeout. However, there are cases where a manual reset is needed instead of an automatic one.

I propose adding a refManualReset utility that allows resetting a ref to its initial value on demand.

Use Cases

- Restoring a ref to its default state after an action is completed.
- Managing UI states that need to be reset programmatically, such as form inputs or temporary status messages.
- Avoiding workarounds like storing the initial value separately.

### Suggested solution

```js
const foo = refManualReset('bar')

foo.value // Normal getter
foo.reset() // Resets to the initial value ('bar')
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vueuse.org/guide).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Issue #1479, opened 2022-04-05
Proposal: refManualReset


pabueco:
### Clear and concise description of the problem

As a developer using VueUse I might want a `refManualReset` function so that I can initialize a ref with a value and reset it later to that initial value.

While building some browser games I constantly had the case that I needed some initial values for variables and had to reset **some** of them later to their initial value. So simply moving the initialization of these variables into a function wasn't really a nice option. Instead I created a really simple function that uses `extendRef` to add the `reset()` method. It works really well for the use cases I have so I wanted to see if this would be helpful to others as well.

#### Usage
```typescript
const myVariable = refManualReset(5)
myVariable.value = 10
myVariable.reset() // value is 5 again
```

### Suggested solution

A really simple implementation could look like this:

```typescript
import { extendRef } from '@vueuse/shared'
import { ref, UnwrapRef } from 'vue'

export function refManualReset<T>(defaultValue?: T): Ref<T> {
  const valueRef = ref<T>(defaultValue)

  const reset = () => {
    valueRef.value = defaultValue as UnwrapRef<T>
  }

  const valueExtendedRef = extendRef(valueRef, {
    reset,
  })

  return valueExtendedRef
}
```

Obviously that would create issues when using it for anything else than primitive values. So to support objects and arrays some kind of cloning would be needed. This could be implemented like in `useManualRefHistory`.

If this idea is something that would make sense to be added I would be happy to create a PR with a better implementation.

### Alternative

You can of course just use a different variable/constant that holds the value and use that to manually change the variable when it needs to be reset, but I found it much cleaner and clearer to just call `myVariable.reset()` instead.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

okxiaoliang4:
Can `useManualRefHistory` meet your needs.

```ts
import { ref } from 'vue'
import { useManualRefHistory } from '@vueuse/core'

const myVariable = ref(0)
const { history, commit, undo, redo } = useManualRefHistory(myVariable)
commit()
myVariable.value = 10

undo()
// myVariable 0
```

Or maybe we could add an option `immediate`, like this:

```ts
[...]

PR #4644, opened 2025-03-05
feat(refManualReset): new function

phk422:
Close: #4564


ferferga:
Instead of a new function... Why we don't adapt the current one so, when an ``undefined`` or ``null`` timeout is passed, it does what's proposed in the issue?

phk422:
> Instead of a new function... Why we don't adapt the current one so, when an `undefined` or `null` timeout is passed, it does what's proposed in the issue?

Thank you for your suggestion! I initially considered modifying the resetAutoRest method to support manual mode, but since its semantics are clearly tied to "Auto," making direct changes might compromise its consistency. Therefore, I opted to add a separate method instead.

Additionally, we can introduce another method, such as resetRest, and use parameters to differentiate between automatic and manual modes, as you mentioned. This approach maintains flexibility while avoiding semantic confusion.

If you agree with this direction, I can further refine the solution and submit an update.

chojnicki:
> Considering that the semantics of `refAutoRest` are clearly tied to "Auto," directly modifying it might not be appropriat

I totally agree with that. I'm actually surprised that there is function for auto option, but not manual one (`refDefault` is something different) :P In my opinion ideally there should be single function for both, where auto with timer would be just optional option. If not, `refManualReset` is still good option. But adding manual option for `refAutoReset` is really not intuitive.

Rhilip:
When defaultValue is object, the reset function will not work.

try [this example](https://play.vuejs.org/#eNp9U8mO2zAM/RVClyRA4KBoT6kTdMEAbYEumG4XX1SbdjwjS4KWNIXhfy9Fj92ksxwMW+R7j48U3YvX1mbHiGIrcl+61gbwGKIFJXWzK0TwhdgXrtBtZ40L0IPDGqSHCtFeY72GYH5IFREGqJ3pYEFii0RITx11GVqjE+mj1FGqayT5ZYW1jCowcQV9ggKURvsARxbbTbKX0HOgw0CwOxtLpq2SMan/vBxxhMhcqke45Qp2+6nSmJoq8XtMDDMxumQ6pDMH0zPW7XxDpP8a6gP6sIXFO1TKwE/jVLUY2O7FGKKtZMDkd+6a5EYnWZIg5UWH3ssG4UDNkDYPc/ZwNlHK3VPi6HLFA5g5+Wa8Wb5HDgTsrCInHAHID8/2fc+dDUO+odNd/FcMgWq9KlVb3tI2zP5pKb7zN9Ah34y4x1iT1ULAnsd1j5RvLhyJNe0dTbtum+zGG03L2SflQpSms61C99mmIdBubql7vtFCSBr97w8cCy7ieoqXByxvH4jf+FOKFeJLMuiOWIg5F6Rr6PY5ffX1E57oe052poqK0E8kqU2jYvI4wt5EXZHtMxy7fc8/Vaubb/7qFFD7qalklBeS8YWgn+rtE63/s/s8ezEush7E8BfBG1Uy) 

image

ilyaliao:
I’d personally like `refManualReset` to mirror `refAutoReset` as much as possible. 

They can live as separate composables, but their API and internal structure should be very similar, not two totally unrelated implementations.

phk422:
> I’d personally like `refManualReset` to mirror `refAutoReset` as much as possible.
> 
> They can live as separate `composables`, but their API and internal structure should be very similar.

I have updated it. Kindly review.

jeffreyvdhondel:
I would really love this feature for reseting an ref back to its default value.

packages/shared/refManualReset/index.ts
```diff
@@ -0,0 +1,50 @@
+import type { MaybeRefOrGetter, Ref } from 'vue'
+import type { Fn } from '../utils'
```
ilyaliao, line 2:
```suggestion
import type { Fn } from '@vueuse/shared'
```

ilyaliao (changes requested):

packages/shared/refManualReset/index.ts
phk422: Base to the ESLint configuration rule: 'import from "@vueuse/shared" is restricted from being used', therefore, here it is changed to import from "@vueuse/core" instead.

packages/shared/refManualReset/index.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Create a ref with manual reset functionality.
+ *
+ * @see https://vueuse.org/refManualReset
+ * @param defaultValue The value which will be set.
+ */
+export function refManualReset<T>(defaultValue: MaybeRefOrGetter<T>) {
+  let value: T = toValue(defaultValue)
+  let trigger: () => void
```
ilyaliao:
```suggestion
  let trigger: Fn
```

ilyaliao (changes requested):

packages/shared/refManualReset/index.ts
ilyaliao: Oops... I realize I was wrong, you're correct - it should import from '../utils' instead. Sorry for the confusion!

ilyaliao (dismissed):
I think this feature makes a lot of sense, and providing both `refAutoReset` and `refManualReset` is a good choice.

Here’s more context:
#1479  
#4564

phk422:
Hi @antfu 👋

I'm just following up on this PR which was recently approved. Whenever you have a moment, could you please take a look and consider merging it?

I'm really excited to contribute to VueUse and believe the function will be a handy

packages/shared/refManualReset/index.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+  }) as ManualResetRefReturn<T>
+
+  refValue.reset = reset
+
+  return refValue
+}
+
+// alias
+export { refManualReset as manualResetRef }
```
antfu: Not sure about if this alias is needed tho

antfu:
OK, I am good to have this new function if people found it useful

packages/shared/refManualReset/index.ts
phk422: I noticed in this PR（#1325） that 'Normalize the function names with the backward compact alias', which is why I wrote it this way. I'm not sure if this is due to historical reasons.

OrbisK: @phk422 https://github.com/vueuse/vueuse/pull/5009

phk422: It seems this is not needed here.

OrbisK (approved):

9romise (commented):
How about including some test cases as well?

9romise (approved):

isimehmeti:
A deep ref option would have been nice, as it’s especially useful for forms or lists.

PR #4650, opened 2025-03-05
chore: switch to yaml for tests

43081j:
Switches to the `yaml` package for tests, since `js-yaml` is no longer maintained and does not support the latest yaml spec.

Note this is low priority since we only use it in tests, but we may as well do the clean up still.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "@rollup/plugin-replace": "catalog:",
     "@shikijs/vitepress-twoslash": "catalog:",
     "@type-challenges/utils": "catalog:",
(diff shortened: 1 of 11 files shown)


antfu:
Ah, I see. There are still other places we use js-yaml, can you help to replace them as well? Thanks

43081j:
Ah my mistake! I thought I found all usages. Looks like I missed a few 😬

I'll take a look soon 👍

43081j:
all usages should be replaced now 👍

43081j:
Bah. Looks like I missed one! Will update 👍

antfu (approved):

PR #4653, opened 2025-03-07
feat(useAxios): Add types and callback to handle AxiosError

michaelcozzolino:

<details>
<br>

>

</details>

---

### Description

This PR introduces a generic type system for the AxiosError, in detail:
- When an Error is thrown by axios we cannot ensure that it is an AxiosError, so I added an `axiosError` ref that will be populated only when the `Error` is an instance of `AxiosError`, furthermore when the error is not AxiosError, the `error` ref will still be populated, and I also added a callback (`onAxiosError`) that will be executed only when we have an `AxiosError`, if that is the case, also the `onError` callback will be executed.
- Now that we have the `AxiosError`, basically, according to me, the developer is mainly looking for the `data` property of the error response, and, also for consistency with the `data` property of a successful `AxiosResponse`,  I added a computed property `axiosErrorResponseData` that returns exactly the data of the error response.

Example usage
```ts
// Some utility types
type DefaultStrictUseAxiosReturn<Response, Payload = unknown, AxiosErrorResponse = unknown> = StrictUseAxiosReturn<
    Response,
    AxiosResponse<Response>,
    Payload,
    AxiosErrorResponse,
    UseAxiosOptionsWithInitialData<Response>
>

export type PromiseLikeStrictUseAxiosReturn<Response, Payload = unknown, AxiosErrorResponse = unknown> =
    Promise<DefaultStrictUseAxiosReturn<Response, Payload, AxiosErrorResponse>>
    & DefaultStrictUseAxiosReturn<Response, Payload, AxiosErrorResponse>

interface MyResponse {
    id: number,
    name: string
}

interface MyPayload {
    name: string
}

interface ErrorResponse {
    errors: { name: string[] }
}

function updateApiData(options: UseAxiosOptionsWithInitialData<MyResponse, ErrorResponse>): PromiseLikeStrictUseAxiosReturn<MyResponse, MyPayload, ErrorResponse> {
    return useAxios(
        '/something',
        { method: 'put' },
        options,
    )
}

const { execute, axiosError, axiosErrorResponseData, data } = updateApiData({
    initialData: { id: 1, name: 'Michael' },
  onAxiosError: (e) => {
    console.log(e?.response?.data.errors.name)
  },
})

```

### Additional context

If the feature is accepted, I will add tests. There is no breaking change in the composable, but only in the types, as it seems unavoidable.
Let me know if the naming is fine, I think they are pretty descriptive, but I'm looking forward to suggestions, if any.

### DIFF
--- a/packages/integrations/useAxios/index.ts
+++ b/packages/integrations/useAxios/index.ts
@@ -98,7 +98,7 @@ export interface UseAxiosOptionsBase<T = any, AxiosErrorResponseData = unknown> 
   /**
    * Callback when axios error is caught.
    */
-  onAxiosError?: (e: AxiosError<AxiosErrorResponseData>) => void
+  onAxiosError?: (e: AxiosError<AxiosErrorResponseData> | undefined) => void
 
   /**
    * Callback when success is caught.
--- a/packages/integrations/useAxios/index.ts
+++ b/packages/integrations/useAxios/index.ts
@@ -1,10 +1,10 @@
 import type { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios'
-import type { Ref, ShallowRef } from 'vue'
+import type { ComputedRef, Ref, ShallowRef } from 'vue'
 import { noop, until } from '@vueuse/shared'
 import axios, { AxiosError } from 'axios'
-import { ref as deepRef, shallowRef } from 'vue'
+import { computed, ref as deepRef, shallowRef } from 'vue'
 
-export interface UseAxiosReturn<T, R = AxiosResponse<T>, _D = any, O extends UseAxiosOptions = UseAxiosOptions<T>> {
+export interface UseAxiosReturn<T, R = AxiosResponse<T>, _D = any, AxiosErrorResponseData = unknown, O extends UseAxiosOptions<T, AxiosErrorResponseData> = UseAxiosOptions<T, AxiosErrorResponseData>> {
   /**
    * Axios Response
    */
@@ -13,7 +13,9 @@ export interface UseAxiosReturn<T, R = AxiosResponse<T>, _D = any, O extends Use
   /**
    * Axios response data
    */
-  data: O extends UseAxiosOptionsWithInitialData<T> ? Ref<T> : Ref<T | undefined>
+  data: O extends UseAxiosOptionsWithInitialData<T, AxiosErrorResponseData> ? Ref<T> : Ref<T | undefined>
+
+  axiosErrorData: ComputedRef<AxiosErrorResponseData | undefined>
 
   /**
    * Indicates if the request has finished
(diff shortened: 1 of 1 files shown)


michaelcozzolino:
@antfu @OrbisK any feedback?

packages/integrations/useAxios/index.ts
```diff
@@ -50,19 +57,19 @@ export interface UseAxiosReturn<T, R = AxiosResponse<T>, _D = any, O extends Use
    */
   isCanceled: Ref<boolean>
 }
-export interface StrictUseAxiosReturn<T, R, D, O extends UseAxiosOptions = UseAxiosOptions<T>> extends UseAxiosReturn<T, R, D, O> {
+export interface StrictUseAxiosReturn<T, R, D, AxiosErrorResponseData = unknown, O extends UseAxiosOptions<T, AxiosErrorResponseData> = UseAxiosOptions<T, AxiosErrorResponseData>> extends UseAxiosReturn<T, R, D, AxiosErrorResponseData, O> {
```
OrbisK, line 60: this type is breaking

michaelcozzolino: @OrbisK  hello, some time has passed, what is the status of the review? would it be possible to merge it as a breaking change after adding tests? I'm relying a lot on useAxios and this feature is really needed and i also think that many other people need it

43081j:
we should be able to simplify this a whole bunch.

i don't think you need to introduce a new type parameter at all.

an axios error is basically typed to the request and response (lets call them `Request` and `Response`). that means you can call `isAxiosError<Response, Request>` using our existing type parameters (`T` and `D` - `isAxiosError<T, D>`).

although im not entirely sure about this change anyway. we already expose the error, we're just doing this so we can filter down to axios errors i suppose?

michaelcozzolino:
> we should be able to simplify this a whole bunch.
> 
> i don't think you need to introduce a new type parameter at all.
> 
> an axios error is basically typed to the request and response (lets call them `Request` and `Response`). that means you can call `isAxiosError<Response, Request>` using our existing type parameters (`T` and `D` - `isAxiosError<T, D>`).
> 
> although im not entirely sure about this change anyway. we already expose the error, we're just doing this so we can filter down to axios errors i suppose?

yes and to be sure that whenever an error is thrown it is related to the response and not to the fact that axios caught an error that is coming from somewhere else E.G the developer uses an interceptor and and a non axios error is thrown there.

43081j:
I think you misunderstood a little. I was telling you two things: 

1. In this change, you don't need a new type parameter (for the reasons I mentioned). You can still introduce the new option and return property without a new type parameter
2. If the change is sensible or not

michaelcozzolino:
> I think you misunderstood a little. I was telling you two things:
> 
> 1. In this change, you don't need a new type parameter (for the reasons I mentioned). You can still introduce the new option and return property without a new type parameter
> 2. If the change is sensible or not

ok i think i understood, i will update the pr soon

michaelcozzolino:
> I think you misunderstood a little. I was telling you two things:
> 
> 1. In this change, you don't need a new type parameter (for the reasons I mentioned). You can still introduce the new option and return property without a new type parameter
> 2. If the change is sensible or not

Updated with your suggestions if I understood correctly. so basically now the final data received from a request could be something like
MyResponse | MyErrorResponse
because we are obliged to assign that error to the same data variable. furthermore I am not sure if we have to assign the error data it only on axios error or any errors, probably the second case is more suitable. If this is the case, also this is a breaking change, indeed a test now fails. Let me know if the approach i applied is right

PR #4658, opened 2025-03-12
feat(useMagicKeys): add `onRegisteredEventFired` option

innocenzi:
This pull request adds a `onRegisteredEventFired` option to `useMagicKeys`. The `onEventFired` callback is fired for all keydown events, while `onRegisteredEventFired` is only fired for registered key presses.

```ts
const { Ctrl_P, Slash } = useMagicKeys({
	passive: false,
	onRegisteredEventFired(e) {
		e.preventDefault() // only called for Ctrl+P and Slash, without the need for a condition
	},
})
```

### DIFF
--- a/packages/core/useMagicKeys/index.ts
+++ b/packages/core/useMagicKeys/index.ts
@@ -37,6 +37,13 @@ export interface UseMagicKeysOptions<Reactive extends boolean> {
   passive?: boolean
 
   /**
+   * Custom event handler for registered keydown/keyup event.
+   *
+   * When using `e.preventDefault()`, you will need to pass `passive: false` to useMagicKeys().
+   */
+  onRegisteredEventFired?: (e: KeyboardEvent) => void | boolean
+
+  /**
    * Custom event handler for keydown/keyup event.
    * Useful when you want to apply custom logic.
    *
@@ -76,6 +83,7 @@ export function useMagicKeys(options: UseMagicKeysOptions<boolean> = {}): any {
     target = defaultWindow,
     aliasMap = DefaultMagicKeysAliasMap,
     passive = true,
+    onRegisteredEventFired = noop,
     onEventFired = noop,
   } = options
   const current = reactive(new Set<string>())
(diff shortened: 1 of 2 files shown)


packages/core/useMagicKeys/index.ts
```diff
@@ -36,6 +36,13 @@ export interface UseMagicKeysOptions<Reactive extends boolean> {
    */
   passive?: boolean
 
+  /**
+   * Custom event handler for registered keydown/keyup event.
+   *
+   * When using `e.preventDefault()`, you will need to pass `passive: false` to useMagicKeys().
+   */
+  onRegisteredEventFired?: (e: KeyboardEvent) => void | boolean
```
OrbisK, line 44: why do we need `boolean` 😄  ?

innocenzi: We don't! It's a typo. Good catch. 😅

OrbisK (approved):

ilyaliao (approved):

43081j (changes requested):
this doesn't seem right

basically the way it works right now is we listen for all keyup/keydown events, and we call `onEventFired` once one of them fires.

_later_, we use a `Proxy` to know which events you care about. we then add those to `refs` (like `refs['A'] = false`, i.e. the `A` key isn't pressed yet).

this code you've added just means it'll do exactly what `onEventFired` does but only for `keydown` events (`value` is just if it was pressed or not).

i think what you actually need to do is update `setRefs` (or do the same check in `updateRefs`), as that is where we actually check that the key you pressed is one of interest (via `if (key in refs)`).

basically, in that condition (or repeat that condition inside `updateRefs`), you would want to do something like this:

```ts
if (key in refs) {
  if (onRegisteredEventFired) {
    onRegisteredEventFired(e); // will be called for keyup and keydown
  }
}
```

your test should actually be dispatching two events:

1. `A` (because thats what you're observing)
2. some other key, e.g. `B` (which we're not observing)

then assert that our callback was only called for `A`

PR #4659, opened 2025-03-14
feat(shared): ensure return types exists

OrbisK:

<details>
<br>

>

</details>

---

### Description

This PR ensures consitent (naming + export) return types for "@vueuse/shared" and "fixes" some minor type issues along the way.

The following composables have still some open questions:

- injectLocal (better type function, to support generics)
- get
- createInjectionState
- isDefined? (typeguard)


packages/shared/createRef/index.ts
```diff
@@ -1,6 +1,8 @@
 import type { Ref, ShallowRef } from 'vue'
 import { ref as deepRef, shallowRef } from 'vue'
 
+export type CreateRefReturn<T = any, D extends boolean = false> = ShallowOrDeepRef<T, D>
```
43081j, line 4: did you mean to update the return type of `createRef` further down too?

packages/shared/provideLocal/index.ts
```diff
@@ -1,6 +1,9 @@
+import type { InjectionKey } from 'vue'
 import { getCurrentInstance, provide } from 'vue'
 import { localProvidedStateMap } from './map'
 
+export type ProvideLocalReturn = void
```
43081j, line 5: do we gain anything in these cases where the return type is `void`?

i wonder if this is a waste until we have a non-void return type

packages/shared/refAutoReset/index.ts
```diff
@@ -2,14 +2,16 @@ import type { MaybeRefOrGetter, Ref } from 'vue'
 import { customRef, toValue } from 'vue'
 import { tryOnScopeDispose } from '../tryOnScopeDispose'
 
+export type RefAutoResetReturn<T = any> = Ref<T>
```
43081j, line 5: i don't think we need a default as we always infer a `T` earlier

```suggestion
export type RefAutoResetReturnType<T> = Ref<T>
```

43081j:
before i continue reviewing, one change across many of these i would make is around default type params

you've changed quite a few places from `T` to `T = any`. but we don't need this as the type will be inferred, i.e. the default will never be hit

e.g. `foo<T>(val: T): boolean` doesn't ever need a default because `T` is inferred, it is whatever type `val` is

packages/shared/reactiveOmit/index.ts
```diff
@@ -1,6 +1,8 @@
 import { toRefs, toValue } from 'vue'
 import { reactiveComputed } from '../reactiveComputed'
 
+export type ReactiveOmitReturn<T extends object, K extends keyof T> = Omit<T, K> | Partial<T>
```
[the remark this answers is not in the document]
OrbisK: yeah. you are right 😅

packages/shared/refAutoReset/index.ts
43081j: they should always be specifying `T`, though. as it should be inferred from any usages

lets move this to the main thread though as i wrote more in my reply there

OrbisK:
> before i continue reviewing, one change across many of these i would make is around default type params
> 
> you've changed quite a few places from `T` to `T = any`. but we don't need this as the type will be inferred, i.e. the default will never be hit
> 
> e.g. `foo<T>(val: T): boolean` doesn't ever need a default because `T` is inferred, it is whatever type `val` is

You are right that we infer this type internally, but if you use the `XyzReturn` outside of `@vueuse/*` means you will always have to pass a generic. Even if it is any or unknown. The `XyzReturn<T = any>` should only simplify the default case. Like the `Ref` type

https://github.com/vuejs/core/blob/021f8f3b690ea875e5602573caf009d796119eab/packages/reactivity/src/ref.ts#L26

Setting this only simplifies the default case.

43081j:
do you have an example of someone using the return type like this today?

i still wouldn't do it. i think people should be declaring the type they want rather than leaning on `any` even more

OrbisK:
> do you have an example of someone using the return type like this today?
> 
> i still wouldn't do it. i think people should be declaring the type they want rather than leaning on `any` even more

Maybe this is just a matter of preference, but I think it is good practice to have any as the default for exported types. Lets take an imaginary example for now.

```ts
export type XyzReturn<T = any> = {
  type: Computed<'a' | 'b' | 'c'>
  data: Ref<T>
  readonly key1: string,
  readonly key2: Readonly<ShallowRef<{a: string, b:string}>>
  key3: number
}
```
If I have a function or whatever type that uses the `XyzReturn`, I can always use it, even if I do not want to use the `data`

```ts
// example 1: dont care about data type
function useExp1(): Pick<XyzReturn, 'key1'|'key2’> {}
// example 2: dont care about data type
function logic2(): XyzReturn {
  const result = useXyz()
  result.key3 = result.key3 +1
  return result
}
```

Overall, I think it feels more natural to just write types like `Ref` instead of having to write `Ref<any>` (some lint rules may not even like this in userspace).

OrbisK:
> All usages of it are inferred though. Nobody within vueuse would ever specify it manually because it would be inferred
> 
> By adding a default, you're implying it can sometimes be unknown/not inferred but that isn't true
> 
> It makes sense where it can't be inferred but that's a rarity in this repo as you pretty much always pass T in. Maybe in fetch or something you don't know up front but it's an edge case and those are the only ones that make sense to have a default
> 
> Your example isn't something that happens. It would be:
> 
> ```ts
> function useExp1<T>(input: T): XyzReturn<T>
> ```
> 
> i.e it always comes from an inferred input, making the default useless

But `XyzReturn` is part of the public API. Internally we do infer it, and we will most likely use each type only once internally (except for overloads), but it is still a public export.

43081j:
It is part of the public API and we infer it publicly, from the input

There's no situation where it wouldn't be set unless someone is trying to work around the type system. In those cases they can pass `typeof input` or some such thing if they must. Or manually pass unknown. But it should be a rarity since all public API we expose infers T

OrbisK:
> I don't know how better to explain this. It doesn't need a default because we should always have a type, usually an inferred one.

We cannot always infer the types in userspace. And sometimes the generic types do not even matter.

```ts
const useMyWebWorker = () => : {whatever: ShallowRef<boolean>} & Pick<UseWebWorkerReturn, 'post' | 'worker'> {
  const whatever = shallowRef(false)
  const {post, worker} = useWebWorker('/path/to.js')

  // ...

  return {
    whatever,
    post,
    worker,
  }
}
```

I do not even care about the data (generic type) here. Sometimes a user just wants to specify something like `UseXyzReturn['whateverKey']` and does not care about the generic, which he does not even use. I can have a function that accepts 

```ts
interface UseXyzReturn<T = any> {
  key1: string
  key2: boolean | ()=>boolean
  key3: T[]
}
const fn = (a: UseXyzReturn['key1'], b: UseXyzReturn['key2']) => /*  */)`
```

fn might not be directly related directly related to the actual data (key3) with the generic type.

Overall, I think this is just a personal preference. I am fine with dropping the `T = any`. We already have this for many composables, if we decide not to use `= any` we could refactor the existing `T = any` to `T`.

43081j:
in your example, `UseWebWorkerReturn` wouldn't be generic as there's no type parameters on the way in

i still don't see a clear example of where the type wouldn't be inferred

in all places we consume a `T`, the return type may or may not also involve it:

```ts
declare function useWhatever<T>(input: T): UseWhateverReturn<T>;
```

there's no situation where the return type doesn't have a `T` already (i.e. `UseWhateverReturn`). the only case is someone working around the type system for some reason. but that should be rare since we would've already inferred the type from proper usage

in places like `useFetch` it may make sense since you would be specifying the return `T` but no input (the type of the response, unrelated to any inputs). in those cases it does make sense to have a sensible default

but thats an edge case. we should have a default when necessary rather than for the sake of it

antfu (approved):

packages/shared/provideLocal/index.ts
OrbisK: I thought its better for docs, because we have always a consitient return type.

ilyaliao:
Sorry for the late reply.

> I think it's better for the docs because we always have a consistent return type.

I completely agree with this; having a unified definition is really nice :)

antfu (approved):

Issue #4669, opened 2025-03-20
`useTimeAgo` | Allow changing case of text in output


moshetanzer:
### Clear and concise description of the problem

When using reactive useTimeAgo due to the reactive output there doesnt seem to be a sane way to capatilize or set case of text output. 

### Suggested solution

Add a way to pass a case type

### Alternative

Havent found another way

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vueuse.org/guide).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

OrbisK:
```ts
const timeAgo = useTimeAgo(new Date(2021, 0, 1))
const capitalize = computed(()=> /* capitalize here */)
```
capitalize like this?

If you want you can also use css [`text-transform: capitalize`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform)

moshetanzer:
Hey @OrbisK,

Dont understand exactly what your solutions is. 

This is exactly my question/suggestion, the only way i see to do this is with a complex regex since there is `just now` and also `10 minutes ago`, `5 years ago`?

Also would be great to be able to customize the word `ago` since in some instances user might want to show just the time ago without the word `ago`!

OrbisK:
> Hey [@OrbisK](https://github.com/OrbisK),
> 
> Dont understand exactly what your solutions is.
> 
> This is exactly my question/suggestion, the only way i see to do this is with a complex regex since there is `just now` and also `10 minutes ago`, `5 years ago`?
> 
> Also would be great to be able to customize the word `ago` since in some instances user might want to show just the time ago without the word `ago`!
> 
> Edit CSS solution also does not help since there is no way to specify the case of the word `ago`

You can customize messages (... ago, in ..., ...hours) by using `options.messages`:

The default is:

https://github.com/vueuse/vueuse/blob/723abab48e297f1e7c4b6b99dc3ace30256ea55a/packages/core/useTimeAgo/index.ts#L91-L119

where you can also set custom case.

moshetanzer:
Will try. This. Was this documented anywhere?

OrbisK:
> Will try. This. Was this documented anywhere?

Partially 

[...]

PR #4670, opened 2025-03-20
feat(useTimeAgo): export `USE_TIME_AGO_MESSAGES_DEFAULT`

OrbisK:
related to https://github.com/vueuse/vueuse/issues/4669

<details>
<br>

>

</details>

---

### Description

This PR exports the default messags to allow partial overwrites.

### DIFF
--- a/packages/core/useTimeAgo/index.md
+++ b/packages/core/useTimeAgo/index.md
@@ -33,3 +33,11 @@ import { formatTimeAgo } from '@vueuse/core'
 
 const timeAgo = formatTimeAgo(new Date(2021, 0, 1)) // string
 ```
+
+## Customize messages
+
+```ts
+import { USE_TIME_AGO_MESSAGES_DEFAULT, useTimeAgo } from '@vueuse/core'
+
+const timeAgo = useTimeAgo(new Date(2021, 0, 1), { messages: { ...USE_TIME_AGO_MESSAGES_DEFAULT, past: n => n.match(/\d/) ? `${n} AGO! Now its gone` : n, } })
+```
--- a/packages/core/useTimeAgo/index.ts
+++ b/packages/core/useTimeAgo/index.ts
@@ -88,7 +88,7 @@ const DEFAULT_UNITS: UseTimeAgoUnit<UseTimeAgoUnitNamesDefault>[] = [
   { max: Number.POSITIVE_INFINITY, value: 31536000000, name: 'year' },
 ]
 
-const DEFAULT_MESSAGES: UseTimeAgoMessages<UseTimeAgoUnitNamesDefault> = {
+export const USE_TIME_AGO_MESSAGES_DEFAULT: UseTimeAgoMessages<UseTimeAgoUnitNamesDefault> = {
   justNow: 'just now',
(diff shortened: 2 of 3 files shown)


OrbisK:
> Wouldnt it be easier to just provide the partial and merge them internaly?

Thought about it, but we dont have deep defaults here. So I think its fine to do it in userspace 😄 and if we want we can still do this at some point, exporting defaults does not deny future default merging.

packages/core/useTimeAgo/index.ts
```diff
@@ -176,11 +178,11 @@ export function formatTimeAgo<UnitNames extends string = UseTimeAgoUnitNamesDefa
     const val = getValue(diff, unit)
     const past = diff > 0
 
-    const str = applyFormat(unit.name as UnitNames, val, past)
+    const str = applyFormat(unit.name as keyof UseTimeAgoMessages<UnitNames>, val, past)
```
OrbisK, line 181: maybe we can clean this up one day 😅

packages/core/useTimeAgo/index.md
```diff
@@ -33,3 +33,11 @@ import { formatTimeAgo } from '@vueuse/core'
 
 const timeAgo = formatTimeAgo(new Date(2021, 0, 1)) // string
 ```
+
+## Customize messages
+
+```ts
+import { USE_TIME_AGO_MESSAGES_DEFAULT, useTimeAgo } from '@vueuse/core'
```
43081j, line 40: this example needs updating too

packages/core/useTimeAgo/index.ts
```diff
@@ -14,9 +14,11 @@ export interface UseTimeAgoMessagesBuiltIn {
   invalid: string
 }
 
+type CustomUnit<UnitNames extends string = UseTimeAgoUnitNamesDefault> = Record<Exclude<UnitNames, keyof UseTimeAgoMessagesBuiltIn>, string | UseTimeAgoFormatter<number>>
```
43081j, line 17: this doesn't need a default (especially since it is never exposed)

packages/core/useTimeAgo/index.ts
```diff
@@ -88,7 +90,7 @@ const DEFAULT_UNITS: UseTimeAgoUnit<UseTimeAgoUnitNamesDefault>[] = [
   { max: Number.POSITIVE_INFINITY, value: 31536000000, name: 'year' },
 ]
 
-const DEFAULT_MESSAGES: UseTimeAgoMessages<UseTimeAgoUnitNamesDefault> = {
+export const USE_TIME_AGO_MESSAGES_DEFAULT: UseTimeAgoMessages = {
```
43081j, line 93: we can stop exporting this since the latest rework

antfu (changes requested):
I think I did that on purpose. The key of the messages will affect how the time are being sliced, and it's better to leave the full control to the users.

Instead I think we should expose `DEFAULT_MESSAGES` (with a better name) and update the docs to guide users to merge the messages if they want to extend it.

43081j:
If you wanted to "disable" one, would you not just replace it with undefined or something?

I don't mind much either way but my preference is still to override rather than expose internals.

PR #4674, opened 2025-03-21
fix: `useDevicesList` in firefox

### LINKED ISSUE #4672: BUG | `useDevicesList` | Doesn't list devices after reload in Firefox
### Describe the bug

Currently, the `useDevicesList` function doesn't load the devices correctly after the page is reloaded. This only happens in Firefox.

Can be reproduced by going to the docs page in Firefox, granting access to the devices (devices should now be listed correctly) and then refreshing the page. We now get only 1 available, but empty, device and not the actual available devices.

The problem seems to lie in the fact that Firefox needs an active mediastream to enumerate the devices (described [here](https://stackoverflow.com/questions/46648645/navigator-mediadevices-enumeratedevices-not-display-device-label-on-firefox)). A fix for this seems to have been put in [previously](https://github.com/vueuse/vueuse/pull/2798), but this will only work when the user hasn't given permission yet. When the user has given permission, the permissions api will say that access to the devices has been grantend, and will not create a mediastream ([the stream is only created when permission is not equal to 'granted'](https://github.com/vueuse/vueuse/blob/72c9aea0ff33105354d000040ec63807420e9d26/packages/core/useDevicesList/index.ts#L86-L89)).

Reproduction in video form:

https://github.com/user-attachments/assets/5639a160-b3c7-4b15-affe-8b3327ccc612

### Reproduction

https://vueuse.org/core/useDevicesList

### System Info

```Shell
System:
    OS: macOS 14.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 333.59 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.18.1 - ~/.nvm/versions/node/v20.18.1/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 11.1.0 - ~/.nvm/versions/node/v20.18.1/bin/npm
    pnpm: 10.6.4 - /opt/homebrew/bin/pnpm
    bun: 1.2.5 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 134.0.6998.118
    Safari: 17.3
  npmPackages:
    @vueuse/core: ^13.0.0 => 13.0.0
    vue: ^3.5.6 => 3.5.13
```

### Used Package Manager

npm

### Validations

markusgeert:

<details>
<br>

>

</details>

---

### Description

Fixes #4672. In Firefox it's not enough to only check if permission has already been granted before devices can be listed. Listing the devices always requires an active stream. This PR makes sure to always create a new stream in Firefox, even when permissions have already been granted. By doing this, listing devices continues to work in FF after a reload.  

### Additional context

I have refactored some logic into the `getCurrentPermission` function, so that it's only ran when the browser is not Firefox. This does require an `await` call inside an if condition, which does feel a bit dirty.

### DIFF
--- a/packages/core/useDevicesList/index.ts
+++ b/packages/core/useDevicesList/index.ts
@@ -93,26 +93,22 @@ export function useDevicesList(options: UseDevicesListOptions = {}): UseDevicesL
         stream = null
         granted = false
       }
+      update()
       permissionGranted.value = granted
     }
     else {
       permissionGranted.value = true
     }
 
-    update()
-
     return permissionGranted.value
   }
 
   if (isSupported.value) {
-    if (requestPermissions) {
+    if (requestPermissions)
       ensurePermissions()
-    }
-    else {
-      update()
-    }
 
     useEventListener(navigator!.mediaDevices, 'devicechange', update, { passive: true })
+    update()
   }
 
   return {
--- a/packages/core/useDevicesList/index.ts
+++ b/packages/core/useDevicesList/index.ts
@@ -71,26 +71,14 @@ export function useDevicesList(
       && navigator.mediaDevices.enumerateDevices,
   )
   const permissionGranted = shallowRef(false)
-  const requiresStreamToEnumerate = navigator!.userAgent.indexOf('Firefox') > 0
+  let stream: MediaStream | null = null
 
   async function update() {
     if (!isSupported.value)
       return
 
-    let stream: MediaStream | null = null
-
-    if (requiresStreamToEnumerate && !stream) {
-      try {
-        stream = await navigator!.mediaDevices.getUserMedia(constraints)
-      }
-      catch {
-        stream = null
-      }
-    }
-
     devices.value = await navigator!.mediaDevices.enumerateDevices()
     onUpdated?.(devices.value)
-
     if (stream) {
       stream.getTracks().forEach(t => t.stop())
       stream = null
@@ -98,28 +86,46 @@ export function useDevicesList(
   }
 
   async function ensurePermissions() {
+    const deviceName = constraints.video ? 'camera' : 'microphone'
+
     if (!isSupported.value)
       return false
 
     if (permissionGranted.value)
       return true
 
-    const deviceName = constraints.video ? 'camera' : 'microphone'
     const { state, query } = usePermission(deviceName, { controls: true })
     await query()
+    const alwaysRequireStreamToEnumerate = navigator!.userAgent.indexOf('Firefox') > 0
+    if (alwaysRequireStreamToEnumerate || state.value !== 'granted') {
+      let granted = true
+      try {
+        stream = await navigator!.mediaDevices.getUserMedia(constraints)
+      }
+      catch {
+        stream = null
+        granted = false
+      }
+      permissionGranted.value = granted
+    }
+    else {
+      permissionGranted.value = true
+    }
 
-    permissionGranted.value = state.value === 'granted'
+    update()
+
     return permissionGranted.value
   }
 
   if (isSupported.value) {
-    if (requestPermissions)
+    if (requestPermissions) {
       ensurePermissions()
+    }
+    else {
+      update()
+    }
 
-    useEventListener(navigator!.mediaDevices, 'devicechange', update, {
-      passive: true,
-    })
-    update()
+    useEventListener(navigator!.mediaDevices, 'devicechange', update, { passive: true })
   }
 
   return {
--- a/packages/core/useDevicesList/index.ts
+++ b/packages/core/useDevicesList/index.ts
@@ -44,7 +44,9 @@ export interface UseDevicesListReturn {
  * @see https://vueuse.org/useDevicesList
  * @param options
  */
-export function useDevicesList(options: UseDevicesListOptions = {}): UseDevicesListReturn {
+export function useDevicesList(
+  options: UseDevicesListOptions = {},
+): UseDevicesListReturn {
   const {
     navigator = defaultNavigator,
     requestPermissions = false,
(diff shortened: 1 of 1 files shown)


ilyaliao:
Do we need to `refactor`? It seems like just `alwaysRequireStreamToEnumerate` is enough to work properly.

markusgeert:
> Do we need to `refactor`? It seems like just `alwaysRequireStreamToEnumerate` is enough to work properly.

Sure, but by doing this we prevent `usePermission` and the subsequent `query` method from being called in Firefox. Just adding `alwaysRequireStreamToEnumerate` without refactor would still query the permissions in Firefox. If that's not a concern I can remove the refactor.

OrbisK:
If we have browser-specific stuff here, we should cover that with browser tests.

markusgeert:
> If we have browser-specific stuff here, we should cover that with browser tests.

I agree that that would be nice. However, you can't grant permissions to camera and microphone in Playwright (https://github.com/microsoft/playwright/issues/5445), which also ties into their lack of support for testing WebRTC in browsers other than Chromium (https://github.com/microsoft/playwright/issues/2973). 

It's possible to [mock](https://github.com/microsoft/playwright/issues/2973#issuecomment-1747373220) the involved API's. However, this kind of circumvents the point of a test, since that test would just be testing our mocks.

In my opinion, adding tests here would be too much work for what it's worth (if even possible).

 @OrbisK let me know what direction you want to take this (closing vs. merging). If you still want to merge, let me know what needs to be done before this can be considered ready to merge.

OrbisK:
> It's possible to [mock](https://github.com/microsoft/playwright/issues/2973#issuecomment-1747373220) the involved API's. However, this kind of circumvents the point of a test, since that test would just be testing our mocks.

agree

Lets test it manually as best as we can 😄

OrbisK:
> > Sure, but by doing this we prevent `usePermission` and the subsequent `query` method from being called in Firefox. Just adding `alwaysRequireStreamToEnumerate` without refactor would still query the permissions in Firefox. If that's not a concern I can remove the refactor.
> 
> @OrbisK What do you think? Do we need to refactor?

We should definitely not call `usePermission` inside the function.  This will cause a memory leak. 

https://github.com/vueuse/vueuse/pull/4674/files#diff-1c6f5be67d3284eb4ebf86f4553255b7271ea803ca1fe5134f397d5c8759933fR77

I wonder if we can detect if we need the stream from something other than the useragent. Like we did with the fallback part of `useClipboard`.

OrbisK:
I am not against relying on the useragent if it solves the problem. Is FF the only browser with this behaviour? And why?

------

A general thing with your PR and proposed changes:

You should not call composables within functions other than composables. Composables should always be called at setup level.

Example

```ts
// useFoo
function useFoo(s: MaybeRef<string>){
  const lowerCase = computed(()=>{
    return toValue(s).toLowerCase()
  })
  
  watch(()=>s,()=>{
    console.log("s changed")
  }

  return {
    lowerCase
  }
} 
```

```ts
// bad usage
const test = ref("abc")

function doSomeThing(){
  const { lowerCase } = useFoo(test)
  /* ... */ 
  return /* ... */ 
}
```

```ts
doSomeThing() // this will call useFoo the first time -> creating a computed and a watcher
test.value = "123" // this will call console.log once -> expected
```

```ts
doSomeThing() // this will call useFoo the first time -> creating a computed and a watcher
doSomeThing() // this will call useFoo the second time -> creating a computed and a watcher again, but does not  dispose the first

test.value = "123" // this will call console.log twice 
```

Alex has a great video on this: https://www.youtube.com/watch?v=njsGVmcWviY (its about useFetch, but can be applied to any composable)

markusgeert:
> I am not against relying on the useragent if it solves the problem. Is FF the only browser with this behaviour? And why?

From what I can gather, this only happens in FF. And while the issue is described on [SO](https://stackoverflow.com/a/69468263/9487833), the only other documentation I could find is the [spec](https://w3c.github.io/mediacapture-main/#microphone-information-can-be-exposed) which states that you need a live `MediaStreamTrack` in order to be able to enumerate devices. If I'm reading that correctly, FF is the only browser properly adhering to the spec. Meanwhile Chrome and Safari allow enumeration of devices without an active mediastream. In that light, checking useragent might seem like a good workaround.

Another solution is to always create a MediaStream before trying to enumerate the devices. This does adhere to the spec, and will work in each browser, but has two important downsides:
- It takes significant time to initiate a MediaStream
- Visual camera-active indicators become active on creation of a MediaStream (eg. my MacBook will show a green indicator next to the camera).

My preference goes out the solution I proposed earlier, first trying to enumerate the devices, and if that fails retry with an active MediaStream. This has the benefit of not impacting the majority of users (on Chrome and Safari) with the downsides mentioned above, no explicit useragent checking, still functioning if behavior of any of the browsers change, with the only downside being having to call `enumerateDevices` twice in FF.

> You should not call composables within functions other than composables. Composables should always be called at setup level.

I'll make sure to update the PR accordingly.

OrbisK:
I have found a discussion on w3c mediacapture https://github.com/w3c/mediacapture-main/issues/1019

markusgeert:
@OrbisK @ilyaliao let's get this PR finished up. If I change the `update` method in the way I describe [here](https://github.com/vueuse/vueuse/pull/4674#issuecomment-2751858280), would you merge? If so, I'll get right on it.

OrbisK:
> @OrbisK @ilyaliao let's get this PR finished up. If I change the `update` method in the way I describe [here](https://github.com/vueuse/vueuse/pull/4674#issuecomment-2751858280), would you merge? If so, I'll get right on it.

Thank you for your commitment. But I think we need to investigate the intended solution for this. From the w3c thread it looks like there should be a proper way to solve this (it also looks like FF is doing it the correct way). I have not had time to look into it yet. Maybe I can do it tomorrow.

TLDR: Let me check the Threads. I will get back to you!

Nikitatopodin:
Hey everyone! Just want to propose a solution with adding some flag that we pass to `useDevicesList` options, for exmaple, `forceGUMToEnumerate` or smth similiar with default false. And if we pass it as `true`  we just modify ensurePermissions this way:

``async function ensurePermissions() {
    const deviceName = constraints.video ? 'camera' : 'microphone'

    if (!isSupported.value)
      return false

    if (permissionGranted.value && !forceGUMToEnumerate)
      return true

    const { state, query } = usePermission(deviceName, { controls: true })
    await query()
    if (state.value !== 'granted' || forceGUMToEnumerate) {
      let granted = true
      try {
        stream = await navigator!.mediaDevices.getUserMedia(constraints)
      }
      catch {
        stream = null
        granted = false
      }
      update()
      permissionGranted.value = granted
    }```

PR #4680, opened 2025-03-27
feat(createGlobalState): prevent memory leakage

dnhy:
Modify createGlobalState to prevent memory leakage.Refer to https://github.com/vuejs/rfcs/blob/master/active-rfcs/0041-reactivity-effect-scope.md.

### DIFF
--- a/packages/shared/createGlobalState/index.ts
+++ b/packages/shared/createGlobalState/index.ts
@@ -13,17 +13,19 @@ export function createGlobalState<Fn extends AnyFn>(
   let initialized = false
   let state: any
   let summary = 0
-  const scope = effectScope(true)
+  let scope = effectScope(true)
 
   function dispose() {
     if (scope && --summary <= 0) {
       scope.stop()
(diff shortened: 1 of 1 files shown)


packages/shared/createGlobalState/index.ts
```diff
@@ -12,13 +12,24 @@ export function createGlobalState<Fn extends AnyFn>(
 ): Fn {
   let initialized = false
   let state: any
+  let summary = 0
```
43081j, line 15: do we need this?

wouldn't `summary` only ever reach `1`? so we'd be imitating the `initialized` flag we already have

dnhy: I've resubmitted to fix the location of the summary flag, which keeps track of how many components use the current hook. Clear the state and scope when no component is in use to prevent memory leaks from data inside the state. Please check

hangboss1761:
This PR seems to fix a necessary issue. Could anyone please explain why it was closed?

PR #4684, opened 2025-03-31
feat(useSortable): Add possibility to use it with Component ref

michaelcozzolino:

<details>
<br>

>

</details>

---

### Description

This PR allows the `useSortable` to be used when the ref is a component instead of a html element. E.G:

```
 <CGallery ref="gallery"
                  v-model="imageUrls"
                  is-deletable
        />

const imageUrls = ref<string[]>({ required: true });

const gallery = useTemplateRef<ComponentPublicInstance>('gallery');

useSortable(gallery, imageUrls, { animation: 200 });

```
https://vuejs.org/guide/typescript/composition-api.html#typing-component-template-refs

### DIFF
--- a/packages/integrations/useSortable/index.ts
+++ b/packages/integrations/useSortable/index.ts
@@ -1,7 +1,7 @@
-import type { ConfigurableDocument, MaybeElementRef } from '@vueuse/core'
+import type { ConfigurableDocument, MaybeElementRefOrGetter } from '@vueuse/core'
 import type { Options } from 'sortablejs'
 import type { MaybeRef, MaybeRefOrGetter } from 'vue'
-import { defaultDocument, tryOnMounted, tryOnScopeDispose } from '@vueuse/core'
+import { defaultDocument, tryOnMounted, tryOnScopeDispose, unrefElement } from '@vueuse/core'
 import Sortable from 'sortablejs'
 import { isRef, nextTick, toValue } from 'vue'
 
@@ -27,7 +27,7 @@ export type UseSortableOptions = Options & ConfigurableDocument
 
 export function useSortable<T>(selector: string, list: MaybeRefOrGetter<T[]>,
   options?: UseSortableOptions): UseSortableReturn
-export function useSortable<T>(el: MaybeRefOrGetter<MaybeElementRef>, list: MaybeRefOrGetter<T[]>,
+export function useSortable<T>(el: MaybeElementRefOrGetter, list: MaybeRefOrGetter<T[]>,
   options?: UseSortableOptions): UseSortableReturn
 
 /**
@@ -37,7 +37,7 @@ export function useSortable<T>(el: MaybeRefOrGetter<MaybeElementRef>, list: Mayb
  * @param options
  */
 export function useSortable<T>(
-  el: MaybeRefOrGetter<MaybeElementRef> | string,
+  el: MaybeElementRefOrGetter | string,
   list: MaybeRefOrGetter<T[]>,
   options: UseSortableOptions = {},
 ): UseSortableReturn {
@@ -52,7 +52,7 @@ export function useSortable<T>(
   }
 
   const start = () => {
-    const target = (typeof el === 'string' ? document?.querySelector(el) : toValue(el))
+    const target = (typeof el === 'string' ? document?.querySelector(el) : unrefElement(el))
     if (!target || sortable !== undefined)
       return
     sortable = new Sortable(target as HTMLElement, { ...defaultOptions, ...resetOptions })
--- a/packages/integrations/useSortable/index.ts
+++ b/packages/integrations/useSortable/index.ts
@@ -27,7 +27,7 @@ export type UseSortableOptions = Options & ConfigurableDocument
 
 export function useSortable<T>(selector: string, list: MaybeRefOrGetter<T[]>,
   options?: UseSortableOptions): UseSortableReturn
-export function useSortable<T>(el: MaybeElementRef, list: MaybeRefOrGetter<T[]>,
+export function useSortable<T>(el: MaybeRefOrGetter<MaybeElementRef>, list: MaybeRefOrGetter<T[]>,
   options?: UseSortableOptions): UseSortableReturn
 
 /**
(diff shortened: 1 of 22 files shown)


packages/integrations/useSortable/index.ts
```diff
@@ -27,7 +27,7 @@ export type UseSortableOptions = Options & ConfigurableDocument
 
 export function useSortable<T>(selector: string, list: MaybeRefOrGetter<T[]>,
   options?: UseSortableOptions): UseSortableReturn
-export function useSortable<T>(el: MaybeRefOrGetter<HTMLElement | null | undefined>, list: MaybeRefOrGetter<T[]>,
+export function useSortable<T>(el: MaybeRefOrGetter<HTMLElement | ComponentPublicInstance | null | undefined>, list: MaybeRefOrGetter<T[]>,
```
OrbisK, line 30:
```suggestion
export function useSortable<T>(el: MaybeElementRef | null | undefined>, list: MaybeRefOrGetter<T[]>,
```

packages/integrations/useSortable/index.ts
```diff
@@ -37,7 +37,7 @@ export function useSortable<T>(el: MaybeRefOrGetter<HTMLElement | null | undefin
  * @param options
  */
 export function useSortable<T>(
-  el: MaybeRefOrGetter<HTMLElement | null | undefined> | string,
+  el: MaybeRefOrGetter<HTMLElement | ComponentPublicInstance | null | undefined> | string,
```
OrbisK, line 40:
```suggestion
  el: MaybeRefOrGetter<MaybeElementRef | null | undefined> | string,
```

OrbisK (changes requested):

packages/integrations/useSortable/index.ts
```diff
@@ -27,7 +27,7 @@ export type UseSortableOptions = Options & ConfigurableDocument
 
 export function useSortable<T>(selector: string, list: MaybeRefOrGetter<T[]>,
   options?: UseSortableOptions): UseSortableReturn
-export function useSortable<T>(el: MaybeRefOrGetter<HTMLElement | null | undefined>, list: MaybeRefOrGetter<T[]>,
+export function useSortable<T>(el: MaybeElementRef<HTMLElement | ComponentPublicInstance | null | undefined>, list: MaybeRefOrGetter<T[]>,
```
OrbisK, line 30:
```suggestion
export function useSortable<T>(el: MaybeElementRef, list: MaybeRefOrGetter<T[]>,
```

packages/integrations/useSortable/index.ts
```diff
@@ -37,7 +37,7 @@ export function useSortable<T>(el: MaybeRefOrGetter<HTMLElement | null | undefin
  * @param options
  */
 export function useSortable<T>(
-  el: MaybeRefOrGetter<HTMLElement | null | undefined> | string,
+  el: MaybeElementRef<HTMLElement | ComponentPublicInstance | null | undefined> | string,
```
OrbisK, line 40:
```suggestion
  el: MaybeElementRef | string,
```

OrbisK (changes requested):
Should work like this no?

michaelcozzolino:
i pushed a change similar to your suggestion, i had to replace unref with toValue because it could be a computed as well with MaybeElementRef, should not bring any backward compatibility change even if tests seem to be missing. let me know what you think!

michaelcozzolino:
actually i think that those overloads are useless:
export function useSortable<T>(selector: string, list: MaybeRefOrGetter<T[]>,
  options?: UseSortableOptions): UseSortableReturn
export function useSortable<T>(el: MaybeElementRef, list: MaybeRefOrGetter<T[]>,
because they are already in the original function as MaybeRefOrGetter, can I remove them?

OrbisK:
> el: MaybeRefOrGetter<MaybeElementRef> | string,

we dont want `MaybeRefOrGetter<MaybeElementRef> `, because this can be `Ref<Ref<Element>>`. What you are suggesting is: `MaybeElementRefOrGetter` but i am not sure if this is what we can do right now. `unrefElement` is not equal to `toValue`. Thats why I think we should only go for `MaybeElementRef`. 

I will have a deeper look into this later this week 👍🏽

43081j:
i've opened #4686 so if we merge that first, you can add further tests for your change here

michaelcozzolino:
@OrbisK I added `MaybeElementRefOrGetter` let me know if anything more is needed

michaelcozzolino:
> ok I had a dig through this. I think it is simpler than we're thinking
> 
> first of all, we need a test to prove this out:
> 
> ```ts
>    it('accepts component refs', () => {
>      const SubComponent = defineComponent({
>        template: '<p>foo</p>',
>      })
>      const wrapper = mount(defineComponent({
>        components: { SubComponent },
>        template: '<SubComponent ref="el"></SubComponent>',
>        setup() {
>          const el = templateRef<InstanceType<typeof SubComponent>>('el')
>          const list = shallowRef<string[]>([])
>          const result = useSortable(el, list, {
>          })
> 
>          return { ...result, el }
>        },
>      }))
>      const vm = wrapper.vm
>      try {
>        const el = unrefElement(vm.el) as HTMLElement
>        const sortable = Sortable.get(el)
>        expect(sortable).toBeDefined()
>      }
>      finally {
>        wrapper.unmount()
>      }
>    })
> ```
> 
> the original problem is that this won't work _within typescript_. the test will pass in main I imagine but won't type check correctly
> 
> this is because the input is `MaybeRefOrGetter<HTMLElement | null | undefined>`, and the instance of a component is actually some `ComponentPublicInstance` thing.
> 
> so the fix would be `MaybeRefOrGetter<HTMLElement | ComponentPublicInstance | null | undefined>`.
> 
> we already do this in other places though, so we already have it in `unrefElement`:
> 
> https://github.com/vueuse/vueuse/blob/a0a7c7557b1c88e8497b4bdff8eab44b90df9512/packages/core/unrefElement/index.ts#L7
> 
> so really it becomes `MaybeRefOrGetter<MaybeElement>`, as that type already accounts for null/undefined too.
> 
> we then continue using `unrefElement`, i.e. we don't change the logic, only the types
> 
> let me know if that makes sense, or I missed something

Thanks a lot for your explanation. I agree 100%. I adjusted it.

43081j (approved):

PR #4698, opened 2025-04-06
fix(nuxt): don't require `@vueuse/core` installed at root

danielroe:

### Description

Currently this will always fail as `@vueuse/core` is a dependency of this package

Nuxt will resolve the paths correcty...


ilyaliao (approved):
Thanks 💚

43081j:
have tried it locally and it worked 👍  so happy to merge this

PR #4716, opened 2025-04-18
feat(useAsyncState): add executeImmediate with the same type as the promise fn

### LINKED ISSUE #4580: `useAsyncState` | execute with the same type as the original function

Regarding [useAsyncState](https://github.com/vueuse/vueuse/blob/main/packages/core/useAsyncState/index.ts) 

I expect the execute function to have the same type as the original function, so it makes migrations and usage much easier.

Currently, we have to change the arguments to remove the 0 delay. Example:
```vue
<Component @click="doSomething" />
<script setup>
const { execute } = useAsyncState(originalDoSomething), null)
const doSomething = (...args) => execute(0, ...args) // converting execute to the original shape.
</script>
```

### Suggested solution

```diff
  export interface UseAsyncStateReturnBase<Data, Params extends any[], Shallow extends boolean> {
    state: Shallow extends true ? Ref<Data> : Ref<UnwrapRef<Data>>
    isReady: Ref<boolean>
    isLoading: Ref<boolean>
    error: Ref<unknown>
-    execute: (delay?: number, ...args: Params) => Promise<Data>
+    execute: (...args: Params) => Promise<Data>
+    executeWithDelay: (delay?: number, ...args: Params) => Promise<Data>
  }
```

Example:
```vue
<Component @click="doSomething" />
<script setup>
const { execute: doSomething } = useAsyncState(originalDoSomething), null)
</script>
```

### Alternative

To avoid breaking changes we could add the `run`, `executeNow` or `executeImmediate` function, but I think that its better default `execute` (now) and `executeWithDelay`/`executeDelayed`/`executeAfter` or other name to include options.

### Additional context

_No response_

### Validations

davidglezz:

<details>
<br>

>

</details>

---

### Description

Implements https://github.com/vueuse/vueuse/issues/4580
Closes #4580

This PR adds a `executeNow` method that has the same input function arguments.
Also adds a example of using `execute` vs `executeNow`
And improves the `delay` config description.

```
const { executeNow, state } = useAsyncState(originalFn), null)
// so that: replacing originalFn('some arg') by executeNow('some arg') to have state ref.
```

### DIFF
--- a/packages/core/useAsyncState/index.md
+++ b/packages/core/useAsyncState/index.md
@@ -28,18 +28,18 @@ You can also trigger it manually. This is useful when you want to control when t
 <script setup lang="ts">
 import { useAsyncState } from '@vueuse/core'
 
-const { state, execute, executeNow } = useAsyncState(logEvent, undefined, { immediate: false })
+const { state, execute, executeImmediate } = useAsyncState(action, '', { immediate: false })
 
-async function logEvent(event) {
-  console.log('Event:', event)
-  await new Promise(resolve => setTimeout(resolve, 50))
+async function action(event) {
+  await new Promise(resolve => setTimeout(resolve, 500))
+  return `${event.target.textContent} clicked!`
 }
 </script>
 
 <template>
   <p>State: {{ state }}</p>
 
-  <button class="button" @click="executeNow">
+  <button class="button" @click="executeImmediate">
     Execute now
   </button>
 
--- a/packages/core/useAsyncState/index.md
+++ b/packages/core/useAsyncState/index.md
@@ -19,3 +19,32 @@ const { state, isReady, isLoading } = useAsyncState(
   { id: null },
 )
 ```
+
+### Manually trigger the async function
+
+You can also trigger it manually. This is useful when you want to control when the async function is executed.
+
+```vue
+<script setup lang="ts">
+import { useAsyncState } from '@vueuse/core'
+
+const { state, execute, executeNow } = useAsyncState(logEvent, undefined, { immediate: false })
+
+async function logEvent(event) {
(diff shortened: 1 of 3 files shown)


davidglezz:
I see that for some people the first parameter of the `execute` method, `delay`, is unexpected. So I have added an example to make it more visible that `exacuteNow` preserves the order of parameters.

43081j:
im not sure about this, given you should already be able to achieve it by `execute(0, fn)`

i understand it is a convenience function though, so not going to block this if other maintainers approve

davidglezz:
@43081j What do you think of a shorter name? like `run`?

For me the best would be "execute" and then add "executeAfter" or "executeWithDelay", but that means breaking change.

davidglezz:
I'm going to change it as you suggest, having 2 different names only increases the confusion.
Having a `executeImmediate` removes the current confusion about `execute`

ferferga (commented):
I would say this should be named trigger to keep it more concise and aligned with [Vue](https://vuejs.org/api/reactivity-advanced.html#triggerref), [other of our composables](https://vueuse.org/shared/watchTriggerable/#watchtriggerable) and even your own docs.

ilyaliao:
This PR looks good. I think it’s a reasonable solution we can offer for now.

IMO, the name should include `execute`, since it’s essentially an alias for `execute(0, ...args)`. I think `executeImmediate` is a good choice, and it’s ready to be merged.

ilyaliao:
The current **Execute** button in the demo calls `execute(2000, args)`.

CleanShot 2025-06-17 at 22 31 20@2x

Could you add another button for `executeImmediate(args)` as well? That way, users can clearly see the difference between the two.

PR #4719, opened 2025-04-21
feat(useDateFormat): add generic to return type

AlwaBysi:

<details>
<br>

>

</details>

---

### DIFF
--- a/packages/shared/useDateFormat/index.ts
+++ b/packages/shared/useDateFormat/index.ts
@@ -109,7 +109,7 @@ export function normalizeDate(date: DateLike) {
   return new Date(date)
 }
 
-export type UseDateFormatReturn = ComputedRef<string>
+export type UseDateFormatReturn<T = string> = ComputedRef<T>
 
 /**
  * Get the formatted date according to the string of tokens passed in.
@@ -119,6 +119,6 @@ export type UseDateFormatReturn = ComputedRef<string>
  * @param formatStr - The combination of tokens to format the date
  * @param options - UseDateFormatOptions
  */
-export function useDateFormat(date: MaybeRefOrGetter<DateLike>, formatStr: MaybeRefOrGetter<string> = 'HH:mm:ss', options: UseDateFormatOptions = {}): UseDateFormatReturn {
+export function useDateFormat<T>(date: MaybeRefOrGetter<DateLike>, formatStr: MaybeRefOrGetter<string> = 'HH:mm:ss', options: UseDateFormatOptions = {}): UseDateFormatReturn<T> {
   return computed(() => formatDate(normalizeDate(toValue(date)), toValue(formatStr), options))
 }


ilyaliao:
Can you explain a bit more about when generics are needed? Thanks.

AlwaBysi:
@ilyaliao Hi, I added generic for better compatibility with other libraries. In particular, I work with primevue and I take some components (for example, datepicker) as the first argument (modelValue) Date and not string, but format as the second argument, which actually equates Date to string.  I understand that this sounds like a primevue problem, but I'm sure that primevue is not the only library that works on a similar principle, so it's easier and more beautiful to add generic to vueuse.&

packages/shared/useDateFormat/index.ts
```diff
@@ -119,6 +119,6 @@ export type UseDateFormatReturn = ComputedRef<string>
  * @param formatStr - The combination of tokens to format the date
  * @param options - UseDateFormatOptions
  */
-export function useDateFormat(date: MaybeRefOrGetter<DateLike>, formatStr: MaybeRefOrGetter<string> = 'HH:mm:ss', options: UseDateFormatOptions = {}): UseDateFormatReturn {
+export function useDateFormat<T>(date: MaybeRefOrGetter<DateLike>, formatStr: MaybeRefOrGetter<string> = 'HH:mm:ss', options: UseDateFormatOptions = {}): UseDateFormatReturn<T> {
```
OrbisK, line 122:
```suggestion
export function useDateFormat<T = string>(date: MaybeRefOrGetter<DateLike>, formatStr: MaybeRefOrGetter<string> = 'HH:mm:ss', options: UseDateFormatOptions = {}): UseDateFormatReturn<T> {
```
this should default. because it cant be infered. Right?

OrbisK (approved):

OrbisK (changes requested):
meant to request changes, not approve 🫠

packages/shared/useDateFormat/index.ts
Kykydysik: I'm not sure, useDateFormat returns UseDateFormatReturn, and the default value is already set in UseDateFormatReturn.

OrbisK: image

https://www.typescriptlang.org/play/?#code/C4TwDgpgBAShwFcBOA7APAFSgXigZ2CQEsUBzAPhygwCg6AzBFAY2CIHsUoE8IAxdu0zkAFAEoAXLHjJ0GcgG8aUKEhmooAIgCGeACaaou6jQC+dZpwL5CVHv0HiaAemcqVAPQD8dUJGmIqABMmFQExGSUuLQMTKwcXPYC7CFYuOEkFOJScIEoqYrKqupcOvqGxrTmNJYo1uFBdrzJQU6u7lDeQA

ilyaliao:
I think this change offers flexibility and won’t affect other users.

maybe I’m missing something. @OrbisK @43081j  wdyt?

43081j:
i'd like to see an example of the usage that lead to this

e.g. you say you bump into it in primevue. can you show us?

`formatDate` will always return a `string`, so what case is there where it'd give us something we can type any stronger than that? a string literal type? if so, how?

Kykydysik:
@43081j Please look at this [primevue file](https://github.com/primefaces/primevue/blob/master/packages/primevue/src/datepicker/DatePicker.d.ts ). on 549 line you can see that modelValue has type ```Date | Array<Date> | Array<Date | null> | undefined | null;```.
So if i do stuff like: 
``` 
const localDate = ref<DatePickerProps['modelValue']>(null);
localDate.value = useDateFormat(new Date()); // here will be error Vue: Type ComputedRef<string> is not assignable to type DatePickerProps['modelValue']

<DatePicker
        :model-value="localDate"
      />

**Comment by 43081j** [comment]
the type error is right, though.

`useDateFormat` resolves to a `string`, not a `Date`

you would need to turn it back into a `Date` or change your `value` to accept strings

**Comment by Kykydysik** [comment]
Yes, i undertstand this but i think a code like this
```localDate.value = useDateFormat<Date>(new Date()); ```
 would look more beautiful.

**Comment by 43081j** [comment]
it would be an error, though:

```ts
localDate.value; // Date | Date[]

useDateFormat(new Date()); // string

localDate.value = useDateFormat(new Date()); // type error, we're trying to assign a string to a Date
```

your change here is to define the _return type_ of `useDateFormat`. but the return type is _always_ a string, so it doesn't make much sense to pretend it is anything else, in fact that'd cover up errors.

when you write `useDateFormat<Date>(val)` what do you think `Date` is being used for? it can't be the return type, because we know this is always a `string`

packages/shared/useDateFormat/index.ts
```diff
@@ -119,6 +119,6 @@ export type UseDateFormatReturn = ComputedRef<string>
  * @param formatStr - The combination of tokens to format the date
  * @param options - UseDateFormatOptions
  */
-export function useDateFormat(date: MaybeRefOrGetter<DateLike>, formatStr: MaybeRefOrGetter<string> = 'HH:mm:ss', options: UseDateFormatOptions = {}): UseDateFormatReturn {
+export function useDateFormat<T = string>(date: MaybeRefOrGetter<DateLike>, formatStr: MaybeRefOrGetter<string> = 'HH:mm:ss', options: UseDateFormatOptions = {}): UseDateFormatReturn<T> {
```
OrbisK, line 122:
```suggestion
export function useDateFormat<T extends string = string>(date: MaybeRefOrGetter<DateLike>, formatStr: MaybeRefOrGetter<string> = 'HH:mm:ss', options: UseDateFormatOptions = {}): UseDateFormatReturn<T> {
```

packages/shared/useDateFormat/index.ts
```diff
@@ -109,7 +109,7 @@ export function normalizeDate(date: DateLike) {
   return new Date(date)
 }
 
-export type UseDateFormatReturn = ComputedRef<string>
+export type UseDateFormatReturn<T = string> = ComputedRef<T>
```
OrbisK, line 112:
```suggestion
export type UseDateFormatReturn<T extends string = string> = ComputedRef<T>
```

OrbisK (changes requested):
I think there is still a use case for this generic. But the return should extend a string.

Example:

```ts
type Month = "jan" | "feb" ...

const month = useDateFormat<Month>(useNow(), 'MMM')
```

43081j:
I don't think that's what the OP was trying to do though 😬 so it's likely this can be closed once there's understanding why it doesn't make sense to do the change

43081j:
i dont think we would achieve that through a type parameter

it would be an overload:

```ts
function formatDate(date: Date, formatStr: "MMM", options: UseDateFormatOptions = {}): Month;
function formatDate(date: Date, formatStr: string, options: UseDateFormatOptions = {}): string;
```

but that is way off on a tangent for this PR. we should open a separate PR for that, if we even bother (as the only format you can strongly type is the month name, so doesn't seem worth it).

what this PR exists for and what the OP originally wanted, i think was a misunderstanding. but lets wait for a response

PR #4721, opened 2025-04-21
feat(useScrollParent): easily get the nearest scrollable parent

yue1123:
easily get the nearest scrollable parent reactively.  

<details>
<br>

>

</details>

---


packages/core/useScrollParent/index.ts
```diff
@@ -0,0 +1,87 @@
...
+        return node
+      }
+      node = node.parentNode as ScrollParent
+      level++
+    }
+
+    return root
+  }
+
+  const update = () => {
```
ferferga, line 69: Any reason why this is not computed?

yue1123: There doesn't really seem to be a reason not to use computed.

packages/core/useScrollParent/index.ts
```diff
@@ -0,0 +1,81 @@
...
+export function useScrollParent(
+  element: MaybeRefOrGetter<ScrollParent>,
+  options: UseScrollParentOptions = {},
+): ComputedRef<ScrollParent> {
+  const {
+    document: root = defaultDocument,
+    lookupLevel = 10,
+    onError = (e) => { console.error(e) },
+  } = options
+  const overflowScrollReg = /scroll|auto|overlay/i
```
ferferga, line 39:
```suggestion
  const baseRegex = 'scroll|overlay|auto'
  const overflowScrollReg = includeHidden
    ? new RegExp(`${baseRegex}|hidden`, 'i')
     : new RegExp(baseRegex, 'i')
```

With this, we can include visually hidden elements, which I think it's a good addition. ``includeHidden`` must be an option.

packages/core/useScrollParent/index.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  function getScrollParent(
+    element: ScrollParent,
+  ) {
+    const rootScrollingElement = root?.scrollingElement || root?.documentElement
+    let node: ScrollParent = element
+    let level = 0
+    while (node && node !== rootScrollingElement && isElement(node as Element) && level < lookupLevel) {
+      const { overflowY } = window.getComputedStyle(node as Element)
+      if (overflowScrollReg.test(overflowY)) {
```
ferferga, line 58: What if we're looking in the x axis? Why we can't return an object like { x, y } from the composable to be usable for those cases as well? So the user consumes whichever node it needs.

packages/core/useScrollParent/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    onError = (e) => { console.error(e) },
+  } = options
+  const overflowScrollReg = /scroll|auto|overlay/i
+
+  function isElement(node: Element) {
+    const ELEMENT_NODE_TYPE = 1
+    return (
+      node.tagName !== 'HTML'
+      && node.tagName !== 'BODY'
+      && node.nodeType === ELEMENT_NODE_TYPE
```
ferferga, line 46:
```suggestion
      && node.nodeType === 1
```

I think this unnecessary assignment can be removed.

ferferga (commented):
Great, much cleaner, thank you! 🎉
Some other points you might have not come across!

Please, also update the tests to take into account the new regexes and x axis.

yue1123:
Apologies for the delay! I was focused on refining the test cases. Really appreciate your detailed code review and thoughtful suggestions!

packages/core/useScrollParent/index.ts
ferferga: @yue1123 You forgot about this one or you don't agree with me in the need for this?

yue1123:
Apologies，I forgot to submit my comments.🤣

packages/core/useScrollParent/index.ts
ferferga: @yue1123 Yes, exactly. I'm not 100% sure of the use case for this as well, but the virtual scroll library I'm using [takes this into account](https://github.com/rocwang/vue-virtual-scroll-grid/blob/main/src/utilites.ts#L98), [jQuery](https://github.com/fis-components/jquery-ui/blob/90fe5b0943d49c2369ad43fe9d6dba26f63f410f/core.js#L42) does as well

yue1123: @ferferga OK, I see. You’ve thought it through thoroughly — this situation can definitely happen, even if it’s a bit of an edge case. I’ll add an option to control it later, as you suggested.

yue1123: @ferferga If there’s any issue with the code in the PR, please point it out and I’ll make adjustments promptly, thank you.

ferferga (approved):
LGTM! 🚀

PR #4723, opened 2025-04-23
fix(useElementVisibility): Resolve possible issue with once watcher inside of callback

### LINKED ISSUE #4704: BUG | `useElementVisibility` | Possible issue with `once` watcher inside of callback
### Describe the bug

### Reproduction

n/a

### System Info

```Shell
n/a
```

### Used Package Manager

npm

### Validations

jiwlee97:

<details>
<br>

>

</details>

---

### Description

fix #4704

Problem solved: The watchOnce call in useElementVisibility is happening inside of the intersection callback. 
Solution: I moved the watchOnce call outside of the intersection callback to ensure it is only registered once when the composable is initialized.

### DIFF
--- a/packages/core/useElementVisibility/index.test.ts
+++ b/packages/core/useElementVisibility/index.test.ts
@@ -6,6 +6,7 @@ describe('useElementVisibility', () => {
 
   beforeEach(() => {
     el = document.createElement('div')
+    vi.clearAllMocks()
   })
 
   it('should work when el is not an element', async () => {
@@ -34,6 +35,19 @@ describe('useElementVisibility', () => {
           return { stop }
         }),
       }))
+
+      vi.mock('../watchOnce', () => {
+        return {
+          watchOnce: vi.fn((source, callback) => {
+            // Immediately execute the callback if the source is true
+            if (typeof source === 'object' && source.value === true) {
+              callback(true)
+            }
+          }),
+        }
+      })
+
+      vi.useFakeTimers()
     })
 
     const { useIntersectionObserver } = await import('../useIntersectionObserver')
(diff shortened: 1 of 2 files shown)


ilyaliao:
So this is more of a structural optimization, right?

jiwlee97:
> So this is more of a structural optimization, right?

Yeah, it was a structural optimization.

ilyaliao:
it's good for me, but I'm not sure if this addresses the requirements described in #4704.

jiwlee97:
> It looks good to me, but I'm not sure if this addresses the requirements described in #4704.

Thanks for the review!
From my understanding of the issue, the concern was that placing watchOnce inside the useIntersectionObserver callback might create a new watcher each time.
To fix that, I moved it outside so it only runs once in the expected scope.
Let me know if there’s anything I can clarify or improve further.

packages/core/useElementVisibility/index.test.ts
```diff
@@ -34,6 +35,19 @@ describe('useElementVisibility', () => {
...
+      vi.mock('../watchOnce', () => {
+        return {
+          watchOnce: vi.fn((source, callback) => {
+            // Immediately execute the callback if the source is true
+            if (typeof source === 'object' && source.value === true) {
+              callback(true)
+            }
+          }),
+        }
+      })
```
OrbisK, line 48: why do we need to mock this? 🤔

jiwlee97: I mock watchOnce to ensure tests run predictably and synchronously. This allows us to control when the callback is triggered, making our tests more reliable and independent of Vue's reactivity system. By doing so, we can focus on testing the logic of useElementVisibility without external timing issues.

PR #4728, opened 2025-04-25
fix(fromEvent): fix type error of element reference

ywenhao:
…e 'HTMLElement'`

---

### Description

image
image

fix typescript error.


packages/rxjs/toObserver/index.md
```diff
@@ -16,6 +16,8 @@ import { shallowRef } from 'vue'
 
 const count = shallowRef(0)
 const button = shallowRef<HTMLButtonElement | null>(null)
```
ilyaliao, line 18: I think we can just change it to `useTemplateRef`.

packages/rxjs/README.md
```diff
@@ -35,7 +35,9 @@ import { concatAll, map, mergeMap, pluck, scan, take } from 'rxjs/operators'
 import { ref } from 'vue'
 
 const BASE_URL = 'https://jsonplaceholder.typicode.com'
-const button = ref<HTMLButtonElement>(null)
+const button = ref<HTMLButtonElement | null>(null)
```
ilyaliao, line 38: same here

ilyaliao:
Do we need to handle the case where the `element` might be `null`? After we remove the type warning

ywenhao:
> Do we need to handle the case where the `element` might be `null`? After we remove the type warning

Yes. Of course.

packages/rxjs/toObserver/_demo.vue
```diff
@@ -29,7 +28,7 @@ useSubscription(
 </script>
 
 <template>
-  <button @click="count++">
+  <button ref="buttonRef" @click="count++">
     count is: {{ count }}
   </button>
   <button ref="button">
```
ilyaliao, line 34: Do we still need the `ref="button"`?

ywenhao: Oh,bro.You're so careful!

ilyaliao (approved):

OrbisK (approved):
lgtm

antfu (dismissed):

PR #4729, opened 2025-04-25
fix: sharedComposable

lin52025iq:

-   [ ] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
-   [ ] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
-   [ ] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
-   [ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
-   [ ] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<br>

>

</details>

---

### Description

Fixed the issue where **createSharedComposable** cannot be properly disposed of when executed in a scope without getCurrentScope.

In the original implementation, when **createSharedComposable** is executed in a scope without getCurrentScope, tryOnScopeDispose directly returns false. This causes the dispose function to not be registered correctly, and thus the resources cannot be cleaned up properly. By directly calling dispose when tryOnScopeDispose returns false, we ensure that resources can be cleaned up correctly even in environments without getCurrentScope.

### **Before**

```
...
tryOnScopeDispose(dispose)
...
```

### **After**

```
...
if(!tryOnScopeDispose(dispose)) {
  dispose()
}
...
```

### DIFF
--- a/packages/shared/createSharedComposable/index.ts
+++ b/packages/shared/createSharedComposable/index.ts
@@ -30,7 +30,9 @@ export function createSharedComposable<Fn extends AnyFn>(composable: Fn): Shared
       scope = effectScope(true)
       state = scope.run(() => composable(...args))
     }
-    tryOnScopeDispose(dispose)
+    if(!tryOnScopeDispose(dispose)) {
+      dispose()
+    }
     return state
   })
 }


ferferga (changes requested):
This composable should return cleanup instead, so users can manually stop the associated effect when desired.

Lots of composables have use cases that are valid outside effect scopes and that need reactivity outside of components.

What I said here is a breaking change, but it's one of the [many inconsistencies](https://github.com/vueuse/vueuse/discussions/4548) we are fighting.

PR #4730, opened 2025-04-26
fix: always mount listeners in useStorage

43081j:
This is _yet another_ possible fix for the various `useStorage` scope issues.

Basically, `onMounted` will _always_ run in the scope of the _component_. This means the event listener will not be torn down unless the element itself is. When the various effect scopes we use are torn down, the listener will remain and cause various memory leaks.

We can solve this by moving the event listeners _outside_ the `onMounted` callback, as they will then teardown when the scope they're in disposes.

However, this then means the event listener could run before the component has mounted, when `initOnMount` is true. To solve this, we check in the event handler and do nothing if it isn't mounted yet.

We only care about the _first_ mount, since all others will have already gone through at least one `update` call.

### DIFF
--- a/packages/core/useStorage/index.ts
+++ b/packages/core/useStorage/index.ts
@@ -197,6 +197,12 @@ export function useStorage<T extends (string | number | boolean | object | null)
     updateFromCustomEvent(ev)
   }
 
+  /**
+   * The custom event is needed for same-document syncing when using custom
+   * storage backends, but it doesn't work across different documents.
+   *
+   * TODO: Consider implementing a BroadcastChannel-based solution that fixes this.
+   */
   if (window && listenToStorageChanges) {
     if (storage instanceof Storage)
       useEventListener(window, 'storage', onStorageEvent, { passive: true })
--- a/packages/core/useStorage/index.ts
+++ b/packages/core/useStorage/index.ts
@@ -181,29 +181,38 @@ export function useStorage<T extends (string | number | boolean | object | null)
 
   watch(keyComputed, () => update(), { flush })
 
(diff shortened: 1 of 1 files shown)


packages/core/useStorage/index.ts
```diff
@@ -181,29 +181,38 @@ export function useStorage<T extends (string | number | boolean | object | null)
...
+      useEventListener(window, customStorageEventName, onStorageCustomEvent)
+  }
 
-      if (initOnMounted)
-        update()
+  if (initOnMounted) {
+    tryOnMounted(() => {
+      firstMounted = true
+      update()
     })
```
ferferga, line 211: This is needed for SSR maybe? Why the hooks need to be added on mount in the first place?

43081j: i'm just maintaining whatever was there before

there's an `initOnMounted` option, that's why this is a thing. don't know why that is an option though

OrbisK (dismissed):

packages/core/useStorage/index.ts
```diff
@@ -181,29 +181,38 @@ export function useStorage<T extends (string | number | boolean | object | null)
...
+
+    updateFromCustomEvent(ev)
+  }
+
   if (window && listenToStorageChanges) {
-    tryOnMounted(() => {
-      /**
-       * Attaching event listeners here should be fine since we are in a mounted hook
-       *
-       * The custom event is needed for same-document syncing when using custom
```
ilyaliao, line 189: As we discussed, these will be retained, right?

OrbisK (approved):

ilyaliao (approved):

PR #4679, opened 2025-03-26
feat(useFileDialog): Allow custom input element for file dialog

ishakhorski:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

Added a new `input` option to the `useFileDialog`, allowing users to specify a custom HTML input element for the file dialog.

### Additional context

Closes #4618

PR #4737, opened 2025-05-05
fix(useRouteParams): solve the problem of the default value remaining in the URL

### LINKED ISSUE #4732: `useRouteQuery/useRouteParam` | Add an option to keep the information in URL for the default value

For now when the option is the default, it's not preserved in the URL.

### Suggested solution

`useRouteQuery('view', 'homepage', { displayDefaultValue: true })`

### Alternative

_No response_

### Additional context

_No response_

### Validations

1-dilikelei:

<details>
<br>

>

</details>

---

### Description
**This PR fixes #4732**
The problem that occurred：

- There is a default value, but it is not retained in the URL.

- The judgment on whether to use the default value can only be made when assigning a value to the ref reference, which triggers the set function (and the judgment logic is incorrect).

solution：

- Perform logical judgment outside the scope of the `set` function. If there is a default value, add it to the `param` and `_paramsQueue`.

- Inside the scope of the `set` function, if a manual assignment is made, replace the default value.

- Move the code logic for updating the current `params` parameter, which is currently placed in the `nextTick` function, outside the `set` function. This way, it can break free from the constraint that the `params` of the current route can only be updated when a value is assigned to the `ref` reference. 

test：
It has been thoroughly tested. When there is a default value and no manual assignment is made, the default value can be added to the `params` parameter in the `route`.
### Additional context
Please review whether my Pull Request (PR) is correct. If it is correct, I will continue to expand it to `useRouteQuery`.


packages/core/useFileDialog/index.ts
```diff
@@ -36,6 +37,10 @@ export interface UseFileDialogOptions extends ConfigurableDocument {
    * @default null
    */
   initialFiles?: Array<File> | FileList
+  /**
+   * Set the id attribute for the input element (for accessibility and testing purposes).
+   */
+  id?: string
 }
```
OrbisK, line 44: we should wait for this: https://github.com/vueuse/vueuse/pull/4679

playgrounds/vite/src/App.vue
```diff
@@ -1,9 +1,28 @@
...
+console.log(route.fullPath);
+console.log(route.path);
+console.log(route.params);
+
+// console.log(userName.value);
+// console.log(route.fullPath);
+// console.log(route.path);
+// console.log(route.params);
+// console.log(route.query);
+}
```
OrbisK, line 23: please remove the playground changes 😅

packages/core/README.md
```diff
@@ -1 +1 @@
-../../README.md
\ No newline at end of file
+../../README.md
```
OrbisK, line 1: 👀

OrbisK (changes requested):

packages/core/README.md
OrbisK: this causes linting to crash

1-dilikelei: Sorry, there might be some issues with the version of the code I submitted. I will adjust it immediately.

PR #4740, opened 2025-05-06
refactor(shared): re-export utilities from Vue (v2) and use utils where it was not being used

ferferga:

---

### Description

Given we dropped Vue 2 support, this is the "resurrection" of #3982.
Given the changeset was so small, I also took the plunge to check where we were not using some of those utilities


antfu (changes requested):
I would prefer not to, because:

- Those utils are very tiny, it doesn't make too much difference by reusing them
- `@vue/shared` would be considered an internal package of Vue, while it's unlikely to happen, but having an extra dependency makes the system more fragile, but with little gain on our side.
- We don't list `@vue/shared` in your dependencies, so it's actually a phantom dep, which would cause issues. Even if we list it, we would have an issue of version duplications etc, that could be de-optimization.

I suggest to close this.

PR #4748, opened 2025-05-13
chore(computedEager): mark for deprecation

ferferga:

<details>
<br>

>

</details>

---

### Description

Marks this composable as deprecated so it can be removed in a future major version.

### Additional context

In my opinion, there is no reason users can't always track latest minor Vue releases. It's been a while since Vue 3.4 has been released and Vue 3.6 is around the corner, so there have been plenty of time for every person to test their `computedEager` usages (plus the introduction in Vue in normal `computed` has hardly been breaking).
Users that stil want to lag behind Vue updates won't also have a problem staying behind in VueUse's

### DIFF
--- a/packages/shared/computedEager/index.ts
+++ b/packages/shared/computedEager/index.ts
@@ -14,6 +14,7 @@ export type ComputedEagerReturn<T = any> = Readonly<ShallowRef<T>>
  * computed, effect, watch, watchEffect, render dependencies will not be triggered.
  * refer: https://github.com/vuejs/core/pull/5912
  *
+ * @deprecated
  * @param fn effect function
  * @param options WatchOptionsBase
  * @returns readonly shallowRef


OrbisK:
LGTM. Maybe we can just alias it too 🤔 

-----

I think `chore(computedEager)!` should be `chore(computedEager)` until we remove it

OrbisK:
> @OrbisK You mean to remove the code (alongside the deprecation marking) and then in next major remove the aliasing as well? Or keep the aliasing indifinitely?

Mark deprecated, and alias (if possible). Remove at some major.

antfu:
Why did we remove the tests?

And if we deprecate a function, we should provide the reason and way to migrate in the docs and the jsdoc. "This function will be removed in future version" is confusing to have the next action even for me.

ferferga:
@antfu I removed the tests because I assume that functionality is now tested on Vue core, so it doesn't make sense to maintain our own suite of tests (specially if we are going to remove the entire source code of this function in a future major). Reintroduced the tests for now (they can always be removed later, when the composable is effectively removed from our source code).

I also thought the JSDoc message above explaining that Vue 3.4's computed worked exactly the same as computedEager was enough, but added it explicitly.

benedictleejh:
I want to chime in a bit here regarding the behaviour of `computedEager` and `computed`. Vue's `computed` does not fully replace the behaviour of `computedEager` at the moment. `computed` is lazy, and the first computation is only be triggered when it is used, whether in the template or later in the `script` block, whereas `computedEager` is triggered immediately.

Is there a way for Vue's `computed` to be triggered immediately? Without it, I am unable to replace `computedEager` with `computed`. I am currently on Vue 3.5.22.

9romise:
Just a note: since [the PR deprecating `computedEager`](#5034) has been merged, this one is likely outdated. Perhaps we could continue the discussion there instead.

PR #4750, opened 2025-05-13
feat(watchOnce): use vue's native once behaviour

ferferga:

<details>
<br>

>

</details>

---

### Description

Updates `watchOnce` logic to match shortcut watchers like `watchImmediate`, so it works as a pure shortcut to Vue's native [once watchers](https://vuejs.org/guide/essentials/watchers.html#once-watchers) feature.

### Additional context

@antfu This requires Vue 3.4, so technically this is a breaking change for users pre-Vue 3.4. Shall we support users staying behind or users are expected to always be in the latest version? According to our semver range, [Vue 3.5](https://github.com/vueuse/vueuse/blob/main/pnpm-workspace.yaml#L43) is the only supported version, but I don't know if this is intended or not.

### DIFF
--- a/packages/shared/watchOnce/index.ts
+++ b/packages/shared/watchOnce/index.ts
@@ -1,23 +1,38 @@
 import type { WatchCallback, WatchOptions, WatchSource, WatchStopHandle } from 'vue'
 import type { MapOldSources, MapSources } from '../utils'
-import { nextTick, watch } from 'vue'
+import { watch } from 'vue'
 
 // overloads
-export function watchOnce<T extends Readonly<WatchSource<unknown>[]>, Immediate extends Readonly<boolean> = false>(source: [...T], cb: WatchCallback<MapSources<T>, MapOldSources<T, Immediate>>, options?: WatchOptions<Immediate>): WatchStopHandle
+export function watchOnce<T extends Readonly<WatchSource<unknown>[]>>(
+  source: [...T],
+  cb: WatchCallback<MapSources<T>, MapOldSources<T, true>>,
+  options?: Omit<WatchOptions<true>, 'once'>
+): WatchStopHandle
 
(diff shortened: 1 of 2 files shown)


packages/shared/watchOnce/index.ts
```diff
@@ -1,23 +1,38 @@
...
+export function watchOnce<T = any>(source: T, cb: any, options?: Omit<WatchOptions, 'once'>) {
+  return watch(
+    source as any,
+    cb,
+    {
+      ...options,
+      once: true,
+    },
+  )
 }
```
OrbisK, line 38:
```suggestion
export function watchOnce<T = any, Immediate extends Readonly<boolean> = false>(source: T | WatchSource<T>, cb: any, options?: Omit<WatchOptions<Immediate>, 'once'>) {
  return watch(
    source,
    cb,
    {
      ...options,
      once: true,
    },
  )
}
```

OrbisK: yeah, and to be as close as possible to native watch

OrbisK: not sure if we need to, but maybe we want to overload like vue does: 

https://github.com/vuejs/core/blob/1f98a9c493d01c21befa90107f0593bc92a58932/packages/runtime-core/src/apiWatch.ts#L87-L140

ferferga: @OrbisK Shall we do the same with the other watch shortcuts I guess.

I wonder as well why we don't do rest parameters and `typeof watch` (with a tail-like helper like the ones in `type-fest` to overwrite just the last parameter) so we don't have to keep this up to date ever...

OrbisK: we can do, but does this work with the generics and overloads?

antfu:
Yes we are requiring 3.5.0+, so it should be fine to do this

antfu (approved):

PR #4749, opened 2025-05-13
feat(useStyleTag): support passing `nonce`

1-dilikelei:
### Before submitting the PR, please make sure you do the following

- [ ] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [ ] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [ ] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description
#4741 Bypass the strict Content Security Policy (CSP) using nonce

// Example of CSP header
Content-Security-Policy: style-src 'self' 'nonce-EDNnf03nceIOfn39fn3e9h3sdfa';

// Pass in the nonce when using it
useStyleTag(css, {
nonce: 'EDNnf03nceIOfn39fn3e9h3sdfa'
})

Issue #4741, opened 2025-05-08
BUG | `useStyleTag,useColorMode` | Breaks CSP


attila-varga-gc:
### Describe the bug

These two hooks forces apps to use `style-src 'self' 'unsafe-inline';` in the CSP settings.

Is there anything that we can do about it?

### Reproduction

Set the CSP header to `style-src 'self';`

### System Info

```Shell
All versions are affected.
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vueuse.org/guide).
- [x] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

1-dilikelei:
Can I use the nonce attribute to improve it?

ilyaliao:
Closed via #4749.

PR #4754, opened 2025-05-14
fix(useColorMode):support passing nonce

1-dilikelei:

<details>
<br>

>

</details>

---

### Description
#4741 
#4749（pr）

Since I'm not sure if the pull request for useColorMode can pass your review, I didn't improve useColorMode at that time. Now, the solution for useColorMode is the same as that for useStyleTag.

### DIFF
--- a/packages/core/useColorMode/index.ts
+++ b/packages/core/useColorMode/index.ts
@@ -90,6 +90,12 @@ export interface UseColorModeOptions<T extends string = BasicColorMode> extends 
    * @default true
    */
   disableTransition?: boolean
+
+  /**
+   * Nonce value for CSP (Content Security Policy)
+   *  @default undefined
+   */
+  nonce?: string
 }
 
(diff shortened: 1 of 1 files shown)


packages/core/useColorMode/index.ts
```diff
@@ -180,10 +187,12 @@ export function useColorMode<T extends string = BasicColorMode>(
       let style: HTMLStyleElement | undefined
       if (disableTransition) {
         style = window!.document.createElement('style')
+        if (nonce) {
+          style.nonce = nonce
+        }
```
OrbisK, line 192: @43081j maybe we want to go for `styleAttributes` here? should not be breaking for this. 
-------------
I wonder if we should use `useStyleTag` here

43081j: that probably makes sense 👍 

`styleAttributes` here, `attributes` (eventually) in `useStyleTag`

and you're right this could use `useStyleTag`, but not necessarily in this PR

PR #4755, opened 2025-05-15
refactor(vOnClickOutside): use weakMap for stop function instead of mutating element

jiangsongyang:
### Description

optimize the vOnClickOutside method , remove handler form element and save handle to weak map. 

This can prevent external operations on the methods mounted on the element, thereby avoiding some unexpected behaviors

### DIFF
--- a/packages/core/onClickOutside/directive.ts
+++ b/packages/core/onClickOutside/directive.ts
@@ -28,6 +28,9 @@ export const vOnClickOutside: ObjectDirective<
     if (stop && typeof stop === 'function') {
       stop()
     }
+    else {
+      stop?.stop()
+    }
     stopClickOutsideMap.delete(el)
   },
 }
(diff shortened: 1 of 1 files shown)


packages/core/onClickOutside/directive.ts
```diff
@@ -1,23 +1,34 @@
...
       const [handler, options] = binding.value
-      ;(el as any).__onClickOutside_stop = onClickOutside(el, handler, Object.assign({ capture }, options))
+      stop = onClickOutside(el, handler, Object.assign({ capture }, options))
     }
+    stopClickOutsideMap.set(el, stop)
   },
   unmounted(el) {
-    (el as any).__onClickOutside_stop()
+    const stop = stopClickOutsideMap.get(el)
+    if (stop && typeof stop === 'function') {
```
43081j, line 28: if it isn't a function, it means `controls` was `true`, right?

so we should just call `stop()` on it I think:

```ts
if (typeof stop === 'function') {
  stop();
} else {
  stop.stop(); // but call the variable something better
}
```

jiangsongyang: thx for your reply！

Yes, you're right. I overlooked this logic branch and I will add this new logic

jiangsongyang: I have added this logic. thx you for your guidance！

PR #4756, opened 2025-05-15
feat(useKeybinds): add useKeybinds

veselj43:

<details>
<br>

>

</details>

---

### Description

The name is quite self-explanatory, so I will give an example use case. Also I've checked all existing key related functions, but I don't think any of them solves the following situation well.

Let's say I have a issue tracking app and I want to have keybinds for some actions like `Ctrl + k` to open command palette, `s - d` (key sequence) to change issue state to done, etc.

Also in VueUse docs, vitepress has `Ctrl/Command + k` for search. Something like that is really easy to implement with `useKeybinds`.

### Additional context

This feature is inspired by [Nuxt UI - defineShortcuts](https://ui.nuxt.com/composables/define-shortcuts). However their implementation lacks a few key pieces like using modifier keys in sequence.

Instead of updating [their implementation](https://github.com/nuxt/ui/blob/d79da9d7b60c9972af64acd8e6eef4ae7d6bc3eb/src/runtime/composables/defineShortcuts.ts) I decided to make this function, so anybody can benefit, including Nuxt UI (if they want) since they use this lib already.

#### For reviewers

Interface, probably the most important part as it is a new function. Would be nice to make sense to others, not just me (even though I think it's quite standard).

Also I have `internal.ts` file with tests, because I wanted to test more than the "public interface" and I don't need every function to be exported from lib. 
I didn't see this approach anywhere else, so let me know if there are conventions about that. I didn't find it.


OrbisK:
Hey, Thank you for your PR

Can you briefly explain the diff to https://vueuse.org/core/useMagicKeys/#usemagickeys?

Thank you!

OrbisK:
Thank you for the clarification.

I think it should be possible to combine both composables into one. I would not have two composables that similar. Maybe we should update `useMagicKeys` to be able to do the same (breaking changes possible). Or migrate `useMagicKeys` to `useShortcut`/`useKeybinds`.

The main diff should be the current handling of the keys.

OrbisK:
> Instead of updating [their implementation](https://github.com/nuxt/ui/blob/d79da9d7b60c9972af64acd8e6eef4ae7d6bc3eb/src/runtime/composables/defineShortcuts.ts) I decided to make this function, so anybody can benefit, including Nuxt UI (if they want) since they use this lib already.

If we want Nuxt UI to be able to use this, I think we should make it compatible with `defineShortcuts`. We should also get @benjamincanac involved to discuss your new feature ideas, as he is the original author of `defineShortcuts`. He may be aware of some pitfalls that we need to be aware of. :) 

-------
I always assumed `defineShortcuts` was some kind of macro, but its just a composable. 🤔 Why is it not `useShortcuts`?

PR #4760, opened 2025-05-20
fix(useSortable): fix type misalignment

michaelcozzolino:

<details>
<br>

>

</details>

---

### Description

The list to be passed must be mutable, that is why if we do
```vue

const { option } = useSortable(el, () => list.value, {
  animation: 150,
})
```
the list in the html won't change, but the data might contain invalid sorting.
Check the following videos:
The first one is with the code above, the second one is with my codebase with more complex objects and conditions, still using hte getter
https://github.com/user-attachments/assets/d999d501-aa37-4bde-8128-6259ad750aba
https://github.com/user-attachments/assets/3693c0fa-96b4-48e0-adba-da02dd4abae1

### DIFF
--- a/packages/integrations/useSortable/index.ts
+++ b/packages/integrations/useSortable/index.ts
@@ -25,9 +25,9 @@ export interface UseSortableReturn {
 
 export type UseSortableOptions = Options & ConfigurableDocument
 
-export function useSortable<T>(selector: string, list: MaybeRefOrGetter<T[]>,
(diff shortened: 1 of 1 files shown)


OrbisK:
If this doesn't work with a getter function, it's not a breaking change but a bug fix. 😄

michaelcozzolino:
> If this doesn't work with a getter function, it's not a breaking change but a bug fix. 😄

fair enough, updated!

PR #4766, opened 2025-05-22
feat(useSortable): add target watch for recreate Sortable instance

nknwns:

<details>
<br>

>

</details>

---

### Description

Right now this composable doesn't work correct when target element rendering later. Example:

```
<script setup lang="ts">
const el = useTemplateRef<HTMLElement>('el')

useSortable(el, list)

const isVisible = ref(false)

setTimeout(() => {
  isVisible.value = true
}, 1000)
</script>

<template>
  <div v-if="isVisible" ref="el">
    <div v-for="item in list" :key="item.id">
      {{ item.name }}
    </div>
  </div>
</template>
```

The solution is simple and it exists.

```
const { start } = useSortable(el, list, {
  animation: 150,
})

const isVisible = ref(false)

setTimeout(() => {
  isVisible.value = true
  nextTick(start)
}, 1000)
```

But I suggest adding `watch` to the `target` element to provide for such options.

### Additional context

Perhaps to test whether there was an attempt to create a `Sortable`, you can make a mock `Sortable` constructor. But I limited myself to checking `isActive` in composable

### DIFF
--- a/packages/integrations/useSortable/index.ts
+++ b/packages/integrations/useSortable/index.ts
@@ -32,7 +32,7 @@ export interface UseSortableOptions extends Options, ConfigurableDocument {
 }
 
 export function useSortable<T>(selector: string, list: MaybeRefOrGetter<T[]>,
-  options: UseSortableOptions): UseSortableReturn
+  options?: UseSortableOptions): UseSortableReturn
 export function useSortable<T>(el: MaybeRefOrGetter<MaybeElement>, list: MaybeRefOrGetter<T[]>,
   options?: UseSortableOptions): UseSortableReturn
 


packages/integrations/useSortable/index.ts
```diff
@@ -70,11 +80,20 @@ export function useSortable<T>(
       return sortable?.option(name)
   }
 
+  watch(() => toValue(el), (value) => {
+    if (isActive.value)
+      stop()
+
+    if (value)
+      start()
+  }, { flush: 'post' })
```
OrbisK, line 89: Thank you for your PR: Please note https://vueuse.org/guidelines.html#watch-options

packages/integrations/useSortable/index.ts
```diff
@@ -23,10 +27,12 @@ export interface UseSortableReturn {
...
 }
 
-export type UseSortableOptions = Options & ConfigurableDocument
+export interface UseSortableOptions extends Options, ConfigurableDocument {
+  watchOptions?: WatchOptions
+}
 
 export function useSortable<T>(selector: string, list: MaybeRefOrGetter<T[]>,
-  options?: UseSortableOptions): UseSortableReturn
+  options: UseSortableOptions): UseSortableReturn
```
OrbisK, line 35: I think we still need this 🤔

OrbisK (changes requested):

packages/integrations/useSortable/index.ts
```diff
@@ -70,11 +82,20 @@ export function useSortable<T>(
       return sortable?.option(name)
   }
 
+  watch(() => toValue(el), (value) => {
```
ilyaliao, line 85: I think using `unrefElement` here would be a better fit, since el could be a `VueInstance`.

ilyaliao:
Not sure if it makes sense to include `watchOptions` in `options`.
For this example, it seems to work fine without `flush: post` (as long as we restart when `el` changes)

/cc @OrbisK  WDYT?

PR #4781, opened 2025-05-28
feat(useIDBKeyval): add `options.serializer`

### LINKED ISSUE #4780: `useIDBKeyval ` | custom serialization as per `useStorage`

The `useIDBKeyval` integration is great, but I was suprised that there was no way to customise the serialization on write/read.

### Suggested solution

It would be good to bring it up to parity with the regular `useStorage`, as per https://vueuse.org/core/useStorage/#custom-serialization

### Alternative

_No response_

### Additional context

_No response_

### Validations

matthewjumpsoffbuildings:
### Description

Attempt to add `serializer` option to `useIDBKeyval` to bring it more in line with `useStorage` as per https://vueuse.org/core/useStorage/#custom-serialization

Closes #4780

### DIFF
--- a/packages/integrations/useIDBKeyval/index.ts
+++ b/packages/integrations/useIDBKeyval/index.ts
@@ -86,12 +86,7 @@ export function useIDBKeyval<T>(
         }
       }
       else {
-        if (serializer) {
-          data.value = serializer.read(rawValue)
-        }
-        else {
-          data.value = rawValue
-        }
+        data.value = serializer.read(rawValue)
       }
     }
     catch (e) {
--- a/packages/integrations/useIDBKeyval/index.ts
+++ b/packages/integrations/useIDBKeyval/index.ts
@@ -68,24 +68,23 @@ export function useIDBKeyval<T>(
       console.error(e)
     },
     writeDefaults = true,
-    serializer: {
-      read: (val)=>val,
-      write: (val)=>val, 
-    }
+    serializer = {
+      read: (raw: any) => raw as T,
+      write: (value: T) => value,
+    },
   } = options
 
   const isFinished = shallowRef(false)
   const data = (shallow ? shallowRef : deepRef)(initialValue) as Ref<T>
 
   const rawInit: T = toValue(initialValue)
-  const serializer = options.serializer
 
   async function read() {
     try {
       const rawValue = await get<T>(key)
       if (rawValue === undefined) {
         if (rawInit !== undefined && rawInit !== null && writeDefaults) {
-          const initValue = serializer ? serializer.write(rawInit) : rawInit
+          const initValue = serializer.write(rawInit)
           await set(key, initValue)
         }
       }
@@ -107,11 +106,9 @@ export function useIDBKeyval<T>(
         await del(key)
       }
       else {
-        let rawValue = toRaw(data.value)
-        if (serializer) {
-          rawValue = serializer.write(rawValue as any)
-        }
-        await update(key, () => rawValue)
+        const rawValue = toRaw(data.value)
+        const serializedValue = serializer.write(rawValue)
+        await update(key, () => serializedValue)
       }
     }
     catch (e) {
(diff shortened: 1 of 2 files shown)


ilyaliao:
Thank you for your PR. Can you add some tests for this feature to prevent future regressions?

packages/integrations/useIDBKeyval/index.ts
```diff
@@ -64,16 +74,24 @@ export function useIDBKeyval<T>(
...
+        }
       }
       else {
-        data.value = rawValue
+        if (serializer) {
+          data.value = serializer.read(rawValue)
+        }
+        else {
+          data.value = rawValue
+        }
```
OrbisK: maybe we should do this via default serializer. Like:

```suggestion
        data.value = serializer.read(rawValue)
```

and

```ts
  const {
    flush = 'pre',
    deep = true,
    shallow = false,
    onError = (e) => {
      console.error(e)
    },
    writeDefaults = true,
    serializer: {
      read: (val)=>val,
      write: (val)=>val, 
    }
  } = options
```

OrbisK (changes requested):

acrobid:
I've encountered this breaking change in production and created minimal reproductions that demonstrate the issue:

Before v10.9.0 (working): https://stackblitz.com/edit/vitejs-vite-cdqkbqje?file=package.json
Current version (broken): https://stackblitz.com/edit/vitejs-vite-hzsdf8sq?file=package.json

This change has been blocking our upgrade from @vueuse/integrations v10.7.2. Since it was released as a minor version update, we didn't anticipate breaking changes.

I came here to propose a serialization option and am glad to see it's already been suggested! The approach in this PR looks great and would solve our use case perfectly.

Is there anything I can do to help move this forward? I'd be happy to:

Test the PR in our production environment
Add test cases for edge cases we've encountered
Help with documentation

Thank you for working on this!

packages/integrations/useIDBKeyval/index.ts
```diff
@@ -4,7 +4,12 @@ import { watchPausable } from '@vueuse/core'
 import { del, get, set, update } from 'idb-keyval'
 import { ref as deepRef, shallowRef, toRaw, toValue } from 'vue'
 
-export interface UseIDBOptions extends ConfigurableFlush {
+interface Serializer<T> {
+  read: (raw: any) => T
+  write: (value: T) => any
```
OrbisK: maybe we need one more generic for the Raw Values

packages/integrations/useIDBKeyval/index.ts
```diff
@@ -90,8 +108,11 @@ export function useIDBKeyval<T>(
         await del(key)
       }
       else {
-        // IndexedDB does not support saving proxies, convert from proxy before saving
-        await update(key, () => toRaw(data.value))
+        let rawValue = toRaw(data.value)
+        if (serializer) {
+          rawValue = serializer.write(rawValue as any)
```
OrbisK: we should replace the cast by the generic

OrbisK (changes requested):

OrbisK:
@matthewjumpsoffbuildings lets get this over the line! Would you be willing to add the suggestions?

OrbisK:
@acrobid You can use `pnpm patch` command to test whether this approach works for you. I think we might merge this soon. But I think we will not backport this to older major versions.

matthewjumpsoffbuildings:
@OrbisK @acrobid I pushed some commits attempting to apply the suggestions, default serializer, better typing etc

I don't have time to write tests right now, if someone else can handle that that would be good

packages/integrations/useIDBKeyval/index.ts
43081j: technically they can be anything, mutated outside of this code. so i don't think we can safely type them

they really are `unknown` since third party libs can save whatever in there, so we should push serializer implementors to properly deal with it

PR #4784, opened 2025-05-29
fix(useStorage): Defaults are updated if the storage gets updated when defaults are object like

michaelcozzolino:

<details>
<br>

>

</details>

---

### Description

Objects are references and when they are passed as defaults, the defaults are directly passed into a ref, and when the ref changes the defaults also change.
E.G:
```
const global = {a: 1};
const components = {b: 1}

    export const initialConfig = { global, components }

    export const config = useLocalStorage(
        'my-config',
        initialConfig,
        { mergeDefaults: true },
    );

    export const resetConfig = (): void => {
        config.value = initialConfig;
    };
```

somewhere else in the codebase:
```
  config.value.global.a = 88
 resetConfig();
 console.log(initialConfig, config.value) // both are {global: { a: 88} components: { b: 1}}
```

a workaround is to have a clone of the object (even a getter returning a non cloned object will make the object to be mutated internally):
```
const initialConfig = () => {
        return structuredClone({ global, components });
    };
```

I tried to implement the structured clone in the composable itself.
```
const data = (shallow ? shallowRef : deepRef)(
    structuredClone(toRaw(unref(typeof defaults === 'function' ? defaults() : defaults))),
  ) as RemovableRef<T>
```
- unRef removes the ref from defaults if any
- toRaw unproxies the proxy because proxies cannot be cloned

a watch is added to check for defaults changes, even tough I don't see a real use case for this feature because i suppose that even if defaults is reactive i am expecting the defaults to be used only when the storage value does not exist, but it seems, that, according to the original implementation, if the defaults change, the storage must be updated for the changed value.

just one small thing had to be changed in a test because now the composable will not return the defaults object when it is a ref but just a clone of it, I don't think this is a breaking change as the final value does not change. 

### Additional context

https://github.com/user-attachments/assets/58c95f5d-fee8-41bd-af32-581b39228ee8

### DIFF
--- a/packages/core/useStorage/index.test.ts
+++ b/packages/core/useStorage/index.test.ts
@@ -1,6 +1,6 @@
 import { debounceFilter } from '@vueuse/shared'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
-import { ref as deepRef, defineComponent, nextTick, toRaw } from 'vue'
+import { ref as deepRef, defineComponent, nextTick, toRaw, toValue } from 'vue'
 import { mount, nextTwoTick, useSetup } from '../../.test'
 import { customStorageEventName, StorageSerializers, useStorage } from './index'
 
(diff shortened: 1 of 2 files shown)


michaelcozzolino:
just realized, that this behavior is documented here https://github.com/vueuse/vueuse/discussions/4548#discussioncomment-12490139
well hope my solution can fit it

Issue #4792, opened 2025-06-03
`useTemporal` | Add temporal support by polyfill


JonathanSchndr:
### Clear and concise description of the problem

As a developer using VueUse, I want to have access to the modern Temporal API through a Vue composable so that I can handle dates, times, and timezones more reliably and consistently across different calendar systems. The current Date object has many limitations and inconsistencies that the Temporal API solves.

This implementation provides a reactive wrapper around the Temporal API using the @js-temporal/polyfill package, offering:

- 🌍 Timezone conversion and management
- 📅 Multiple calendar system support (Gregorian, Islamic, Hebrew, etc.)
- ⚡ Reactive updates with configurable intervals
- 🎯 Full TypeScript support
- 🔧 Rich set of date/time manipulation utilities

I have already implemented this feature and am submitting a PR (#4793).

### Suggested solution

In the VueUse integrations module, we have implemented `useTemporal` composable that provides a reactive interface to the Temporal API:

1. Core Features:
   - Reactive Temporal.ZonedDateTime with auto-updates
   - Timezone conversion utilities
   - Calendar system switching
   - Duration-based calculations
   - Formatting options

2. Implementation Details:
   - Uses @js-temporal/polyfill for browser compatibility
   - Provides both composable (useTemporal) and utility (createTemporal) functions
   - Full TypeScript support with proper types
   - Comprehensive test coverage
   - Demo component showcasing main features

3. API Structure:
   - Simple configuration options (timezone, calendar, interval, immediate)
   - Intuitive methods for common operations
   - Consistent with Vue's reactivity system
   - Proper cleanup and lifecycle management

The implementation is complete with full documentation, tests, and a demo component showing real-world usage scenarios.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vueuse.org/guide).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

ilyaliao:
Closed via #4793

PR #4793, opened 2025-06-03
feat(useTemporalNow): add useTemporalNow

JonathanSchndr:
### Description

Issue: #4792 

This PR introduces `useTemporal`, a new composable that provides a reactive wrapper around the modern Temporal API (TC39 proposal) using the `@js-temporal/polyfill` package. This implementation addresses the limitations of the legacy `Date` object by providing:

- 🌍 Robust timezone handling with conversion utilities
- 📅 Support for multiple calendar systems (Gregorian, Islamic, Hebrew, etc.)
- ⚡ Configurable reactive updates
- 🎯 Comprehensive TypeScript support
- 🔧 Rich set of date/time manipulation functions

The implementation includes:
- Full test coverage (index.test.ts)
- Comprehensive documentation (index.md)
- Interactive demo component (demo.vue)
- Two main exports:
  - `useTemporal()`: Reactive composable
  - `createTemporal()`: Static utility function

I have:
- [x] Read the Contributing Guidelines
- [x] Read the Pull Request Guidelines
- [x] Verified no similar PR exists
- [x] Included comprehensive tests
- [x] Added detailed documentation
- [x] Created a demo component

### Additional context

This implementation is particularly focused on:
1. **Type safety**: Full TypeScript support with proper Temporal types
2. **Flexibility**: Support for different timezones and calendar systems
3. **Performance**: Efficient updates using shallowRef
4. **Testing**: Comprehensive test coverage including edge cases
5. **Documentation**: Detailed API documentation with examples

The demo component showcases real-world usage scenarios including:
- World clock functionality
- Calendar system conversion
- Duration-based calculations
- Formatting options


JonathanSchndr:
I would like to discuss again whether https://github.com/fullcalendar/temporal-polyfill is not the better polyfill? What is your opinion?

Screenshot 2025-06-04 at 08 48 15

antfu (commented):
We are on hold for introducing new functions, and also that Temporal is still in Stage 3, https://github.com/tc39/proposal-temporal. I think we'd better leave it to community until the propose is land.

JonathanSchndr:
> We are on hold for introducing new functions, and also that Temporal is still in Stage 3, https://github.com/tc39/proposal-temporal. I think we'd better leave it to community until the propose is land.

My reason was that Firefox is now the first browser to support this natively and I have the hope that the others will soon follow suit. Therefore, my idea was to support this transition with the polyfill. But let’s wait until it has a better native coverage, then the polyfill would be removed and the PR would revise it. What do you think of the procedure? I think everyone hates new date 😅

antfu:
Stage 3 is not stable enough. Once the proposal is landed, I think we can have it in the core package and without the polyfill (polyfill should be something for users to install themselves if needed)

JonathanSchndr:
Great, then I would wait for it and adjust the PR as soon as it has increased the coverage and stability. You are also welcome to ping me if you think it is the right time. Thank you 😊👍

9romise:
Now that [Temporal](https://github.com/tc39/proposal-temporal) has reached [Stage 4](https://github.com/tc39/proposals/blob/main/finished-proposals.md), I think it's time to move this PR forward!

antfu (changes requested):
I would love to have this landed. Thank you

But as this is something going to be standard, I don't think we should deps on the ployfill, instead we can write in the docs to for users to isntall the polyfill on their own.

antfu (approved):
I have pushed some changes:
- Decoupled the polyfill, allow passing the Termporal object
- removed `createTemporal` for now as I am not sure how much value it provides at the current stage, we can keep discussing and bring it back in another PR if we found it useful.
- renamed the function to `useTemporalNow` to be more clear. Later we might think to have a more general `useTemporal` but right now it seems the composable is focusing on the current time.

JonathanSchndr:
Thank you 🙏 if I can still support you somewhere, please let me know ☺️

9romise (approved):

packages/core/useTemporalNow/index.ts
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * @default 'gregory'
+   */
+  calendar?: string
+  /**
+   * Update interval in milliseconds
+   *
+   * @default 1000
+   */
+  interval?: MaybeRefOrGetter<number>
```
9romise, line 24: Maybe we could follow the same approach as `useNow` and inherit from `ConfigurableScheduler`, rather than offering fixed `useIntervalFn` options.

antfu (approved):

PR #4797, opened 2025-06-05
feat(useAsyncState): cancel previous promise and use latest promise

snowbitx:
Add a “cancellable” parameter to ensure that only the latest Promise takes effect, canceling the previous ones. The Promise can also be manually canceled by calling cancel().

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
   "name": "@vueuse/monorepo",
   "type": "module",
-  "version": "13.3.0",
+  "version": "13.4.0",
   "private": true,
-  "packageManager": "pnpm@10.11.0",
+  "packageManager": "pnpm@10.12.1",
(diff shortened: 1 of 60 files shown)


ilyaliao:
Thanks for your PR, would you mind adding some unit tests?

snowbitx:
> Thanks for your PR, would you mind adding some unit tests?

Thanks for the suggestion! I've added the unit tests and updated the PR. Please let me know if any further changes are needed.

snowbitx:
Based on testing, I found that by default, the result of a slower previous request might override the latest one. 
To prevent this, a cancellable option was introduced to ensure only the latest request's result is retained.
A cancel method is also provided to allow manual cancellation of pending promises.

f589f79a190163cdf253c8816c4dd673

snowbitx:
Closed in favor of #4874 due to branch name conflict.

PR #4802, opened 2025-06-07
perf(useTemplateRefsList): Optimize with shallowRef

ilyaliao:

<details>
<br>

>

</details>

---

### Description

This PR optimizes `useTemplateRefsList` performance by switching to `shallowRef` to avoid unnecessary deep reactivity.

### DIFF
--- a/packages/core/useTemplateRefsList/index.ts
+++ b/packages/core/useTemplateRefsList/index.ts
@@ -1,12 +1,12 @@
-import type { Ref } from 'vue'
-import { ref as deepRef, onBeforeUpdate } from 'vue'
+import type { ShallowRef } from 'vue'
+import { onBeforeUpdate, shallowRef } from 'vue'
 
 export type TemplateRefsList<T> = T[] & {
   set: (el: object | null) => void
 }
 
-export function useTemplateRefsList<T = Element>(): Readonly<Ref<Readonly<TemplateRefsList<T>>>> {
-  const refs = deepRef<unknown>([]) as Ref<TemplateRefsList<T>>
+export function useTemplateRefsList<T = Element>(): Readonly<ShallowRef<Readonly<TemplateRefsList<T>>>> {
+  const refs = shallowRef<unknown>([]) as ShallowRef<TemplateRefsList<T>>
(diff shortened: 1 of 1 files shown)


packages/core/useTemplateRefsList/index.ts
```diff
@@ -1,12 +1,12 @@
...
+import { onBeforeUpdate, shallowRef } from 'vue'
 
 export type TemplateRefsList<T> = T[] & {
   set: (el: object | null) => void
 }
 
-export function useTemplateRefsList<T = Element>(): Readonly<Ref<Readonly<TemplateRefsList<T>>>> {
-  const refs = deepRef<unknown>([]) as Ref<TemplateRefsList<T>>
+export function useTemplateRefsList<T = Element>(): Readonly<ShallowRef<Readonly<TemplateRefsList<T>>>> {
+  const refs = shallowRef<unknown>([]) as ShallowRef<TemplateRefsList<T>>
```
OrbisK, line 9: do we still need that cast?
```suggestion
  const refs = shallowRef<TemplateRefsList<T>>([])
```

ilyaliao: Unfortunately, it seems so :(

packages/core/useTemplateRefsList/index.ts
```diff
@@ -1,12 +1,12 @@
...
   refs.value.set = (el: object | null) => {
```
OrbisK, line 10: if we use shallow ref does this trigger the ref?

packages/core/useTemplateRefsList/index.ts
43081j: Because `[]` doesn't have a `set` method?

So the type checker is right to raise that

You could instead create the empty array first, cast it and add the set method onto it
Then pass that in and get rid of line 10

packages/core/useTemplateRefsList/index.ts
ilyaliao: You’re right; this won’t work :(

ilyaliao:
Closing this as we’re considering dropping `templateRef`.

Issue #4814, opened 2025-06-12
BUG | `useTransition` | passing an unsupported type will result in a runtime exception


Unknown51335:
### Describe the bug

When an unsupported parameter is passed as a source, an error is thrown and the page fails to render.

`TypeError: v.map is not a function`

This happens because the source value is assumed to be an array or number, and the following line fails:

`return typeof v === 'number'
      ? v
      : v.map(toValue<number>)`

I know it's wrong to pass an unsupported type, but I've actually run into situations where the backend accidentally returned an empty object in production, and passing that into this function caused the page to not render anything, and it was very hard to troubleshoot.

The `sourceVal` function can be modified slightly to return 0 when it receives an unexpected type.

### Reproduction

[...]

PR #4815, opened 2025-06-12
fix(useTransition): prevent runtime error when source is not number o…

Unknown51335:
…r number[]

<details>
<br>

>

</details>

---

### Description

Fixes https://github.com/vueuse/vueuse/issues/4814

This PR fixes a potential runtime error in useTransition when the source is not a number or array of numbers.

Currently, passing a string or other unsupported type as source will result in:

TypeError: v.map is not a function
This can crash the application or prevent rendering when unexpected types are passed in.

### Additional context

The fix introduces a simple type check to ensure the value is either:

- a single number, or

- an array (e.g., number[] or MaybeRef<number>[])

Otherwise, it falls back to a safe default (0), avoiding the .map() call on unsupported types.

### DIFF
--- a/packages/core/useTransition/index.ts
+++ b/packages/core/useTransition/index.ts
@@ -213,9 +213,13 @@ export function useTransition(
   const sourceVal = () => {
     const v = toValue(source)
 
-    return typeof v === 'number'
-      ? v
-      : v.map(toValue<number>)
+    if (typeof v === 'number')
+      return v
+
+    if (Array.isArray(v))
+      return v.map(toValue)
+
+    return 0
   }
 
   const outputRef = deepRef(sourceVal())


9romise:
Actually, I think throwing an error here is expected because this will remind you of the reason.
According to your description, you defined the data type wrong and need to fix it.
Doing handling here will cause harder to troubleshoot non-expected cases.

9romise:
Actually, I think throwing an error here is expected because this will remind you of the reason.

From what I've seen in the [documentation](https://vueuse.org/core/useTransition/#type-declarations), the first argument to `useTransition` is of type `MaybeRefOrGetter<number>` or `MaybeRefOrGetter<number>[]`, and I don't think it should handle cases outside of these two types. Doing handling here will cause harder to troubleshoot non-expected cases.

According to your description, you defined the data type wrong and need to fix it.

Unknown51335:
@9romise Thank you for your reply — I completely agree with your point. You're right that my commit could hide the real issue and make debugging harder.

But I noticed that other functions like `useCountdown` and `useCounter` don’t crash the app when they receive unexpected input. I wonder if this could be an area for improvement in `useTransition`?

If you think it makes sense, I’d be happy to revise my code to improve the robustness of `useTransition` without compromising the clarity of the error.

scottbedard:
Throwing an error is definitely the correct result here. I don't think changes are necessary, but if anyone disagrees, maybe an explicit `TypeError` would be helpful?

```js
throw new TypeError('useTransition must be called with a number or array of numbers')
```

scottbedard:
I think we can close this PR now. If you'd like to implement this "0 on error" behavior, you can use a custom interpolation function.

```ts
useTransition({
  interpolation(from, to, alpha) {
    try {
      // ...
    } catch {
      return 0
    }
  }
})
```

See more info here: https://github.com/vueuse/vueuse/issues/5006, https://github.com/vueuse/vueuse/pull/5011

Issue #3034, opened 2023-04-27
Localized useTimeAgo()


AbdallahAlhaddad:
### Clear and concise description of the problem

As for now, [useTimeAgo()](https://vueuse.org/core/useTimeAgo/#usetimeago) composable only supports english.

### Suggested solution

Using [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat)  to support different languages, I think it has a [good browser support](https://caniuse.com/?search=Intl.RelativeTimeFormat) now.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
I think that's a good idea but I think it's better to have it in a separate function to avoid breaking changes and compatibility issues.

michealroberts:
@antfu I believe we could introduce this in a non-breaking way ...

antfu:
Sure thing. PR welcome :)

michealroberts:
> Sure thing. PR welcome :)

@antfu I desperately need bandwidth! :) but will do soon (how soon is now) haha

shaokeyibb:
Is there any updated? I'm very happy to working on it if there's nobody doing this.

mml00:
Here's a quick example of configuring `messages` object to use Intl.RelativeTimeFormat.

```javascript
const locale = "en"
const date = new Date("2024-02-29T18:00:00+00:00")
const rtf = new Intl.RelativeTimeFormat(locale, { numeric: "auto" })

const timeAgo = useTimeAgo(date, {
  messages: {
    justNow: t('just now'), // i18n, but you can use showSecond option to hide this message entirely
    past: n => n,
    future: n => n,
    ...Object.fromEntries([
      "year", "month", "week", "day", "hour", "minute", "second",
    ].map(
      t => [t, (n, past) => rtf.format(past ? 0 - n : n, t)]
    )),
    invalid: '',
  },
})
```

MickL:
Also a duplicate of #1592

[...]

Issue #3586, opened 2023-11-25
useTimeAgo join Chinese unit


facexyy:
### Clear and concise description of the problem

useTimeAgo adds Chinese units. This is a very practical function. I hope you will consider adding it.

### Suggested solution

Add unit configuration item options

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Issue #3356, opened 2023-08-30
How to use useTimeAgo to display in Chinese


facexyy:
### Clear and concise description of the problem

![image](https://github.com/vueuse/vueuse/assets/125525983/a0c7d5d6-7389-4449-a2c6-5df666bbffc6)

Can be changed to Chinese restrictions

### Suggested solution

Can be changed to Chinese restrictions

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Jannchie:
You can use [Intl RelativeTimeFormat API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat), which is built-in in browser.

Issue #1592, opened 2022-05-12
Translations for useTimeAgo


robheymann:
### Clear and concise description of the problem

Hi,

Is there any functionality planned (or existing) around translations for the English strings represented in useTimeAgo output?

e.g. "Yesterday", "Just Now", "6 months ago" ...

Thanks in advance,
Rob

### Suggested solution

Possibly pass through reactive locale ISO-code property. There is a finite list of strings, so would make sense to just keep a store of translations, or pass though an array of key+strings by locale to the useTimeAgo function (less clean)

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

JMaylor:
Have you considered using [Vue i18n](https://vue-i18n.intlify.dev/guide/) would do the job for you I think

wvffle:
Has anyone translated timeago using vue-i18n yet?

robheymann:
> 

I dont think so, hopefully something that will be looked at in the future.

ManUtopiK:
Here is the code to translate messages and formatter with Vue i18n using `t` :

```js
const timeago = useTimeAgo(dateValue, {
    fullDateFormatter: (date: Date) => date.toLocaleDateString(),
    messages: {
      justNow: t('just now'),
      past: n => n.match(/\d/) ? t('ago', [n]) : n,
      future: n => n.match(/\d/) ? t('in', [n]) : n,
      month: (n, past) => n === 1
        ? past
          ? t('last month')
          : t('next month')
        : `${n} month${n > 1 ? 's' : ''}`,
      year: (n, past) => n === 1
        ? past
          ? t('last year')
          : t('next year')
        : `${n} year${n > 1 ? 's' : ''}`,
      day: (n, past) => n === 1
        ? past
          ? t('yesterday')
          : t('tomorrow')
        : `${n} day${n > 1 ? 's' : ''}`,
      week: (n, past) => n === 1
        ? past
          ? t('last week')
          : t('next week')
        : `${n} ${t(`week${n > 1 ? 's' : ''}`)}`,
[...]

PR #4821, opened 2025-06-15
feat(useTimeAgoIntl): add `useTimaAgoIntl`

northword:

<details>
<br>

>

</details>

---

### Description

resolve: #3034

Added `useTimeAgoIntl` feature following the suggestion by @AbdallahAlhaddad and the comments by @antfu, using `Intl.RelativeTimeFormat` to enable `useTimeAgo` with i18n support.

### Additional context

This is a frequently requested feature, as many people have asked for it in the following issues. I hope this PR can be considered for merging.

ref: #3586 
ref: #3356 
ref: #2142 
ref: #1592
ref: https://github.com/vueuse/vueuse/discussions/2469

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Refer to [functions list](https://vueuse.org/functions) or [documentations](http
 ## 📦 Install
 
 > From v12.0, VueUse no longer supports Vue 2. Please use v11.x for Vue 2 support.
+> From v13.0, VueUse requires Vue v3.3+
 
 ```bash
 npm i @vueuse/core
--- a/packages/guidelines.md
+++ b/packages/guidelines.md
@@ -17,7 +17,7 @@ You can also find some reasons for those design decisions and also some tips for
 - Use `configurableWindow` (etc.) when using global variables like `window` to be flexible when working with multi-windows, testing mocks, and SSR.
 - When involved with Web APIs that are not yet implemented by the browser widely, also outputs `isSupported` flag
 - When using `watch` or `watchEffect` internally, also make the `immediate` and `flush` options configurable whenever possible
-- Use `tryOnUnmounted` to clear the side-effects gracefully
+- Use `tryOnScopeDispose` to clear the side-effects gracefully
 - Avoid using console logs
 - When the function is asynchronous, return a PromiseLike
 


packages/core/useTimeAgoIntl/index.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    expect(formatTimeAgoIntlParts(parts1)).toEqual('5 days')
+
+    const parts2: Intl.RelativeTimeFormatPart[] = [
+      { type: 'integer', value: '5', unit: 'day' },
+      { type: 'literal', value: '天后' },
+    ]
+    // autocorrect-disable
+    expect(formatTimeAgoIntlParts(parts2)).toEqual('5天后')
+    // autocorrect-enable
```
OrbisK:
```suggestion
    expect(formatTimeAgoIntlParts(parts2)).toEqual('5天后')
```

packages/core/useTimeAgoIntl/index.ts
```diff
@@ -0,0 +1,125 @@
...
+  return formatTimeAgoIntlParts(parts, insertSpace)
+}
+
+export function formatTimeAgoIntlParts(parts: Intl.RelativeTimeFormatPart[], insertSpace = false): string {
+  if (!insertSpace)
+    return parts.map(part => part.value).join('')
+
+  return parts
+    .map(part => part.value.trim())
+    .join(' ')
```
OrbisK, line 175: what if we let the user controll how he wants to joint the parts? We can default to trim + join(" ")

OrbisK: we can do this in favor of `insertSpace`

OrbisK (changes requested):
Sorry for the long delay.

packages/core/useTimeAgoIntl/index.ts
northword: This option was primarily designed with CJK languages in mind.
For example, in a Chinese locale, the default `Intl.RelativeTimeFormat` outputs `5天后`. However, in Chinese typography, it’s common to insert a space (such as a Pangu space) between numbers and other characters, as well as between CJK and English text.

Outside of CJK usage, I rarely see other symbols used as connectors — for example, `5-days` or `in-100-days ago` is uncommon.

Add a customizable `separator` would only affect the gaps between the parts returned by `formatToParts`. In English, the ` days ago` segment comes back as a single literal part that already contains the space between `days` and `ago`. For example:

```js
const rtf = new Intl.RelativeTimeFormat("en", { numeric: "auto" });
const parts = rtf.formatToParts(-10, "days");

console.log(parts)
// [
//   { type: "integer", value: "10", unit: "day" },
//   { type: "literal", value: " days ago" }
// ]
```

So a separator can change only the boundary between `10` and that `days`, but it cannot change the internal spacing in `days ago` to `days-ago`—unless we explicitly post-process and replace the content of the literal part.


If you think it’s still worthwhile, i can expose a hook like `joinParts?: (parts) => string` to allow users to transform/replace literal content as well—but that would go beyond a simple separator and might risk locale-specific grammar. I’m happy to proceed if that direction makes sense to you.

packages/core/useTimeAgoIntl/index.test.ts
northword: removed and rebased.

packages/core/useTimeAgoIntl/index.ts
OrbisK: With different locales in mind, I think there is an even stronger argument for `joinParts`. We could have `joinParts?: (parts, locale) => string`.

Also, partially related to this, we could consider also exposing the plain parts to the user.

```ts
const { parts, timeAgoIntl } = useTimeAgoIntl(past)
```

northword: Added a option `joinParts`; and when `controls=true`, `parts` will be returned to user.

packages/core/useTimeAgoIntl/demo.vue
```diff
@@ -0,0 +1,49 @@
+<script setup lang="ts">
+import { timestamp } from '@vueuse/core'
+import { computed, shallowRef } from 'vue'
+import { useTimeAgoIntl } from './index'
```
OrbisK:
```suggestion
import { useTimeAgoIntl } from '@vueuse/core'
```

northword: fixed

PR #4836, opened 2025-06-25
docs: onClickOutside ignore by CSS Selectors

### LINKED ISSUE #4827: onClickOutside: Document ignore option better
The `ignore` option of `onClickOutside` hook accepts an array of strings, which are then [used as selector strings to query for elements to be ignored](https://github.com/vueuse/vueuse/blob/main/packages/core/onClickOutside/index.ts#L99). But, this isn't mentioned in the docs, there is only an example with an array of refs.

Using an array of selector strings is quite helpful, as it gives more flexibility with more complex cases. However, since it's not documented, it's not clear if it's perhaps a deprecated feature that will be removed in the future.

If you confirm that this is actually a feature that is intended by the team to be publicly visible and supported long-term, I'd be happy to open a PR to update the docs.

vitomir-knime:

### Description

Resolves #4827

### DIFF
--- a/packages/core/onClickOutside/index.md
+++ b/packages/core/onClickOutside/index.md
@@ -44,6 +44,19 @@ useEventListener('pointermove', (e) => {
 })
 ```
 
+If you want to ignore certain elements, you can use the `ignore` option. Provide the elements to ignore as an array of Refs or CSS Selectors.
+
+```ts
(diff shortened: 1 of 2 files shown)


OrbisK:
Thank you!

ilyaliao (approved):

Issue #4828, opened 2025-06-21
BUG | `useCssVar` | The css variables are added to document.documentElement when the initial value of target is undefined


CaroleLIANG:
### Describe the bug

There is a modal component that uses `v-if` to control the rendering of the modal element. The goal is to use `useCssVar` to add CSS variables to the modal element when the modal opens. However, even though a `target` is bound, the CSS variable is still being added to `document.documentElement` instead of the expected element.

After examining the source code, I found that when the `target` cannot retrieve an element, `elRef` always returns document.documentElement. Therefore, I would like to suggest adding a check logic for `target`, such as:

```vue
const elRef = computed(() => {
  return target === undefined
    ? window?.document?.documentElement
    : unrefElement(target)
})

```

### Reproduction

[...]

PR #4841, opened 2025-06-27
Fix useCssVar untarget style pollution 

ArthurDarkstone:

<details>
<br>

>

</details>

---

### Additional context

https://github.com/vueuse/vueuse/issues/4828


ilyaliao (approved):

43081j:
does this mean there's no way to get back to the fallback?

maybe we should just treat `null` and `undefined` differently

if the target is `null`, someone must've explicitly set it to that or failed to find the element. if it is `undefined`, they just didn't set a target

so we can just fallback when it is `undefined`?

ilyaliao:
Yes, another issue is that when the target gets replaced, the new one might not have been rendered yet.
(Also, I don’t think we should cause side effects inside a computed)

packages/core/useCssVar/index.ts
```diff
@@ -30,11 +30,27 @@ export function useCssVar(
 ) {
   const { window = defaultWindow, initialValue, observe = false } = options
   const variable = shallowRef(initialValue)
-  const elRef = computed(() => unrefElement(target) || window?.document?.documentElement)
+
+  // Track if target has ever had a truthy value
+  const targetHadValue = shallowRef(false)
```
OrbisK: I think this can be 

```suggestion
  let targetHadValue = false
```

ArthurDarkstone: thanks for the reminder !

ArthurDarkstone:
What if v-if is `false` at first?

then element = null, then targetHadValue = false use root?

v-if="false" case: 
https://playground.vueuse.org/#eNp9U01v2zAM/SucLk6BJN6wnQJ36Fb0sB22oel60sVz6EStLBmi7GYI8t9HSbXjFGtv/Hh8fCSlg/jStsu+Q7ESBVVOtR4IfdeCLs32UgpPUnyWRjWtdR4O0BFeE92XDo5QO9tAdsXVHM0r6zCbIGlXam2fbrGeh6o7bFpdemR/LOVKrpCmsoY8VLZpQ/byBXyWhUx2MUVq6xIuiZlli0UMZvOBZg4HaQCUUV6V+r7UHa4gc7jhjscJl6L1zj4x2UnvrC41IWOKPK2EF8COf5bEHkCxUX002PzTeW8NXFVaVY+8s5HyXbJ4gzEFDRZ5Ag+lJ5bkQKVLIuZo/OLDeymgX6h6pGTfYXDDjOyQ/6sxujz6CnpexPMeLuLVEi/AumTlCB73fmyWp9Znzc/5XqX7TQjOWr4DEYNeUp7bg1Xkk/1JI+b8tPgCtdouH8gafn/xXmk2pdH9bL3iC0mxSpcMuXih7zHmXYfzIV7tsHr8T/yB9iEmxS+HhK5HKcacL90WfUrfrH+E3ZySjd10mtFvJG+RrO6CxgT72pkNy57gotpv8T8os72jm71HQ8NQQWhAHiNeCv4M12+MfpL7cfkp1vE7Fsd/HnM82Q==

PR #4844, opened 2025-06-28
test: refactor base url

iiio2:
Actually there are https://example.com is used many places as base url. That is why I refactor it. But I am not sure it would be better or not though.


antfu (approved):

OrbisK (approved):

ilyaliao:
Spend some time investigating the cause of the previous test failures, now ready for re-review

cc: @iiio2, I made some changes, hoping it helps get this PR merged sooner!

packages/core/useFetch/index.test.ts
```diff
@@ -17,14 +18,21 @@ function fetchSpyHeaders(idx = 0) {
   return (fetchSpy.mock.calls[idx][1]! as any).headers
 }
 
+function createFetch(options: Parameters<typeof _createFetch_>[0] = {}) {
+  return _createFetch_({
+    baseUrl,
+    ...options,
+  })
+}
+
```
ilyaliao, line 27: I added an extra wrapper around `createFetch` in `useFetch`, so we don't need to specify `baseUrl` every time.

This should make test writing much simpler.

antfu (approved):

iiio2:
Thanks @antfu . :blush:

PR #4849, opened 2025-06-30
feat(useFocusTrap): expose updateContainerElements for dynamic contai…

### LINKED ISSUE #4848: `useFocusTrap` | Implement `updateContainerElements()`

Summary:
The focus-trap library provides a useful `trap.updateContainerElements()` method, allowing consumers to update the container elements of a focus trap without needing to destroy and recreate it. This is especially helpful in dynamic UIs where the trap boundaries may change during the component lifecycle.

Use Case:
In applications where the content within a modal, drawer, or any focus-trapped container changes (e.g., conditional rendering, dynamic portals), being able to manually update the container elements is essential to maintain correct focus behavior.

Thank you for considering this enhancement!

### Suggested solution

Expose the `updateContainerElements` method via the useFocusTrap hook so users can call it directly when needed. https://github.com/focus-trap/focus-trap?tab=readme-ov-file#trapupdatecontainerelements

Suggested API:

```ts
const { activate, deactivate, updateContainerElements } = useFocusTrap(ref, options);
```

### Alternative

Currently, developers must recreate the trap if they need to update the container elements, which is not efficient and can lead to undesirable side effects.

### Additional context

Benefits:

- Aligns useFocusTrap closer to the full capabilities of the underlying focus-trap library.
- Reduces boilerplate and avoids unnecessary unmounting/mounting cycles.
- Improves accessibility in dynamic content scenarios.

### Validations

PeikyLiu:
### Description
This PR exposes the `updateContainerElements` method from the underlying focus-trap library, allowing users to dynamically update focus trap containers while the trap is active.

### Related Issues
resolve #4848

### 📋 Changes
- **Added** `updateContainerElements` method to `UseFocusTrapReturn` interface
- **Created** `ContainerElements` type alias using `Parameters<FocusTrap['updateContainerElements']>[0]`
- **Updated** return type to handle `undefined` case when trap is not initialized
- **Added** basic container switching demo in demo.vue

### 🔧 Technical Details
- Uses TypeScript's `Parameters` utility type to automatically inherit parameter types from focus-trap library
- Ensures type consistency with the underlying focus-trap library

### 📚 Example Usage
```typescript
const { updateContainerElements, activate } = useFocusTrap(target)
activate()
updateContainerElements(target2.value) // Switch to different container
```

### 🔄 Breaking Changes
None - this is a purely additive change.


ilyaliao (approved):

packages/integrations/useFocusTrap/index.ts
```diff
@@ -87,6 +95,13 @@ export function useFocusTrap(
     }
   }
 
+  const updateContainerElements = (el: ContainerElements) => {
+    if (trap) {
+      trap.updateContainerElements(el)
+      return trap
+    }
+  }
```
OrbisK: I think we dont want to expose this, but watch the target and `updateContaienrElements` when target changes (instead of activate/deactivate)? wdyt @ilyaliao

ilyaliao: > I think we dont want to expose this, but watch the target and `updateContaienrElements` when target changes (instead of activate/deactivate)? wdyt @ilyaliao

Yes, that's what I initially had in mind as well. We could initialize once with `createFocusTrap`, and then update the element via `updateContainerElements` afterwards. But I haven’t looked into the feasibility of this approach in depth yet.

OrbisK: @PeikyLiu wdyt?

PeikyLiu: ok,I'll try based your idea,but it will be take some time

PeikyLiu: > I think we dont want to expose this, but watch the target and `updateContaienrElements` when target changes (instead of activate/deactivate)? wdyt @ilyaliao

To summarize: you suggest using watch to monitor target changes inside the hook, while keeping the activate/deactivate methods exposed, right?

OrbisK: we always want to use the target thats passed to the composable. If we want to change the target we want to update the target ref. So we can use updateContainerElements, we should do this while watching target 😄

packages/integrations/useFocusTrap/index.md
```diff
@@ -76,6 +76,41 @@ const { hasFocus, activate, deactivate } = useFocusTrap([targetOne, targetTwo])
...
+const target = computed(() => {
+  return currentRef.value === 'left' ? left : currentRef.value === 'right' ? right : null
+})
+const { hasFocus, activate } = useFocusTrap(target)
+
+activate()
+
+setTimeout(() => {
+  currentRef.value = 'right'
+}, 3000)
```
ilyaliao:
```suggestion
import { computed, shallowRef, useTemplateRef } from 'vue'
import { useFocusTrap } from '@vueuse/integrations/useFocusTrap'

const left = useTemplateRef('left')
const right = useTemplateRef('right')
const currentRef = shallowRef<'left' | 'right'>('left')

const target = computed(() =>
  currentRef.value === 'left'
    ? left
    : currentRef.value === 'right'
      ? right
      : null,
)

const { activate } = useFocusTrap(target)
```

PeikyLiu: yes,that's be better,thanks for your review! ^^

ilyaliao (approved):

antfu (approved):

PR #4857, opened 2025-07-06
fix(useVirtualList): resolve invalid watch source

### LINKED ISSUE #4854: BUG | `useVirtualList` | Invalid watch source with doc example
### Describe the bug

When using the "simple list" example in the `useVirtualList` docs:

```js
import { useVirtualList } from '@vueuse/core'

const { list, containerProps, wrapperProps } = useVirtualList(
  Array.from(Array.from({ length: 99999 }).keys()),
  {
    // Keep `itemHeight` in sync with the item's row.
    itemHeight: 22,
  },
)
```
console warnings appear about the provided data not being a valid watch source.

### Reproduction

https://play.vueuse.org/#N4IgDghgxg1hDmBTAziAXAbVAOwgW0XRADcBXRAWgBNE8BLEAGhGQHtSAnKQtEU7MDHhMQNZFA50wAFzqtsRAGrkABABFadFQAoAFhAA2AMxV1sKgGIdE2KLoCUp5CogqaxRAdZgz8FaVkDOmkATxcDLwB3ZxD2FWlWFUjJaUQVAFVsOg8OZEMVZTSAGToAIw4ISRQVI1YOAtUAJhUAMhUAZhFOAyJdaWkwZDQAemH+QXgAOihWPGGyShp6YaDS4bMaAA9JvAArVABfRhx8HhAAAQXSZERh5H1rKhE2Tm4icaERMQkpWXkiADKD0QVAaiHSN38gWCdBQky6HB6vD6AyGow+Uxmc0u5Gut3ulRB5wAjAAGSbEyak9bYLY7fYgI4nAhEHGIPHDGbWZ7sLhnDFfFA-GRyBS8ADCrAiiCgf3MrBMKBu2Fk+UKKkleDArGQMPkGWhshQCKRIBRgxGYwEQmms2GbI5XMQJPJlOpG0Q2z2hwAuswjHQDMa0KAAIJgMCTBboUCpLUGCCpIgAHio2RUxAopQ2AF4ADogGYqiBmRAcAAKHG8yALKmQoSD+bNiDo8D6aA6pNJYE2BYAfHnzCpU+nM9naU3khAI2XK9X+4OVEvh2niBmKLUOE3grRTOYgvXa2gYIgQtu45MPb2QHWG4gm7oW23pB3Go0ewuh8uVAAlViRDtgGAUwLyoRNXAOA5F2XZNhlXAch1g+DByQ7I+2eYVpCIOgtTqaQVGAvFFDoDhpFIQwSnrFQDhqKs8BUAByB0bk5OpEAYwdByLKjgIPaRGBUItpBLbBZyrQYBKnGcK3E5waJzfwbmI0jyIMSjpG0aDQw4CoQkmIw6O0bTdP0wzeJseBpF0DsAE47Ls6j7EmE8QmQbR7HsRhoOAaCl1GFQAGlEEQMAVAAAx3PAAAknz6MK9zrEJbCSYJdHiR8QNoBjnCrSJJl8zLotil8VDfLyhyOQd7EZSCgA

### System Info

```Shell
System:
    OS: Linux 6.6 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
    Memory: 7.47 GB / 9.71 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 22.17.0 - ~/.nvm/versions/node/v22.17.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v22.17.0/bin/npm
    pnpm: 10.7.0 - ~/.local/share/pnpm/pnpm
  npmPackages:
    @vueuse/core: ^13.5.0 => 13.5.0
    vue: ^3.5.17 => 3.5.17
```

### Used Package Manager

npm

### Validations

rzzf:

<details>
<br>

>

</details>

---

### Description

Ensure that the `list` being listened to is always reactive data.

### Additional context

fixes #4854

### DIFF
--- a/packages/core/useVirtualList/index.test.ts
+++ b/packages/core/useVirtualList/index.test.ts
@@ -7,7 +7,7 @@ describe('useVirtualList', () => {
     expect(useVirtualList).toBeDefined()
   })
 
-  it('should not warn when list is a plain array', () => {
+  it('should accept plain arrays as input', () => {
     const {
       list,
       containerProps: { ref: containerRef },


packages/core/useVirtualList/index.ts
```diff
@@ -201,7 +202,7 @@ function createGetDistance<T>(itemSize: UseVirtualListItemSize, source: UseVirtu
 }
 
 function useWatchForSizes<T>(size: UseVirtualElementSizes, list: MaybeRef<readonly T[]>, containerRef: Ref<HTMLElement | null>, calculateRange: () => void) {
-  watch([size.width, size.height, list, containerRef], () => {
+  watch([size.width, size.height, toRef(list), containerRef], () => {
```
OrbisK:
```suggestion
  watch([size.width, size.height, () => toValue(list), containerRef], () => {
```

OrbisK: or `unref`, because we dont have `MaybeRefOrGetter`

rzzf:
@OrbisK Thank you for your review, I have updated the code.

OrbisK (dismissed):

ilyaliao (approved):

packages/core/useVirtualList/index.ts
43081j: is this any slower than using the original ref if it was one?

i.e. is it slower to watch a function than a ref?

what is the benefit of `() => toValue(list)` over `toRef(list)`? (not saying either is better, just wondering if there's a gain)

43081j:
can you add a test too?

just set the list as a fixed array or something instead of a ref, in the test

if you also catch up from main, the docs build will succeed (although happy to merge regardless of that in this case)

rzzf:
> can you add a test too?
> 
> just set the list as a fixed array or something instead of a ref, in the test
> 
> if you also catch up from main, the docs build will succeed (although happy to merge regardless of that in this case)

Okey, Tonight I will increase the test cases.

packages/core/useVirtualList/index.ts
OrbisK: Maybe we should benachmark this.

rzzf:
> can you add a test too?
> 
> just set the list as a fixed array or something instead of a ref, in the test
> 
> if you also catch up from main, the docs build will succeed (although happy to merge regardless of that in this case)

Thank you for your review, I have added test case.

packages/core/useVirtualList/index.ts
rzzf: I am sorry, I am not familiar with benchmark tests. Can I do anything else?

packages/core/useVirtualList/index.test.ts
```diff
@@ -1,11 +1,21 @@
...
   it('should be defined', () => {
     expect(useVirtualList).toBeDefined()
   })
+
+  it('should not warn when list is a plain array', async () => {
+    const spy = vi.spyOn(console, 'warn').mockImplementation(() => {})
+
+    useVirtualList(['a', 'b', 'c', 'd', 'e', 'f'], { itemWidth: () => 50, overscan: 1 })
+
+    expect(spy).not.toHaveBeenCalled()
```
43081j: i dont think this'll ever be called since we don't console.warn in there

either way, we should test that mutations are observed instead i think

something like this:

```ts
    const items = shallowRef(['a', 'b'])
    const {
      list,
      containerProps: { ref: containerRef },
      scrollTo,
    } = useVirtualList(items, { itemHeight: () => 50 })
    const div = { ...document.createElement('div'), clientHeight: 100 }

    containerRef.value = div
    scrollTo(0)

    items.value = ['c', 'd']
    await nextTick()

    expect(list.value.map(i => i.data)).toEqual(['c', 'd'])

```

the test is that non-ref lists are observed

rzzf: Umm, But the issue we need to fix is that when passing a plain list to watch, it sends a warning, and console.warn is called within watch.

I restored the code and retained the test code, this test case will not pass.
image


Perhaps we should keep them both at the same time?

something like this?
```ts
it('should not warn when list is a plain array', async () => {
    const spy = vi.spyOn(console, 'warn').mockImplementation(() => {})
    const items = shallowRef(['a', 'b'])
    const {
      list,
      containerProps: { ref: containerRef },
      scrollTo,
    } = useVirtualList(items.value, { itemHeight: () => 50 })
    const div = { ...document.createElement('div'), clientHeight: 100 }

    containerRef.value = div
    scrollTo(0)

    items.value = ['c', 'd']
    await nextTick()

    expect(spy).not.toHaveBeenCalled()
    expect(list.value.map(i => i.data)).toEqual(['a', 'b'])
  })
```

43081j: the warning is just a side effect of us doing something wrong internally, though

if we were writing this function from scratch, we would just test that you can use it with a ref and without a ref

we're getting a warning because there's a bug, not because this function is meant to warn

that warning could go away any day, unrelated to this function. so it wouldn't be a concrete assertion

rzzf: Thank you very much for pointing that out. I agree with your perspective.

rzzf: I'm not very experienced in writing tests, so please feel free to point out any issues. I really appreciate your patience and understanding.

packages/core/useVirtualList/index.test.ts
```diff
@@ -6,6 +6,19 @@ describe('useVirtualList', () => {
   it('should be defined', () => {
     expect(useVirtualList).toBeDefined()
   })
+
+  it('should not warn when list is a plain array', () => {
```
43081j: this should now be something like:

> should accept plain arrays as input

rzzf: done, Thanks!

packages/core/useVirtualList/index.ts
43081j: Don't worry about it. This is fine as it is 👍

ilyaliao (approved):

PR #4873, opened 2025-07-12
test: add edge case tests for useCountdown

snowbitx:

<details>
<br>

>

</details>

---

### Description

add edge case tests for useCountdown

### DIFF
--- a/packages/core/useCountdown/index.test.ts
+++ b/packages/core/useCountdown/index.test.ts
@@ -148,13 +148,14 @@ describe('useCountdown', () => {
     start()
     start()
     start()
-    vi.advanceTimersByTime(310)
+    vi.advanceTimersByTime(1000)
     expect(tickCallback).toHaveBeenCalledTimes(3)
   })
 
   it('isActive becomes false after countdown completes', async () => {
     const { isActive } = useCountdown(2, options)
-    vi.advanceTimersByTime(210)
+    expect(isActive.value).toBe(true)
+    vi.advanceTimersByTime(1000)
     expect(isActive.value).toBe(false)
   })
 })


packages/core/useCountdown/index.test.ts
```diff
@@ -142,4 +142,19 @@ describe('useCountdown', () => {
...
+    start()
+    start()
+    start()
+    vi.advanceTimersByTime(310)
+    expect(tickCallback).toHaveBeenCalledTimes(3)
+  })
+
+  it('isActive becomes false after countdown completes', async () => {
+    const { isActive } = useCountdown(2, options)
+    vi.advanceTimersByTime(210)
```
OrbisK: can we check if it is true before advancing the timers? 😄

packages/core/useCountdown/index.test.ts
```diff
@@ -142,4 +142,19 @@ describe('useCountdown', () => {
...
     expect(completeCallback).toHaveBeenCalledTimes(3)
   })
+
+  it('multiple start() calls only create one timer', async () => {
+    const { start } = useCountdown(3, options)
+    start()
+    start()
+    start()
+    vi.advanceTimersByTime(310)
+    expect(tickCallback).toHaveBeenCalledTimes(3)
```
OrbisK: why 310? should'nt we advance by +1000ms?

OrbisK (changes requested):

packages/core/useCountdown/index.test.ts
snowbitx: > why 310? should'nt we advance by +1000ms?

Good catch — I originally used 310ms by referencing the 'start can provide a custom interval' test case (which also uses short intervals), but you’re right that 1000ms makes it much clearer. I've updated it accordingly — thanks! 🙌

packages/core/useCountdown/index.test.ts
snowbitx: > can we check if it is true before advancing the timers? 😄

Thanks for the suggestion — added the initial check for isActive.

packages/core/useCountdown/index.test.ts
```diff
@@ -142,4 +142,20 @@ describe('useCountdown', () => {
...
+  it('multiple start() calls only create one timer', async () => {
+    const { start } = useCountdown(3, options)
+    start()
+    start()
+    start()
+    vi.advanceTimersByTime(1000)
+    expect(tickCallback).toHaveBeenCalledTimes(3)
+  })
+
+  it('isActive becomes false after countdown completes', async () => {
```
Alfred-Skyblue, line 155: This feature has corresponding test cases.
https://github.com/vueuse/vueuse/blob/d23812dbebece77e5e0bd6d9d3574a3637ac31b0/packages/core/useCountdown/index.test.ts#L58-L67

snowbitx: > This feature has corresponding test cases.
> 
> https://github.com/vueuse/vueuse/blob/d23812dbebece77e5e0bd6d9d3574a3637ac31b0/packages/core/useCountdown/index.test.ts#L58-L67

Confirmed — existing tests already cover it. Sorry for the confusion and thanks! 🙏

PR #4917, opened 2025-07-27
fix(docs): useStorage wrong parameters order in docs for mergeDefaults

michaelcozzolino:

<details>
<br>

>

</details>

---

### Description
According to the deepmerge docs https://www.npmjs.com/package/deepmerge

> merge(x, y, [options])
> Merge two objects x and y deeply, returning a new merged object with the elements from both x and y.
> 
> **If an element at the same key is present for both x and y, the value from y will appear in the result.**
> 
> Merging creates a new object, so that neither x or y is modified.

the params in the useStorage example should be inverted as we want to merge the defaults to the storage ones

### DIFF
--- a/packages/core/useStorage/index.md
+++ b/packages/core/useStorage/index.md
@@ -68,7 +68,7 @@ const state = useStorage(
   'my-store',
   { hello: 'hi', greeting: 'hello' },
   localStorage,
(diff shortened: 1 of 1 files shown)


antfu:
This deepMerge function is a pseudo code, it's not necessary refers to [npmjs.com/package/deepmerge](https://www.npmjs.com/package/deepmerge). I guess we made the args name clear already, so I don't think this change is necessary. Thanks.

Issue #4236, opened 2024-09-20
useSortable moving item another list crashes 


azadbilge:
### Describe the bug

Hello VueUse Team,

I'm encountering an issue while using the useSortable function in my Vue 3 project. The problem occurs when dragging an item from one list to another. After the item is successfully moved, a duplicate DOM element remains, causing unwanted behavior in the UI. After I reconstructed the simplified version on stackblitz, duplicating on DOM is not occured. However, after a dragging process from a list to another list, flicking behaviour beginning for dragging process, which I face this issue on my code when I tried to remove the duplicated block with evt.item.remove(). I didn't understand these 2 issue are from same problem.
Basically I am trying to build a block editor. SectionPage and GroupPage are sample blocks and RenderBlock.vue is the renderer component for blocks. My data and data flow is the same and begins in App.vue and helper functions are in useBlocks.js

### Reproduction

https://stackblitz.com/edit/vitejs-vite-4qbt23?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: Windows 11 10.0.22631
    CPU: (12) x64 AMD Ryzen 5 5500U with Radeon Graphics
    Memory: 8.15 GB / 23.33 GB
  Binaries:
    Node: 20.9.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.1.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    @vueuse/core: ^10.9.0 => 10.9.0
    @vueuse/integrations: ^10.9.0 => 10.9.0
    @vueuse/motion: ^2.2.3 => 2.2.3
    vue: ^3.4.21 => 3.4.21
```

### Used Package Manager

npm

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vueuse.org/guide).
- [X] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
[...]

PR #4922, opened 2025-07-29
perf(useSortable): Show warn if useSortable is not called within a Ef…

CaptainWang98:
…fectScope

<details>
<br>

>

</details>

---

### Description

In #4236 , `useSortable` is [called](https://stackblitz.com/edit/vitejs-vite-4qbt23?file=src%2Fcomponents%2FRenderBlock.vue) within `onMounted(() => nextTick(...))`, causing here failed to handle `sortable.destroy`. https://github.com/vueuse/vueuse/blob/2e347bd50d5ac13057de70b5a14a6fe5ba3ee1bf/packages/integrations/useSortable/index.ts#L61-L75 
And the bug disappears when handle it manually, here is the [example](https://stackblitz.com/edit/vitejs-vite-1f8sa6be?file=src%2Fcomponents%2FRenderBlock.vue)

### Additional context

Since we can't resctict users where to call `useSortable`, we can show warns if the usage may cause unexpected behaviors.

### DIFF
--- a/packages/integrations/useSortable/index.ts
+++ b/packages/integrations/useSortable/index.ts
@@ -72,7 +72,9 @@ export function useSortable<T>(
 
   tryOnMounted(start)
 
-  tryOnScopeDispose(stop)
+  const registSuccess = tryOnScopeDispose(stop)
+  if (!registSuccess && process.env.NODE_ENV !== 'production')
+    console.warn('useSortable is not called within a EffectScope, causing failure of managing sortable.destory. You may have to call stop() manually.')
 
   return {
     stop,


packages/integrations/useSortable/index.ts
```diff
@@ -72,7 +72,9 @@ export function useSortable<T>(
 
   tryOnMounted(start)
 
-  tryOnScopeDispose(stop)
+  const registSuccess = tryOnScopeDispose(stop)
+  if (!registSuccess)
+    console.warn('useSortable is not called within a EffectScope, causing failure of managing sortable.destory. You may have to call stop() manually.')
```
OrbisK: if we want this, we should limit this to development

CaptainWang98: Dev only warn is right.
I'm wondering if there is way to refact `useSortable` to avoid `'unhandling destroy'`, do you have any ideas?
I'll create a new PR if I get the refact way.

antfu:
Thank you, but we don't do this (VueUse is very strict on console.log), it's users' responsibility to know how effects work, and if they want to have the effect globally or do manual clean-up, we shouldn't prevent them from doing that.

PR #4710, opened 2025-04-14
docs: remove docs-utils and inline stringify function into demos

OrbisK:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

This PR inlines the strigify function in every demo.vue. It then removes the docs-utils

Fixes #4705

PR #4928, opened 2025-07-30
feat(yamlStringify): extraction of yaml.stringify

doyuli:
### Description

I found this code repeated 22 times in vueuse, I think it should be extracted to utils, This is more conducive to centralized management.

```ts
import YAML from 'yaml'

const stringify = reactify(
  (input: any) => YAML.stringify(input, (k, v) => {
    if (typeof v === 'function') {
      return undefined
    }
    return v
  }, {
    singleQuote: true,
    flowCollectionPadding: false,
  }),
)
```


packages/shared/utils/stringify.ts
```diff
@@ -0,0 +1,22 @@
...
+  return reactify(
+    (input: any) => _stringify(
+      input,
+      (_, v) => typeof v === 'function' ? undefined : v,
+      options,
+    ),
+  )
+}
+
+export const yamlStringify = createReactifyStringify()
```
Alfred-Skyblue: You can directly write `export const stringify = createReactifyStringify()`

doyuli: Thanks, I've submitted the changes

OrbisK:
Hey, Thank you for your PR, BUT this is intentional: https://github.com/vueuse/vueuse/pull/4710/

We aim to remove as much dependencies from the demos as possible to always be compartible with the playground links.

I have to close this one.

PR #4932, opened 2025-07-31
docs: add `whitelistCommits` for Changelog

serkodev:

---

### Description

This PR adds commit d32f80ca4e0f7600b68cbca05341b351e31563c1 to the whitelist in `changelog.ts` so that it appears in the Changelog.

The reason is that the changes in that PR represent a feature-level update instead of just `refactor` from the user’s perspective.

### DIFF
--- a/scripts/changelog.ts
+++ b/scripts/changelog.ts
@@ -10,6 +10,8 @@ const git = Git({
 let cache: CommitInfo[] | undefined
 
 const whitelistCommits = new Set<string>([
+  // This commit adds NO_SIDE_EFFECTS annotations to pure functions.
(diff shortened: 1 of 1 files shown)


scripts/changelog.ts
```diff
@@ -9,6 +9,10 @@ const git = Git({
 })
 let cache: CommitInfo[] | undefined
 
+const whitelistCommits = new Set<string>([
+  'd32f80ca4e0f7600b68cbca05341b351e31563c1',
```
antfu, line 15: Let's have a comment to explain the purpose of the hash

serkodev: Added, thanks!

antfu (approved):

Issue #4930, opened 2025-07-31
BUG | `useElementVisibility` | `rootMargin` marked as responsive, but not implemented


akfakf0509:
### Describe the bug

Although the `rootMargin` value in the `UseElementVisibilityOptions` type appears to support responsiveness, the actual implementation does not take responsiveness into account.
and personally, considering the purpose of this composable, I think it might make more sense not to accept the rootMargin property at all.

On StackBlitz, `rootMargin` doesn't seem to work correctly—possibly due to viewport issues. I recommend running the reproduction locally for more accurate results.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-eeuwym9e

### System Info

```Shell
System:
  OS: Linux 5.0 undefined
  CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  Memory: 0 Bytes / 0 Bytes
  Shell: 1.0 - /bin/jsh
Binaries:
  Node: 20.19.1 - /usr/local/bin/node
  Yarn: 1.22.19 - /usr/local/bin/yarn
  npm: 10.8.2 - /usr/local/bin/npm
  pnpm: 8.15.6 - /usr/local/bin/pnpm
npmPackages:
  @vueuse/core: ^13.6.0 => 13.6.0 
  vue: ^3.5.18 => 3.5.18
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vueuse.org/guide).
- [x] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

akfakf0509:
@PeikyLiu
Thank you for the PR! What are your thoughts on possibly removing the rootMargin option?

[...]

PR #4933, opened 2025-07-31
feat: support reactive options

PeikyLiu:

<details>
<br>

>

</details>

---

### Description

Before:
useElementVisibility accepted a reactive parameter **rootMargin**,and the it typed as **MaybeRefOrGetter<string>**,but it didn't support dynamic updates.

After:
Support dynamic updates

### Solution

use watchEffect for  monitoring changes to all reactive options, including **rootMargin**, **threshold** and **scrollTarget**.When these values change, the old IntersectionObserver is cleaned up and a new instance is created, ensuring the configuration is always up to date.

related #4930

### DIFF
--- a/packages/core/useElementVisibility/index.ts
+++ b/packages/core/useElementVisibility/index.ts
@@ -3,7 +3,7 @@ import type { ConfigurableWindow } from '../_configurable'
 import type { MaybeComputedElementRef } from '../unrefElement'
 import type { UseIntersectionObserverOptions } from '../useIntersectionObserver'
 import { watchOnce } from '@vueuse/shared'
-import { shallowRef, toValue } from 'vue'
+import { shallowRef, toValue, watchEffect } from 'vue'
 import { defaultWindow } from '../_configurable'
 import { useIntersectionObserver } from '../useIntersectionObserver'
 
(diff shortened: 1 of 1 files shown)


doyuli:
I think we can modify its underlying `useIntersectionObserver` to support responsive rootMargin, What do you think?
```ts
const stopWatch = isSupported.value
    ? watch(
        () => [targets.value, unrefElement(root as MaybeComputedElementRef), isActive.value, toValue(rootMargin)] as const,
        ([targets, root]) => {
          cleanup()
          if (!isActive.value)
            return

          if (!targets.length)
            return

          const observer = new IntersectionObserver(
            callback,
            {
              root: unrefElement(root),
              rootMargin: toValue(rootMargin) ?? '0px',
              threshold,
            },
          )

          targets.forEach(el => el && observer.observe(el))

          cleanup = () => {
            observer.disconnect()
            cleanup = noop
          }
        },
        { immediate, flush: 'post' },
      )
    : noop
```
In this way, you can just pass the responsive rootMargin directly into `useElementVisibility`

PeikyLiu:
> I think we can modify its underlying `useIntersectionObserver` to support responsive rootMargin, What do you think?
> 
> ```ts
> const stopWatch = isSupported.value
>     ? watch(
>         () => [targets.value, unrefElement(root as MaybeComputedElementRef), isActive.value, toValue(rootMargin)] as const,
>         ([targets, root]) => {
>           cleanup()
>           if (!isActive.value)
>             return
> 
>           if (!targets.length)
>             return
> 
>           const observer = new IntersectionObserver(
>             callback,
>             {
>               root: unrefElement(root),
>               rootMargin: toValue(rootMargin) ?? '0px',
>               threshold,
>             },
>           )
> 
>           targets.forEach(el => el && observer.observe(el))
> 
>           cleanup = () => {
>             observer.disconnect()
>             cleanup = noop
>           }
>         },
>         { immediate, flush: 'post' },
>       )
>     : noop
> ```
> 
> In this way, you can just pass the responsive rootMargin directly into `useElementVisibility`

First, I'd like to clarify that this bug originates in the useElementVisibility hook rather than in useIntersectionObserver. Looking at the documentation, I noticed that rootMargin in useElementVisibility is defined as a MaybeRefOrGetter<string>, indicating it should support reactive values, while in useIntersectionObserver it's defined as a simple string type. This type definition difference suggests that the original design intention was to keep useIntersectionObserver simpler with static configuration, while enabling reactive options in the higher-level useElementVisibility hook. Therefore, I believe modifying useIntersectionObserver would go against its original design principles and could potentially introduce unexpected behavior in other components that rely on it.

packages/core/useElementVisibility/index.ts
```diff
@@ -40,36 +40,51 @@ export function useElementVisibility(
...
     once = false,
   } = options
+
   const elementIsVisible = shallowRef(false)
 
-  const { stop } = useIntersectionObserver(
-    element,
-    (intersectionObserverEntries) => {
-      let isIntersecting = elementIsVisible.value
+  const cleanup = shallowRef<() => void>(() => {})
```
antfu: This doesn't have to be a ref I think

PeikyLiu: ok!I got it! thanks for your CR

packages/core/useElementVisibility/index.ts
```diff
@@ -40,36 +45,51 @@ export function useElementVisibility(
...
-    },
-    {
-      root: scrollTarget,
-      window,
-      threshold,
-      rootMargin: toValue(rootMargin),
-    },
-  )
+    cleanup = stop
+    onCleanup(() => stop())
```
antfu: Is onCleanup and cleanup duplicated?

PeikyLiu: yes, it's look like duplicated, the onCleanup will be called before next watchEffect execution, so i just need keep the onCleanup and remove cleanup, right? and I saw the documentation mentioned we can use onWatcherCleanup on vue3.5+, may i ? Or we need maintain compatibility with 3.4. Finally, thank you for your code review

antfu: I think we are targeting to have this PR in v14, which would require Vue 3.5, so it would be fine to use new features

PeikyLiu: I'm very sorry. I've been quite busy recently.I will done at this week for this pr, and I found another bug during the test,when switch the target element visible(such like scroll),it will be recreating a new observer, because I change the reactive state 'elementIsVisble' in watchEffct callback, Obviously, this is incorrect. And I think i should use 'watch' to replace 'watchEffect'. 
image

packages/core/useElementVisibility/index.md
```diff
@@ -4,7 +4,7 @@ category: Elements
 
 # useElementVisibility
 
-Tracks the visibility of an element within the viewport.
+Tracks the visibility of an element within the viewport11.
```
OrbisK, line 7: 👀

PeikyLiu: > 👀

It has been corrected. Thanks

PR #4936, opened 2025-07-31
feat(useAsyncState): cancel previous promise and use latest promise

snowbitx:

<details>
<br>

>

</details>

---

### Description

This PR fixes an issue where, due to network delays, older requests could overwrite the latest one — like when switching tabs quickly. It now ensures only the latest result is used, and also adds support for manual cancellation.

### Additional context

Replaces https://github.com/vueuse/vueuse/pull/4797 — reopened because the previous branch name conflicted with an existing one, and includes updated test cases.


snowbitx:
[demo](https://stackblitz.com/edit/efeb8ik7?file=index.js)

packages/core/useAsyncState/index.ts
```diff
@@ -108,10 +117,17 @@ export function useAsyncState<Data, Params extends any[] = any[], Shallow extend
...
-    const _promise = typeof promise === 'function'
+    let _promise = typeof promise === 'function'
       ? promise(...args as Params)
       : promise
-
+    if (cancellable) {
+      const rawPromise = _promise
+      _promise = new Promise((resolve, reject) => {
+        cancel()
+        cancel = () => (resolve = reject = noop)
```
ilyaliao, line 127: But this doesn't actually cancel the previous promise; it still continues executing in the background, right?

ilyaliao, line 127: But this doesn't actually cancel the previous promise; it still continues executing in the background, right?

snowbitx: > But this doesn't actually cancel the previous promise; it still continues executing in the background, right?

Thanks for the review 🙏 — you're totally right, the previous promise isn't truly cancelled; it still runs in the background.
What the code does is invalidate the resolve and reject of the previous promise whenever execute is called, so that only the latest one takes effect.
This way, we avoid having any result from the old promise interfere with the new one

ilyaliao: Maybe you could look into AbortController

OrbisK: We can wait for https://github.com/nuxt/nuxt/pull/32531 and adapt some changes

snowbitx: > Maybe you could look into AbortController

Got it, thanks！

snowbitx: > We can wait for [nuxt/nuxt#32531](https://github.com/nuxt/nuxt/pull/32531) and adapt some changes

I see see. Thanks for the reply

PR #4938, opened 2025-08-04
docs(useCycleList): modify demo

pkc918:
https://github.com/user-attachments/assets/23dab65f-af11-4d4a-8e15-d0048b861f8f

This seems to fit the demo better, Aha.


antfu:
Thank you, but what's the purpose of this change? Is it just about styling? While I do like the new design, I am just a bit concerned if that would be too much and may give users the wrong impression that the list has to be rounded? I feel this would be a bit too specific for what that function is supposed to do

pkc918:
Thank you for your reply. I agree with what you said: "I feel this would be a bit too specific for what that function is supposed to do". I will close this PR.

PR #4953, opened 2025-08-10
feat(useEventSource): add `serializer`

imddc:

<details>
<br>

>

</details>

---

### Description

- JSON Parsing: Automatically parse JSON strings from EventSource into JavaScript objects
- Custom Transformations: Apply user-defined transformation functions to incoming data
- Backward Compatibility: All existing functionality remains unchanged, serialization is opt-in

### Additional context

This enhancement addresses a common pain point when working with EventSource APIs that send JSON data as strings, requiring manual parsing in every usage. The implementation follows VueUse's existing patterns and maintains full backward compatibility.


packages/core/useEventSource/index.ts
```diff
@@ -48,6 +48,24 @@ export interface UseEventSourceOptions extends EventSourceInit {
...
+     */
+    parseJSON?: boolean
+
+    /**
+     * Custom data transformer function
+     * @param data Raw data from EventSource
+     * @returns Transformed data
+     */
+    transform?: (data: any) => any
+  }
```
OrbisK: IMO we should only have the transform/serializer. 

we should add an optional generic to the options and use this for the serializer
```
  serializer?: (v: unknown) => T
```

the serializer should default to:

```
  serializer: (v) => v
```

The reason I don't like the parseJSON option is that it doesn't save much.

```
  serialization?: {
    parseJSON: true
  }
```
vs 
```
  serializer: (v) => JSON.parse(v)
```
It is not clear to the user what should happen if we pass both.

```
  serialization?: {
    parseJSON: true,
    transform: (v)=> /* ... */
  }
```

OrbisK:
Thank you for your PR.

packages/core/useEventSource/index.ts
imddc: thank you for the feedback!

I modified the code according to your suggestion

how is it this time?

packages/core/useEventSource/index.ts
```diff
@@ -131,8 +138,15 @@ export function useEventSource<Events extends string[], Data = any>(
     immediate = true,
     autoConnect = true,
     autoReconnect,
+    serialization,
```
OrbisK:
```suggestion
    serialization: (v)=>v,
```

packages/core/useEventSource/index.ts
```diff
@@ -131,8 +138,15 @@ export function useEventSource<Events extends string[], Data = any>(
...
   } = options
 
+  const processData = (rawData?: Data) => {
+    if (rawData === undefined)
+      return null
+    return serialization ? serialization(rawData) : rawData
+  }
+
```
OrbisK:
```suggestion
```

packages/core/useEventSource/index.ts
```diff
@@ -183,14 +197,14 @@ export function useEventSource<Events extends string[], Data = any>(
 
     es.onmessage = (e: MessageEvent) => {
       event.value = null
-      data.value = e.data
+      data.value = processData(e.data)
```
OrbisK:
```suggestion
      data.value = serialization(e.data)
```

packages/core/useEventSource/index.ts
```diff
@@ -183,14 +197,14 @@ export function useEventSource<Events extends string[], Data = any>(
...
       lastEventId.value = e.lastEventId
     }
 
     for (const event_name of events) {
       useEventListener(es, event_name, (e: Event & { data?: Data, lastEventId?: string }) => {
         event.value = event_name
-        data.value = e.data || null
+        data.value = processData(e.data)
```
OrbisK:
```suggestion
        data.value = serialization(e.data)
```

imddc:
> Looks good so far. Some small adjustments.
> 
> Maybe we should rename it to: `serializer` like `useBase64`

Thank you！ 
I have adopted your suggestion, and it looks much better!

OrbisK:
There are no new commits. Maybe you forgot to push 😅

imddc:
> There are no new commits. Maybe you forgot to push 😅

I forgot last time. Could you please do the code review again?

packages/core/useEventSource/index.ts
```diff
@@ -183,14 +186,14 @@ export function useEventSource<Events extends string[], Data = any>(
 
     es.onmessage = (e: MessageEvent) => {
       event.value = null
-      data.value = e.data
+      data.value = serialization(e.data) || null
```
OrbisK:
```suggestion
      data.value = serialization(e.data) ?? null
```

packages/core/useEventSource/index.ts
```diff
@@ -183,14 +186,14 @@ export function useEventSource<Events extends string[], Data = any>(
...
       lastEventId.value = e.lastEventId
     }
 
     for (const event_name of events) {
       useEventListener(es, event_name, (e: Event & { data?: Data, lastEventId?: string }) => {
         event.value = event_name
-        data.value = e.data || null
+        data.value = serialization(e.data) || null
         lastEventId.value = e.lastEventId || null
```
OrbisK:
```suggestion
        data.value = serialization(e.data) ?? null
        lastEventId.value = e.lastEventId ?? null
```

packages/core/useEventSource/index.ts
```diff
@@ -110,13 +112,13 @@ function resolveNestedOptions<T>(options: T | true): T {
...
+export function useEventSource<Events extends string[], Data = any, TransformedData = Data>(
   url: MaybeRefOrGetter<string | URL | undefined>,
   events: Events = [] as unknown as Events,
-  options: UseEventSourceOptions = {},
-): UseEventSourceReturn<Events, Data> {
+  options: UseEventSourceOptions<Data, TransformedData> = {},
+): UseEventSourceReturn<Events, Data, TransformedData> {
   const event: ShallowRef<string | null> = shallowRef(null)
-  const data: ShallowRef<Data | null> = shallowRef(null)
+  const data: ShallowRef<Data | TransformedData | null> = shallowRef(null)
```
OrbisK:
```suggestion
  const data: ShallowRef<TransformedData | null> = shallowRef(null)
```

OrbisK (changes requested):

imddc:
how about now?

packages/core/useEventSource/index.ts
```diff
@@ -48,14 +48,16 @@ export interface UseEventSourceOptions extends EventSourceInit {
...
 }
 
-export interface UseEventSourceReturn<Events extends string[], Data = any> {
+export interface UseEventSourceReturn<Events extends string[], Data = any, TransformedData = Data> {
   /**
    * Reference to the latest data received via the EventSource,
    * can be watched to respond to incoming messages
    */
-  data: ShallowRef<Data | null>
+  data: ShallowRef<Data | TransformedData | null>
```
OrbisK:
```suggestion
  data: ShallowRef<TransformedData | null>
```

sorry, overlooked this one 🫠

packages/core/useEventSource/index.md
```diff
@@ -76,3 +76,22 @@ const { status, data, close } = useEventSource(
...
+Apply custom transformations to incoming data using a serialization function.
+
+```ts
+import { useEventSource } from '@vueuse/core'
+// ---cut---
+const { data } = useEventSource(
+  'https://event-source-url',
+  [],
+  {
+    serialization: (rawData) => JSON.parse(rawData),
```
OrbisK: We held an internal discussion. We want to use the serialiser to ensure consistency with the other composables.

Could you please adjust the PR?

OrbisK (changes requested):

packages/core/useEventSource/index.md
imddc: sorry, I'm not very clear about that,

should i modify the code? 
cause the position you adjusted in the code is in the md file.

or, do you mean that the default value of the serialization function is 
```js 
(rawData) => JSON.parse(rawData)
```

OrbisK: Just reword from `serialization` to `serializer`. The rest is fine

imddc: sure!

packages/core/useEventSource/index.ts
```diff
@@ -48,14 +48,16 @@ export interface UseEventSourceOptions extends EventSourceInit {
    * @default true
    */
   autoConnect?: boolean
+
+  serialization?: (v?: T) => R
```
OrbisK: we have to reword this in all places 😅

imddc: hah,
other files have also been modified now

43081j:
this looks good, but its actually **de**serialising, right?

it is taking the data from the event and deserialising it. afaik, event source data is always a string and its upto the consumer to deserialise it

if this is the case, we should probably either call it `deserializer` or we should follow the structure `useStorage` has (`serializer: {read, write}`) - but just not use the `write` part

@OrbisK thoughts?

OrbisK:
> this looks good, but its actually **de**serialising, right?

Yeah, you're right.

> if this is the case, we should probably either call it `deserializer` or we should follow the structure `useStorage` has (`serializer: {read, write}`) - but just not use the `write` part

I like the idea of having only the `read` part. If we use this approach, we should update `useBase64` to only include the `write` part (v14).

43081j:
👍  that works for me, these 3 functions can share roughly the same serialisation type then, which would be nice

OrbisK:
@imddc would you mind changing this one more time? 😅

imddc:
> @imddc would you mind changing this one more time? 😅


I have modified the code. Please review it again

ilyaliao (approved):

PR #4954, opened 2025-08-11
feat(useClipboardItems): add `options.immediate` and expose `read`

OrbisK:

<details>
<br>

>

</details>

---

### Description

This PR adds `options.immediate` to allow read onMounted. It does also exposes the `read` function for manual reads.

ToDos:

- tests
- discussion about `options.immediate`. This option is only used when `options.read` is set to true. Should we reuse `read` for this?
```
  /**
   * Enabled reading for clipboard
   *
   * @default false
   */
  read?: boolean | { immediate: boolean}
```
- had to use `shallowReadonly` for content.. `readonly` did not work, possibly because some keys of `ClipboardItem` are already readonly?

@43081j @ilyaliao wdyt?

### DIFF
--- a/packages/core/useClipboardItems/demo.vue
+++ b/packages/core/useClipboardItems/demo.vue
@@ -4,7 +4,7 @@ import { effect, shallowRef } from 'vue'
 
 const input = shallowRef('')
 
-const { content, isSupported, copy } = useClipboardItems()
+const { content, isSupported, copy, read } = useClipboardItems()
 const computedText = shallowRef('')
 const computedMimeType = shallowRef('')
 effect(() => {
@@ -37,11 +37,14 @@ function createClipboardItems(text: string) {
     </p>
     <input v-model="input" type="text">
     <button
-      @click="
-        copy([createClipboardItems(input)])
-      "
+      @click="() => copy([createClipboardItems(input)])"
     >
       Copy
+    </button>
+    <button
+      @click="() => read()"
+    >
+      Read
     </button>
   </div>
   <p v-else>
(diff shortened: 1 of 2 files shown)


43081j:
do you have an example of where we might want to do this?

as far as i understand, we react to cut/copy events so, after mount, the value should always reflect whats in clipboard. this might mean you don't need a `read` function?

OrbisK:
> do you have an example of where we might want to do this?

https://deploy-preview-4954--vueuse.netlify.app/core/useClipboardItems/#useclipboarditems

I think its a permission thing. If you have the exposed read, you can control when you want to prompt for the permission.

https://github.com/user-attachments/assets/127e1514-6803-4a8a-9f36-e25fd87996e8

Or we have a bug, because copy/cut does not update it for me

OrbisK:
If you want to manually read copy from outside the window

OrbisK:
```ts
document.addEventListener('copy', (e) => {
    console.log("COPY");
    navigator.clipboard.read().then(()=>{console.log("bope")})
})
```
navigator.clipboard read does not trigger if ctrl + c, but "copy" does


---------------

```ts
document.addEventListener('copy', async (e) => {
    console.log("COPY");
    await new Promise((res)=>setTimeout(()=>res(1), 1))
    navigator.clipboard.read().then(()=>{console.log("aha")})
})
```
this does work

----- 

maybe we should use `e.clipboardData`

OrbisK:
Maybe we should not go for mounted but for focus change 🤔 but I think it would be nice to have the opt in into manual mode. This way you can control if you want to prompt for the permission

OrbisK:
FYI @43081j I have removed `"on-focus"` again. Mozilla has some restrictions regarding reading without explicit interaction. We should make the read() function available to allow users to manually read if they want to sync manually.

43081j (approved):

ilyaliao (approved):

PR #4960, opened 2025-08-14
feat(useSpeechSynthesisOptions): add option `onBoundary`

pkc918:
I want to know the current position of the speech being read in useSpeechSynthesis. There should be a callback function onBoundary supported in the options.

https://github.com/user-attachments/assets/2c3d912c-e2ff-42db-9e8f-09aac4b2b729


ilyaliao (approved):
Thanks 💚

PR #4969, opened 2025-08-15
fix(useUrlSearchParams): restore proper history and navigation behavior

### LINKED ISSUE #4968: BUG | `useUrlSearchParams` | `writeMode: 'push'` breaks browser history navigation
### Describe the bug

```ts
useUrlSearchParams(undefined, { writeMode: "push" });
```
breaks browser history navigation - this is because the callback of the `pausableWatch` gets run by the `write` call in `onChange`, which in turn breaks browser history

### Reproduction

https://stackblitz.com/edit/vitejs-vite-epnsxtm6?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: Linux 6.15 cpe:/o:nixos:nixos:25.11 25.11 (Xantusia)
    CPU: (16) x64 AMD Ryzen 7 5700G with Radeon Graphics
    Memory: 17.54 GB / 30.69 GB
    Container: Yes
    Shell: 4.0.2 - /run/current-system/sw/bin/fish
  Binaries:
    Node: 20.19.3 - /nix/store/lz7iav1hd92jbv44zf2rdd7b2mj23536-nodejs-20.19.3/bin/node
    Yarn: 1.22.22 - /nix/store/v3gr7h5266sgf8ylg8x7jxw81z56sf3w-corepack-nodejs-22.17.0/bin/yarn
    npm: 10.8.2 - /nix/store/lz7iav1hd92jbv44zf2rdd7b2mj23536-nodejs-20.19.3/bin/npm
    pnpm: 10.11.1 - /nix/store/v3gr7h5266sgf8ylg8x7jxw81z56sf3w-corepack-nodejs-22.17.0/bin/pnpm
  npmPackages:
    @vueuse/core: ^13.5.0 => 13.5.0 
    vue: ^3.5.17 => 3.5.17
```

### Used Package Manager

pnpm

### Validations

yosong-github:
fix #4968 browser's history recording and navigation functions

<details>
<br>

>

</details>

---

### Description

fix #4968


patienry:
Hello, I am also trying to handle this issue today. I followed the steps below for debugging.

In the example of the issue:
1. Enter some values one by one in the input box.
2. Click the Back button.

image

I found that `state` would be reset once and then become the correct value. The callback in `pausableWatch` is also executed as a result.This reset phenomenon confused me. I tried to prevent the reset, but all attempts failed.May I ask if you've encountered the same problem?

ilyaliao:
Thank you for your PR. Can you add some tests for this feature to prevent future regressions?

yosong-github:
> Thank you for your PR. Can you add some tests for this feature to prevent future regressions?

No problem. I have already added the unit tests.

ilyaliao:
I noticed that the test you added would still pass even without your fix. This means the test might not be covering the actual bug.

Could you double-check the test logic and make sure it would fail without your patch?

yosong-github:
> I noticed that the test you added would still pass even without your fix. This means the test might not be covering the actual bug.
> 
> Could you double-check the test logic and make sure it would fail without your patch?

I revised the unit tests to ensure that without my patch, the program would fail the tests.

ilyaliao (approved):

packages/core/useUrlSearchParams/index.test.ts
```diff
@@ -245,6 +283,95 @@ describe('useUrlSearchParams', () => {
             break
         }
       })
+
+      it('when the user presses the forward or backward button of the browser, the parameters should be updated', async () => {
```
ilyaliao: Should we add issue numbers here like Vue or Nuxt do? Maybe it would make it easier to understand the purpose of this test case.

/cc @OrbisK @43081j wdyt?

43081j: personally i prefer having well written test descriptions rather than deferring to an issue

often issue-driven tests can be far too specific imo too

yosong-github: have any good ideas or anything that needs to be revised?

43081j: you're basically testing that it mutates the history stack, so i'd probably name this test something like:

> should push history state when value is updated

antfu (approved):

PR #4974, opened 2025-08-19
fix(useEventSource): replace wrong `Data` generic with `string`

Menci:

<details>
<br>

>

</details>

---

### Description

Remove the `Data` generic from `useEventSource` and just use `string` instead, as SSE's data could only be string.

https://html.spec.whatwg.org/multipage/server-sent-events.html#the-eventsource-interface

### Additional context

In [the original PR](https://github.com/vueuse/vueuse/pull/4598#issuecomment-2661102701), the author referenced MDN's docs:

> The data sent by the message emitter; this can be any data type, depending on what originated this event

Yes, this applies to `MessageEvent`, because that `MessageEvent` is not only born for `EventSource`, but also used in non-SSE scenarios like `postMessage`. This is possible to be other types in other scenarios, but only `string` is possible for SSE.

### DIFF
--- a/packages/core/useEventSource/index.ts
+++ b/packages/core/useEventSource/index.ts
@@ -50,12 +50,12 @@ export interface UseEventSourceOptions extends EventSourceInit {
   autoConnect?: boolean
 }
 
-export interface UseEventSourceReturn<Events extends string[], Data = any> {
+export interface UseEventSourceReturn<Events extends string[]> {
   /**
    * Reference to the latest data received via the EventSource,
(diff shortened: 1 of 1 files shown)


OrbisK:
I think this is obsolete with https://github.com/vueuse/vueuse/pull/4953

@Menci can you confirm this?

Menci:
@OrbisK Oh, that makes sense!

OrbisK:
Thank you anyway ❤️

PR #4977, opened 2025-08-21
refactor(docs): replace `twoslash` with own function for comparing js snippet

serkodev:

---

### Description

In `MarkdownTransform` plugin for vitepress, replace the `twoslash` dependency with a custom function for comparing JavaScript snippets, enhancing performance and memory usage.

The build time was originally `115.32s`, and is now `79.42s`, which is `31.13%` faster!

### DIFF
--- a/packages/.vitepress/plugins/markdownTransform.ts
+++ b/packages/.vitepress/plugins/markdownTransform.ts
@@ -2,7 +2,6 @@ import type { Plugin } from 'vite'
 import { existsSync } from 'node:fs'
 import { join, resolve } from 'node:path'
 import { format } from 'prettier'
-import { createTwoslasher } from 'twoslash'
 import ts from 'typescript'
 import { packages } from '../../../meta/packages'
 import { version as currentVersion } from '../../../package.json'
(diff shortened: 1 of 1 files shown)


packages/.vitepress/plugins/markdownTransform.ts
```diff
@@ -264,6 +252,51 @@ function replaceAsync(str: string, match: RegExp, replacer: (substring: string,
...
+  const formattedTS = (await format(snippet.replace(/\n+/g, '\n'), { semi: false, singleQuote: true, parser: 'typescript' })).trim()
+  const js = ts.transpileModule(formattedTS, {
+    compilerOptions: { target: 99 },
+  })
+  const formattedJS = (await format(js.outputText, { semi: false, singleQuote: true, parser: 'typescript' })).trim()
+  return formattedTS === formattedJS ? null : formattedJS
+}
+
+function removeTwoslashNotations(snippet: string): string {
+  function cutAboveMarker(input: string): string {
```
43081j: nit: can these functions be moved outside of `removeTwoslashNotations` since they can just be pure operations?

and the two `marker` consts can be outside too (like `MARKER_CUT` and `MARKER_CUT_AFTER` or some such thing)

43081j:
so basically we're using twoslash in main just to strip the syntax, and actually not making any use of twoslash itself

this change makes sense to me, given that 👍  since we just want to drop the cut sections

packages/.vitepress/plugins/markdownTransform.ts
serkodev: Just did some refactoring, readability and performance should be improved. 😃

PR #4997, opened 2025-08-29
feat(createSharedComposable)!: return only the sharedComposable on client side

### LINKED ISSUE #4991: `createSharedComposable` | SSR Handlers Support

by @atinux:
> I see people using createSharedComposable from VueUse but it’s dangerous on SSR at it can be shared across requests

> Most of the time the fix I was doing is to return only the sharedComposable on client side: https://github.com/nuxt/nuxt.com/blob/2269bdf2c332c3d465c24aaebe2865bed7d58d10/app/composables/useNavigation.ts#L352

### Suggested solution

Support SSR handler for `createSharedComposable` and implement the handler in the Nuxt module to support SSR properly

### Alternative

_No response_

### Additional context

_No response_

### Validations

9romise:

<details>
<br>

>

</details>

---

### Description

This PR introduces `packages/**/*.server.test.ts` files dedicated to server-side testing, along with simple scripts and CI configurations.

Closes #4991

### DIFF
--- a/packages/shared/createSharedComposable/index.ts
+++ b/packages/shared/createSharedComposable/index.ts
@@ -2,6 +2,7 @@ import type { EffectScope } from 'vue'
 import type { AnyFn } from '../utils'
 import { effectScope } from 'vue'
 import { tryOnScopeDispose } from '../tryOnScopeDispose'
+import { isClient } from '../utils'
 
 export type SharedComposableReturn<T extends AnyFn = AnyFn> = T
 
@@ -13,6 +14,9 @@ export type SharedComposableReturn<T extends AnyFn = AnyFn> = T
  * @__NO_SIDE_EFFECTS__
  */
 export function createSharedComposable<Fn extends AnyFn>(composable: Fn): SharedComposableReturn<Fn> {
+  if (!isClient)
+    return composable
+
   let subscribers = 0
   let state: ReturnType<Fn> | undefined
   let scope: EffectScope | undefined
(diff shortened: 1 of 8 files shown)


packages/shared/createSharedComposable/index.ts
```diff
@@ -13,6 +14,9 @@ export type SharedComposableReturn<T extends AnyFn = AnyFn> = T
  * @__NO_SIDE_EFFECTS__
  */
 export function createSharedComposable<Fn extends AnyFn>(composable: Fn): SharedComposableReturn<Fn> {
+  if (!isClient)
+    return composable
```
antfu, line 18: I am not sure if this is the most expected behavior. But at least we should mention this behavior in the docs. (also better consider this breaking to bring more awareness)

9romise:
I believe `createGlobalState` would likely need this too.

Additionally, I noticed we already have an [`ssr-plugin`](https://github.com/vueuse/vueuse/blob/main/packages/nuxt/ssr-plugin.mjs) in `@vueuse/nuxt` – should we handle this within that plugin?

antfu (approved):

packages/shared/createSharedComposable/index.ts
Tofandel: Couldn't we use getCurrentInstance if available and keep a Weakmap based on the root app of which instance is requesting the sharedComposable to avoid the cross request pollution while still preserving the desired behavior?

Issue #5006, opened 2025-09-01
`useTransition` | Support for custom interpolators


scottbedard:
### Clear and concise description of the problem

The capability of `useTransition` and `executeTransition` could be greatly improved with minimal api changes.

Currently, transitions [only operate via linear interpolation](https://github.com/vueuse/vueuse/blob/2c6c4f5fee523852fb5261faa4776ba04aa52108/packages/core/useTransition/index.ts#L123-L125). This is fine for text displays, and simple number transitions, but can be limiting in animation and 3D use cases.

For example, imagine a vector transition from `[0, 1]` to `[1, 0]`. When the source changes, our transitioned value will take the following green path. The circular red path however, is beyond the capability of our function.

![Image](https://github.com/user-attachments/assets/1d7d0ffb-53ad-46c7-924c-242095a3d1ee) ![Image](https://github.com/user-attachments/assets/13d94134-4310-46a7-8fbc-1aa39dfc32e4)

To create the red path, more reactivity is needed. One ref for the transition progress, and another to compute output value.

### Suggested solution

I'd like to make the interpolation function customizable, and leave linear interpolation the default. For example, we could use a spherical linear interpolation to produce the red path.

```js
import { slerp } from 'some-math-library'

const source = ref([0, 1])

const output = useTransition(source, {
  interpolation: slerp
})
```

And taking things further, now that the interpolator is customizable, we can remove the type restrictions on `source`. Our source refs could be more sophisticated objects, like a threejs [`Quaternion`](https://threejs.org/docs/?q=quaternion#api/en/math/Quaternion) or [`Color`](https://threejs.org/docs/?q=color#api/en/math/Color).

```js
const source = ref(new Quaternion())

const output = useTransition(source, {
  interpolation: (q1, q2, t) => q1.slerp(q2, t)
})
```

```js
const source = ref(new Color())

const output = useTransition(source, {
  interpolation: (c1, c2, t) => c1.lerpHSL(c2, t)
})
```

To implement this, we'd only need add an option, and adjust [these lines of code](https://github.com/vueuse/vueuse/blob/2c6c4f5fee523852fb5261faa4776ba04aa52108/packages/core/useTransition/index.ts#L175-L178) to utilize it. We can also simplify the function signature by declaring the source type as a generic.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
[...]

PR #5011, opened 2025-09-03
Add support for custom interpolator functions

scottbedard:

<details>
<br>

>

</details>

---

### Description

Extracted linear interpolation function to option, and allow it to be overwritten

Implements https://github.com/vueuse/vueuse/issues/5006

### Additional context

I chose to demo a word-interpolation just to show that _anything_ can be transitioned.

### DIFF
--- a/packages/core/useTransition/index.test.ts
+++ b/packages/core/useTransition/index.test.ts
@@ -338,4 +338,25 @@ describe('useTransition', () => {
 
     expectBetween(transition.value, 0, 0.5)
   })
+
+  it('supports custom interpolator functions', async () => {
+    const source = shallowRef('')
+    const interpolator = vi.fn((_a, _b, n) => n < 1 / 2 ? 'foo' : 'bar')
+    const transition = useTransition(source, {
+      duration: 100,
+      interpolator,
+    })
+
+    source.value = 'test'
+
+    await promiseTimeout(25)
+    expect(interpolator).toBeCalled()
+    expect(transition.value).toBe('foo')
+
+    await promiseTimeout(50)
+    expect(transition.value).toBe('bar')
+
+    await promiseTimeout(50)
+    expect(transition.value).toBe('test')
+  })
 })
--- a/packages/core/useTransition/index.ts
+++ b/packages/core/useTransition/index.ts
@@ -242,7 +242,7 @@ export function useTransition<T>(
       : val as SourceVal
   }
 
-  const outputRef = shallowRef(sourceVal()) as Ref<T>
+  const outputRef = shallowRef(sourceVal())
 
   watch(sourceVal, async (to) => {
     if (toValue(options.disabled))
--- a/packages/core/useTransition/index.ts
+++ b/packages/core/useTransition/index.ts
@@ -17,7 +17,7 @@ export type EasingFunction = (n: number) => number
 /**
  * Transition options
  */
-export interface TransitionOptions extends ConfigurableWindow {
+export interface TransitionOptions<T> extends ConfigurableWindow {
 
   /**
    * Manually abort a transition
@@ -29,7 +29,7 @@ export interface TransitionOptions extends ConfigurableWindow {
    */
   duration?: MaybeRef<number>
 
-  interpolator?: (a: any, b: any, t: any) => any
+  interpolator?: (a: T, b: T, t: number) => T
 
   /**
    * Easing function or cubic bezier points for calculating transition values
@@ -37,7 +37,7 @@ export interface TransitionOptions extends ConfigurableWindow {
   transition?: MaybeRef<EasingFunction | CubicBezierPoints>
 }
 
-export interface UseTransitionOptions extends TransitionOptions {
+export interface UseTransitionOptions<T> extends TransitionOptions<T> {
   /**
    * Milliseconds to wait before starting transition
    */
@@ -143,11 +143,11 @@ function defaultInterpolator(a: unknown, b: unknown, t: number) {
  * @param to
  * @param options
  */
-export function executeTransition<T = number>(
+export function executeTransition<T>(
   source: Ref<T>,
   from: MaybeRefOrGetter<T>,
   to: MaybeRefOrGetter<T>,
-  options: TransitionOptions = {},
+  options: TransitionOptions<T> = {},
 ): PromiseLike<void> {
   const {
     window = defaultWindow,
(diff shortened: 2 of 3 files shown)


packages/core/useTransition/index.ts
```diff
@@ -193,14 +208,14 @@ export function executeTransition<T extends number | number[]>(
   })
 }
 
-// option 1: reactive number
-export function useTransition(source: MaybeRefOrGetter<number>, options?: UseTransitionOptions): ComputedRef<number>
```
scottbedard, line 197: The first option was simplified and moved to the end to allow for transitioning unknown types.

packages/core/useTransition/index.test.ts
```diff
@@ -338,4 +338,25 @@ describe('useTransition', () => {
...
+    const source = shallowRef('')
+    const interpolation = vi.fn((_a, _b, n) => n < 1 / 2 ? 'foo' : 'bar')
+    const transition = useTransition(source, {
+      duration: 100,
+      interpolation,
+    })
+
+    source.value = 'test'
+
+    await promiseTimeout(25)
```
OrbisK, line 386: can we use `vi.advanceTimersByTime(25)` here? Maybe its not possible 🤔

packages/core/useTransition/index.ts
```diff
@@ -29,13 +29,18 @@ export interface TransitionOptions extends ConfigurableWindow {
    */
   duration?: MaybeRef<number>
 
+  /**
+   * Custom interpolation function
+   */
+  interpolation?: (a: T, b: T, t: number) => T
```
OrbisK: `(a: T, b: T, t: number) => T` maybe we should expose/export this type

packages/core/useTransition/index.test.ts
scottbedard: The alpha value is determined by `Date.now()`, so I expect using mock timers would work! I'll make a refactor later today to update all transition tests 🙂

https://github.com/vueuse/vueuse/blob/ac0fc1189c56b7431a7a2e8f28177b48b6c4a4e9/packages/core/useTransition/index.ts#L192-L193

packages/core/useTransition/index.ts
scottbedard: Sure, how does the name `InterpolationFunction` sound? I'll export it alongside our `EasingFunction` type.

https://github.com/vueuse/vueuse/blob/ac0fc1189c56b7431a7a2e8f28177b48b6c4a4e9/packages/core/useTransition/index.ts#L12-L15

**Update:** Done! 8d890b40e443e866ad44a5967b94230ed5a26bcc

packages/core/useTransition/index.test.ts
scottbedard: @OrbisK Actually, this is a bit trickier than I thought because we rely on promises that tick via a `requestAnimationFrame` loop. Once we hit `vi.useFakeTimers(), does time stop ticking? I played with it a bit, but no luck.

scottbedard:
@OrbisK I've made one of the refactors you suggested, but using fake timers was a bit more complex than expected so I left the tests as they were.

I've also pushed 02560c93861ee57c6902c6f3bd5145705a7336d2, which makes the deprecations I suggested. Having clear names for each half of the transition will be much easier to read than of a generic "transition" term. I chose "easing" over "timing" because it's the term many other tools use when referring to this function.

I've forwarded the old APIs, but is anything other than a `@deprecated` JSDoc comment needed? My guess is that we make these sort of changes over two major releases, with the first release (14.x) deprecating, and the second one (15.x) removing. If there is any more that needs to happen please let me know! 🙂 

CC @antfu

antfu:
~~Tests are failing, can you have a look? Thanks~~

Updated snapshot, nevermind

antfu (approved):

scottbedard:
Thanks for the review everyone!

scottbedard:
~~@antfu I've removed this from the auto-merge queue because I just realized I didn't update the docs to reflect the new terms. Give me a sec and I'll push another commit! 🏃‍♂️~~

Ok, we should be good to go now! 🙂

scottbedard:
@OrbisK @antfu this PR is ready again, I've updated the documentation. After canceling the auto-merge before, it changed to require approval again. Sorry about that, if either of you can approve this again I think it's good to go.

packages/core/useTransition/index.ts
```diff
@@ -136,25 +163,32 @@ function toVec(t: number | number[] | undefined) {
...
+    ? options.interpolation
+    : defaultInterpolation
+
+  const trans = typeof options.easing !== 'undefined'
+    ? typeof options.easing === 'function'
+      ? options.easing
+      : (toValue(options.easing) ?? linear)
+    : typeof options.transition === 'function'
+      ? options.transition
+      : (toValue(options.transition) ?? linear)
```
OrbisK, line 191: I don't think we need all of those checks.

If `options.easing` and `options.transition` are `MaybeRef<Fn>`, we can do the following (maybe):

```ts
  const trans = toValue(options.easing) ?? toValue(options.transition) ?? linear
```

OrbisK (changes requested):

packages/core/useTransition/index.ts
scottbedard: I can make this code a bit cleaner, but we cannot use `toValue(easing)` / `toValue(transition)`, because this may be a custom easing function. We ran into a regression here when we tried to do this in the past.

https://github.com/vueuse/vueuse/pull/2973

I'll push a commit that normalizes this though, I agree the nested ternaries are a bit weird looking.

scottbedard: @OrbisK Done, I think this removes a branch of redundant logic. When `options.transition` is removed, we can move the logic back into the `transition` function to live alongside `interpolation`

b49631e04df41ff07e8ce1405d514e4088c8b094

OrbisK: Oh! Is it because `toValue` calls it as if it were a getter function?

But shouldnt `unref()` work?

```ts
  const trans = unref(options.easing) ?? unref(options.transition) ?? linear

OrbisK: ([b49631e](https://github.com/vueuse/vueuse/pull/5011/commits/b49631e04df41ff07e8ce1405d514e4088c8b094) does work for me)

scottbedard: It looks like we have eslint guards around importing the `unref` function,
https://github.com/vueuse/vueuse/blob/cf905ccfb5dd7a6a3e65aa087a034b5157c9f9fb/eslint.config.js#L20

There are a couple places where we use `eslint-disable no-restricted-imports`, but I'm not sure what the nuance is there.

scottbedard: Oh, also it looks like the auto-merge took effect anyway 😆

And thanks for your help with these reviews! This will make the function much more powerful in 3D use cases 🙌

PR #5021, opened 2025-09-08
build: simplify baseConfig setup and improve `dist` output

9romise:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

In PR #5004, to ensure consistency of the output code, when bundling packages with `submodules` set as true, multiple configs are generated based on functions as inputs for `tsdown`. This causes each module to be bundled independently, and repeated logic gets bundled multiple times.

This PR will directly add `key-value` pairs in the `entry` for:
- Faster bundling speed
- Smaller output size.

Below is an example for reference:

#### Build time
[Before](https://github.com/vueuse/vueuse/actions/runs/17378607349/job/49330556476?pr=5004):
image

[After](https://github.com/vueuse/vueuse/actions/runs/17557498035/job/49865312905?pr=5021):
image

#### Output
Before:
image

After:
image

PR #5025, opened 2025-09-10
fix(integrations): use relative import in components

9romise:

<details>
<br>

>

</details>

---

### Description

https://github.com/vueuse/vueuse/pull/5021#issuecomment-3272591761

### DIFF
--- a/packages/integrations/useFocusTrap/component.ts
+++ b/packages/integrations/useFocusTrap/component.ts
@@ -1,8 +1,8 @@
 import type { RenderableComponent } from '@vueuse/core'
-import type { UseFocusTrapOptions, UseFocusTrapReturn } from '@vueuse/integrations/useFocusTrap'
 import type { Reactive, SlotsType } from 'vue'
-import { useFocusTrap } from '@vueuse/integrations/useFocusTrap'
+import type { UseFocusTrapOptions, UseFocusTrapReturn } from './index'
 import { defineComponent, h, reactive, shallowRef } from 'vue'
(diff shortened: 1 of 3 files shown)


packages/integrations/useFocusTrap/component.ts
```diff
@@ -1,8 +1,8 @@
 import type { RenderableComponent } from '@vueuse/core'
-import type { UseFocusTrapOptions, UseFocusTrapReturn } from '@vueuse/integrations/useFocusTrap'
 import type { Reactive, SlotsType } from 'vue'
-import { useFocusTrap } from '@vueuse/integrations/useFocusTrap'
+import type { UseFocusTrapOptions, UseFocusTrapReturn } from './index'
 import { defineComponent, h, reactive, shallowRef } from 'vue'
+import { useFocusTrap } from './index'
```
antfu, line 5: I think this was on purpose, otherwise the functions would be duplicated in the `/components` bundle. Instead, we should limit to not use `./index` in component.ts

9romise: Actually, it won’t be bundled repeatedly — it imports a hashed file where duplicate logic has been extracted. See https://github.com/vueuse/vueuse/pull/5021#issuecomment-3272591761.
Since `@vueuse/integrations` isn’t part of `@vueuse/components` and only references content within its own package, I think it doesn’t need to import using a full package name.

antfu: Ah I see.

PR #5034, opened 2025-09-13
feat(computedEager): deprecate `computedEager`

ilyaliao:
### Description

This PR deprecates `computedEager`, as since Vue 3.4 the built-in `computed` already avoids triggering dependents when the value is unchanged, and with VueUse v14 requiring Vue 3.5+, users can use the native `computed` directly.

### DIFF
--- a/packages/shared/computedEager/index.md
+++ b/packages/shared/computedEager/index.md
@@ -7,6 +7,10 @@ alias: eagerComputed
 
 Eager computed without lazy evaluation.
 
+::: info
+This function will be removed in future version.
+:::
+
 ::: tip
 Note💡: If you are using Vue 3.4+, you can use `computed` right away, you no longer need this function.
 In Vue 3.4+, if the computed new value does not change, `computed`, `effect`, `watch`, `watchEffect`, `render` dependencies will not be triggered.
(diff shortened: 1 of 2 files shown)


benedictleejh:
I'm posting here as suggested by @9romise, and hopefully there can be a discussion on a way forward. To repost my comment from #4748:

> I want to chime in a bit here regarding the behaviour of `computedEager` and `computed`. Vue's `computed` does not fully replace the behaviour of `computedEager` at the moment. `computed` is lazy, and the first computation is only be triggered when it is used, whether in the template or later in the script block, whereas `computedEager` is triggered immediately.
>
>Is there a way for Vue's `computed` to be triggered immediately? Without it, I am unable to replace `computedEager` with computed. I am currently on Vue 3.5.22.

While Vue 3.4's `computed` may already avoid triggering dependents (and I'm sure some people were using `computedEager` to do so previously), as things currently stand, `computed` does not work as a complete replacement for the behaviour of `computedEager` due to the continued lazy evaluation of `computed`.

packages/shared/computedEager/index.md
OrbisK, line 4: @ilyaliao we forgot to deprecate it in frontmatter 😄

PR #5042, opened 2025-09-16
feat!: drop watchPausable

ilyaliao:
### Description

Implement (wip) #5040


Tofandel:
I think the deprecation and removal of this util is a mistake because vue's pause and resume control on `watch` have a different behavior that doesn't make it practical to use.

When you pause the watcher and resume it in vue, if the value was changed while it was paused the watcher will fire when it's resumed with the changed value. This is not the case with this util, and allows skipping the watcher altogether.

https://play.vuejs.org/#eNp9Uk1P3DAQ/StTX5KVorRVe6IsUlshtT20qKzEAXMwySRrcGzLH9lFUf47Y4dlQUJcrPHMvOf3Zjyx79bWY0R2wk5946QN4DFEC0rofs1Z8JydcS0Ha1yACRx2FexEaLYVxaIJckSYoXNmgIJ4Cq65boz2AQbfwzoByuIXKmXgyjjVfihWh4ZOOmxzy8JTXt+A8OCDk7q/vll9O3JNVkSP6UkfB5wJlDWU9EYF5biC9RlMXMPCWdvot5Qlgjkd1FWPQkUkXNEZU1AuE5apSn43ckATQ1keiV5hboVLGEizeaMXnnRlOoC5gs+f8uMpyNHpx2W4NEq6BBysEgHpBjBNWfM8p8tmKz3sTFQtGK0egNwHIfWiGijIti9Iu7hVmHhfcLGK1kWITvb1nTeadprlcdaYwUqF7p8NkubJ2clBOGeCVrP7k3PBRawO+WaLzf0b+Tu/TznOLsg0uhE5e64F4XoMS/n88i/uKX4uDqaNpPm94n/0RsWkcWn7EXVLsl/0ZbW/82+kT7Lx5/uA2h9MJaHLBhbf9CF/vmP9KPdL/TXjaGdsfgTUSgZS

OrbisK:
> I think the deprecation and removal of this util is a mistake because vue's pause and resume control on `watch` have a different behavior that doesn't make it practical to use.
> 
> When you pause the watcher and resume it in vue, if the value was changed while it was paused the watcher will fire when it's resumed. This is not the case with this util, and allows skipping the watcher.

Fair point. I think we should still recommend the native pausable watch and maybe extract the behavior you want into a separate util? 🤔

Tofandel:
The `watchPausable` of vueuse is how you would intuitively expect the pause/resume to work. Not sure why we'd have a separate util since the whole util itself is the expected behavior, so it would just be renaming it, causing an unnecessary breaking change.

Deprecating and recommending to use an alternative as if it's a drop in replacement, while it's not, will lead to bugs.

I think it should just be undeprecated.

packages/shared/syncRef/index.ts
Tofandel, line 158: For example even in the core, this would create an unwanted behavior here, when `resume` is called, the left watcher will be firing and then this right watcher will fire again.

So it could cause an infinite watcher loop right there if the transforms do not converge to the same value.

Here is the demo: https://play.vuejs.org/#eNq9VVFv0zAQ/isnP6wpKhlsPFVtJUB7AFWAysTLvIeQOm2G40S203bq8t+5s500HaOAhJaXOufPn7/77nLds7dVFW9qwcZsYlKdVxaMsHUFMlGrKWfWcDbjKi+qUlvYwzax6XoEC5GNwKwTKcstrqGBTJcFDJBpwBVXWa1Sm5cKzL1KETCZ45lZxBWAFJkdE8FkPhtRQOerdYgsfKSs6KwZQ6LuuRrCnoIpRlABLQEyWZs1TGFA/AN3CGApRIWxLJFGhFBeFGKZJ1Zg3Oq6DS9zLbw+pPhe2nVLYXWiTFbqAjf2jQs2uAyC4OEBo5RfK8fZIbRBzM0tJIYU39z2ER3j/HqBqGggrR5ArnpXnZ0dXmLcHtI90QamM9gMn2BaXM8dk7byFBNuP2IirjyDqJf+tDWAkI/ipDSYD12mcYXOR+7NlZMeKmkwECBSYvstkbUY0rXh9BEDqrtK8PiWANu4SmojoqHL1D+uI+INkbi6HRzskaPZiz+Ta2Hqos/ui0rP3nfRyPXNqNcqLcQfCvX+B9+oLn/lm8vzfxpHhfjVN+yXY9/mz+ib711jS/o0o35qv7uasOHiwKFxImmcJbjBlYt5Bs9NSSP3YRpFr18F2R7gbD5GXISvAdoBFbkeDhU5mN95OO4XBD+Mcbh25n9fwEXnD+q1ctzeOguLc7jsEK1RDaqYnPuxi0MWX6woKole4hsavSfupoGXuHQsDSY/Oe+B2AgnNGaZ5av4zpQKx7jTyVlaFlUuhf7sRxdnXQacORs+ulhvKOKZtUh/PBG/MzuKcfYF20LojeCs27OJXgnrt6++fhI7XHebRbmsJaJPbC6EKWVNGj3sXa2WKLuHc2o/uD+gXK2uzdXOCmXapEioaxWH5wz/gd6fSP0g9zJ+E1qsYc1PA/o85w==

packages/core/useStorage/index.ts
Tofandel, line 319: And again here, this would create an infinite loop with vue's behavior or a different output than expected

ilyaliao:
Thanks for pointing this out.

I also noticed the inconsistency while refactoring, through the CI failure, so I temporarily put this PR on hold afterward.

As OrbisK mentioned, we should recommend users to use the native pausable watch. However, I think this needs enough discussion before moving forward.

Sector6759:
I've stumbled across this unexpected behaviour of Vue's own pausable watch for the second time now, wasting time figuring out why my watcher callback was triggered even when I didn't expect it to, so I am pro "undeprecating" VueUse's `watchPausable` as the hint to use Vue's own pausable watch (which behaves differently) has the potential to create a lot of frustration. Maybe you could include an info note explaining the difference so users who read the VueUse documentation know why `watchPausable` exists despite Vue's own pausable watch.

oliver139:
So given that the behaviour is not the same, I think PR #5040 should be reverted to cancel marking it deprecated?
Besides, the message should be changed for avoiding wrong message.
Otherwise, there may be more PR trying to help replace `watchPausable` to native `watch` like #5197

oliver139:
> So given that the behaviour is not the same, I think PR #5040 should be reverted to cancel marking it deprecated? Besides, the message should be changed for avoiding wrong message. Otherwise, there may be more PR trying to help replace `watchPausable` to native `watch` like #5197

I think this need to be done asap?
There is a new related issue again #5634

ilyaliao:
> > So given that the behaviour is not the same, I think PR #5040 should be reverted to cancel marking it deprecated? Besides, the message should be changed for avoiding wrong message. Otherwise, there may be more PR trying to help replace `watchPausable` to native `watch` like #5197
> 
> I think this need to be done asap? There is a new related issue again #5634

Hi, thanks for staying on this.

The team hasn't reached a conclusion yet. We're focused on the v15 release right now, so I expect we'll pick this up in v16.

Personally, I still prefer the native `watch` API. As discussed above, having two pausable-watch mechanisms tends to confuse people about which one they should use — or even which one they're already using. I think we need to unify on one.

cc @OrbisK

oliver139:
> > > So given that the behaviour is not the same, I think PR #5040 should be reverted to cancel marking it deprecated? Besides, the message should be changed for avoiding wrong message. Otherwise, there may be more PR trying to help replace `watchPausable` to native `watch` like #5197
> > 
> > 
> > I think this need to be done asap? There is a new related issue again #5634
> 
> Hi, thanks for staying on this.
> 
> The team hasn't reached a conclusion yet. We're focused on the v15 release right now, so I expect we'll pick this up in v16.
> 
> Personally, I still prefer the native `watch` API. As discussed above, having two pausable-watch mechanisms tends to confuse people about which one they should use — or even which one they're already using. I think we need to unify on one.
> 
> cc @OrbisK

I see. Looking forward to v15 :D

OrbisK:
I think we should aim to implement a change (an option) upstream in Vue. If Vue declines this change, we can consider having a utility that covers the different behaviour. If we need this utility, we should strictly separate it from Vue's pausable watcher. We could consider renaming `watchPausable` to something more descriptive to highlight the differences.

oliver139:
> I think we should aim to implement a change (an option) upstream in Vue. If Vue declines this change, we can consider having a utility that covers the different behaviour. If we need this utility, we should strictly separate it from Vue's pausable watcher. We could consider renaming `watchPausable` to something more descriptive to highlight the differences.

Indeed. Actually two behavior can be merged into one `watchPausable` by providing an option.

PR #5049, opened 2025-09-19
feat(useMouseInElement): add support for tracking inline-level elements

siavava:

<details>
<br>

>

</details>

---

#### Current Behavior

[`useMouseInElement`][docs] works perfectly for block elements, but does not work correctly for inline elements. This is because it uses [`el.getBoundingClientRect()`][bounding-rect], which returns a single rect that wraps the _entire_ element. For inline elements, this means that this rect will potentially include parts that are not in the element itself.

For example, the regions marked with `x` below would be detected as being in the inline element when hovered.

```
--------------------------------------
| xxxxxxxxxxxxxxx here is the inline |
| element which wraps onto this line |
| and ends here. xxxxxxxxxxxxxxxxxxx | 
--------------------------------------
```

#### Proposed Behavior

Instead of using [`el.getBoundingClientRect()`][bounding-rect], we can use [`el.getClientRects()`][client-rects] to get more precise bounding.

Per [the docs][client-rects]:

> The getClientRects() method of the [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) interface returns a collection of [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect) objects that indicate the bounding rectangles for each [CSS border box](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_box_model/Introduction_to_the_CSS_box_model) in a client.
>
> Most elements only have one border box each, but a multiline [inline-level element](https://developer.mozilla.org/en-US/docs/Glossary/Inline-level_content) (such as a multiline [<span>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/span) element, by default) has a border box around each line.

For example, in the above scenario, we'll get three rects returned, the first one containing `"here is the inline"`, the second rect containing `"element which wraps onto this line"`, and the third element containing `"and ends here."`  
And hovering on the `x` regions will no longer be detected as being _still_ within the inline element.

```
                ----------------------
 xxxxxxxxxxxxxx | here is the inline |
                ----------------------
--------------------------------------
| element which wraps onto this line |
--------------------------------------
------------------
| and ends here. | xxxxxxxxxxxxxxxxx
------------------
```

> [!NOTE]
> Since `getClientRects` returns a single `DOMRect` for block elements (see second paragraph in the quoted text above, or check the [documentation source][client-rects]), there should be no regression in performance for block elements.

[docs]:          https://vueuse.org/core/useMouseInElement/
[bounding-rect]: https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect
[client-rects]:  https://developer.mozilla.org/en-US/docs/Web/API/Element/getClientRects


packages/core/useMouseInElement/demo.vue
```diff
@@ -19,11 +19,37 @@ const stringify = reactify(
...
+    >
+      <div m="auto">
+        <span>
+          Hover <span ref="inline-target" style="background: rgb(36, 146, 82, 0.5)">
+            the highlighted text to see the</span> tracking.
+        </span>
+      </div>
+    </div>
+    <pre lang="yaml">{{ inlineText }}</pre>
+  </div>
```
siavava: > [!NOTE]
> This is near identical to the `Area` component used above,
> but I needed to inline this to be able to apply the `ref` property on the `span`.
>
> Please let me know if there is a better or preferred way to handle this situation.

packages/core/useMouseInElement/demo.vue
```diff
@@ -19,11 +19,37 @@ const stringify = reactify(
...
+      bg="gray-400/20"
+      border="rounded"
+      flex="~"
+      place="content-center"
+      select="none"
+      style="text-align: center"
+    >
+      <div m="auto">
+        <span>
+          Hover <span ref="inline-target" style="background: rgb(36, 146, 82, 0.5)">
```
siavava: background color is the green color used for highlights on `vueuse` page, just lowered the opacity to not mask the actual text.

siavava:
Added an [inline-element demo][inline-demo]

(or if link is dead, go to the latest deploy-preview and search for `useMouseInElement` or paste `/core/useMouseInElement/#usemouseinelement` after the base URL.

[inline-demo]: https://deploy-preview-5049--vueuse.netlify.app/core/useMouseInElement/#usemouseinelement

packages/core/useMouseInElement/index.ts
```diff
@@ -71,27 +71,32 @@ export function useMouseInElement(
...
+        || elX < 0 || elY < 0
+        || elX > width || elY > height
+
+      if (handleOutside || !isOutside.value) {
+        elementX.value = elX
+        elementY.value = elY
+      }
+
+      if (!isOutside.value)
+        break
```
siavava, line 99: For inline elements, if we're inside the current rect,
there's no need to check the others coming after it.

This also avoids the state of the other rects from overwriting
the fact that we're hovering on the current rect.

siavava:
...not sure who is the right person to reach out to for approval or feedback on if this change is desirable.

9romise (approved):

9romise:
Updated the demo. For preview:
image

siavava:
@9romise could you merge this if it looks good to you?
I don't have the right permissions to run two of the workflows needed to merge (rerun needed because of rebase)

9romise:
To prevent regressions, I think it would be best to add some test cases to this PR after #5101 has been merged. Once other members also approve this request, we can proceed with merging.

siavava:
@9romise thanks for adding the tests! 🎉

siavava:
> To prevent regressions, I think it would be best to add some test cases to this PR after #5101 has been merged. Once other members also approve this request, we can proceed with merging.

cc @9romise —— any chance we could get someone to review/merge this? since it's been stalled for over a month now and we already added the tests.

PR #5051, opened 2025-09-21
test(useKeyModifier): add test to cover all features

NoiseFan:

<details>
<br>

>

</details>

---

### Description

What this PR solves:
- Add 7 test cases covering all functionality of `useKeyModifier`.
- Move `dispatchKeyboardEvent` to packages/.test directory for unified management.
- Remove duplicate `dispatchKeyboardEvent` implementations in useMagicKeys.

### Additional context
This PR only adds tests without modifying the existing `useKeyModifier` implementation.
The test file follows the project's testing conventions and uses the same testing utilities.
No breaking changes or new dependencies are introduced.

### DIFF
--- a/packages/core/useKeyModifier/index.test.ts
+++ b/packages/core/useKeyModifier/index.test.ts
@@ -1,0 +1,78 @@
+import type { KeyModifier } from './index'
+import { describe, expect, it } from 'vitest'
+import { dispatchKeyboardEvent } from '../../.test'
+import { useKeyModifier } from './index'
+
+describe('useKeyModifier', () => {
+  it('should be defined', () => {
+    expect(useKeyModifier).toBeDefined()
+  })
+
+  it('all events', () => {
+    const cases: Array<{ key: KeyModifier, options: KeyboardEventInit }> = [
+      { key: 'Alt', options: { altKey: true } },
+      { key: 'AltGraph', options: { modifierAltGraph: true } },
+      { key: 'CapsLock', options: { modifierCapsLock: true } },
+      { key: 'Control', options: { ctrlKey: true } },
+      { key: 'Fn', options: { modifierFn: true } },
+      { key: 'FnLock', options: { modifierFnLock: true } },
+      { key: 'Meta', options: { metaKey: true } },
+      { key: 'NumLock', options: { modifierNumLock: true } },
+      { key: 'ScrollLock', options: { modifierScrollLock: true } },
+      { key: 'Shift', options: { shiftKey: true } },
+      { key: 'Symbol', options: { modifierSymbol: true } },
+      { key: 'SymbolLock', options: { modifierSymbolLock: true } },
+    ]
+    function assertModifierState(key: KeyModifier, options: KeyboardEventInit) {
+      const state = useKeyModifier(key)
+      expect(state.value).toBe(null)
+
+      dispatchKeyboardEvent({ key, ...options })
+      expect(state.value).toBeTruthy()
+
+      dispatchKeyboardEvent({ key, type: 'keyup' })
+      expect(state.value).toBeFalsy()
+    }
+
+    cases.forEach(({ key, options }) => {
+      assertModifierState(key, options)
+    })
+  })
+
+  describe('params', () => {
+    describe('events', () => {
+      it('should be work with custom events', () => {
+        const state = useKeyModifier('Alt', { events: ['mousedown'] })
+        expect(state.value).toBe(null)
+
+        document.dispatchEvent(new MouseEvent('mousedown', { altKey: true }))
+        expect(state.value).toBeTruthy()
+
+        document.dispatchEvent(new MouseEvent('mousedown', { altKey: false }))
+        expect(state.value).toBeFalsy()
+      })
+    })
+
+    describe('initial', () => {
+      it('should be null when initial value is default', () => {
+        const state = useKeyModifier('Alt')
+        expect(state.value).toBeNull()
+      })
+
+      it('should be true when initial value is true', () => {
+        const state = useKeyModifier('Alt', { initial: true })
+        expect(state.value).toBeTruthy()
+      })
+
+      it('should be false when initial value is false', () => {
+        const state = useKeyModifier('Alt', { initial: false })
+        expect(state.value).toBeFalsy()
+      })
+
+      it('should be null when initial value is null', () => {
+        const state = useKeyModifier('Alt', { initial: null })
+        expect(state.value).toBeFalsy()
+      })
+    })
+  })
+})
--- a/packages/.test/index.ts
+++ b/packages/.test/index.ts
@@ -4,3 +4,4 @@ export const isBelowNode18 = Number(process.version.slice(1).split('.')[0]) < 18
 
 export * from './mount'
 export * from './nextTick'
+export * from './keyboardEvent'
--- a/packages/.test/keyboardEvent.ts
+++ b/packages/.test/keyboardEvent.ts
@@ -1,0 +1,13 @@
+interface dispatchKeyboardEventOptions extends Partial<KeyboardEvent> {
+  key: string
+  target?: HTMLElement
+  eventType?: 'keydown' | 'keyup'
+}
+
+/**
+ * Dispatch keyboard event
+ */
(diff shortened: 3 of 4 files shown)


NoiseFan:
I don't think [CI / test (20.x)](https://github.com/vueuse/vueuse/actions/runs/17889131381/job/50867212372?pr=5051) error has anything to do with this modification. Can you help me retry it?

ilyaliao (approved):

OrbisK (approved):

packages/core/useKeyModifier/index.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    it.for(cases)('should track state for %s', ({ key, options }) => {
+      assertModifierState(key, options)
+    })
+  })
+
+  describe('params', () => {
+    describe('events', () => {
+      it('should be work with custom events', () => {
+        const state = useKeyModifier('Alt', { events: ['mousedown'] })
```
43081j, line 49: in this test can you use an actual custom event to prove it works with arbitrary ones?

e.g. `testevent` and `new CustomEvent('testevent')`

NoiseFan: It's resolve!

packages/core/useKeyModifier/index.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      it('should be work with non-default event', () => {
+        const state = useKeyModifier('Alt', { events: ['click'] })
+        expect(state.value).toBe(null)
+
+        document.dispatchEvent(new MouseEvent('click', { altKey: true }))
+        expect(state.value).toBeTruthy()
+
+        document.dispatchEvent(new MouseEvent('click', { altKey: false }))
+        expect(state.value).toBeFalsy()
+      })
```
43081j:
```suggestion
      it('should be work with non-default event', () => {
        const state = useKeyModifier('Alt', { events: ['foo'] })
        expect(state.value).toBe(null)

        document.dispatchEvent(new MouseEvent('foo', { altKey: true }))
        expect(state.value).toBeTruthy()

        document.dispatchEvent(new MouseEvent('foo', { altKey: false }))
        expect(state.value).toBeFalsy()
      })
```

something like this should work

just to prove that custom event names will still work

NoiseFan: Does this have to be the case? I think events has type constraints.

43081j: the default events are these:

```ts
const defaultEvents: WindowEventName[] = ['mousedown', 'mouseup', 'keydown', 'keyup']
```

so the main thing is to test what happens when you pass an event which isn't one of these.

you've done that by using `click` (which is correct). but this and the other test are now testing the same thing really

my original comment was trying to point out that we should prove truly custom event names also work. not only known ones.

the function constrains around the type of event fired, not the event name. we should be able to bind to any event name, as long as it emits a mouse or keyboard event

NoiseFan: I'm ready to rollback. If I want to solve this problem, I 'd better open a new pr.

43081j: for example that means your test:

```ts
      it('should be work with custom events', () => {
        const state = useKeyModifier('Alt', { events: ['mousedown'] })
```

this is actually just testing that you can pass in events. but `mousedown` is a default anyway so we're not testing that non-default events work

which is what i was originally trying to point out

43081j: you do not need to roll anything back.

packages/core/useKeyModifier/index.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    ]
+
+    it.for(cases)('should track state for %s', ({ key, options }) => {
+      assertModifierState(key, options)
+    })
+  })
+
+  describe('params', () => {
+    describe('events', () => {
+      it('should be work with custom events', () => {
```
43081j:
```suggestion
      it('should allow events to be specified', () => {
```

packages/core/useKeyModifier/index.test.ts
43081j: you can forget this comment.

just do my other comment and we can leave this as it is 👍

packages/core/useKeyModifier/index.test.ts
NoiseFan: wdyt?

```ts
  it('should allow event to be specified', () => {
    const state = useKeyModifier('Alt', { events: ['mousedown'] })
    expect(state.value).toBe(null)

    document.dispatchEvent(new MouseEvent('mousedown', { altKey: true }))
    expect(state.value).toBeTruthy()

    document.dispatchEvent(new MouseEvent('mousedown', { altKey: false }))
    expect(state.value).toBeFalsy()
  })

  it('should be work with custom event', () => {
    const state = useKeyModifier('Alt', { events: ['foo' as keyof WindowEventMap] })
    expect(state.value).toBe(null)

    document.dispatchEvent(new MouseEvent('foo', { altKey: true }))
    expect(state.value).toBeTruthy()

    document.dispatchEvent(new MouseEvent('foo', { altKey: false }))
    expect(state.value).toBeFalsy()
  })
```

43081j: don't worry about changing the last one

we can leave it as `click`

i think as long as the names of the tests are clear then everything is fine 👍

NoiseFan: hold on, I'm a little confused now. I'm reading it carefully.(English and Chinese have the same meaning)
不好意思，稍等一会，我有点迷茫了，我需要仔细在看一下。

NoiseFan: Like this?
```ts
  it('should allow event to be specified', () => {
    const state = useKeyModifier('Alt', { events: ['mousedown'] })
    expect(state.value).toBe(null)

    document.dispatchEvent(new MouseEvent('mousedown', { altKey: true }))
    expect(state.value).toBeTruthy()

    document.dispatchEvent(new MouseEvent('mousedown', { altKey: false }))
    expect(state.value).toBeFalsy()
  })

  it('should be work with custom event', () => {
    const state = useKeyModifier('Alt', { events: ['click'] })
    expect(state.value).toBe(null)

    document.dispatchEvent(new MouseEvent('click', { altKey: true }))
    expect(state.value).toBeTruthy()

    document.dispatchEvent(new MouseEvent('click', { altKey: false }))
    expect(state.value).toBeFalsy()
  })
```

packages/core/useKeyModifier/index.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        document.dispatchEvent(new MouseEvent('mousedown', { altKey: false }))
+        expect(state.value).toBeFalsy()
+      })
+
+      it('should be work with non-default event', () => {
+        const state = useKeyModifier('Alt', { events: ['click'] })
+        expect(state.value).toBe(null)
+
+        document.dispatchEvent(new MouseEvent('click', { altKey: true }))
```
9romise, line 63:
```suggestion
        document.dispatchEvent(new CustomEvent('foo'))
```

packages/core/useKeyModifier/index.test.ts
43081j: yes we can go with that 👍 

sorry for confusing you. i wanted to test that we can use custom events, but i think you're right - click is still a non-default so should be fine

packages/core/useKeyModifier/index.test.ts
9romise: @43081j  If I understand correctly, are you suggesting we test with an incorrect event type like this?

NoiseFan: I understand, this will not work. `getModifierState()` is the method of `KeyboardEvent` and `MouseEvent`

43081j: do exactly what is in this comment:
https://github.com/vueuse/vueuse/pull/5051#discussion_r2372649956

it will be ok

i just wanted to be sure we could use a custom event. but `click` will test the same code path since its a non-default

NoiseFan:
Thank you all!

PR #5059, opened 2025-09-23
fix(useStorage): preserve updates when key changes

### LINKED ISSUE #4767: BUG | `useStorage` | The new value is ignored when changing both the key and value synchronously
### Describe the bug

When changing the storage key and then the ref containing the value synchronously, the new value is ignored and the ref contains the value from storage.

Here's what happens when changing both the `useStorage` key and value synchronously:

1. Set the `key` and `value`.
2. The [`key` watcher](https://github.com/vueuse/vueuse/blob/v13.2.0/packages/core/useStorage/index.ts#L182) triggers [`update`](https://github.com/vueuse/vueuse/blob/v13.2.0/packages/core/useStorage/index.ts#L289).
3. The `update` function loads the value from the storage and sets the `value` (`data`).
4. The [`data` watcher](https://github.com/vueuse/vueuse/blob/v13.2.0/packages/core/useStorage/index.ts#L176) triggers [`write`](https://github.com/vueuse/vueuse/blob/v13.2.0/packages/core/useStorage/index.ts#L242).
5. The `write` function updates the storage with the value loaded from the storage in step 3 (not expected), rather than with the value explicitly set in step 1 (expected).

### Reproduction

https://playground.vueuse.org/#eNqVVcFu2zgQ/RVCFyWAK223PXndINkih+wu2kVb7F50YcSJxIQiCXLk2DD87x2SsiQ7beGcLM48vvc4HI532Y21xbqHbJmtfO2kReYBe8sU182HKkNfZVeVlp01DtmO9R7+MTVXX9E43gDbswdnOpZfEwflyto4yGd433KlzPMXeBihhCREpWujPbIn2N5piZIr9oHlMn3mf8zSH1uyAiKk6/R5lKb4JHIx0V0eMGuueiDUifUAXbA8vySySjugY1+k74de1yiNZkOQ7SrNglRxoJpUaAMLFbtD6GbiRCydAwVrrjFKHKGGIxHq0+dvDDYWaqRjBdz+yEE68N+wvdHivyD+EzcDYdSJ4TGZj+wvyGeGlsyjk7pZMA3PUegQGfRSKX0q3Y0DTtRqVs2onEBGJavhYiZ80UxqQz1myZmVyUKCPUstzHMhpLcc6/Z2DRovCMIG5RTIB7Z8kfzGCi3S18HQsDzQD8uZjRjZXw7XsCrTg6D2pwW5s4oj0IqxlZDrq/9bGC6IysTuDbYMW4gtybUY2s5vdd06o03vFR0tAIL3lJSeaYPMgvPS0xUVqzIQjwqr+x6RLuq6VrJ+otf4oh3ocaarP5ZdlWnnS4bY0rTrS/idYMe6sSF2u8i5358ko0BMp0NMgFU5KxItPW4VVaA2FgRFCJTuxnIhqGRL9t5uUqkHlzF7z+unxpleiyVTsmmxcbCNnXBvnAC3ZG/thnmjpGBj6oTygH3juJC9nynRpQZXV9kiSzPqTcdt8eiNpgkY9ashQYOPTpl6pMoO860D5IIjD8kqaxGtX5ZlLTRxCFBy7QoNWGrbladbrt++K34vfiupn2FTdI8kMLTgRE+TzIF4BXnacA51mMyvIA7wc2ilRmgcDwMlVuxM+vm2c2Q6ju1rak7wc2jdJoTPpg3wH9DGsUHtRU2F1O76QTYnLVWbzkoF7rMdCzW2Vvzv+ivG0I1jKTx2qJ9+EH/0m2T53/CY3ZpudcwhdzRnU/r26yfY0PeY7Izo1dADP0nSWDCqDx4T7E96hmR7hotu7+ILoQf3zd9uELQ/HCoYjdWI+Cqjun38xdEnu++K92MV998BRGL3Nw==

### System Info

```Shell
System:
    OS: Linux 6.8 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
    Memory: 4.10 GB / 15.34 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 22.14.0 - /run/user/1000/fnm_multishells/5676_1747894241665/bin/node
    npm: 10.9.2 - /run/user/1000/fnm_multishells/5676_1747894241665/bin/npm
    pnpm: 10.10.0 - /run/user/1000/fnm_multishells/5676_1747894241665/bin/pnpm
  Browsers:
    Chrome: 135.0.7049.114
```

### Used Package Manager

pnpm

### Validations

ZKunZhang:

- [x] Read the Contributing Guidelines (https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the Pull Request Guidelines (https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Provide a description in this PR that addresses what the PR is solving, or reference the issue that it solves
(e.g. fixes #4767).

<details>
<br>

>

</details>

---

### Description

- adjust the useStorage key watcher to flush in the post phase (sync mode unchanged), ensuring we don’t read from
storage before same-tick explicit writes land.
- add a regression test covering “change key and value within the same tick” to prove the new value now sticks, and
update the prior expectation that relied on the old behaviour.

### Additional context

- tests: pnpm vitest run packages/core/useStorage/index.test.ts

### DIFF
--- a/packages/core/useStorage/index.test.ts
+++ b/packages/core/useStorage/index.test.ts
@@ -571,7 +571,8 @@ describe('useStorage', () => {
     key.value = KEY
     data.value = 3
     await nextTick()
-    expect(storage.getItem(KEY)).toBe('2')
+    expect(data.value).toBe(3)
+    expect(storage.getItem(KEY)).toBe('3')
     expect(storage.getItem(ANOTHER_KEY)).toBe('1')
   })
 
@@ -591,6 +592,19 @@ describe('useStorage', () => {
     key.value = KEY
     await nextTick()
     expect(data.value).toBe(0)
+    expect(storage.getItem(KEY)).toBe('0')
+    expect(storage.getItem(ANOTHER_KEY)).toBe('2')
+  })
+
+  it('keeps the new value when key and value change in the same tick', async () => {
+    const key = deepRef(KEY)
+    const data = useStorage(key, 0, storage)
+
+    key.value = ANOTHER_KEY
+    data.value = 2
+    await nextTick()
+
+    expect(data.value).toBe(2)
     expect(storage.getItem(KEY)).toBe('0')
     expect(storage.getItem(ANOTHER_KEY)).toBe('2')
   })
--- a/packages/core/useStorage/index.ts
+++ b/packages/core/useStorage/index.ts
@@ -179,7 +179,8 @@ export function useStorage<T extends (string | number | boolean | object | null)
     { flush, deep, eventFilter },
   )
 
-  watch(keyComputed, () => update(), { flush })
+  const keyWatcherFlush = flush === 'sync' ? 'sync' : 'post'
+  watch(keyComputed, () => update(), { flush: keyWatcherFlush })
 
   let firstMounted = false
   const onStorageEvent = (ev: StorageEvent): void => {


ilyaliao:
Also, I'm wondering if we should never allow users to pass `'pre'`, or maybe we just need to simply change the default value to `'sync'`.

ZKunZhang:
> Also, I'm wondering if we should never allow users to pass , or maybe we just need to simply change the default value to .`'pre'``'sync'`

1. Type restriction approach: While it provides type safety, it would force users to modify existing code, making it a Breaking Change.
2. Default value modification approach: Although it seems simple, it changes the default behavior for all users, and testing has confirmed it would break existing logic.

ilyaliao:
> 1. Type restriction approach: While it provides type safety, it would force users to modify existing code, making it a Breaking Change.
> 2. Default value modification approach: Although it seems simple, it changes the default behavior for all users, and testing has confirmed it would break existing logic.

oh, I see, because `flush` is also used elsewhere

Then I just want to confirm if we can always set it to `sync` or `post`

ZKunZhang:
> > 1. Type restriction approach: While it provides type safety, it would force users to modify existing code, making it a Breaking Change.
> > 2. Default value modification approach: Although it seems simple, it changes the default behavior for all users, and testing has confirmed it would break existing logic.
> 
> oh, I see, because `flush` is also used elsewhere
> 
> Then I just want to confirm if we can always set it to `sync` or `post`

Current Logic
`const keyWatcherFlush = flush === 'sync' ? 'sync' : 'post'`

 Why we cannot always set it to 'sync':
1. Performance issues: 'sync' executes immediately without batching optimizations
2. Violates user intent: When users set flush='pre', there's a reason - forcing 'sync' goes against their intention
3. Vue reactivity best practices: Vue defaults to 'pre' for better performance and batching

Why we cannot always set it to 'post':

```
  // If we always use 'post', when user sets flush='sync':
  const { flush = 'sync' } = options
  // Data watcher: 'sync' (immediate execution)
  // Key watcher: 'post' (delayed execution)
  // This would cause execution order issues
```

Specific scenario problems:

When key and data change simultaneously:
  - User sets sync: Expects immediate synchronous execution
  - If key watcher uses post: Will execute in next tick, breaking sync semantics

Edge case example:

```
 // User expects synchronous behavior
  const storage = useStorage('key', 'default', storage, { flush: 'sync' })
  storage.value = 'newValue' // Data watcher runs immediately ('sync')
  key.value = 'newKey'       // If key watcher is 'post', runs later - wrong!
```


Conclusion:
The conditional logic is necessary because it ensures:
- Respects user's flush configuration intent
- Maintains correct execution order
- Prevents key changes from overwriting data changes

Did my answer solve your problem?

9romise:
Based on your changes, passing `pre` to the `flush` option would become meaningless. This could break existing usage for users who use the default value or explicitly set `pre`, and likely cause unnecessary confusion.
Also, changing the default value would be a breaking change.

I'm leaning towards refining the type definitions for the `flush` option.

ZKunZhang:
> Based on your changes, passing `pre` to the `flush` option would become meaningless. This could break existing usage for users who use the default value or explicitly set `pre`, and likely cause unnecessary confusion. Also, changing the default value would be a breaking change.
> 
> I'm leaning towards refining the type definitions for the `flush` option.

Thanks for the feedback! You raise valid concerns about the potential impact on existing usage.

  Looking at this more carefully, I think there are a few ways to approach this:

  1. **Refine type definitions approach**: As you suggested, we could create more specific flush option types that better communicate the actual behavior and constraints.

  2. **Alternative solution**: Instead of changing the flush timing, we could explore using a different approach like:
     - Using a single watcher that handles both key and data changes
     - Or using a micro-task queue to ensure proper execution order without changing the flush API

  3. **If we keep the current approach**: We could consider this as a bug fix rather than a breaking change, since the current behavior (data updates being overwritten by key changes in the same tick) is
  arguably incorrect. However, I understand the concern about user expectations.

  The core issue we're solving is that when both key and data change in the same tick, we want to preserve the data update rather than having it overwritten by the key change.

  Would you prefer to explore the type definition approach first, or would you like to see an alternative implementation that doesn't modify the flush behavior?

PR #5066, opened 2025-09-24
feat(useIntersectionObserver): add more IntersectionObserver options

wbolster:
### Description

This adds options for:

- delay
- scrollMargin
- trackVisibility

See https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

### Additional context

Note that upstream typescript (5.9) does not include these options though multiple browsers have supported them for years; see https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#browser_compatibility

### DIFF
--- a/packages/core/useIntersectionObserver/index.ts
+++ b/packages/core/useIntersectionObserver/index.ts
@@ -8,6 +8,16 @@ import { defaultWindow } from '../_configurable'
 import { unrefElement } from '../unrefElement'
 import { useSupported } from '../useSupported'
 
+declare global {
+  // Augment the type b/c typescript 5.9 lib.dom does not include
+  // definitions for the these options yet.
+  interface IntersectionObserverInit {
+    delay?: number
+    scrollMargin?: string
+    trackVisibility?: boolean
+  }
+}
+
 export interface UseIntersectionObserverOptions extends ConfigurableWindow {
   /**
    * Start the IntersectionObserver immediately on creation
@@ -15,6 +25,11 @@ export interface UseIntersectionObserverOptions extends ConfigurableWindow {
    * @default true
    */
   immediate?: boolean
+
+  /**
+   * A number specifying the minimum permitted delay between notifications from the observer, in milliseconds.
+   */
+  delay?: number
 
   /**
    * The Element or Document whose bounds are used as the bounding box when testing for intersection.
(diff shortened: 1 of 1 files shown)


packages/core/useIntersectionObserver/index.ts
```diff
@@ -8,6 +8,16 @@ import { defaultWindow } from '../_configurable'
...
+declare global {
+  // Augment the type b/c typescript 5.9 lib.dom does not include
+  // definitions for the these options yet.
+  interface IntersectionObserverInit {
+    delay?: number
+    scrollMargin?: string
+    trackVisibility?: boolean
+  }
+}
+
```
OrbisK, line 20: I think this should be solved upstream

wbolster: yes, eventually. upstream issue for that: https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/2156

note that `useScreenOrientation`, for example, also adds (back) some type definitions that are not available from typescript; see https://github.com/vueuse/vueuse/blob/main/packages/core/useScreenOrientation/index.ts#L7

wbolster: MR for adding `scrollMargin` upstream: https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2157

re `delay` and `trackVisibility`: this would need local definitions for the foreseeable future. i think the approach in this MR should be fine for that, no?

antfu: Yeah I think VueUse as a library shouldn't augement global types

wbolster: what's the vueuse policy for supporting browser apis that are not (yet) available in all browsers?

e. g. two of the flags in this pull request. without support inside vueuse itself, these cannot be used. the change itself won't break other browsers, but typescript checks for vueuse itself won't pass.

what's the recommended approach here? new custom types and some ‘as any’ here and there?

antfu: It's not about policy; it's just that arguing global types would introduce unexpected side effects and pollution, and might be conflicted when the built-in types are implemented.

I personally would prefer to wait until the types landed in TypeScript first, at least (as a sign of API maturity to me). Before that, I'd suggest you vendor the function to have it in your own codebase for now.

PR #5067, opened 2025-09-24
refactor(useStorage): simplify guessSerializerType for clarity and readability

snowbitx:
…adability

<details>
<br>

>

</details>

---

### DIFF
--- a/packages/core/useStorage/guess.ts
+++ b/packages/core/useStorage/guess.ts
@@ -1,4 +1,8 @@
-export function guessSerializerType<T extends (string | number | boolean | object | null)>(rawInit: T) {
+export type SerializerType = 'boolean' | 'object' | 'number' | 'any' | 'string' | 'map' | 'set' | 'date'
+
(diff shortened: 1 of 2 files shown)


antfu (commented):
Why is this refactor necessary? I think extracting `SerializerType` is a good call, but the refactor of `guessSerializerType` doesn't feel necessary to me. Like creating an internal map inside would reduce the performance a bit but doesn't bring much value honestly.

snowbitx:
> Why is this refactor necessary? I think extracting `SerializerType` is a good call, but the refactor of `guessSerializerType` doesn't feel necessary to me. Like creating an internal map inside would reduce the performance a bit but doesn't bring much value honestly.

Ah, the original intent of this PR was actually from learning VueUse — I found the nested ternary here was getting really hard to read 😅. I refactored it for clarity, and only extracted the SerializerType afterward because type errors popped up during the refactor. If this doesn't add real value, feel free to ignore it! 🙈

PR #5043, opened 2025-09-16
feat: add new function `useWebWorkerIntervalFn`

bprag:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

fix #4671

PR #5070, opened 2025-09-25
feat: add custom scheduler option to `useIntervalFn-based` composables and unify options

9romise:

<details>
<br>

>

</details>

---

### Description

This PR unifies the options of composables built on top of `useIntervalFn` and introduces a customizable scheduler option:

- [`useInterval`](https://vueuse.org/useInterval)
- [`useCountDown`](https://vueuse.org/useCountDown)
- [`useMemory`](https://vueuse.org/useMemory)
- [`useVibrate`](https://vueuse.org/useVibrate)
- [`useWebSocket`](https://vueuse.org/useWebSocket) - a bit special

There are also a few functions that currently allow choosing between useRaf and useIntervalFn through an interval option. I didn’t make changes to those in this PR, and I’d like to hear everyone’s thoughts on this.

- [`useElementByPoint`](https://vueuse.org/useElementByPoint)
- [`useNow`](https://vueuse.org/useNow)
- [`useTimestamp`](https://vueuse.org/useTimestamp)

We can update the [configurations documentation](https://vueuse.org/guide/config.html#configurations) once we’ve discussed and agreed on a suitable approach.

### Additional context

Related: #5043

### DIFF
--- a/packages/core/useCountdown/index.ts
+++ b/packages/core/useCountdown/index.ts
@@ -1,13 +1,9 @@
-import type { Pausable } from '@vueuse/shared'
+import type { ConfigurableScheduler, Pausable } from '@vueuse/shared'
 import type { MaybeRefOrGetter, ShallowRef } from 'vue'
 import { useIntervalFn } from '@vueuse/shared'
 import { shallowRef, toValue } from 'vue'
 
-export interface UseCountdownOptions {
-  /**
-   *  Interval for the countdown in milliseconds. Default is 1000ms.
-   */
-  interval?: MaybeRefOrGetter<number>
+export interface UseCountdownOptions extends ConfigurableScheduler {
   /**
    * Callback function called when the countdown reaches 0.
    */
@@ -16,12 +12,6 @@ export interface UseCountdownOptions {
    * Callback function called on each tick of the countdown.
    */
   onTick?: () => void
-  /**
-   * Start the countdown immediately
-   *
-   * @default false
-   */
-  immediate?: boolean
 }
 
 export interface UseCountdownReturn extends Pausable {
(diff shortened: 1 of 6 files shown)


antfu:
I would be happy to see `useRafFn` to be a valid scheduler

packages/shared/utils/scheduler.ts
```diff
@@ -0,0 +1,67 @@
...
+   * @default useIntervalFn
+   */
+  scheduler?: Scheduler
+}
+
+export type ConfigurableRafScheduler = ConfigurableScheduler & {
+  /**
+   * Custom scheduler to use for interval execution.
+   *
+   * @default useRafFn
```
9romise, line 45: The only difference is the default value in the JSDoc; I'm not sure if there's a better way to make it more noticeable to users.

packages/shared/utils/scheduler.ts
```diff
@@ -0,0 +1,67 @@
...
+   */
+  scheduler?: Scheduler
+}
+
+export type ConfigurableSchedulerLazy = ConfigurableScheduler & {
+  /**
+   * Start the interval immediately
+   *
+   * @default false
```
9romise, line 54: Same as above.

packages/core/useCountdown/index.ts
```diff
@@ -51,18 +41,34 @@ export interface UseCountdownReturn extends Pausable {
...
-    remaining.value = value < 0 ? 0 : value
-    options?.onTick?.()
-    if (remaining.value <= 0) {
-      intervalController.pause()
-      options?.onComplete?.()
-    }
-  }, options?.interval ?? 1000, { immediate: options?.immediate ?? false })
+  const {
+    scheduler = useIntervalFn,
+    interval = 1000,
```
9romise, line 49: I’d like to get some opinions on whether it’s necessary to explicitly set default values in the top-level functions, rather than passing `undefined` and letting the underlying scheduler handle it.

Just want to note that different functions handle this differently, and all properties in `SchedulerOptions` have this situation.

9romise:
> I would be happy to see `useRafFn` to be a valid scheduler

Absolutely agree! I did run into some issues earlier, but after some more thought, I realized I was overcomplicating it. I've now found a clean solution and have pushed the updates.

packages/shared/useIntervalFn/index.ts
```diff
@@ -1,40 +1,35 @@
...
  * Wrapper for `setInterval` with controls
  *
  * @see https://vueuse.org/useIntervalFn
- * @param cb
- * @param interval
- * @param options
  */
-export function useIntervalFn(cb: Fn, interval: MaybeRefOrGetter<number> = 1000, options: UseIntervalFnOptions = {}): UseIntervalFnReturn {
+export function useIntervalFn(cb: Fn, interval: MaybeRefOrGetter<number>, options?: UseIntervalFnOptions): UseIntervalFnReturn
+export function useIntervalFn(cb: Fn, options?: SchedulerOptions): UseIntervalFnReturn
```
9romise, line 18: Overloaded the method and implemented the `Scheduler` interface.

9romise:
Since this PR touches multiple files, I'd recommend adding tests to prevent regressions before we merge.

9romise:
I'd like to gather some input on:

1. Whether we should unify the default values for each method
2. Whether we should explicitly set default values for each method at the top level
3. Whether there’s a better way to handle different default values across methods without breaking the user experience

PR #5071, opened 2025-09-29
feat: moving away from array return types

ArthurDarkstone:

<details>
<br>

>

</details>

---


packages/shared/createInjectionState/index.ts
```diff
@@ -18,6 +19,20 @@ export type CreateInjectionStateReturn<Arguments extends Array<any>, Return> = R
   useInjectedState: () => Return | undefined,
 ]>
 
+export interface CreateInjectionStateObject<Arguments extends Array<any>, Return> extends Record<string | number, unknown> {
```
43081j, line 22: do we need to extend `Record` here?

the shape is defined by us so shouldn't need to be extensible afaict

packages/shared/createInjectionState/index.ts
```diff
@@ -48,5 +68,17 @@ export function createInjectionState<Arguments extends Array<any>, Return>(
...
+  const obj: CreateInjectionStateObject<Arguments, Return> = {
+    useProvidingState,
+    useInjectedState,
+    provide: useProvidingState,
+    inject: useInjectedState,
+  }
+
+  const destructurable = makeDestructurable<CreateInjectionStateObject<Arguments, Return>, CreateInjectionStateTuple<Arguments, Return>>(
+    obj,
+    [useProvidingState, useInjectedState] as CreateInjectionStateTuple<Arguments, Return>,
```
43081j, line 80: curious - does it not infer the right type here without the cast? 🤔

packages/shared/useToggle/index.test.ts
```diff
@@ -94,4 +88,36 @@ describe('useToggle', () => {
...
+
+  it('should return a function when passing a ref', () => {
+    const r = shallowRef(false)
+    const toggle = useToggle(r)
+    toggle()
+    expect(r.value).toBe(true)
+  })
+
+  it('should respect custom truthy/falsy values', () => {
+    const { state, toggle } = useToggle<'ON' | 'OFF', 'OFF' | 'ON', string>('OFF' as any, { truthyValue: 'ON', falsyValue: 'OFF' }) as any
```
43081j: similar question as before - does it not work without the `as any`? that seems surprising

packages/shared/createInjectionState/index.ts
ArthurDarkstone: image

It should because this generic type is Record<string, unknown>

PR #5088, opened 2025-10-06
fix(useElementSize): width & height should take watched element sizes

palagadec:

<details>
<br>

>

</details>

---

### Description

Fix a bug where width and height were set to initial size (0 by default) instead of watched element sizes.

This caused issues when trying to access these values on mount as the watcher watches for a template ref.

### DIFF
--- a/packages/core/useElementSize/index.ts
+++ b/packages/core/useElementSize/index.ts
@@ -70,8 +70,14 @@ export function useElementSize(
   const stop2 = watch(
     () => unrefElement(target),
     (ele) => {
-      width.value = ele ? initialSize.width : 0
-      height.value = ele ? initialSize.height : 0
+      width.value =
+                ele && "offsetWidth" in ele
+                    ? ele.offsetWidth
+                    : initialSize.width;
+      height.value =
+          ele && "offsetHeight" in ele
+              ? ele.offsetHeight
+              : initialSize.height;
     },
   )
 
--- a/packages/core/useElementSize/index.ts
+++ b/packages/core/useElementSize/index.ts
@@ -70,14 +70,14 @@ export function useElementSize(
   const stop2 = watch(
     () => unrefElement(target),
     (ele) => {
-      width.value =
-                ele && "offsetWidth" in ele
-                    ? ele.offsetWidth
-                    : initialSize.width;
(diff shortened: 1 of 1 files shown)


packages/core/useElementSize/index.ts
```diff
@@ -70,8 +70,14 @@ export function useElementSize(
...
-      width.value = ele ? initialSize.width : 0
-      height.value = ele ? initialSize.height : 0
+      width.value =
+                ele && "offsetWidth" in ele
+                    ? ele.offsetWidth
+                    : initialSize.width;
+      height.value =
+          ele && "offsetHeight" in ele
+              ? ele.offsetHeight
+              : initialSize.height;
```
43081j: i wonder if this was `0` on purpose

when you mount an element, it makes sense it is `initialSize`.

however, this code path happens when changing target. that means the only way we hit this fallback is if we change to a nullish target or a broken element. in those cases, we probably already observed some dimensions before we got here, so it doesn't seem to make sense going _back_ to initial size

its just a weird state, so id probably leave it at 0

palagadec: No, this code path runs on initial render because the watched target changes from `null` (or `undefined`) in setup to the template ref on mount.

Here is a log of the sequence where you can see the width is set to `0` (initial size) after tryOnMounted set it previously to `300` ([useElementSize tryOnMounted]) because the template ref got set (`null` to `<div>`) which triggered the "stop2" watcher ([useElementSize stop2 watch])

image

And I'm not sure why you'd want to have different value than the actual size of the element you're using this composable on.

43081j: bleh, that's unfortunate

the initial target should be the initial size. later ones (i.e. if you change target) probably shouldn't be

> And I'm not sure why you'd want to have different value than the actual size of the element you're using this composable on.

neither of us are suggesting this. when changing target (_not_ the initial target, literally when changing target), 0 isn't the actual size, and initialSize isn't the actual size.

if i change target later on and it somehow doesn't have an `offsetHeight`, it must be nullish or a really funky element (one that somehow doesn't have a built in property). that's the only way we'd hit the condition _other than the initial render/target_ as you pointed out

PR #5089, opened 2025-10-07
feat(useWebSocket): `autoConnect.delay` support function

YuchenWell:

### Description

This PR adds support for an exponential backoff–like strategy in useWebSocket’s reconnection

### DIFF
--- a/packages/core/useWebSocket/index.browser.test.ts
+++ b/packages/core/useWebSocket/index.browser.test.ts
@@ -89,5 +89,84 @@ describe('useWebSocket', () => {
 
       expect(status.value).toBe('CONNECTING')
     })
+
+    it('should support delay as a function', () => {
+      const delayFn = vi.fn((retries: number) => retries * 1000)
+      const { ws, status } = useWebSocket('ws://localhost', {
+        autoReconnect: {
+          retries: 3,
+          delay: delayFn,
+        },
+      })
+
+      ws.value?.onopen?.(new Event('open'))
+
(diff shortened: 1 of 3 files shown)


OrbisK (dismissed):

antfu (dismissed):

9romise (approved):

packages/core/useWebSocket/index.browser.test.ts
```diff
@@ -89,5 +89,84 @@ describe('useWebSocket', () => {
...
+          delay: delayFn,
+        },
+      })
+
+      ws.value?.onopen?.(new Event('open'))
+
+      // First reconnection: delay should be 1000ms (retries = 1)
+      ws.value?.onclose?.(new CloseEvent('close'))
+      expect(status.value).toBe('CLOSED')
+      expect(delayFn).not.toHaveBeenCalled()
```
9romise: CI fails at this line.

9romise (changes requested):

packages/core/useWebSocket/index.browser.test.ts
9romise, line 1: We don't need to test how to use function-type `delay` in `VueUse` — `VueUse` should only test whether the function-type `delay` option executes correctly, not its usage examples.

9romise (changes requested):

packages/core/useWebSocket/index.browser.test.ts
YuchenWell: Thank you for pointing this out!
I have fixed it it in the latest commit.

PR #5092, opened 2025-10-11
feat(useTextSelection): Set initial value for use text selection

kevinluo201:

<details>
<br>

>

</details>

---

### Description

The initial value of `selection` is `null`, which means the text selection must be after this composable is run, then the `selection` can be updated via the `selectionchange` event. It is not practical in a real scenario because it can already have text selection in the browser.

1. The logic for text selection cannot be encapsulated in a single component. To detect the text selection, we have to run this composable in a parent component and then pass the values exported from this composable to the child component, which truly uses the selection. Eventually, we will put this composable in the root component to be injected in the descendants.
2. Sometimes we don't render the Vue app as an SPA or in Nuxt. We just mount Vue apps only for a small part on an HTML page. The text selection can happen before the Vue app is even mounted, so there's no way to detect the initial selection

There will be more and more usage of this composable because people want to use LLM AI's power to do something for the selected text on the page.

This PR also changes the `deepRef` to `shallowRef` because we don't need to deep reactive of the Selection object, we always update it whenever `selectionchange` happens, which can be triggered by any Range changes in it.

### DIFF
--- a/packages/core/useTextSelection/index.browser.test.ts
+++ b/packages/core/useTextSelection/index.browser.test.ts
@@ -1,4 +1,3 @@
-import { page, userEvent } from '@vitest/browser/context'
 import { afterEach, beforeEach, describe, expect, it } from 'vitest'
 import { useTextSelection } from './index'
 
@@ -17,12 +16,15 @@ describe('useTextSelection', () => {
     window.getSelection()?.removeAllRanges()
   })
 
-  const selectHelloWorldNode = async () => {
-    // triple click to select the whole text node
-    await page.getByText('Hello World').tripleClick()
+  const selectHelloWorldNode = () => {
+    window.getSelection()?.selectAllChildren(node)
+    const event = new Event('selectionchange')
+    document.dispatchEvent(event)
   }
-  const clickOutside = async () => {
-    await userEvent.click(document.body)
+  const removeSelection = () => {
+    window.getSelection()?.removeAllRanges()
+    const event = new Event('selectionchange')
+    document.dispatchEvent(event)
   }
 
   it('selection should always return the singleton Selection object no matter how the selection changes', async () => {
@@ -32,10 +34,10 @@ describe('useTextSelection', () => {
     const result = useTextSelection()
     expect(result.selection.value).toBe(windowSelection)
 
-    await selectHelloWorldNode()
+    selectHelloWorldNode()
     expect(result.selection.value).toBe(windowSelection)
 
-    await clickOutside()
+    removeSelection()
     expect(result.selection.value).toBe(windowSelection)
   })
 
@@ -49,29 +51,29 @@ describe('useTextSelection', () => {
   })
 
   it('should initialize with an existing range', async () => {
-    await selectHelloWorldNode()
+    selectHelloWorldNode()
 
     const result = useTextSelection()
 
     expect(result.text.value).toBe('Hello World')
     expect(result.ranges.value.length).toBe(1)
     expect(result.rects.value.length).toBe(1)
-    expect(result.selection.value!.anchorNode).toBe(node.firstChild)
-    expect(result.selection.value!.focusNode).toBe(node.firstChild)
+    expect(result.selection.value!.anchorNode).toBe(node)
+    expect(result.selection.value!.focusNode).toBe(node)
   })
 
   it('should update on selectionchange', async () => {
     const result = useTextSelection()
 
-    await selectHelloWorldNode()
+    selectHelloWorldNode()
 
     expect(result.text.value).toBe('Hello World')
     expect(result.ranges.value.length).toBe(1)
     expect(result.rects.value.length).toBe(1)
-    expect(result.selection.value!.anchorNode).toBe(node.firstChild)
-    expect(result.selection.value!.focusNode).toBe(node.firstChild)
+    expect(result.selection.value!.anchorNode).toBe(node)
+    expect(result.selection.value!.focusNode).toBe(node)
 
-    await clickOutside()
+    removeSelection()
 
     expect(result.text.value).toBe('')
     expect(result.ranges.value.length).toBe(0)


43081j (approved):

43081j (approved):
seems good to me

once #5100 lands, can you add a test for an initial selection? 

ill make sure that PR is merged before this one

kevinluo201:
> seems good to me
> 
> once #5100 lands, can you add a test for an initial selection?
> 
> ill make sure that PR is merged before this one

Thank you! Yes I will try to add test for that. Does it use jsdom in vitest?

43081j:
it runs in a real browser (via playwright). so you should be able to just setup a selection before calling `useTextSelection` and observe that the resulting value is non-null 👍 

basically the same test as whats in my branch but in a different order (setup the range first)

edit: tests are in main now. if you catch up you should see them

antfu (approved):

kevinluo201:
I added tests; however, I don't know why the `selectionchange` needs to be dispatched manually. It should be triggered automatically in browsers. I am familiar with the playwright. I can dig deeper on that. Nonetheless, you can merge this PR if you think it is good enough

packages/core/useTextSelection/index.browser.test.ts
```diff
@@ -16,27 +16,51 @@ describe('useTextSelection', () => {
...
+    document.dispatchEvent(new Event('selectionchange'))
+  }
+
+  it('should initialize with window.getSelection(), which is always present in browsers', async () => {
     const result = useTextSelection()
     expect(result.text.value).toBe('')
     expect(result.rects.value).toEqual([])
     expect(result.ranges.value).toEqual([])
-    expect(result.selection.value).toBe(null)
+    expect(result.selection.value).toBe(window.getSelection())
```
43081j: I think this needs to be explicit since getSelection can return null. We should probably check that it is the selection we expected

43081j:
a browser emits the event when a user selects text. you can't do that through automation, so you have to emit the event yourself

just the same as a click. clicking on an element through playwright etc, will just send the correct events (mousedown, mouseup, click). it isn't actually simulating an OS level mouse click

the right thing to do is emit the event

packages/core/useTextSelection/index.browser.test.ts
43081j: i think this test was probably right before - since we want to check that if there is no initial selection, the value is null

so we should assert against `null`, not `window.getSelection()` since that could return non-null values too.

at the start of the test we should probably just ensure nothing is selected, then run this code, and assert the value is `null`

packages/core/useTextSelection/index.browser.test.ts
```diff
@@ -16,27 +16,51 @@ describe('useTextSelection', () => {
...
-    range.selectNodeContents(node)
-    const selection = window.getSelection()!
-    selection.addRange(range)
 
-    const event = new Event('selectionchange')
-    document.dispatchEvent(event)
+    expect(result.text.value).toBe('Hello World')
+    expect(result.ranges.value.length).toBe(1)
+    expect(result.rects.value.length).toBe(1)
+    expect(result.selection.value).toBe(window.getSelection())
```
43081j: this should also assert against an exact selection. otherwise we could've broken the test one day and `getSelection` returns `null` but the test still passes

we should assert the value is an exact thing rather than the result of a function call

43081j:
the tests seem good but all of them assert against `window.getSelection()` which doesn't seem right

if we expect a test case to result in `getSelection` being non-null, we need to assert that. blindly asserting that it is equal to `getSelection()` means it would still pass if it was non-null or vice versa (even though we expected it not to be)

you should at the very least check that `getSelection()` is non-null or null for each case because of this
or assert that the value is an expected object shape rather than whatever `getSelection` returned

packages/core/useTextSelection/index.browser.test.ts
kevinluo201: In the browser, it always returns a Selection object, it is just an empty selection. 
CleanShot 2025-10-16 at 09 59 20@2x

It also says a selection objet must be returned if there's browsing context.
https://w3c.github.io/selection-api/#extensions-to-document-interface

https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection mentions that if a <iframe> doesn't attach to a document can make it return `null`. I don't know how to reproduce that in Playwright, but I can try to figure it out.
CleanShot 2025-10-16 at 10 10 03


Whatever, if I put a test in `index.server.test.ts` then it will be `null` because it has no browser. I can do that later today if you agree

43081j: no we don't need to do that. these belong in the browser

im just pointing out that testing against a dynamic call (`getSelection()`) isn't right

from what you said, we should be asserting that the value is an empty selection

and in the opposite tests, test that the selection is of something (not empty)

otherwise its possible tests pass when `getSelection` isn't what we expect, since we're just checking if they are equal blindly

kevinluo201: sorry, I misunderstood you. I added some assertions, like
```ts
expect(result.selection.value!.anchorNode).toBe(null)
expect(result.selection.value!.focusNode).toBe(null)
```
because an empty Selection doesn't have either `anchorNode` or `focusNode`

For a non-empty Selection, I also added
```ts
expect(result.selection.value!.anchorNode).toBe(node.firstChild)
expect(result.selection.value!.focusNode).toBe(node.firstChild)
```
Although it feels redundant (because the result.selection should be updated if result.text, result.ranges and result.rects have been updated and asserted), it is just to make the assertions more consistent

packages/core/useTextSelection/index.browser.test.ts
```diff
@@ -16,27 +17,66 @@ describe('useTextSelection', () => {
     window.getSelection()?.removeAllRanges()
   })
 
-  it('should initialise', async () => {
+  const selectHelloWorldNode = async () => {
+    // triple click to select the whole text node
+    await page.getByText('Hello World').tripleClick()
```
43081j: FYI this will just dispatch events inside playwright the same as we were doing here. like i mentioned before, there isn't a way to _actually_ click or cause a selection like a user would. its event simulation all the way down

if there's ever more text on the same line, triple click will highlight the line and not only this text too iiuc

happy to leave this but its not really doing anything more than what we had originally, and can be less accurate thanks to relying on the triple click instead of an exact selection

e.g. you could avoid the complexity and just select the exact node:

```ts
window.getSelection().selectAllChildren(node)
```

43081j (dismissed):
the assertions make much more sense now 👍 

thanks for sticking with it

packages/core/useTextSelection/index.browser.test.ts
kevinluo201: All I want to do is to simulate a real user interaction in the browser to trigger "selectionchange" event automatically. 

Sorry for the mess and for bothering you more time 😫 May I clarify a few things that still confuse me?

I noticed that when I do anything related to text selection directly in Chrome’s console, the "selectionchange" event fires automatically. For example:

I am still confused about the code in the browser tests. First, when I do anything to the selection in Chrome's console, the `selectionchange` event will be fired, for example:
```ts
document.addEventListener("selectionchange", () => {
  console.log('selectionchange!')
});
// assume we have text selection at this point
window.getSelection().removeAllRanges()
// selectionchange!
```
So I thought the event should also be triggered automatically in browser tests, for example, if I wait some time like 1 second after `window.getSelection()!.removeAllRanges()`:
```js
 it('should update on selectionchange', async () => {
    const result = useTextSelection()

    selectHelloWorldNode()

    expect(result.text.value).toBe('Hello World')
    expect(result.ranges.value.length).toBe(1)
    expect(result.rects.value.length).toBe(1)

    window.getSelection()!.removeAllRanges()
    await new Promise(resolve => setTimeout(resolve, 1000)) // <= wait 1 second after changing the selection

    expect(result.text.value).toBe('')
    expect(result.ranges.value.length).toBe(0)
    expect(result.rects.value.length).toBe(0)
    expect(result.selection.value!.anchorNode).toBe(null)
    expect(result.selection.value!.focusNode).toBe(null)
  })
```
then the test will pass. It doesn't pass if it doesn;t wait that 1 second. So I thought the event would be triggered, but it's just too quick to assert


> a browser emits the event when a user selects text. you can't do that through automation, so you have to emit the event yourself
just the same as a click. clicking on an element through playwright etc, will just send the correct events (mousedown, mouseup, click). it isn't actually simulating an OS level mouse click
the right thing to do is emit the event

However, according to your previous comment, do you mean the whole "selectionchange" event is also simulated in Playwright? If it is true, then simulating a real user interaction doesn't make sense anymore...🥲

If it's not to use a browser to do an e2e test, I would definitely choose to emit the event manually. I am just very confused about why I can't expect the event to be triggered naturally in the browser tests

OrbisK (dismissed):

packages/core/useTextSelection/index.browser.test.ts
43081j: pretty much all of the playwright interactions are emitting events under the hood. a click will emit a mousedown/mouseup/click, moving the mouse will emit a mousemove, etc. puppeteer, playwright, and every other browser automation tool will do it this way.

the only way you'd manage to simulate a "real" click would be to get the OS to click in the browser (so chrome receives an OS level click, and itself fires events, etc). but nothing does that since there's probably no API for it from each OS

so in this case, you can do one of a few things:

- simulate a mouse interaction through events that results in selection (**what you're doing now**)
- select an exact element through the selection API (what im suggesting)
- emit the selection change event manually and don't simulate any interactions

my previous comment is trying to explain that _interactions_ are simulated through events (the above). using the selection API isn't an interaction, and will indeed emit events, as you've seen

hopefully this makes sense

kevinluo201: I updated the code to use your suggestion👍 Sorry for dismissing the approval again

43081j (approved):

PR #5094, opened 2025-10-14
feat(usePermission): support more permissions

luoxiangfan:
### Description
usePermissions was lacking some APIs, this PR should fix that.

---

* [ ] Bug fix
* [x] New Feature
* [ ] Documentation update
* [ ] Other
* [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
* [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
* [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
* [x] Provide a description in this PR that addresses what the PR is solving, or reference the issue that it solves (e.g. fixes #123).
* [x] Ideally, include relevant tests that fail without this PR but pass with it.


packages/core/usePermission/index.ts
OrbisK, line 9: is this polyfill still needed? Is there no official type?

43081j: it seems to be that typescript only defines some of these (in `PermissionDescriptor`).

so this isn't really a polyfill type. its extending the built in type with extra unknowns

not sure why typescript doesn't have these. we forcefully cast it as the official type to make the call work

PR #5097, opened 2025-10-15
feat(useStorageAsync): allow disabling serializer

### LINKED ISSUE #5079: `useStorage` & `useStorageAsync` | Allow storing JS values in storage as-is without using a serializer

I am developing a browser extension, and the commonly used storage method is `chrome.storage.local`. This method can store ordinary objects, not just strings, so it does not require a serializer at all.

Please note that chrome.storage.local can be accessed in the background. This is not a web page in the usual sense and cannot use vueuse. Therefore, if useStorage uses serialization, then I need to handle the serialization logic of the vueuse code myself when accessing chrome.storage.local in the background, which is unnecessary.

I now use the following solution:
```ts
import type { SerializerAsync, StorageLikeAsync } from "@vueuse/core";
import { useStorageAsync } from "@vueuse/core";
import { toRaw } from "vue";

/**
 * The purpose is just to use `as` to resolve type errors.
 */
export function useBrowserStorage<T>(key: StorageItemKey, initialValue: T) {
  return useStorageAsync<T>(
    key,
    initialValue,
    {
      async setItem(key: StorageItemKey, value: T) {
        // chrome.storage stores the proxy array and converts it to a normal object. This breaks everything.
        // We need to store the original array in the proxy.
        await storage.setItem(key, toRaw(value));
      },
      async getItem(key: StorageItemKey) {
        return await storage.getItem(key);
      },
    } as StorageLikeAsync,
    {
      mergeDefaults: true,
      serializer: {
        read: v => v as unknown as T,
        write: v => v as unknown as string,
      } as SerializerAsync<T>,
    },
  );
}
```

### Suggested solution

I hope I don't need to write serializer-related code. Allowing it to be set to false is a good idea.

### Alternative

_No response_

### Additional context

IndexedDB can also store objects as is without serialization, which is also a common use case.

### Validations

lanxiong-015:

<details>
<br>

>

</details>

---

### Description

close #5079

### DIFF
--- a/packages/core/useStorage/index.ts
+++ b/packages/core/useStorage/index.ts
@@ -120,6 +120,14 @@ export interface UseStorageOptions<T> extends ConfigurableEventFilter, Configura
    * @default false
    */
   initOnMounted?: boolean
+
+  /**
+   * Skip serialization for storage backends that support native object storage
+   * (e.g., chrome.storage.local, IndexedDB)
+   *
+   * @default false
+   */
+  noSerialization?: boolean
 }
 
 export function useStorage(key: MaybeRefOrGetter<string>, defaults: MaybeRefOrGetter<string>, storage?: StorageLike, options?: UseStorageOptions<string>): RemovableRef<string>
@@ -152,6 +160,7 @@ export function useStorage<T extends (string | number | boolean | object | null)
       console.error(e)
     },
     initOnMounted,
+    noSerialization = false,
   } = options
 
   const data = (shallow ? shallowRef : deepRef)(typeof defaults === 'function' ? defaults() : defaults) as RemovableRef<T>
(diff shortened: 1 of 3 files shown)


43081j:
ill wait for other maintainer opinions, but i'm not sure we should do this.

i think we should document that a user can set the serializer to an identity function as a way to apply "no serialization".

we probably shouldn't introduce a standalone option for it, in my opinion.

packages/core/useStorage/index.ts
antfu, line 101: I think we should support `serializer: false` instead of adding a new option. Otherwise I am ok with this direction.

(and this is not blocking as you can already pass your custom serializer)

antfu (changes requested):
As comment above

PR #5100, opened 2025-10-15
test(useTextSelection): add initial tests

43081j:

---

### Description

Adds some basic tests for `useTextSelection`.

### DIFF
--- a/packages/core/useTextSelection/index.browser.test.ts
+++ b/packages/core/useTextSelection/index.browser.test.ts
@@ -1,7 +1,21 @@
-import { describe, expect, it } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
 import { useTextSelection } from './index'
 
 describe('useTextSelection', () => {
+  let node: HTMLElement
+
+  beforeEach(() => {
+    node = document.createElement('div')
+    node.id = 'test'
+    node.textContent = 'Hello World'
+    document.body.appendChild(node)
+  })
+
+  afterEach(() => {
+    node.remove()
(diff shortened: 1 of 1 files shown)


packages/core/useTextSelection/index.browser.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    const div = document.createElement('div')
+    div.id = 'test'
+    div.textContent = 'Hello World'
+    document.body.appendChild(div)
+
+    try {
+      const range = document.createRange()
+      range.selectNodeContents(div)
+      const selection = window.getSelection()!
+      selection.removeAllRanges()
```
9romise: I think it would be cleaner to hoist the `selection` variable to the top scope and reset it in `beforeEach` for better test isolation.

```ts
const selection = window.getSelection()!
beforeEach(() => {
	selection.removeAllRanges()
	document.body.innerHtml = '' // extra cleanup
})
```

43081j: for the element, that makes sense. for the selection/range, not so much as we need to create it at different points in time

ill update

9romise:
Oh, that reminds me — maybe we could also add a test case for when `removeAllRanges()` is called?

43081j:
> Oh, that reminds me — maybe we could also add a test case for when `removeAllRanges()` is called?

im not sure we need to, as we know mutating ranges will always emit a selectionchange event - which we have tested

otherwise you're just testing that browsers are working properly (i.e. that they emit the event as expected)

PR #5102, opened 2025-10-16
Generate a random number or randomly select one from multiple numbers

lychee712:

<details>
<br>

>

</details>

---

### Description

Provides a readable Ref<number> for use in Vue composition: when using two parameters, generates a random integer within a closed interval; when using multiple parameters or an array, randomly selects from the candidate values;


packages/math/useRandomInt/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  return customRef((track, trigger) => ({
+    get() {
+      track()
+      const values = toValueArgsFlat(args)
+
+      // If exactly 2 numbers, return random integer between them
+      if (values.length === 2) {
+        const min = Math.ceil(Math.min(values[0], values[1]))
+        const max = Math.floor(Math.max(values[0], values[1]))
+        return Math.floor(Math.random() * (max - min + 1)) + min
```
OrbisK: This will result in a hydration mismatch with SSR. We need to handle this.

OrbisK (changes requested):

packages/math/useRandomInt/index.ts
OrbisK: I am not sure about `useRandomInt`, but it might be beneficial to have a lowlevel SSR-safe `useMathRandom`.

lychee712:
I have listened to your advice and adopted it , have a nice day

9romise:
Hi! Just curious — was this PR written with the help of a LLM?

lychee712:
> 你好！我只是好奇——这份PR是在LLM的帮助下写的吗？

I designed the API, implemented the logic, wrote the tests, and responded to review feedback manually. That said, like many developers today, I may have used an LLM occasionally for small things

9romise (commented):
I believe the values generated on the server should also match those rendered on the client-side during the initial render.

Additionally, in my view, `useMathRandom` (I'd personally prefer `useRandom`) should behave similarly to other methods in `@vueuse/math` — staying close to browser APIs.
The current implementation feels a bit custom but only works with numbers.
If it's meant to belong to `@vueuse/math`, I think its parameters should align with browser APIs; If it's intended as a custom utility, then the parameters shouldn't be limited to just numbers.
Looking forward to hearing other opinions!

lychee712:
> 我相信服务器上生成的值也应该与初始渲染期间客户端上渲染的值相匹配。
> 
> 此外，在我看来，`useMathRandom`（我个人更喜欢`useRandom`）应该与 中的其他方法类似`@vueuse/math`——保持与浏览器 API 的相似性。 当前的实现感觉有点自定义，但只适用于数字。 如果它属于`@vueuse/math`，我认为它的参数应该与浏览器 API 保持一致；如果它旨在作为自定义实用程序，那么参数不应该仅限于数字。 期待听到更多意见！

You are right. I was thinking today that we should allow strings in multiple choice questions. It shouldn't just be useMathRandom, but more like useRandom. This would fit more scenarios.

OrbisK:
I am not sure about setting the value on mounted. Let me try a different approach.

OrbisK:
I think we want to have a pseudo `Math.random` generator that can be used with a seed. We can save the random seed on window. For the server-side and the initial client-side rendering, we can use the pseudo `Math.random` with the seed. After that, we can use either the real or the pseudo `Math.random` for client-side regenerations. wdyt? @9romise @john-li-creater

lychee712:
> I think we want to have a pseudo `Math.random` generator that can be used with a seed. We can save the random seed on window. For the server-side and the initial client-side rendering, we can use the pseudo `Math.random` with the seed. After that, we can use either the real or the pseudo `Math.random` for client-side regenerations. wdyt? @9romise @john-li-creater

Thanks for the great suggestion! The seeded PRNG approach is indeed more robust than simply returning fixed values for SSR.Better randomness than fixed values。I will modify the code according to your suggestions and add the following features: 1. Allow users to input a string. 2. When users input an array, allow them to input a second value N, which is used to generate N different random strings.

OrbisK:
IMO we should keep it as low level as possible.

```ts
const {random} = useMathRandom()
console.log(random.value) // i.e. 0.38238
```

We should not provide a start or end value.

OrbisK:
> Allow users to input a string

I am not sure what you mean. If you mean the seed, this should not be exposed. It is an internal value that guarantees the first random value on the client and server.

lychee712:
> > 允许用户输入字符串
> 
> 我不太明白你的意思。如果你指的是种子，那么它不应该暴露出来。它是一个内部值，用于保证客户端和服务器上的第一个随机值。

What I mean is to allow users to input multiple strings and implement random multiple-choice selection, rather than just returning numbers

lychee712:
> 我认为我们应该将其保持在尽可能低的水平。
> 
> ```ts
> const {random} = useMathRandom()
> console.log(random.value) // i.e. 0.38238
> ```
> 
> 我们不应该提供开始值或结束值。

I will modify it and return undefined

OrbisK:
> What I mean is to allow users to input multiple strings and implement random multiple-choice selection, rather than just returning numbers

I think this should not be part of `useRandom`. This should be implemented on top of `useRandom` 😅

lychee712:
If you have any questions, please leave a message.I will deal with the problem tomorrow. Now I have to accompany my family.have a nice day @OrbisK @9romise

lychee712:
How was your day today

OrbisK:
@john-li-creater If you agree, I will discuss this internally and, if necessary, make changes to your branch.

lychee712:
> @john-li-creater如果您同意，我将在内部讨论此事，并在必要时对您的分支进行更改。

Thank you for reviewing my PR! I really appreciate you taking the time to look at it. Please feel free to make any changes you see fit - I'm happy to collaborate and learn from your suggestions. Looking forward to seeing this feature in vueuse! 🚀

Issue #4762, opened 2025-05-20
BUG | `useInfiniteScoll` | does not work if `canLoadMore` is initially `false`


maximepvrt:
### Describe the bug

When using the `useInfiniteScroll` via the Nuxt integration (`@vueuse/nuxt`), if `canLoadMore` is initially `false `(e.g., due to a condition not being met on mount), and later becomes a valid function, infinite scrolling never triggers.

### Reproduction

https://stackblitz.com/edit/github-wqevg5tx?file=app.vue

### System Info

```Shell
System:
    OS: macOS 15.3.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 95.20 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Yarn: 1.22.21 - /usr/local/bin/yarn
    npm: 11.3.0 - ~/Library/Application Support/Herd/config/nvm/versions/node/v20.18.0/bin/npm
    pnpm: 10.10.0 - ~/Library/Application Support/Herd/config/nvm/versions/node/v20.18.0/bin/pnpm
    bun: 1.1.32 - ~/.bun/bin/bun
  Browsers:
    Chrome: 136.0.7103.114
    Safari: 18.3
```

### Used Package Manager

npm

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vueuse.org/guide).
- [x] Check that there isn't [already an issue](https://github.com/vueuse/vueuse/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a VueUse issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vueuse/vueuse/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

OrbisK:
[...]

PR #5110, opened 2025-10-21
fix(useInfiniteScroll): make canLoadMore reactive

nhquyss:

---

### Description

Fixes https://github.com/vueuse/vueuse/issues/4762.

### DIFF
--- a/packages/core/useInfiniteScroll/index.test.ts
+++ b/packages/core/useInfiniteScroll/index.test.ts
@@ -56,6 +56,74 @@ describe('useInfiniteScroll', () => {
     expect(mockHandler).toHaveBeenCalledTimes(1)
   })
 
+  it('should not call loadMore when canLoadMore returns false', async () => {
+    const mockHandler = vi.fn()
+    const mockElement = givenMockElement()
+    givenElementVisibilityRefMock(true)
+
+    useInfiniteScroll(mockElement, mockHandler, {
+      canLoadMore: () => false,
+    })
+
+    expect(mockHandler).not.toHaveBeenCalled()
+  })
+
+  it('should call loadMore when canLoadMore returns true', async () => {
+    const mockHandler = vi.fn()
+    const mockElement = givenMockElement()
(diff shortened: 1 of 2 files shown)


antfu (approved):

nhquyss:
Hi @antfu 

Sorry for the delay - I made some additional improvements yesterday but forgot to push them until now.

## Additional Changes

While testing the reactive `canLoadMore` fix in my project, I noticed a couple of issues and added improvements:

**1. Moved `promise.value` check to early return:**
- Prevents unnecessary execution of the logic below when a load is already in progress
- Simplifies the code by avoiding nested `if` conditions

**2. Added `flush: 'post'` to the watch:**
- I encountered a bug in my project when using `document`/`window` as the element and testing on mobile screen sizes, where `onLoadMore` was being called unexpectedly even without scrolling
- Not entirely sure why this happens, and I haven't been able to reproduce it in a minimal example yet
- Adding `flush: 'post'` resolved the issue in my case
- This ensures the watcher runs after DOM updates, which might help with timing-sensitive scenarios

**Code changes:**
```ts
// Before:
if (state.arrivedState[direction] || isNarrower) {
  if (!promise.value) {
    promise.value = Promise.all([...])
  }
}
....
const stop = watch(
  () => [state.arrivedState[direction], isElementVisible.value],
  checkAndLoad,
  { immediate: true }
)

// After:
if (!observedElement.value || !isElementVisible.value || !canLoad.value || promise.value)
  return  // Early return - no need to run logic below

if (state.arrivedState[direction] || isNarrower) {
  promise.value = Promise.all([...])
}
....
const stop = watch(
  () => [state.arrivedState[direction], isElementVisible.value, canLoad.value],
  checkAndLoad,
  { immediate: true, flush: 'post' }
)
```

Could you please review these additional changes? All tests are still passed.

If you think `flush: 'post'` might cause issues or if you'd like me to check further, please let me know!

Thanks!

antfu:
@nhquyss you can open a new PR and we will review it. Thanks!

Issue #5112, opened 2025-10-22
BUG | `refManualReset` | The reactive behavior is inconsistent with "ref" and the deep reactivity has been lost.


eric-gitta-moore:
### Describe the bug

The reactive behavior is inconsistent with "ref" and the deep reactivity has been lost.

```tsx
import { refManualReset } from '@vueuse/core'
import { isReactive, ref } from 'vue'

const test = refManualReset({ level1: { level2: '' } })
console.log(isReactive(test.value.level1)) // false, but expect true
const test2 = ref({ level1: { level2: '' } })
console.log(isReactive(test2.value.level1)) // true
```

Temporary repair plan

```tsx
// temp fixed
const test3 = refManualReset(reactive({ level1: { level2: '' } }))
//                           ^~ Wrap the initial value with the reactive package
console.log(isReactive(test3.value.level1)) // true
test3.reset();
console.log(isReactive(test3.value.level1)) // true
```

### Reproduction

[...]

PR #5115, opened 2025-10-23
feat(refManualReset): add shallow parameter

phk422:
via: #5112


antfu:
I think this is out-of-scope. This function is designed to be shallow and does not handle object cloning. We should update the docs instead. Thanks for the PR.

PR #5122, opened 2025-10-26
fix(useInfiniteScroll): improve promise handling and add flush post to watch

nhquyss:

### Description

Follow-up improvements to #5110 

For detailed, see [this comment in PR](https://github.com/vueuse/vueuse/pull/5110#issuecomment-3448076861)

### Changes

- Move `promise` check to early return to prevent unnecessary logic execution

- Add `flush: 'post'` to watch to ensure it runs after DOM updates


OrbisK:
Thank you 🙏🏽 

Would you mind adding some tests to cover the new handling?

maximepvrt:
@nhquyss

nhquyss:
> Thank you 🙏🏽
> 
> Would you mind adding some tests to cover the new handling?

Hi @OrbisK, sorry for the late reply.

I’ve tried adding tests for this change, but I still haven’t been able to write one that reproduces the issue.​

So I created a small live demo to show the behavior before and after this change.

[Stackblitz live demo](https://stackblitz.com/~/github.com/nhquyss/infinite-scroll-flush-demo)

Without `flush: 'post'`, the API is called **twice** on the initial load even though the first call already returns enough items to fill the screen.

Hope this demo helps clarify the issue!!

nhquyss:
> @nhquyss

@maximepvrt, thanks for the reminder :)

I've added a demo above.

Let me know if you need anything else!!

maximepvrt:
Hi @nhquyss, thanks for the update!

I’ve just tested with the newly released v14.1.0 (which includes https://github.com/vueuse/vueuse/pull/5110) as well as with the PR build from [pkg.pr.new](https://github.com/vueuse/vueuse/pull/5122/checks?check_run_id=56163169022).

I’m running into an issue: it triggers **infinite loading of my data**.

Here is a reproduction of the issue:
https://stackblitz.com/edit/github-ostrvl2k

Do you have any idea what might be causing this?

nhquyss:
> Hi @nhquyss, thanks for the update!
> 
> I’ve just tested with the newly released v14.1.0 (which includes #5110) as well as with the PR build from [pkg.pr.new](https://github.com/vueuse/vueuse/pull/5122/checks?check_run_id=56163169022).
> 
> I’m running into an issue: it triggers **infinite loading of my data**.
> 
> Here is a reproduction of the issue: https://stackblitz.com/edit/github-ostrvl2k
> 
> Do you have any idea what might be causing this?

@maximepvrt

I think the scroll container in your `Items.vue` doesn't have a fixed height or max-height set.

This causes `useInfiniteScroll` to continuously trigger the load function because of this condition check in the composable:

```
const isNarrower =
  direction === "bottom" || direction === "top"
    ? scrollHeight <= clientHeight
    : scrollWidth <= clientWidth;
```
Without a defined height, `scrollHeight` always equals `clientHeight`, making the condition for loading the next page always true, which leads to **infinite loading**.

I've created [a demo based on your reproduction](https://stackblitz.com/edit/github-ostrvl2k-6wzfebap) with a `max-height` applied to the scroll container. This way, the next page will only be loaded when you actually scrolls to the bottom of the container.

Or if you prefer full-page scrolling, you can use `useInfiniteScroll` composable with `document` or `window` as the scroll element instead of a specific container:

```
useInfiniteScroll(
  document,
  async () => {
    await refresh();
  },
  {
    canLoadMore: () => canLoadMore.value && props.infiniteScroll,
    distance: 100,
  }
);
```

Hope this helps resolve your issue!!

maximepvrt:
@nhquyss A huge thank you for your help! 🙏
I thought the composable just added a div at the bottom and that if it entered the visible area, it would trigger the next load. I didn’t realize that not having a fixed height on the container caused `scrollHeight === clientHeight` all the time, which made it load infinitely.

Thanks again!

maximepvrt:
@OrbisK @antfu can you review this PR ? 🙏

PR #5125, opened 2025-10-28
Update classes for layout and textarea elements

josueayala27:
### Description

Update classes for layout and textarea elements

### DIFF
--- a/packages/core/useBase64/demo.vue
+++ b/packages/core/useBase64/demo.vue
@@ -17,55 +17,54 @@ function onFileInput(e: Event) {
 </script>
 
 <template>
-  <div space-y-4>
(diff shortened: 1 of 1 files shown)


9romise:
Hi! This PR doesn't seem to modify any actual styles — is there something specific we should be aware of? 🤔

josueayala27:
> Hi! This PR doesn't seem to modify any actual styles — is there something specific we should be aware of? 🤔

`css` classes wasn't inside `class=""`

9romise:
We are using UnoCSS's [attributify preset](https://unocss.dev/presets/attributify), so it's possible to style elements without adding classes.

PR #5129, opened 2025-10-31
feat: support configurable scheduler for timed composables

9romise:
### Description

This PR is redesigned with #5070.

For those interested in this configuration option:
Please refer to the preview version at: https://deploy-preview-5129--vueuse.netlify.app/guide/config.html#custom-scheduler
For the official documentation, see: https://vueuse.org/guide/config.html#custom-scheduler


9romise:
I'd like to confirm if this aligns with what you had in mind — perhaps we no longer need to worry about scheduler options like `immediate` and `interval`?
The CI failures can be ignored for now — I'll fix the errors once we confirm the correct approach.

/cc @43081j  @antfu

packages/core/useCountdown/index.ts
```diff
@@ -1,13 +1,29 @@
...
+      immediate = false,
+    } = options
+
+    return (cb: AnyFn) => useIntervalFn(cb, interval, { immediate })
+  }
+
+  return (cb: AnyFn) => useIntervalFn(cb, 1000, { immediate: false })
+}
+
+export interface UseCountdownOptions<Legacy = false> extends ConfigurableScheduler {
```
43081j: are you doing this as a type parameter for backwards compat?

i.e. why not two interfaces, `UseCountdownOptions` and `UseCountdownOptionsLegacy`

personally i'd just have one though with the two deprecated properties, keep it simple. in the implementation we can ignore them if the newer options are set anyway

9romise: Do you mean
```ts
interface UseCountdownOptionsLegacy {
  /**
   *  Interval for the countdown in milliseconds. Default is 1000ms.
   *
   * @deprecated
   */
  interval?: MaybeRefOrGetter<number>
  /**
   * Start the countdown immediately
   *
   * @deprecated
   * @default false
   */
  immediate?: boolean
}
```

This would make it harder for TypeScript to infer the type of `options` correctly, and we need to add type assertions or `@ts-expect-error` to make it happy.

9romise: Or
```ts
export interface UseCountdownOptionsLegacy extends UseCountdownOptions {
...
```
?

43081j: im suggesting it doesn't need to infer the type of options.

since there's no crossover between old and new options, what's stopping us from doing this:

```ts
export interface UseCountdownOptions extends SchedulerOptions {
  /** @deprecated */
  interval?: MaybeRefOrGetter<number>;

  // ...
}

// ...

if (options.scheduler) {
  // do the new way
} else {
  // old way
}

// or the other way around in case new options end up having more properties in future

if (options.interval || options.immediate !== undefined) {
  // old way
} else {
  // new way
}
```

9romise: Oh, the reason for distinguishing the `legacy` type is to enable better editor support in overloaded methods. It helps make users aware of deprecation, since marking options with `@deprecated` alone isn’t always noticeable enough.
image

9romise: For now:
image

43081j: it seems a shame to have what are basically hacky types (using a boolean type param as a interface brand) to get a nicer deprecation notice

personally i'd just have the one interface, as the properties are what are deprecated, not any overloads.

but im not going to block the PR. do as you wish 👍  if you think the bigger deprecation message is worth it, go for it

43081j (approved):

OrbisK (approved):

OrbisK:
Partially related to this PR:

We should update the guidelines at some point:
image

antfu (approved):

PR #5130, opened 2025-10-31
feat(useWebsocket): add exponential backoff when auto reconnect

luoling8192:
When retrying with exponential backoff, the retry time will become 2^n times the delay.

References:
https://docs.cloud.google.com/memorystore/docs/redis/exponential-backoff
https://en.wikipedia.org/wiki/Exponential_backoff

### DIFF
--- a/packages/core/useWebSocket/index.browser.test.ts
+++ b/packages/core/useWebSocket/index.browser.test.ts
@@ -89,5 +89,95 @@ describe('useWebSocket', () => {
 
       expect(status.value).toBe('CONNECTING')
     })
+
+    it('should use fixed delay when backoff is false', () => {
+      const { ws, status } = useWebSocket('ws://localhost', {
+        autoReconnect: {
+          retries: 3,
+          delay: 1000,
+          backoff: false,
+        },
+      })
+
+      ws.value?.onopen?.(new Event('open'))
+
+      // First reconnect - should use fixed delay of 1000ms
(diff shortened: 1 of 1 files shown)


9romise:
I believe #5089 can achieve the same outcome that this PR is aiming for.

luoling8192:
> I believe #5089 can achieve the same outcome that this PR is aiming for.

This press release is straightforward and doesn't clash with #5089. Adding an extra option could be beneficial. 

Many may not know how to implement exponential backoff, but with this built-in feature, it's as simple as toggling a switch, eliminating the need for a custom anonymous function.

9romise:
> > I believe #5089 can achieve the same outcome that this PR is aiming for.
> 
> This press release is straightforward and doesn't clash with #5089. Adding an extra option could be beneficial.
> 
> Many may not know how to implement exponential backoff, but with this built-in feature, it's as simple as toggling a switch, eliminating the need for a custom anonymous function.

That makes sense. However, I think it would be better to either:

1. Allow `autoConnect.delay` to accept a string like 'backoff', or
2. Provide a preset backoff implementation that users can pass to the `delay` option.

The second approach would be more maintainable and enable better tree-shaking (despite its small size), though we'd need to wait for #5089 to be merged first. What are your thoughts?

agroebe:
Personally I would prefer #5089 because it allows you to implement all sorts of reconnect strategies, including exponential backoff.

PR #5132, opened 2025-10-31
chore: remove gitpod

iiio2:
Gitpod is now Ona. (https://ona.com/stories/gitpod-is-now-ona)
This is the migration setup. (https://ona.com/docs/ona/configuration/migrate-from-classic)


antfu:
I think we should remove GitPod then. We are not collaborating with them, and I don't think we should bring traffic to them as an open source project.

iiio2:
@antfu , I've removed gitpod.

iiio2:
Thanks @antfu .

PR #5136, opened 2025-11-01
test(useIntersectionObserver): Add browser tests for useIntersectionObserver

kevinluo201:

<details>
<br>

>

</details>

---

### Description

Add the tests for `useIntersectionObserver`, and it should be easier for developers to add new features to this composable. 

### Additional context

This PR add both unit test and browser test. 

#### Unit Test
Why does it still have a unit test while JSDom doesn't have `IntersectionObserver`? First, it can check if `isSupported` works. Second, I think Vueuse doesn't need to know if `IntersectionObserver` works correctly; that should be the browser's responsibility. Therefore, verifying the messages sent to the `IntersectionObserver` from `useIntersectionObserver` makes more sense.

#### Browser Test
Again, while we don't need to check the correctness of the `IntersectionObserver`, it should check the reactivity and the exposed properties, which are implemented uniquely by `useIntersectionObserver`

### DIFF
--- a/packages/core/useIntersectionObserver/index.browser.test.ts
+++ b/packages/core/useIntersectionObserver/index.browser.test.ts
@@ -1,7 +1,7 @@
 import type { UseIntersectionObserverReturn } from '.'
-import { page } from '@vitest/browser/context'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { cleanup } from 'vitest-browser-vue'
+import { page } from 'vitest/browser'
 import { defineComponent, shallowRef, toValue, useTemplateRef } from 'vue'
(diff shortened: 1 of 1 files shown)


kevinluo201:
I think the way `Component` is created in each test may be simplified. It is because the template and the options of `useIntersectionObserver` change almost in every case, so I just left them like that. However, I think the most important thing is, do you think the test cases make sense and cover the usage of the composable?

kevinluo201:
btw, I keep seeing the warning
```
[@vue/compiler-core] decodeEntities option is passed but will be ignored in non-browser builds.
```
in raised in every test, but I have no idea what it is...

PR #5137, opened 2025-11-01
chore: patch google-font-installer

userquin:
### Before submitting the PR, please make sure you do the following

- [ ] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [ ] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [ ] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

supersedes #5134

### Additional context

Beware running the `install-fonts` on Windows, will show a popup to install the fonts, running the script again will ask if you want to override each font previously installed:

image

PR #5138, opened 2025-11-02
chore(dev-deps): migrate to `rolldown-vite`

9romise:

<details>
<br>

>

</details>

---

### Description

This PR replaces `vite` with `rolldown-vite` in the whole repo.

### Additional context

https://github.com/vueuse/vueuse/pull/5137#issuecomment-3476672874
Thanks @brc-dd

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -314,16 +314,16 @@ importers:
         version: 7.7.1
       '@vite-pwa/vitepress':
         specifier: catalog:docs
-        version: 1.0.1(vite-plugin-pwa@1.1.0(rolldown-vite@7.2.5(@types/node@24.9.1)(esbuild@0.25.11)(jiti@2.6.1)(terser@5.24.0)(tsx@4.20.6)(yaml@2.8.1))(workbox-build@7.0.0)(workbox-window@7.0.0))
+        version: 1.0.1(vite-plugin-pwa@1.1.0(rolldown-vite@7.2.7(@types/node@24.9.1)(esbuild@0.25.11)(jiti@2.6.1)(terser@5.24.0)(tsx@4.20.6)(yaml@2.8.1))(workbox-build@7.0.0)(workbox-window@7.0.0))
       '@vitejs/plugin-vue':
         specifier: catalog:dev
-        version: 6.0.1(rolldown-vite@7.2.5(@types/node@24.9.1)(esbuild@0.25.11)(jiti@2.6.1)(terser@5.24.0)(tsx@4.20.6)(yaml@2.8.1))(vue@3.5.22(typescript@5.9.3))
+        version: 6.0.1(rolldown-vite@7.2.7(@types/node@24.9.1)(esbuild@0.25.11)(jiti@2.6.1)(terser@5.24.0)(tsx@4.20.6)(yaml@2.8.1))(vue@3.5.22(typescript@5.9.3))
       '@vitest/browser-playwright':
         specifier: catalog:test
-        version: 4.0.8(msw@2.3.0(typescript@5.9.3))(playwright@1.49.0)(rolldown-vite@7.2.5(@types/node@24.9.1)(esbuild@0.25.11)(jiti@2.6.1)(terser@5.24.0)(tsx@4.20.6)(yaml@2.8.1))(vitest@4.0.8)
+        version: 4.0.8(msw@2.3.0(typescript@5.9.3))(playwright@1.49.0)(rolldown-vite@7.2.7(@types/node@24.9.1)(esbuild@0.25.11)(jiti@2.6.1)(terser@5.24.0)(tsx@4.20.6)(yaml@2.8.1))(vitest@4.0.8)
       '@vitest/coverage-v8':
         specifier: catalog:test
-        version: 4.0.8(@vitest/browser@4.0.8(msw@2.3.0(typescript@5.9.3))(rolldown-vite@7.2.5(@types/node@24.9.1)(esbuild@0.25.11)(jiti@2.6.1)(terser@5.24.0)(tsx@4.20.6)(yaml@2.8.1))(vitest@4.0.8))(vitest@4.0.8)
+        version: 4.0.8(@vitest/browser@4.0.8(msw@2.3.0(typescript@5.9.3))(rolldown-vite@7.2.7(@types/node@24.9.1)(esbuild@0.25.11)(jiti@2.6.1)(terser@5.24.0)(tsx@4.20.6)(yaml@2.8.1))(vitest@4.0.8))(vitest@4.0.8)
       '@vitest/ui':
         specifier: catalog:test
         version: 4.0.8(vitest@4.0.8)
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "dev": "nr update && nr docs",
     "docs": "vitepress dev packages --open",
     "docs:build": "nr update:full && nr build:packages && turbo run docs:build:vitepress && nr build:redirects",
-    "docs:build:vitepress": "NODE_OPTIONS=\"--max-old-space-size=5120\" ROLLDOWN_OPTIONS_VALIDATION=loose vitepress build packages",
+    "docs:build:vitepress": "vitepress build packages",
```
userquin: Add back the node options here, we need some more memory to build the docs, just the memory heap

userquin:
We should update playgrounds,  maybe updating nuxt  also to v4. I guess adding vite with rolldown-vite at resolutions should be better.

Maybe we should check first tests, vitest tests maybe broken using rolldown.

userquin:
We should update vitest to v4 or apply this patch to fix unit tests using node 24, will check if ported to v3: https://github.com/vitest-dev/vitest/pull/8390#issuecomment-3330925212

9romise:
> We should update playgrounds, maybe updating nuxt also to v4. I guess adding vite with rolldown-vite at resolutions should be better.

Oh, I did miss those, but we seem not to use `playgrounds/nuxt`. See: https://github.com/vueuse/vueuse/blob/main/playgrounds/build.sh

> Maybe we should check first tests, vitest tests maybe broken using rolldown.

The errors in CI appear to be related to vitest: https://github.com/vitest-dev/vitest/issues/8374. Upgrading to vitest v4 might resolve the issue.

package.json
9romise: Do we also need to specify it in `package.json` if it's already defined in [`netlify.toml`](https://github.com/vueuse/vueuse/blob/main/netlify.toml#L6)?

9romise:
The `playgrounds` are only used for smoke testing.
- https://github.com/search?q=repo%3Avueuse%2Fvueuse+playgrounds&type=code
- https://github.com/vueuse/vueuse/blob/c33aaca64582f419efe00d00fb50e96da5a4df90/.github/workflows/ci.yml#L95

packages/.vitepress/vite.config.ts
```diff
@@ -70,31 +70,28 @@ export default defineConfig({
...
-    include: [
-      'axios',
-      'yaml',
-      'nprogress',
-      'qrcode',
-      'tslib',
-      'fuse.js',
-      'universal-cookie',
-    ],
-    disabled: 'dev',
```
9romise, line 88: Related to https://vite.dev/config/dep-optimization-options.html#optimizedeps-disabled

9romise: Seems like we can remove it directly, because it's only used in `dev` now and we disabled it in `dev`.

userquin:
> The errors in CI appear to be related to vitest: [vitest-dev/vitest#8374](https://github.com/vitest-dev/vitest/issues/8374). Upgrading to vitest v4 might resolve the issue.

Beware, there are some changes we need for browser tests and coverage, browser provider now must be imported from `@vitest/browser-<provider>`, check: https://vitest.dev/guide/migration.html#vitest-4

packages/.vitepress/vite.config.ts
userquin: iirc the include options should be an empty array or including just CJS-only dependencies: https://vite.dev/config/dep-optimization-options.html#optimizedeps-nodiscovery

userquin:
> Oh, I did miss those, but we seem not to use `playgrounds/nuxt`. See: https://github.com/vueuse/vueuse/blob/main/playgrounds/build.sh

Where is the sh being used (I use Windows 10/11)? forgot it, it is being used in the CI workflow arggg

9romise:
> > The errors in CI appear to be related to vitest: [vitest-dev/vitest#8374](https://github.com/vitest-dev/vitest/issues/8374). Upgrading to vitest v4 might resolve the issue.
> 
> Beware, there are some changes we need for browser tests and coverage, browser provider now must be imported from `@vitest/browser-<provider>`, check: https://vitest.dev/guide/migration.html#vitest-4

Thanks for the heads-up! 💚 I've noticed it and am already working on the upgrade in a separate branch. I'll submit a PR once it's ready.

package.json
userquin: in both, if running local docs build we'll have memory heap error, at netlify.toml for the same reason

9romise: My initial thought was to keep the script clean, since we don't usually build docs locally 😄 — but it doesn't really matter, I'll add it back here.

userquin: I build the docs on my local to test the CI doesn't break, and takes  ~1minute (the VP build process takes a lot of time)

packages/.vitepress/vite.config.ts
9romise: Leaving it `undefined` is fine. https://vite.dev/config/dep-optimization-options.html#optimizedeps-disabled
> To disable the optimizer completely, use `optimizeDeps.noDiscovery: true` to disallow automatic discovery of dependencies and leave `optimizeDeps.include` undefined or empty.

userquin: why removing `exclude` in https://github.com/vueuse/vueuse/pull/5138/commits/1b0d1fd6f56968849dac08b94048f73978ea5b1b? now not required?

9romise: [According to the documentation](https://vite.dev/config/dep-optimization-options.html#optimizedeps-nodiscovery):
> When set to `true`, automatic dependency discovery will be disabled and only dependencies listed in `optimizeDeps.include` will be optimized.

Since we set `optimizeDeps.noDiscovery` to `true`, making `optimizeDeps.exclude` meaningless. Please correct me if I'm misunderstanding.

userquin: not sure, I haven't used `noDiscovery` yet, that's why I'm asking

9romise: I'll leave a comment here specifically so the reviewers can take note of this.

packages/.vitepress/vite.config.ts
```diff
@@ -70,31 +70,22 @@ export default defineConfig({
...
+    noDiscovery: true,
   },
```
9romise, line 89: Please note: `optimizeDeps.disabled` has been deprecated, so I've modified this section accordingly. I'm not entirely sure if the change is correct, so I'd appreciate it if you could pay extra attention during review.

Here is the documentation: https://vite.dev/config/dep-optimization-options.html#optimizedeps-disabled

43081j: doesn't this mean we should still have `include`?

it sounds like `noDiscovery` just means anything _not inside include_ will not automatically be optimised during dev

so we could still populate include and get rid of `exclude` maybe?

9romise: IMO, we previously had `optimizeDeps.disabled: 'dev'`, which meant the configuration only took effect during build. However, since Vite 5.1, Vite no longer performs the optimization during build, meaning we now need to disable this feature entirely.

BTW, I also referenced [`isDepOptimizationDisabled`](https://github.com/vitejs/vite/blob/e3a6a83406943bc59a9916cae3f25ab33c2b5802/packages/vite/src/node/optimizer/index.ts#L168-L176) in Vite.

43081j: makes sense 👍

43081j (approved):
looks good

blocked until the vitest 4 upgrade i guess (which is why ci fails)

OrbisK (approved):

PR #5151, opened 2025-11-05
feat(whenever): allow array sources

Flambe:

<details>
<br>

>

</details>

---

### Description

The vue `watch` function allows an array of `WatchSource`s to be passed as the watch source, and it then triggers on any of the items in the array changing. `whenever` directly passes the `source` to `watch` but the typing & subsequent callback do not support arrays.

This PR adds typing for arrays and only fires the callback if **any** of the array items are truthy. If none of the items in the array are truthy then it does not fire the callback.

```typescript
// before
function doSomething() {
  // ...
}

whenever(valueA, doSomething)
whenever(valueB, doSomething)

// after
whenever([valueA, valueB], () => {
  // ...
})
```

### Additional context

Should there be an option to only fire the callback if every array item is truthy?
If so, which should be the default?

There is a similar PR #5096 focused on the `oldValue` typing. Although my PR doesn't make the same changes, it does conflict so (presuming both PRs are accepted) it might be worth me and @VChet getting our changes together into a single PR?

### DIFF
--- a/packages/shared/whenever/index.ts
+++ b/packages/shared/whenever/index.ts
@@ -23,11 +23,7 @@ export function whenever<T extends object>(source: T, cb: WatchCallback<T>, opti
  *
  * @see https://vueuse.org/whenever
  */
-export function whenever(
-  source: Readonly<WatchSource<unknown>[]> | WatchSource<unknown> | object,
-  cb: WatchCallback,
-  options?: WheneverOptions,
-) {
+export function whenever(source: object, cb: WatchCallback, options?: WheneverOptions) {
   const stop = watch(
     source,
     (v, ov, onInvalidate) => {


packages/shared/whenever/index.ts
```diff
@@ -12,16 +13,21 @@ export interface WheneverOptions extends WatchOptions {
...
+export function whenever<T>(source: WatchSource<T | false | null | undefined>, cb: WatchCallback<T>, options?: WheneverOptions): WatchHandle
+export function whenever<T extends object>(source: T, cb: WatchCallback<T>, options?: WheneverOptions): WatchHandle
+
 /**
  * Shorthand for watching value to be truthy
  *
  * @see https://vueuse.org/whenever
  */
-export function whenever<T>(source: WatchSource<T | false | null | undefined>, cb: WatchCallback<T>, options?: WheneverOptions) {
+export function whenever(source: any, cb: WatchCallback, options?: WheneverOptions) {
```
43081j: can we keep this strongly typed? or at least use `unknown`

`any` is basically a source for bugs and "turn off the type checker"

either we should use `unknown` or type it properly (`WatchSource<...> | Array<WatchSource<...>>`)

Flambe: i've tried switching it out for more explicit typing and i believe it's fairly accurate now. My knowledge of generics is not enough to know if i can properly use them to remove the `unknown`

packages/shared/whenever/index.ts
```diff
@@ -12,16 +13,25 @@ export interface WheneverOptions extends WatchOptions {
...
+export function whenever<T extends object>(source: T, cb: WatchCallback<T>, options?: WheneverOptions): WatchHandle
+
 /**
  * Shorthand for watching value to be truthy
  *
  * @see https://vueuse.org/whenever
  */
-export function whenever<T>(source: WatchSource<T | false | null | undefined>, cb: WatchCallback<T>, options?: WheneverOptions) {
+export function whenever(
+  source: Readonly<WatchSource<unknown>[]> | WatchSource<unknown> | object,
```
43081j: this can probably still be `T` i suppose, like:


```suggestion
  source: Readonly<WatchSource<T>[]> | WatchSource<T> | object,
```

although it won't matter much since all of it is widened by `object` anyway

packages/shared/whenever/index.ts
```diff
@@ -12,16 +13,25 @@ export interface WheneverOptions extends WatchOptions {
   once?: boolean
 }
 
+// overloads
+export function whenever<T extends Readonly<WatchSource<unknown>[]>>(sources: readonly [...T] | T, cb: WatchCallback<MapSources<T>, MapOldSources<T, true>>, options?: WheneverOptions): WatchHandle
+export function whenever<T>(source: WatchSource<T | false | null | undefined>, cb: WatchCallback<T>, options?: WheneverOptions): WatchHandle
+export function whenever<T extends object>(source: T, cb: WatchCallback<T>, options?: WheneverOptions): WatchHandle
```
43081j, line 19: what is this one accounting for? do you have an example, or can you point me at the test that covers this one?

Flambe: i was copying the syntax from some of the other watch functions to be consistent https://github.com/vueuse/vueuse/blob/main/packages/shared/watchThrottled/index.ts#L15 https://github.com/vueuse/vueuse/blob/main/packages/shared/watchWithFilter/index.ts#L11

from my understanding of typescript, `object` should be covered by the `WatchSource<T | false | null | undefined>` but i didn't want to deviate from the others linked above. I can't point to any explicit test for this usecase because of that reasoning

43081j: yeah i see. its because `watch` can work against arbitrary objects. so whoever wrote these other ones was trying to account for that

fair enough!

the implementation (L26) should probably just be `source: object` in that case, given the type system will widen it to that anyway

Flambe: makes sense to me, thank you for the explanation! 

i've switched it for `source: object` which i presume makes this review outdated? https://github.com/vueuse/vueuse/pull/5151#discussion_r2533845762

antfu:
I am not so sure if we should do this. It's implicit to say if we want the array items to behave like an AND or OR. I think user should better use `logicAnd` or `logicOr` to compose explicitly.

Flambe:
> I am not so sure if we should do this. It's implicit to say if we want the array items to behave like an AND or OR. I think user should better use `logicAnd` or `logicOr` to compose explicitly.

Thank you for considering it! I didn't realise that these existed honestly and they're definitely clearer to use.

PR #5175, opened 2025-11-19
fix(useWebSocket): race condition caused by onopen/onclose events.

DanCardin:

---

### Description

Certain timings of onclose/onopen on the underlying websocket mutate the persistent `status` state while other asynchronous actions may have already replaced the `wsRef`.

This race is **very** frequent in our use case where the websocket URL is reactive and can be changed by user action relatively frequently. It only takes a couple of user actions to encounter timing where the mutations of the shared `status` state can become out of sync with the actual status of the websocket. In our cases, it's typically that the composable status will become `CLOSED` while the `wsRef`'s status is 1 (open)

### Additional context

It maybe seems like `status` could just be derived from the underlying websocket status instead of being separate ref state that needs to be kept in sync, but that'd have been a lot more involved and there's probably a reason it is the way it is.


DanCardin:
i'm not certain how to make the unit tests work. it seems to be related to equality among the mocks in a way that does not happen in the browser.

PR #5193, opened 2025-11-30
fix(useStorage): cross-document sync solution for custom storage backends

Promise2679:

<details>
<br>

>

</details>

---

### Description

When using custom storage backends, state changes cannot be synchronized across different documents.

Here is a simple [example](https://stackblitz.com/edit/vitejs-vite-k46mds4h?file=src%2FApp.vue). If you open two tabs and click the button in tab A, tab B will not update the value instantly.

As mentioned in notes, I solved this by using a BroadcastChannel object to send and receive data.

### DIFF
--- a/packages/core/useStorage/index.ts
+++ b/packages/core/useStorage/index.ts
@@ -2,10 +2,12 @@ import type { Awaitable, ConfigurableEventFilter, ConfigurableFlush, RemovableRe
 import type { MaybeRefOrGetter } from 'vue'
 import type { ConfigurableWindow } from '../_configurable'
 import type { StorageLike } from '../ssr-handlers'
-import { pausableWatch, tryOnMounted, tryOnScopeDispose } from '@vueuse/shared'
+import type { UseBroadcastChannelReturn } from '../useBroadcastChannel'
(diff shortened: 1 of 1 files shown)


Promise2679:
> I think the content added in this PR goes outside the responsibility of `useStorage`. We also provide `useBroadcastChannel`, so your use case can be achieved by using `useStorage` and `useBroadcastChannel` together.

I think since `useStorage` supports cross-document sync for `localStorage` (although it is handled by the browser), it should also offer the same capability when using `customStorage`, rather than a more complicated approach just because a custom storage is used.

Anyway, I am just following the TODO's instruction, I'll respect your final decision.

9romise:
> Anyway, I am just following the TODO's instruction, I'll respect your final decision.

Oh, I see. In this case, would it be possible to use `useBroadcastChannel` directly instead of implementing it manually?

Promise2679:
yeah I'll have a try

Promise2679:
`useBroadcastChannel` internally listens to the `message` event, but it doesn't offer a custom callback, so I use a watcher to call the `update` function.

I'm wondering whether it would make sense to add an `onMessage` option to `useBroadcastChannel`.

PR #5196, opened 2025-12-02
fix(useStorageAsync): ensure that data is resolved properly

### LINKED ISSUE #5195: Promise returned by `useStorageAsync` never resolves
The example [from the docs of `useStorageAsync`](https://github.com/vueuse/vueuse/blob/8ff6f1948b68aab4e840de24678a38452d1e0c2d/packages/core/useStorageAsync/index.md?plain=1#L36-L41) does not actually work because the returned Promise does never resolve:

```ts
// Use top-level await if your environment supports it
const accessToken = await useStorageAsync('access.token', '', SomeAsyncStorage)

console.log(accessToken.value) // "the real value stored in storage"
```

See [full reproducer example in playground](https://playground.vueuse.org/?vueuse=14.1.0#eNqFU9Fq2zAU/ZWLKNiBYD9sT1lS1o0+dIxtLH3UQzX7xnFjS0aS3RTjf9+VFMduWlpIwLr36NxzjqSe3TRN0rXIVmxtMl02FgzatoFKyGLDmTWcXXNZ1o3SFnrYWqVFgT/LA96YZ5ktoTV4KvoCDLDTqoboK7FSL82UxmjGoHF3xhAk+sIll5mSxsJW1YHkRLh6NQ420HMJIPyiQHtnsY4P+LwIdQDHpCpMKlXEDxpFXsoCrnqCDA8LP8uhNHnUEiJD/JhDJ6qgBGBYTvyUxMi/DJi3xzzp0k5jSONV79Fh4ktOjbXq8CPZhLnQHViGKSxjhW0NDaNA46hB6ZxGDhr6IsvQmHt1QEmgi1OKo9BOrOtHS4jofxn/gssZSWL3KON4AZvrIDpNwe5LA/ST2KEGPGLWWorzH2aCBr6QkCsccRrJaoeOY257PiuE7W0Hn6FCRqLT7twd1+Ag6zRcXLqmtKBgm0pYpBVA348xDZTcOp012ZLuNo3flUXyaJSkB+BdcZapuikr1L8bW5I8zlbjIXEmqko9/fA1q1v05+r37DE7vFF/NEdX4+wPyUZNptm5Z4WmCxzat9tfeKTvc7NWeVsR+p3mXxdE6zQG2LdW5iR7hvNq7/zDo7txb26PFqUZTTmh0+3kjB7j93esT3I/JZ/9PsqfDf8Bq3B3Yg==).

The returned Promise is meant to resolve with the data after the first/initial read: https://github.com/vueuse/vueuse/blob/8ff6f1948b68aab4e840de24678a38452d1e0c2d/packages/core/useStorageAsync/index.ts#L104-L109

The issue here is though, that `data` - having a `then` and `catch` method is a Promise itself: https://github.com/vueuse/vueuse/blob/8ff6f1948b68aab4e840de24678a38452d1e0c2d/packages/core/useStorageAsync/index.ts#L136-L139

So calling `resolve(data)` does not resolve with the _value_ `data` but instead first tries to resolve `data` as a Promise - which circularly depends on itself...

patienry:

<details>
<br>

>

</details>

---

### Description

fix #5195

There is already a test case for this scenario--'onReadyByPromise'. But `(result) => { expect(result.value).toBe('AnotherValue') }` won't execute before or after the modification, so I modified this test case.

### DIFF
--- a/packages/core/useStorageAsync/index.ts
+++ b/packages/core/useStorageAsync/index.ts
@@ -104,6 +104,12 @@ export function useStorageAsync<T extends(string | number | boolean | object | n
   const promise = new Promise((resolve) => {
     read().then(() => {
       onReady?.(data.value)
+
+      Object.assign(data, {
+        then: undefined,
+        catch: undefined,
+      })
+
       resolve(data)
     })
   })
--- a/packages/core/useStorageAsync/index.test.ts
+++ b/packages/core/useStorageAsync/index.test.ts
@@ -66,16 +66,12 @@ describe('useStorageAsync', () => {
   it('onReadyByPromise', async () => {
     localStorage.setItem(KEY2, 'AnotherValue')
 
-    const storage = useStorageAsync(
+    const storage = await useStorageAsync(
       KEY2,
       '',
       new AsyncStubStorage(),
     )
 
-    expect(storage.value).toBe('')
-
-    storage.then((result) => {
-      expect(result.value).toBe('AnotherValue')
-    })
+    expect(storage.value).toBe('AnotherValue')
   })
 })


packages/core/useStorageAsync/index.ts
```diff
@@ -104,6 +104,12 @@ export function useStorageAsync<T extends(string | number | boolean | object | n
   const promise = new Promise((resolve) => {
     read().then(() => {
       onReady?.(data.value)
+
+      Object.assign(data, {
```
acran, line 108: This mutates the `data` returned by `useStorageAsync()` and thus creates race conditions where `data` is not a Promise anymore, i.e.:
~~~js
const data = useStorageAsync('example', 5);

// do something async
await new Promise(resolve => setTimeout(resolve, 1000));

await data; // errors when data already did resolve and `then` was unset
~~~

patienry: Thank you for your feedback. Personally, I feel that most use cases should be like the following.

```
const accessToken = await useStorageAsync('access.token', '', SomeAsyncStorage)
// do something

// or

const accessToken = useStorageAsync('access.token', '', SomeAsyncStorage)
accessToken.then(() => {
  // do something
})
```

In what scenarios would your example be used? 🤔

acran: I agree that that is the most common and expected use case. My example was only to illustrate the possible race condition here - which will result in a breaking exception.

(although my example actually was wrong since `await` will automatically wrap plain values, so no issue here, but only when using `then()`, see below)

The issue is that the returned data will be a `Promise` or not-a-`Promise` depending on _when_ you use it; which is really unexpected behavior and impossible to type correctly.
To make a even compacter example, this is totally valid code:
~~~ts
const data = useStorageAsync('example', 5);
await data;
data.then(() => {
  // do something
});
~~~

And although this passes all TypeScript type checks it results in `TypeError: data.then is not a function`.

As long as it _can_ be used this way without TypeScript complaining this is a possible source of pretty nasty to debug issues^^

patienry: Thank you for your explanation. Unfortunately, I haven't come up with a way to solve all the problems at once. Let's wait for the suggestions from the team members. 😂

OrbisK: Maybe we should do the oposite. Nuxt `useAsyncData` assigns the return values to the promise instead of "mocking" the promise. https://github.com/nuxt/nuxt/blob/8f4fbecd02ce50c75fc2fe83a2442b551f0dbb86/packages/nuxt/src/app/composables/asyncData.ts#L426-L429

acran: Using `Object.assign()` will probably always cause some headaches since it mutates the object itself.
But maybe just wrapping  the `data` with a `Proxy` would work.

This worked well in my tests:
~~~ts

const promiseHandlers = {
	then: promise.then.bind(promise),
	catch: promise.catch.bind(promise),
	finally: promise.finally.bind(promise),
};

return new Proxy(data, {
	get(target, prop, receiver) {
		if (Object.hasOwn(promiseHandlers, prop)) {
			return promiseHandlers[prop];
		}

		return Reflect.get(...arguments);
	},
});
~~~

So `useStorageAsync()` only returns the `data` wrapped with the additional promise functions but the `promise` will resolve unaltered `data`.

OrbisK: I wonder if it would work the other way around. But testing locally it seems that `Proxy(promise, ...)` doesnt work. 🤔

43081j: i've had a read through it, i think we should basically do the following:

```ts
let promisified: RemovableRef<T> & Promise<RemovableRef<T>>;

// bunch of logic

promisified = /* SOMEHOW create a mirror ref of `data`, explained below */
Object.assign(promisified, {
  then: /* ... */,
  catch: /* ... */
});
return promisified;
```

then where we resolve, just return `data` instead of `promisified`

the "mirror" - it can't be `data` itself since that means we're still mutating it (bad). it can't just be a new ref, since then it won't later be in sync with the original `data`.

one option is a custom ref:

```ts
promisified = customRef((track, trigger) => ({
  get() {
    track()
    return data.value
  },
  set(newValue) {
    data.value = newValue
    trigger()
  }
}));
```

otherwise, **if we really want it to be the same ref**, i'd argue we should just make this always need to be awaited. i.e. return `Promise<RemovableRef<T>>` always, instead of the union.

ill chase up on discord with the other maintainers about this too

jsjames:
Is this one close to getting resolved?  Trying to use this function and not able to because of the issue.  Thanks!

patienry:
> Is this one close to getting resolved? Trying to use this function and not able to because of the issue. Thanks!

The current changes may not be the best approach, and I also don't have much time to follow up on this issue at the moment. So I'll close this PR for now. Perhaps another contributor can address this issue in the future.

Thank you everyone for the discussion! 🙏 ❤️

PR #5205, opened 2025-12-09
fix(useMagicKeys): KeyboardEvent with .key == undefined throws an error.

koglerch13:
### Description

Autocomplete in chromium can trigger a KeyboardEvent where the `key` property is undefined (as suggested here: https://issues.chromium.org/issues/40539270 and here: https://github.com/microsoft/TypeScript/issues/59631).

In this case, (useMagicKeys/index.ts:122) throws an error:
```typescript
const depsMapKey = `${key[0].toUpperCase()}${key.slice(1)}`
```

### Additional context

Line :139 (in `updateRefs`) already suggests that an undefined value is actually anticipated:
```typescript
const key = e.key?.toLowerCase()
```

#### Fixed existing test

While fixing this, I also noticed that the test _'should handle empty key events without errors'_ probably does not work as intended:
- It does not fail the `expect().not.toThrow`
- An unhandled error is registered by vitest instead

This is fixed in this PR as well.

### DIFF
--- a/packages/core/useIntersectionObserver/demo.vue
+++ b/packages/core/useIntersectionObserver/demo.vue
@@ -1,17 +1,23 @@
 <script setup lang="ts">
 import { useIntersectionObserver } from '@vueuse/core'
-import { shallowRef, useTemplateRef } from 'vue'
+import { computed, shallowRef, useTemplateRef } from 'vue'
 
 const root = useTemplateRef('root')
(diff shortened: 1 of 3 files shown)


NoiseFan:
This pull request is a duplicate of #5225.

9romise:
Thanks for the PR! Closing since #5225 has been merged.

PR #5209, opened 2025-12-10
ensure reactivity by using router.currentRoute in useRouteParams Fixes #5160

### LINKED ISSUE #5160: BUG | `useRouteParams` | navigate back breaks the route param
### Describe the bug

When I navigate away from a page with useRouteParams and navigating back via the browsers back button the route param looses reactivity somehow.

This does not work

```
<script setup lang="ts">
const foo = useRouteParams('foo', '', { transform: String })

const { data, status } = await useAsyncData(`some-key:${foo.value}`, () =>
  GqlSomeFetch({
    foo: foo.value,
  }),
)
</script>

<template>
  <UButton to="some-route" label="Some Route" />
</template>
```

This works:

```
const route = useRoute()
const foo = computed(() => route.params.foo as string)

```

### Reproduction

WIP

### System Info

```Shell
WIP
```

### Used Package Manager

pnpm

### Validations

- [ ] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

kaiketsu927:
Fixes #5160

---

### Description

This PR fixes a reactivity loss issue in `useRouteParams` when navigating back in history (common in Nuxt or SSR scenarios).

**The Problem:**
Previously, `useRouteParams` relied on `route.params`. In some cases (like navigating back), the `route` object might become stale or lose reactivity connection, causing the params not to update.

**The Solution:**
I updated the logic to prioritize `router.currentRoute.value.params` as the source of truth when available. This ensures we are always listening to the latest state from the Router instance.

I also added a safety check `(router && router.currentRoute)` to ensure existing unit tests (which mock the router without `currentRoute`) continue to pass without errors.

### Additional context

Verified locally. The changes ensure that unit tests pass by gracefully falling back to `route.params` if `router.currentRoute` is undefined (which happens in the current test environment mocks).

### DIFF
--- a/packages/router/useRouteParams/index.ts
+++ b/packages/router/useRouteParams/index.ts
@@ -53,7 +53,8 @@ export function useRouteParams<
 
   const _paramsQueue = _queue.get(router)!
 
-  let param = route.params[name] as any
+  const getParams = () => (router && router.currentRoute) ? router.currentRoute.value.params : route.params
+  let param = getParams()[name] as any
 
   tryOnScopeDispose(() => {
     param = undefined
@@ -104,7 +105,7 @@ export function useRouteParams<
   })
 
   watch(
-    () => route.params[name],
+    () => getParams()[name],
     (v) => {
       if (param === transformGet(v as T))
         return


43081j:
do you know why the current route becomes stale?

it seems like `useRoute` should be tracking the current route already. so if you're seeing that not being the case, maybe its a bug in `useRoute` or how we're observing it?

i.e. the thing `useRoute` is observing should be `router.currentRoute`, no?

kaiketsu927:
@43081j 

I have some updates after deep verification.

To ensure this fix is valid, I created a minimal reproduction in the local Nuxt playground (specifically by setting up a dynamic route page like `pages/[id].vue` to mimic the original issue exactly).

**The Result:**
I was **unable to reproduce the bug** on the current environment (Nuxt 3.20.2). The `useRouteParams` synced correctly with the URL even after back navigation. This suggests the original reactivity issue might have been fixed upstream in recent Nuxt/Vue Router versions.

**Technical justification for merging:**
The core reason this is safer is the difference between **Mutation** and **Replacement**:
1. `useRoute()` returns a reactive proxy. If the router or framework (e.g. Nuxt during hydration) replaces the underlying route object entirely instead of mutating it, the component might hold onto a stale proxy reference.
2. `router.currentRoute` is a `Ref`. Even if the route object is replaced entirely, `router.currentRoute.value` will always point to the new instance.

By switching to `router.currentRoute`, `useRouteParams` becomes resilient to "object replacement" bugs in upstream frameworks. It bypasses the potential "stale proxy" issue entirely.

**Conclusion:**
Since the bug is not reproducible anymore, this PR technically shifts from a "Bug Fix" to a "Defensive Refactor" (switching to the router's source of truth instead of the proxy).

**Proposal:**
1. If you believe relying on `router.currentRoute` is structurally better/safer for VueUse long-term, we can merge this.
2. If you prefer to keep the code simple and rely on standard `useRoute` behavior (since it works now), feel free to close this PR.

I'll leave the final decision to you. Thanks!

43081j:
> the component might hold onto a stale proxy reference

might it?

i'm going to close this since there seems to be no evidence that this is a thing. looks like it was fixed upstream

if someone runs into this as a bug again, we can reopen

PR #5210, opened 2025-12-11
chore(useIntersectionObserver): Align rootBounds assertion to the same callback entry in useIntersectionObserver test

E66Crisp:
### Description
In it('can specify different root'), read rootBounds from the same (second) callback entry as the isIntersecting check (mock.calls[1][0][0]), ensuring both assertions refer to the same invocation and avoiding mixing data across calls.


coderwhytop:
Reorder overload 6 (custom event target with event type infer) and 7 (custom event target fallback) to ensure type inference takes higher priority during function resolution.

PR #5214, opened 2025-12-13
test: update `msw`

9romise:

<details>
<br>

>

</details>

---

### DIFF
--- a/taze.config.ts
+++ b/taze.config.ts
@@ -5,7 +5,6 @@ export default defineConfig({
   exclude: [
     'rxjs',
     'electron',
-    'msw',
   ],
   packageMode: {
     vue: 'minor',


packages/core/useFetch/index.test.ts
```diff
@@ -112,11 +112,12 @@ describe('useFetch', () => {
   it('should throw error', async () => {
     const options = { immediate: false }
     const error1 = await useFetch(`${baseUrl}?status=400`, options).execute(true).catch(err => err)
-    const error2 = await useFetch(`${baseUrl}?status=600`, options).execute(true).catch(err => err)
```
9romise, line 115: In the previous version, the error here originated from an `unhandled rejection` in the upstream `vitest`. After upgrading `msw`, this error disappeared.
Also, `600` is not a valid HTTP status code, so I changed it to `500`. See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status

taze.config.ts
```diff
@@ -5,7 +5,6 @@ export default defineConfig({
   exclude: [
     'rxjs',
     'electron',
-    'msw',
```
9romise, line 8: If there's a specific reason to exclude this from `taze`, please let me know! 💚

PR #5221, opened 2025-12-16
feat(useThrottleFn): add `flush` and `isPending` to filter system

### LINKED ISSUE #5215: No way to cancel `useThrottleFn` when scope gets disposed?
Hi there!

I was trying to see if there's a way to cancel an operation that got triggered by calling a throttled function that got created by using `useThrottleFn`.

We only use this for throttling visual stuff, so if a component gets dismounted, there's no need for the callback to run 'long' after it's no longer actually around (and visible).

As a workaround, I was thinking that a custom version using the component-/scope-aware `useTimeout` could help with?

Thanks in advance.

snowbitx:

<details>
<br>

>

</details>

---

### Description

This PR resolves the issue where a pending trailing invocation scheduled by `useThrottleFn` may still run after the current Vue scope is disposed.
The fix keeps the public API unchanged and clears the internal pending throttle timer via `tryOnScopeDispose(...)`, so trailing invocations are automatically cancelled when the scope is stopped.
### Additional context

fixed: #5215


OrbisK:
we could add `cancel`, but why are you using `useThrottleFn` outside of vue scopes?

OrbisK:
and this is not really solving https://github.com/vueuse/vueuse/issues/5215

snowbitx:
> and this is not really solving #5215

Thanks for the review, and you are right, so I changed the approach to keep the API unchanged and instead clear pending trailing invocations automatically via `tryOnScopeDispose(...) `when `useThrottleFn` is created inside a Vue scope.

9romise (commented):
I think we should leave it up to users to decide whether to cancel or not.
We just need to expose a `cancel` method, like we did in #5259.
BTW, we could probably reuse the `filter` from #5259 as well.

snowbitx:
> I think we should leave it up to users to decide whether to cancel or not. We just need to expose a `cancel` method, like we did in #5259. BTW, we could probably reuse the `filter` from #5259 as well.

Thanks for the explanation, I get it now. I’ll wait for [#5259 ](https://github.com/vueuse/vueuse/pull/5259)to be merged, then I’ll continue and reuse the `filter`.

9romise:
Hi, since #5259 has been merged, would you like to update this PR?

snowbitx:
> Hi, since #5259 has been merged, would you like to update this PR?

Thanks for the reminder. I’ve updated the implementation accordingly based on PR #5259—please take a look when you have time.

PR #5223, opened 2025-12-17
feat(useAsyncState): add `status` property for state machine pattern

### LINKED ISSUE #5019: `useAsyncState` | Add status to UseAsyncStateReturnBase

I discovered this util because Nuxt's `useFetch > useAsyncData` only works in Nuxt/Vue context and throws an error everywhere else, but I'm missing the `status` property instead of booleans `isLoading`. `isReady`.

### Suggested solution

useAsyncState should have a `status` property 
`'idle' | 'loading' | 'error' | 'ready'` for us who prefer state machine instead of booleans.

### Alternative

_No response_

### Additional context

_No response_

### Validations

xiaoxiaohuayu:
## Description

Add a `status` computed property to `useAsyncState` that provides a state machine style API for async state management.

## Motivation

Currently `useAsyncState` uses multiple boolean flags (`isLoading`, `isReady`) and `error` to represent state. While this works well, developers who prefer state machine patterns often need to derive a single status value themselves. This PR adds a built-in `status` property that returns `'idle' | 'loading' | 'success' | 'error'`.

## Changes

- Add `UseAsyncStateStatus` type: `'idle' | 'loading' | 'error' | 'success'`
- Add `status` computed property to the return value
- Add unit tests and property-based tests for the new feature

## Additional context
fixed: https://github.com/vueuse/vueuse/issues/5019

## Usage

```ts
const { status, execute } = useAsyncState(fetchData, null, { immediate: false })

// status.value is 'idle' | 'loading' | 'success' | 'error'

// Use in template
<div v-if="status === 'loading'">Loading...</div>
<div v-else-if="status === 'error'">Error occurred</div>
<div v-else-if="status === 'success'">Data loaded</div>
## Description

Add a `status` computed property to `useAsyncState` that provides a state machine style API for async state management.

## Motivation

Currently `useAsyncState` uses multiple boolean flags (`isLoading`, `isReady`) and `error` to represent state. While this works well, developers who prefer state machine patterns often need to derive a single status value themselves. This PR adds a built-in `status` property that returns `'idle' | 'loading' | 'success' | 'error'`.

## Changes

- Add `UseAsyncStateStatus` type: `'idle' | 'loading' | 'error' | 'success'`
- Add `status` computed property to the return value
- Add unit tests and property-based tests for the new feature

## Usage

```ts
const { status, execute } = useAsyncState(fetchData, null, { immediate: false })

// status.value is 'idle' | 'loading' | 'success' | 'error'

// Use in template
<div v-if="status === 'loading'">Loading...</div>
<div v-else-if="status === 'error'">Error occurred</div>
<div v-else-if="status === 'success'">Data loaded</div>


9romise (commented):
I personally don’t mind doing it this way, but I’d prefer to wait for input from other team members first.

If we do move forward, I think it would be better to make `isLoading` and `isReady` computed properties based on `status` (rather than having `status` derived from them). In that case, `isLoading` and `isReady` could be deprecated and removed in a future version.

OrbisK:
I think status is great. It aligns well with `useAsyncData` in Nuxt. 👍🏽 

> In that case, isLoading and isReady could be deprecated and removed in a future version.

👍🏽

PR #5225, opened 2025-12-21
fix(useMagicKeys): handle undefined key in keyboard events

### LINKED ISSUE #5224: BUG | `useMagicKeys ` | error when key is undefined
### Describe the bug

The branch babu-ch:fix/bab/key_default includes a fix in updateRefs: `if (key === "") return;`.
However, in my scenario, the key is undefined: clicking the submit button triggers a custom keyup event, which throws the following error:
```
Uncaught TypeError: Cannot read properties of undefined (reading '0') 
at clearDeps (index.js:4717:25) 
at updateRefs (index.js:4742:3) 
at useEventListener.passive.passive (index.js:4756:3) 
clearDeps @ index.js:4717 
updateRefs @ index.js:4742 
useEventListener.passive.passive @ index.js:4756
```
Should we take these edge cases into consideration?

### Reproduction

https://github.com/dwjohnston/vue-use-bug-use-magic-keys

### System Info

```Shell
System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
    Memory: 1.34 GB / 15.71 GB
  Binaries:
    Node: 20.19.2 - D:\nodejs\node.EXE
    Yarn: 1.22.22 - D:\nodejs\yarn.CMD
    npm: 10.8.2 - D:\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (135.0.3179.85)
    Firefox: 144.0 - C:\Program Files (x86)\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.19041.5794
  npmPackages:
    @vueuse/core: ^14.1.0 => 14.1.0
    vue: ^3.5.17 => 3.5.26
```

### Used Package Manager

yarn

### Validations

LouisLau-art:
## Description

Fixes #5224

When a custom keyup event is dispatched without a key property, the current code throws:
\`\`\`
TypeError: Cannot read properties of undefined (reading '0')
at clearDeps (index.js:4717:25)
\`\`\`

## Changes

Changed the condition from \`key === ''\` to \`!key\` to handle both empty string and undefined/null cases, preventing the TypeError when processing keyboard events with undefined keys.

## Testing

- Enhanced the existing test case to cover both empty string and undefined key scenarios
- All 12 tests pass

## Checklist

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md)
- [x] Read the [docs](https://vueuse.org/guide)
- [x] Added tests for the changes


43081j:
in what situation is the `key` set to `undefined`?

the tests don't test this because `KeyboardEvent` sets it to `""` when you don't specify it.

this seems like a sensible change just to simplify the code, but nobody seems to have a working repro of this ever being a problem.

afaict the only way to hit this is by emitting a fake event under `keyup` (i.e. something which isn't a `KeyboardEvent`)

dwjohnston:
@43081j 

```
const e = new KeyboardEvent({}); 
console.log(e.key); // ''
e.key = undefined; 
console.log(e.key) // ''

Object.defineProperty(e, "key", {value: undefined}); 
console.log(e.key); // undefined
```

dwjohnston:
@43081j  See also this comment: https://github.com/vueuse/vueuse/issues/5224#issuecomment-3726556925

When you use the browser's autocomplete functionality it occurs. Might be browser specific - I could reproduce it on Chrome and Brave, but not Firefox.

LouisLau-art:
Thanks for the review!

As mentioned by @dwjohnston, this issue is reproducible when using browser autocomplete (e.g., in Chrome/Brave) or when custom events are dispatched without a `key` property. The `!key` check covers both empty strings and `undefined`/`null` values, making the hook more robust against these edge cases.

I've added a test case to cover this scenario as requested.

OrbisK (approved):

PR #5085, opened 2025-10-03
fix(useSortable): restart instance if provided element changes

haykkh:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

Fixes #5084

PR #5228, opened 2025-12-22
fix(useSortable): fix target resolution and support dynamic elements

Codfisher:

---

### Description

This PR fixes an issue where useSortable would fail to initialize if the target element was not available immediately upon mounting (e.g., when the element is rendered conditionally using v-if).

Refactored the target resolution to use computed and watch. Instead of only initializing inside tryOnMounted, the Sortable instance is now created or destroyed reactively whenever the target element changes.

### Additional context

The previous implementation relied on tryOnMounted, which meant if the el ref was null initially (common in async data loading or conditional rendering), Sortablejs would never attach. The new watch implementation ensures robustness across component lifecycles.

I have also added a unit test to verify the fix for dynamic elements (e.g., using v-if) ensuring Sortable initializes correctly when the element appears later in the DOM.

### DIFF
--- a/packages/integrations/useSortable/index.ts
+++ b/packages/integrations/useSortable/index.ts
@@ -3,7 +3,7 @@ import type { Options } from 'sortablejs'
 import type { MaybeRef, MaybeRefOrGetter } from 'vue'
 import { defaultDocument, tryOnMounted, tryOnScopeDispose, unrefElement } from '@vueuse/core'
 import Sortable from 'sortablejs'
-import { computed, isRef, nextTick, toValue, watch } from 'vue'
+import { isRef, nextTick, toValue, watch } from 'vue'
 
 export interface UseSortableReturn {
   /**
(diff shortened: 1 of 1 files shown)


packages/integrations/useSortable/index.ts
```diff
@@ -51,11 +51,13 @@ export function useSortable<T>(
     },
   }
 
+  const target = computed(() => (typeof el === 'string' ? undefined : unrefElement(el)))
+
   const start = () => {
-    const target = (typeof el === 'string' ? document?.querySelector(el) : unrefElement(el))
-    if (!target || sortable !== undefined)
+    const targetElement = (typeof el === 'string' ? document?.querySelector(el) : target.value)
```
43081j: maybe we should just make `target` a computed of `Element | string`?

then this usage goes away as we'd unconditionally use `target.value`

Codfisher: Good call on cleaning up the start function!

I hesitated to use computed though, because if el is a static string, Vue won't track any dependencies. This could lead to a 'dead cache' where it stays null if resolved too early.

I went with a helper function instead to keep start clean but ensure we get the fresh DOM element every time.

43081j (approved):

ilyaliao:
It seems a bit similar to #5085 and #5189?

Codfisher:
Ah, I see. I missed that there was already an active PR for this. I'll close this one in favor of #5189. Thanks!

PR #5250, opened 2026-01-14
fix(useColorMode): mock localStorage for unit tests

ishechuan:
- Replace unreliable jsdom localStorage with controlled mock
- Add storageMock with getItem, setItem, removeItem, clear, length, key
- Use vi.stubGlobal to provide mock in test environment
- Fix 12 failing tests caused by localStorage.clear() not being a function

### Description

Add global localStorage mock in test setup for Node.js 25+ compatibility.

**Problem:**
- Node.js 25+ introduces an incomplete global `localStorage` object (exists but missing `clear`/`getItem`/`setItem` methods)
- This breaks jsdom's localStorage simulation, causing 40+ unit tests to fail with `TypeError: localStorage.clear is not a function`
- Affected tests: `useColorMode` (12), `useDark` (6), `useStorage` (35), etc.
- **Note:** CI uses Node.js 20/22 where this issue doesn't occur, which is why no failing tests were visible in CI

**Solution:**
- Add a conditional localStorage mock in `packages/.test/setup.ts`
- The mock only activates when native localStorage is incomplete (`typeof localStorage.clear !== 'function'`)
- Implements all Storage interface methods: `getItem`, `setItem`, `removeItem`, `clear`, `length`, `key`
- No changes to individual test files needed - the global mock handles all cases

### Additional context
- Only modifies test setup file, no production code changes
- Forward-compatible: won't affect Node.js 20/22 (CI environment)
- Fixes compatibility for developers using Node.js 25+

### DIFF
--- a/packages/.test/setup.ts
+++ b/packages/.test/setup.ts
@@ -1,2 +1,31 @@
 import './polyfillPointerEvents'
 import './mockServer'
+
+// Node.js 25+ introduces an incomplete global localStorage object
+// (exists but missing clear/getItem/setItem methods).
+// This breaks jsdom's localStorage simulation.
+// We provide a complete mock here for compatibility.
+if (
+  typeof localStorage !== 'undefined'
+  && typeof localStorage.clear !== 'function'
+) {
+  const storageState = new Map<string, string>()
+  const storageMock: Storage = {
+    getItem: (key: string) => storageState.get(key) ?? null,
+    setItem: (key: string, value: string) => {
+      storageState.set(key, value)
+    },
+    removeItem: (key: string) => {
+      storageState.delete(key)
+    },
+    clear: () => {
+      storageState.clear()
+    },
+    get length() {
+      return storageState.size
+    },
+    key: (index: number) => Array.from(storageState.keys())[index] ?? null,
+  }
+
+  globalThis.localStorage = storageMock
+}


9romise:
> Fix 12 failing unit tests in useColorMode caused by jsdom's incomplete `localStorage` implementation.

I don’t see any failing test cases in CI. Where are you seeing the 12 failed tests you mentioned?

If we do make this change, I'd recommend moving it to vitest's browser mode.

alexyunxiang:
> I don’t see any failing test cases in CI. Where are you seeing the 12 failed tests you mentioned?
> 
> If we do make this change, I'd recommend moving it to vitest's browser mode.

Thanks for the review!

**About the failing tests:**

The 12 failing tests I mentioned occur specifically with **Node.js 25+**, which introduces an incomplete global `localStorage` object (exists but missing methods like `clear()`, `getItem()`, etc.). CI uses Node.js 20/22 where this issue doesn't exist, which is why you don't see any failures there.

I've updated the PR with a different approach based on your feedback:

- Instead of modifying individual test files, I added a conditional localStorage mock in `packages/.test/setup.ts`
- The mock only activates when `typeof localStorage.clear !== 'function'` (i.e., on Node.js 25+)
- This fixes 40+ failing tests (useColorMode, useDark, useStorage, etc.) without affecting the CI environment

**About browser mode:**

I considered moving to vitest's browser mode, but the current tests heavily rely on `vi.mock('../usePreferredDark')` to simulate dark mode preference changes. Converting to browser mode would require significant refactoring of the test structure. The global setup mock is a minimal, forward-compatible solution that doesn't change any test logic.
Let me know if you'd prefer a different approach!

PR #5257, opened 2026-01-25
fix(useUrlSearchParams): Prevent "params.keys() is not iterable" in Firefox content scripts

bperel:

### Description

In the context of Firefox content scripts (for instance Firefox extensions), [`UrlSearchParams.keys()` is not iterable](https://bugzilla.mozilla.org/show_bug.cgi?id=1023984), making `useUrlSearchParams()` throw the error `TypeError: params.keys() is not iterable`.

This PR avoids the need to use `URLSearchParams.prototype.keys`. The behavior of the composable stays the same.

### DIFF
--- a/packages/core/useUrlSearchParams/index.test.ts
+++ b/packages/core/useUrlSearchParams/index.test.ts
@@ -384,4 +384,27 @@ describe('useUrlSearchParams', () => {
     window.location.hash = newHash
     expect(window.location.hash).toBe(newHash)
   })
+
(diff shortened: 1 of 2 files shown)


43081j:
this is an awkward one. we shouldn't really have to workaround this since its clearly a bug in firefox.

who knows how many places we and other libraries use iterators from various things, which don't work in firefox content scripts?

ill wait and see what other maintainers think. its an easy enough change that'll unblock you presumably, but we should also ping someone at firefox to sort this

PR #5259, opened 2026-01-26
feat(useDebounceFn): add flush and isPending to filter system

pcoterecollective:
## Summary

- Add `flush()` and `isPending` (reactive `Ref<boolean>`) to `debounceFilter` and `CancelableEventFilter`
- `useDebounceFn` stays a thin wrapper around `createFilterWrapper(debounceFilter(...), fn)`
- Return value remains a callable function (backward-compatible), with `isPending`, `cancel()`, and `flush()` as properties

### API

```ts
const debouncedFn = useDebounceFn(() => {}, 1000)

debouncedFn()               // call it directly
debouncedFn.isPending.value  // Ref<boolean> — reactive pending state
debouncedFn.cancel()         // cancel pending execution
debouncedFn.flush()          // execute pending invocation immediately
```

## Changes

- **`filters.ts`**: `CancelableEventFilter` and `CancelablePromisifyFn` now include `flush` and `isPending` (replaces `pending: boolean`). `debounceFilter` uses `shallowRef` and tracks `lastResolve` for flush support.
- **`useDebounceFn/index.ts`**: Reverted to thin wrapper delegating to `createFilterWrapper` + `debounceFilter`
- **`useDebounceFn/index.md`**: Documented `flush()`, `isPending`, and `cancel()`
- **`utils/index.test.ts`**: Updated tests for `isPending.value`, added 7 tests for `flush()` behavior

## Test plan

- [x] All unit tests pass (`pnpm vitest packages/shared/utils/index.test.ts` — 47 passed)
- [x] Typecheck passes (`pnpm typecheck`)
- [x] Flush: immediately executes pending invocation and resolves promise
- [x] Flush: no-op when nothing is pending
- [x] Flush: clears maxWait timer
- [x] isPending/flush forwarded through `createFilterWrapper`

Closes #4561


OrbisK:
Thank you for your PR🙏🏽! Some composables depend on `useDebounceFn` (e.g. `refDebounced`). Would you mind adjusting these too?

pcoterecollective:
@OrbisK Done!

OrbisK:
I think we should talk about the api. Maybe (IMO) the api should look like this:

```ts
const debouncedFn = useDebounceFn(() => {}, 1000)

debounceFn() // debounced call of the fn

debounceFn.isPending // isPending Ref
debounceFn.cancel() // cancels all pending calls
debounceFn.flush() // executes the latest pending call immediatly
``` 

inspired by: https://github.com/unjs/perfect-debounce

43081j:
im not a fan of adding custom properties to functions, but if it is generally agreed thats the way we want to go, i won't block it 👍 

these changes look good either way

OrbisK:
> these changes look good either way

Agreed. I just want to avoid any more breaking changes in future versions.

> im not a fan of adding custom properties to functions, but if it is generally agreed thats the way we want to go, i won't block it 👍

```ts
const {execute, isPending, cancel, flush}= useDebounceFn(() => {}, 1000) 
// or
const {debounce, isPending, cancel, flush}= useDebounceFn(() => {}, 1000) 

debounce() // debounced call of the fn

isPending // isPending Ref
cancel() // cancels all pending calls
flush() // executes the latest pending call immediatly
```

43081j:
either of those looks good to me and feels aligned to other functions we've turned into APIs like this in the past

OrbisK:
@pcoterecollective would you mind transfering the api to one of those?

packages/shared/useDebounceFn/index.ts
```diff
@@ -1,8 +1,22 @@
-import type { MaybeRefOrGetter } from 'vue'
+import type { ComputedRef, MaybeRefOrGetter } from 'vue'
 import type { DebounceFilterOptions, FunctionArgs, PromisifyFn } from '../utils'
+import { computed } from 'vue'
 import { createFilterWrapper, debounceFilter } from '../utils'
 
-export type UseDebounceFnReturn<T extends FunctionArgs> = PromisifyFn<T>
+export interface UseDebounceFnReturn<T extends FunctionArgs> {
```
9romise:
```suggestion
export type UseDebounceFnReturn<T extends FunctionArgs> = PromisifyFn<T> & {
```
Would you be open to this approach? https://github.com/vueuse/vueuse/pull/5259#issuecomment-3891795763
Ideally, it wouldn't introduce any breaking changes, and composables that internally rely on `useDebounceFn` wouldn’t need to be modified.

pcoterecollective:
Makes sense — I’ll rework this to return `PromisifyFn<T> & { isPending, cancel, flush }` so it stays backward-compatible. Will also add `flush` per @OrbisK’s suggestion and revert the changes to dependents since they won’t need them anymore.

API will now be:

```
const debouncedFn = useDebounceFn(() => {}, 1000)

debouncedFn()           // call it directly
debouncedFn.isPending   // Ref<boolean>
debouncedFn.cancel()    // cancel pending
debouncedFn.flush()     // execute immediately
```

Will be able to complete this on Monday

FYI @9romise @43081j

pcoterecollective:
## Why callable-with-properties instead of a returned object

Following the discussion with @OrbisK and @43081j about the API shape, here's the rationale for keeping the callable approach:

**Backward compatibility is the deciding factor.** `useDebounceFn` and `useThrottleFn` consumers (`refDebounced`, `useScroll`, `useScreenSafeArea`, etc.) all call the return value directly as a function. Switching to `{ execute, isPending, cancel, flush }` would break every single one of them.

That said, the object return pattern is the dominant convention in the codebase (48+ composables vs ~6 callable):

| Pattern | Count | Examples |
|---|---|---|
| Object with action + state | 48+ | `useAsyncState`, `useFetch`, `useTimeoutFn`, `useCounter`, ... |
| Callable with properties | ~6 | `useDebounceFn`, `useThrottleFn`, `useMemoize`, `computedWithControl`, ... |

The callable-with-properties pattern is a minority, but it's specifically the pattern that `useDebounceFn`/`useThrottleFn` and the filter system already use. Changing it now would be a breaking change with no functional benefit.

One possible future improvement: adding an `execute` self-reference property (i.e. `debouncedFn.execute === debouncedFn`) would let people who prefer destructuring do so without breaking existing consumers. That said, typing a circular self-reference like that in TypeScript is messy (recursive type alias or intersection with itself), so it may not be worth the complexity. Either way, out of scope for this PR.

rubenhazelaar:
We are eagerly anticipating the `flush()` feature. Seeing the discussion on the API shape for which has been chosen as argued by @pcoterecollective [here](https://github.com/vueuse/vueuse/pull/5259#issuecomment-3909854164), I wanted to ask:

1. Is this feature as implemented now still a breaking change (as the PR is labeled)?
2. When could we expect for this feature to be released?

If I missed anything regarding releasing or anything else, I apologize in advance :).

43081j (dismissed):

packages/shared/useDebounceFn/index.ts
```diff
@@ -12,17 +12,12 @@ export type UseDebounceFnReturn<T extends FunctionArgs> = PromisifyFn<T>
  * @param  ms          A zero-or-greater delay in milliseconds. For event callbacks, values around 100 or 250 (or even higher) are most useful.
  * @param  options     Options
  *
- * @return A new, debounce, function.
- *
- * @__NO_SIDE_EFFECTS__
```
OrbisK, line 17: are there now sideeffects?

9romise (dismissed):

OrbisK (dismissed):

OrbisK:
@pcoterecollective could you please resolve the conflicts 🙏🏽

felixdenoix:
Hi there !
This feature is really going to be super useful, thanks a lot for the PR :)

antfu (dismissed):

OrbisK (approved):

PR #5267, opened 2026-01-29
fix(useVirtualList): react to changes made in mutable arrays properly

dcherman:

---

### Description

Due to the use of `toValue` in the original watch code, the watcher that updates the list returned to the caller would not execute unless the actual value of the array changed (meaning it was treated as immutable), or if the change to the array also resulted in a change to the size of the container.

Since we already are converting the caller's inputs to a Ref, we can directly watch that and allow the caller to optimize their own code by choosing the correct ref implementation for their use case.

### DIFF
--- a/packages/core/useVirtualList/index.ts
+++ b/packages/core/useVirtualList/index.ts
@@ -1,5 +1,5 @@
 import type { ComputedRef, MaybeRef, Ref, ShallowRef, StyleValue } from 'vue'
-import { computed, ref as deepRef, shallowRef, toValue, watch } from 'vue'
+import { computed, ref as deepRef, shallowRef, watch } from 'vue'
 import { useElementSize } from '../useElementSize'
 
 type UseVirtualListItemSize = number | ((index: number) => number)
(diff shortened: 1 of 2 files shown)


dcherman:
Anything I can do to help move this along?  Thanks!

packages/core/useVirtualList/index.ts
```diff
@@ -200,8 +200,8 @@ function createGetDistance<T>(itemSize: UseVirtualListItemSize, source: UseVirtu
   }
 }
 
-function useWatchForSizes<T>(size: UseVirtualElementSizes, list: MaybeRef<readonly T[]>, containerRef: Ref<HTMLElement | null>, calculateRange: () => void) {
-  watch([size.width, size.height, () => toValue(list), containerRef], () => {
+function useWatchForSizes<T>(size: UseVirtualElementSizes, listRef: Ref<readonly T[]>, containerRef: Ref<HTMLElement | null>, calculateRange: () => void) {
```
OrbisK, line 203: I think we can leave the type as `MaybeRef`. Passing a non reactive array will not impact the watcher.

dcherman: I can revert this change if you'd like, but this is a non-exported function that will only ever receive a `ref` since the the caller's input is converted to a ref always [here](https://github.com/dcherman/vueuse/blob/96885ccf92b91213ef72cb0a4071f49cd2ee4c74/packages/core/useVirtualList/index.ts#L117).  The `source` variable will never be a non-reactive array.

OrbisK: Thats true. You are right.

PR #5284, opened 2026-02-05
fix(useRafFn): resolve reactive null fpsLimit not being handled

### LINKED ISSUE #5273: BUG | `useRafFn` | Reactive value of null not handled for fpsLimit
### Describe the bug

In useRafFn passing a null value as fpsLimit works fine to have it run at max fps.

However passing a ref which is set to null (or false, or 0) for this purpose (to later potentially limit the fps) fails. Due to this line

`return fpsLimit ? 1e3 / toValue(fpsLimit) : null;`

fpsLimit is truthy because it's a ref, so this then returns Infinity, so never runs.
Should instead check `toValue(fpsLimit) ?` most likely.

I'd be happy to submit a pull request for this, never contributed to a public repo before so new to me what the etiquette is.

### Reproduction

https://playground.vueuse.org/#eNp9Ustu20AM/BViL7EB2zq0J0cO+kqBFkVbpM1Nl82KstfeF/YhOxD07+FKluNDkNtyZkgOuezYZ+dWbUK2ZmUQXroIAWNycFcZqZ31ETpIAR94891AD423Gm4+UQaBhbAeb26vlB6bi4g0mauMsCZEiFIjbLJi9o1HXBl7nM3nJBjpxoUza5JS81soChA7brYI0cIT9wiZAOtJxEWUbQb0E/rMa35AOFp/yP0ufmezOWzuoCNoaL9quUrZxKsB6t8voKPuv6SWcZ199BmtTFmMC6FNUBBRO0VpFAGUtWxBKB7CpmKNwhNI4sNSoIlkaJ9ClM3zFB6XVAjRwO78qNhQZaxzfgI8uprq14PTNZTBcXPpEfEUl85Lzf0zZXfduM6+J5Okm8oVU73pVRZXvtmCxUDbbuR2tQ/W0J8Pq6mYsNpJhf6Pi5J+o2LrcWmZ40rZ488Biz7hYsLFDsXhDXwfThmr2F+PAX2LFbtwkfstxpG+//ebhroita2TIvU75AMGq1L2OMq+JFOT7Svd4PbHcI7SbP+H+1NEE6ahstGs7Ad9xehEv74z+qvdD6uPQx5dC+tfAJ+iE2w=

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz
    Memory: 18.28 GB / 31.85 GB
  Binaries:
    Node: 22.13.1
    Yarn: 1.22.22
    npm: 11.3.0
    bun: 1.2.0
  Browsers:
    Chrome: 144.0.7559.110
    Firefox: 147.0.2 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.26100.7309
```

### Used Package Manager

npm

### Validations

nemanjamalesija:
Fixes #5273

When you pass `fpsLimit` as a `ref(null)` (so you can toggle the limit on/off later), the callback never fires. 

The problem is that the truthiness check is on the ref object itself, which is always truthy — so `toValue()` resolves to `null` and `1000 / null` gives `Infinity`. The interval limit is never satisfied and the loop just skips every frame.

The fix is simple — unwrap with `toValue()` first, then check:

```ts
// before
return fpsLimit ? 1000 / toValue(fpsLimit) : null

// after
const limit = toValue(fpsLimit)
return limit ? 1000 / limit : null
```

Also added a test for this case. All 12 tests passing.

### DIFF
--- a/packages/core/useRafFn/index.ts
+++ b/packages/core/useRafFn/index.ts
@@ -28,7 +28,7 @@ export interface UseRafFnOptions extends ConfigurableWindow {
    * The maximum frame per second to execute the function.
    * Set to `null` to disable the limit.
    *
-   * @default undefined
+   * @default null
    */
   fpsLimit?: MaybeRefOrGetter<number | null>
   /**
@@ -49,7 +49,7 @@ export interface UseRafFnOptions extends ConfigurableWindow {
 export function useRafFn(fn: (args: UseRafFnCallbackArguments) => void, options: UseRafFnOptions = {}): Pausable {
   const {
     immediate = true,
-    fpsLimit = undefined,
+    fpsLimit = null,
     window = defaultWindow,
     once = false,
   } = options


OrbisK:
Thank you for your PR 🙏🏽!

Can you briefly explain why we should handle `ref(null)` if the types only accepts `MaybeRef<number>`?

9romise:
> Can you briefly explain why we should handle `ref(null)` if the types only accepts `MaybeRef<number>`?

Oh, I confirmed this as a bug in #5273. I hadn't noticed the type restriction before.🤦‍♂️

But if someone turns off `strictNullChecks` in their `tsconfig.json`, TypeScript won't report an error. For example:

```ts
const fpsLimit = ref<number>(null)

useRafFn(() => {

}, { fpsLimit })
```

nemanjamalesija:
You're right that the types only accept `MaybeRefOrGetter<number>`, but I think that's actually the root of the problem. The option defaults to `undefined` (no limit), so the API already supports the concept of "no limit." But if a user wants to reactively toggle between limited and unlimited at runtime, they can't express that with the current type.

As @9romise mentioned, this also surfaces with `strictNullChecks` off, but beyond that I think the type should be widened to `MaybeRefOrGetter<number | null | undefined>` to officially support the reactive toggle use case. I'll update the PR with that change.

OrbisK:
> As @9romise mentioned, this also surfaces with `strictNullChecks` off, but beyond that I think the type should be widened to `MaybeRefOrGetter<number | null | undefined>` to officially support the reactive toggle use case. I'll update the PR with that change.

Honestly, I don't think we should allow `MaybeRefOrGetter<null | undefined>`. You should always handle what you expect null to be in your code. I think @43081j might have a stronger opinion on that 😅

nemanjamalesija:
That's a fair point.

That said, I think the runtime fix itself still stands on its own. It's not really about supporting null, it's about fixing the `toValue()` ordering. When fpsLimit was made reactive in #4409, the truthiness check stayed on the raw fpsLimit parameter instead of the unwrapped value. Since a ref object is always truthy, `toValue()` then resolves to whatever the ref holds, and 1000 / null gives `Infinity`  (silently breaking the loop with no errors).

The fix just moves `toValue()`,  **before** the truthiness check.

9romise:
From my perspective, the core need here seems to be the ability to control the limit reactively (enable/disable).
See: https://github.com/vueuse/vueuse/blob/main/packages%2Fcore%2FuseRafFn%2Findex.ts#L27-L33

However, while we treat `undefined` as no limit, the type is currently restricted to `MaybeRefOrGetter<number>`, which makes it impossible to support this use case.

43081j:
change makes sense to me but i agree, i would be explicit. is `null` "off", or is `undefined` "off"? we should pick one and only accept that. in this case i'd lean towards `null` since it implies someone had to actively set the value, rather than just mistakenly passing an undefined variable.

note that this is just a typescript decision, the underlying code outside of TS will handle null _and_ undefined still

9romise:
> change makes sense to me but i agree, i would be explicit. is `null` "off", or is `undefined` "off"? we should pick one and only accept that. in this case i'd lean towards `null` since it implies someone had to actively set the value, rather than just mistakenly passing an undefined variable.
> 
> note that this is just a typescript decision, the underlying code outside of TS will handle null _and_ undefined still

I'll cast my vote for `null`, for the same reason.

packages/core/useRafFn/index.ts
9romise:
```suggestion
    fpsLimit = null,
```

nemanjamalesija: Ok. `fpsLimit` now defaults to `null`, and updated the `@default` tag to match.

PR #5286, opened 2026-02-08
feat(skills): transfer `@vueuse/skills`

serkodev:
### Description

Trasfer skills package from https://github.com/vueuse/skills

### Additional context

needs to install via [skills-npm](https://github.com/antfu/skills-npm)


antfu:
@serkodev Thanks! I think we could also have the `skills` folder at the root so it also supports the `skills` cli installation style.

Issue #5254, opened 2026-01-22
`useMounted` | Support awaitable return value


ZiuChen:
### Clear and concise description of the problem

As a developer using VueUse, I want the return value of `useMounted` to be awaitable, so that I can seamlessly pause asynchronous execution until the component is mounted, without resorting to manual watchers or workarounds. This would improve code readability and simplify asynchronous logic that depends on mount state. If the maintainers are open to this feature, I am willing to submit a PR to implement it.

### Suggested solution

In the `useMounted` function, we could extend the returned reactive boolean to also implement a `then` method (making it "thenable"), allowing it to be used with `await`. The implementation could internally resolve a promise when the component mounts, while maintaining backward compatibility for existing reactive usage.

```ts
const isMounted = useMounted()
const foo = async () => {
  // This should work after the fix
  await isMounted
  // ... logic waiting for mounted
}
```

### Alternative

An alternative would be to create a separate function like `awaitUntilMounted` that returns a promise, but this would introduce additional API surface. Extending `useMounted` itself feels more intuitive and aligned with common async/await patterns.

Now, we have to:

```ts
const isMounted = useMounted()
const foo = async () => {
  // Current workaround (cumbersome)
  const cancel = watch(isMounted, (val) => {
    // ... logic waiting for mounted
    cancel()
  })
  // Desired behavior (clean and intuitive)
  await isMounted
  // ... logic waiting for mounted
}
```

### Additional context

This feature is particularly useful in setups where asynchronous operations (e.g., fetching data, interacting with DOM elements) must wait until the component is fully mounted. Currently, developers need to write custom watchers or timeouts, which adds complexity. Making `useMounted` awaitable would provide a clean and declarative way to handle such cases.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vueuse.org/guide).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #5289, opened 2026-02-09
feat(useMounted): support awaitable return value

ZiuChen:
- Fixed: #5254

---

### Description

This PR adds awaitable support to `useMounted`, allowing developers to use `await` syntax to pause async execution until the component is mounted.

**Changes:**

- **Implementation**: Extended the return type from `ShallowRef<boolean>` to `ShallowRef<boolean> & PromiseLike<boolean>` by injecting a `then` method using `Object.assign`, following the same pattern used in `useStorageAsync`
- **Tests**: Added comprehensive test suite covering both reactive and awaitable usage
- **Documentation**: Updated the function documentation with an "Awaitable" section showing the new usage pattern

**Usage:**

```ts
const isMounted = useMounted()

// Reactive usage (backward compatible)
watch(isMounted, (val) => { /* ... */ })

// New: Awaitable usage
await isMounted
// Component is now mounted
```

**Motivations:**

Previously, developers had to write manual watchers or workarounds when async operations needed to wait for component mount:

```ts
// Before: cumbersome
const isMounted = useMounted()
const cancel = watch(isMounted, (val) => {
  // logic waiting for mounted
  cancel()
})
```

Now with awaitable support, the code is much cleaner:

```ts
// After: clean and intuitive
await useMounted()
// logic waiting for mounted
```

This change maintains full backward compatibility while providing a more ergonomic API for async scenarios.

### Additional context

- The implementation follows existing patterns in VueUse (e.g., `useStorageAsync`, `useAsyncState`, `useFetch`)
- All existing functionality remains unchanged
- When no component instance exists, the promise resolves immediately with `false`

### DIFF
--- a/packages/core/useMounted/index.md
+++ b/packages/core/useMounted/index.md
@@ -23,3 +23,18 @@ onMounted(() => {
   isMounted.value = true
 })
 ```
+
+### Awaitable
+
+The return value of `useMounted` is also a `PromiseLike`, so you can `await` it to wait until the component is mounted:
+
+```ts
+import { useMounted } from '@vueuse/core'
+
+const isMounted = useMounted()
+
+async function doSomething() {
+  await isMounted
+  // Component is now mounted, safe to access DOM etc.
+}
+```
--- a/packages/core/useMounted/index.test.ts
+++ b/packages/core/useMounted/index.test.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from 'vitest'
+import { useSetup } from '../../.test'
+import { useMounted } from './index'
+
+describe('useMounted', () => {
+  it('should be defined', () => {
+    expect(useMounted).toBeDefined()
+  })
+
+  it('should return true after mount', () => {
+    const vm = useSetup(() => {
+      const isMounted = useMounted()
+      return { isMounted }
+    })
+
+    expect(vm.isMounted).toBe(true)
+    vm.unmount()
+  })
+
+  it('should be awaitable', async () => {
+    let resolved = false
+
+    const vm = useSetup(() => {
+      const isMounted = useMounted()
+
+      ;(async () => {
+        const result = await isMounted
+        resolved = true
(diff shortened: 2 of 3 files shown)


9romise (commented):
I don’t quite see the purpose of an async `useMounted`. If async logic has to run after mount, why not just place it inside `onMounted` instead?
IMO, this feels quite different from how Vue’s reactivity is usually written.

ZiuChen:
As shown in the example in the Issue:

An asynchronous method `foo` within a component may be called at any time, and the logic of `foo` involves the component's mount state. However, the component does not want the `foo` call to be rejected when it is not mounted, so it is necessary to support "waiting for the component to be fully mounted" within this asynchronous method.

The `useMounted` return value becoming `awaitable` can make the original large block of code simpler and more elegant:

Before:
```ts
const isMounted = useMounted()

let resolve, reject;
const promise = new Promise((res, rej) => {
  resolve = res;
  reject = rej;
});

watch(isMounted, (val) => {
  val && resolve()
})

const foo = async () => {
  await promise
  // ... logic waiting for mounted
}
```

After:
```ts
const isMounted = useMounted()
const foo = async () => {
  await isMounted
  // ... logic waiting for mounted
}
```

9romise:
So why not:
```ts
const foo = async () => {
  // ... logic waiting for mounted
}

onMounted(() => {
  foo()
})
```

ZiuChen:
`foo` is not a method used to initialize the component, but rather a method that can be called at any time.

Please let me give you a more specific example from business code:

```ts
const inputRef = useTemplateRef('inputRef')
const focus = async () => {
  await isMounted
  // ... visit `inputRef` safely
}

defineExpose({
  focus
})
```

antfu:
async/await is not a common practice of VueUse / Vue Composition API. I don't think this is general enough to be included in VueUse, unfortunately. I am closing this PR. Thank you!

PR #5300, opened 2026-02-12
feat(createReusableTemplate): add support for specifying component names

wbolster:
This adds support for passing a component name to createReusableTemplate().

This is useful for development/debugging, e.g. in the component view in the Vue DevTools:

- Before: each instance is listed twice (once for the ‘define’ and once for the ‘reuse’ part) as ‘Anonymous Component’. This quickly becomes unwieldy and hard to navigate.

- After this change: components are shown in DevTools as ‘MyName.define’ / ‘MyName.reuse’ if specified, and ‘ReusableTemplate.define’ / ‘ReusableTemplate.reuse’ if no name was given, which is still more useful than ‘Anonymous Component’ (which also applies to functional components, for example).

### DIFF
--- a/packages/core/createReusableTemplate/index.ts
+++ b/packages/core/createReusableTemplate/index.ts
@@ -41,6 +41,10 @@ export interface CreateReusableTemplateOptions<Props extends Record<string, any>
    */
   inheritAttrs?: boolean
   /**
+   * Name for the reuse component (useful for devtools).
+   */
+  name?: string
+  /**
    * Props definition for reuse component.
    */
   props?: ComponentObjectPropsOptions<Props>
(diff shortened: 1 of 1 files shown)


wbolster:
example devtools component inspector pane: before (left) and after (right)

image

note: the screenshot shows `AnonymousReusableTemplate` but in reality it will show `ReusableTemplate` – the picture was taken using an earlier version of my patch that used a longer name)

wbolster:
i see this is still open after 2 weeks and multiple approvals. is there anything blocking merging this PR?

OrbisK:
> i see this is still open after 2 weeks and multiple approvals. is there anything blocking merging this PR?

we usually aim for 3 approvals before merging.

wbolster:
thanks!

PR #5305, opened 2026-02-17
test(core): migrate to browser tests

OrbisK:

<details>
<br>

>

</details>

---

### Description

This PR migrates most of the core function tests to browser tests. I have already replaced some mocks.

Follow-up Todos:

- replace more mocks
- migrate shared etc.
- replace missing jsdom tests


OrbisK:
there are warnings:

```
[@vue/compiler-core] decodeEntities option is passed but will be ignored in non-browser builds.
```

packages/core/useFocusWithin/index.test.ts
9romise, line 1: Should this also migrate to browser tests?

packages/core/onStartTyping/index.browser.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const functionKeys = range(32, 112)
+
+    onStartTyping(callBackFn)
+
+    for (let i = 0; i < arrows.length; i++) {
+      await userEvent.fill(element, String.fromCharCode(arrows[i]))
+    }
+
+    for (let i = 0; i < functionKeys.length; i++) {
+      await userEvent.fill(element, String.fromCharCode(functionKeys[i]) + String.fromCharCode(functionKeys[i]))
```
9romise: I might be missing something. Why is it added twice?

OrbisK: I think i migrated this at some point from `.keyboard` to `.fill` 

```ts
await userEvent.keyboard('{{a[[') // translates to: {, a, [
```

OrbisK:
```suggestion
      await userEvent.fill(element, String.fromCharCode(functionKeys[i]))
```

packages/core/useFocusWithin/index.test.ts
OrbisK: I think I tried, but without success.

OrbisK: planned this as follow up 😅

43081j (dismissed):
very nice work 🎉 

looks good to me.

it may be worth checking that the coverage is roughly the same too, just to be sure everything we think is running, is running.

OrbisK:
> it may be worth checking that the coverage is roughly the same too, just to be sure everything we think is running, is running.

image


Coverage looks the same!

9romise (dismissed):

43081j:
the useTransition test seems to fail in CI. i haven't read it but if we're looking at a range, maybe its a naturally flaky test and we need a better assertion?

43081j (dismissed):

9romise (dismissed):

43081j (dismissed):

9romise (dismissed):

43081j (approved):

Tofandel:
I can understand the need to add some tests to browser, or migrate a few that needed a lot of mocking to work in unit tests.

But if the files are renamed without changes, then it means those tests already supported being run in a node env. So won't the tests just run slower for those, and developers without a local playwright env won't be able to run those unit tests anymore?

Wouldn't it just make sense to leave those files as is and just add the additional index.test.browser.ts for the rest of the tests?

For example `useMediaQueries` test were entirely SSR tests with window explicitely null, so wouldn't it make more sense to have one index.test.ts for the SSR tests (as it had) and one new `index.test.browser.ts` file for the new tests that can test with the actual window.

OrbisK:
> Wouldn't it just make sense to leave those files as is and just add the additional index.test.browser.ts for the rest of the tests?

I think @43081j summarised it somewhere, but the TL;DR is that if your code is meant to run in a browser, you should test it in a browser environment. As of today, browser testing does not result in much overhead compared to happy-dom tests. https://bsky.app/profile/43081j.com/post/3m56plwodbs2y

> But if the files are renamed without changes, then it means those tests already supported being run in a node env.

Most of them were not run in a node env, but in jsdom: https://github.com/vueuse/vueuse/pull/5305/changes#diff-2ee894bf23aa44ff4ce12a3da8af19ab20180474ebf2176dbe5f2eea3f96dc92L99

The goal is to run all tests in the browser. Some of them still use mocks after the migration because it was not always straightforward. I removed or replaced the mocks when it was "easy". The migration is not yet complete, but I wanted to establish the basis as soon as possible so that we can have new tests in the browser only (if possible). The rest could be done in followups.

43081j:
> But if the files are renamed without changes, then it means those tests already supported being run in a node env. So won't the tests just run slower for those, and developers without a local playwright env won't be able to run those unit tests anymore?

They worked in node because of various shims and compat layers. Browser code should be tested against browsers - not against a mock environment (e.g. jsdom). Otherwise, we're testing our code against unrealistic environments and never against what our users actually use.

> For example useMediaQueries test were entirely SSR tests with window explicitely null, so wouldn't it make more sense to have one index.test.ts for the SSR tests (as it had) and one new index.test.browser.ts file for the new tests that can test with the actual window.

This isn't representative of the whole PR and just sounds like a basic miss. What you suggested makes sense to me, just have two test files (and we do that elsewhere already). SSR in node, rest in browser.

OrbisK:
> This isn't representative of the whole PR and just sounds like a basic miss. What you suggested makes sense to me, just have two test files (and we do that elsewhere already). SSR in node, rest in browser.

We do have `*.server.test.ts` https://github.com/vueuse/vueuse/blob/382d7e5509632c71a647485665cc14b957d2bd9c/vitest.config.ts#L93-L98

Once we can drop jsdom, we can have `*.browser.test.ts` for browser and `*.test.ts` for ssr/node. Or `*.test.ts` for browser and  `*.browser.test.ts` for ssr/node (given we mostly have browser).

PR #5315, opened 2026-03-03
feat(useLiveAnnouncer): new function

IceMooncake:

<details>
<br>

>

</details>

---

### Description

This PR introduces a new function `useLiveAnnouncer`.

This function provides an accessible way to announce messages to screen reader users via **ARIA live regions**. It is essential for notifying users about dynamic content changes (like form errors, success messages, or loading states) without moving the keyboard focus.

Features:
- **Automatic Setup**: Automatically creates and manages the necessary DOM elements (visually hidden) for live regions.
- **Polite & Assertive Modes**: Supports both `polite` (waits for the user to finish) and `assertive` (interrupts the user) announcement modes.
- **Auto Cleanup**: Automatically clears the message after a timeout to ensure the same message can be announced again if triggered consecutively.
- **Flexible**: Custom timeouts and ID prefixes.

Resolves #5314 

#### Usage Example

```ts
import { useLiveAnnouncer } from '@vueuse/core'

const { polite, assertive } = useLiveAnnouncer()

// Polite: "Saved successfully" (Will speak when the user pauses)
polite('Saved successfully')

// Assertive: "Network Error" (Will interrupt the user immediately)
assertive('Network Error')

### CODE REVIEW

File: packages/core/useLiveAnnouncer/index.ts [hunk 1]
```diff
@@ -0,0 +1,105 @@
...
+
+export interface UseLiveAnnouncerReturn {
+  announce: (message: string, mode?: 'polite' | 'assertive', timeout?: number) => void
+  polite: (message: string, timeout?: number) => void
+  assertive: (message: string, timeout?: number) => void
+}
+
+export function useLiveAnnouncer(options: UseLiveAnnouncerOptions = {}): UseLiveAnnouncerReturn {
+  const {
+    idPrefix = 'vueuse-announcer',
```
yamanoku:
```suggestion
    idPrefix = 'vueuse-live-announcer',
```
Match the wording to the composable name

IceMooncake: Thank you for your feedback. I have completed the corrections to the documentation, code, and unit tests.

yamanoku: Thank you ! 👍🏼

packages/core/useLiveAnnouncer/index.ts
```diff
@@ -0,0 +1,105 @@
...
+      container.style.height = '1px'
+      container.style.padding = '0'
+      container.style.margin = '-1px'
+      container.style.overflow = 'hidden'
+      container.style.clip = 'rect(0, 0, 0, 0)'
+      container.style.whiteSpace = 'nowrap'
+      container.style.border = '0'
+      container.style.wordWrap = 'normal'
+      container.style.clipPath = 'inset(50%)'
+      document.body.appendChild(container)
```
OrbisK, line 83: We should discuss, if we want to clean up the container, if **all** `useLiveAnnouncer` are disposed

IceMooncake: If cleaning up, components like 'Message' or 'Toast' would require developers to maintain a global useLiveAnnouncer rather than creating and disposing of it at will. This point should be added to the documentation: 'When all components using useLiveAnnouncer with the same idPrefix are destroyed, the ongoing announcement will be interrupted.

IceMooncake: I've added a commit to handle the automatic cleanup. Happy to revert it if you'd prefer to keep the container persistent. 😄

IceMooncake:
I reorganized the branches without any changes

packages/core/useLiveAnnouncer/index.ts
```diff
@@ -0,0 +1,132 @@
...
+      container.style.position = 'absolute'
+      container.style.width = '1px'
+      container.style.height = '1px'
+      container.style.padding = '0'
+      container.style.margin = '-1px'
+      container.style.overflow = 'hidden'
+      container.style.clip = 'rect(0, 0, 0, 0)'
+      container.style.whiteSpace = 'nowrap'
+      container.style.border = '0'
+      container.style.wordWrap = 'normal'
```
9romise, line 81: `clip` and `wordWrap` are deprecated.

packages/core/useLiveAnnouncer/index.ts
```diff
@@ -0,0 +1,132 @@
...
+    ensureAnnouncer()
+
+    const element = document.getElementById(`${idPrefix}-${mode}`)
+
+    if (element) {
+      element.textContent = ''
+      nextTick(() => element.textContent = message)
+
+      if (timeout && timeout > 0) {
+        setTimeout(() => {
```
9romise:
```suggestion
        window.setTimeout(() => {
```

packages/core/useLiveAnnouncer/index.browser.test.ts
9romise, line 1: Can we move to the browser test?

9romise (changes requested):
I’m not very familiar with a11y, but I noticed a few minor things.
Regarding the method name, as someone not familiar with a11y, I couldn’t immediately tell what it’s for. That said, if this is a common and appropriate term in the a11y domain, feel free to ignore my comment.

packages/core/useLiveAnnouncer/index.ts
yamanoku: Although `clip` and `word-wrap` are indeed deprecated properties today, some screen-readers still fail to detect certain elements unless these properties are present. That’s why we add them as a matter of convention — you can think of them as a kind of “incantation” or workaround.

packages/core/useLiveAnnouncer/index.ts
IceMooncake: A new commit has been pushed. Thank you for your review

packages/core/useLiveAnnouncer/index.browser.test.ts
IceMooncake: Resolved

9romise (approved):
I tested this using VoiceOver on macOS. I don't want to block this PR on my end, so I'm approving it.

One note: the `assertive` live region currently only announces once. It likely doesn’t re-trigger because the text content doesn’t change. I'm not sure if we need to update the node to force a new announcement.

antfu (approved):

PR #5321, opened 2026-03-13
fix(useIntersectionObserver): accept useTemplateRef with array type as target

### LINKED ISSUE #4712: BUG | `useIntersectionObserver` | `useTemplateRef` of array type cannot be passed as argument
### Describe the bug

Given
```
<template>
<div v-for="event in events" ref="$events"></div>
</template>

<script setup lang="ts">
...
const $events = useTemplateRef('$events');
useIntersectionObserver($events, () => {});
</script>
```
Image

where `$events` is a ref with array of elements and `ref="$events"` bound to element which has `v-for` it will mark template ref's value as `Readonly<ShallowRef<HTMLDivElement[] | null>>` which is incompatible with `MaybeComputedElementRef | MaybeRefOrGetter<MaybeElement[]> | MaybeComputedElementRef[]`.

Of course it can be mitigated by using `computed(() => $events.value ?? [])` and it will work fine, but it would propably be possible to improve the input types for the composable itself (did not have time yet to look closer into it).

For reproduction type `vue-tsc -p tsconfig.app.json` in terminal.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-cadjhx57?file=src%2Fviews%2FAbout.vue,src%2Fcomponents%2FCounter.vue,src%2Fcomponents%2FHeader.vue,src%2Fviews%2FHome.vue,src%2FApp.vue

### System Info

```Shell
Windows 11
Vue 3.5.13
VueUse 13.1.0
```

### Used Package Manager

pnpm

### Validations

isaackaara:
Fixes #4712

## Problem

When using `useTemplateRef` with a `v-for` binding in Vue 3.5+, the returned type is `Readonly<ShallowRef<HTMLDivElement[] | null>>`. Passing this to `useIntersectionObserver` fails at compile time:

```ts
const $events = useTemplateRef('$events') // Readonly<ShallowRef<HTMLDivElement[] | null>>
useIntersectionObserver($events, () => {}) // TS error: not assignable to target type
```

The error comes from the `MaybeRefOrGetter<MaybeElement[]>` variant in the target union -- it does not allow `null` as the ref's value type.

## Fix

Change `MaybeRefOrGetter<MaybeElement[]>` to `MaybeRefOrGetter<MaybeElement[] | null>` in the target parameter.

The runtime already handles null correctly: the computed getter calls `filter(notNullish)` which drops null/undefined elements before passing them to the observer.

## Change

One line change in the function signature. No logic changes.

### DIFF
--- a/packages/core/useIntersectionObserver/index.ts
+++ b/packages/core/useIntersectionObserver/index.ts
@@ -47,7 +47,7 @@ export interface UseIntersectionObserverReturn extends Supportable, Pausable {
  * @param options
  */
 export function useIntersectionObserver(
-  target: MaybeComputedElementRef | MaybeRefOrGetter<MaybeElement[]> | MaybeComputedElementRef[],
+  target: MaybeComputedElementRef | MaybeRefOrGetter<MaybeElement[] | null> | MaybeComputedElementRef[],
   callback: IntersectionObserverCallback,
   options: UseIntersectionObserverOptions = {},
 ): UseIntersectionObserverReturn {


OrbisK:
LGTM, but maybe we have more composables that should be effected by this? Maybe we can have a general solution for this?

ilyaliao:
Thanks for the PR. Closing as fixed via https://github.com/vueuse/vueuse/pull/5514

PR #5338, opened 2026-03-27
feat(useIDBKeyval)!: sync changes across tabs

Slessi:
Add `listenToStorageChanges` option to `useIDBKeyval` to sync value changes across tabs like in `useStorage`. Uses `BroadcastChannel` and not `StorageEvent`, but kept the same option name for consistency.

Useful for a project I have where I store most things in local storage, but I have one instance of `useIDBKeyval` to store webcrypto keys in an un-exportable state. I want to have the same cross-tab syncing that `useStorage` has.

<details>
<br>

>

</details>

---

### DIFF
--- a/packages/integrations/useIDBKeyval/index.ts
+++ b/packages/integrations/useIDBKeyval/index.ts
@@ -1,15 +1,16 @@
+import type { ConfigurableWindow } from '@vueuse/core'
 import type { ConfigurableFlush, RemovableRef } from '@vueuse/shared'
-import type { MaybeRefOrGetter, Ref, ShallowRef } from 'vue'
-import { watchPausable } from '@vueuse/core'
+import type { ComputedRef, MaybeRefOrGetter, Ref, ShallowRef } from 'vue'
+import { defaultWindow, tryOnScopeDispose, useSupported, watchPausable } from '@vueuse/core'
 import { del, get, set, update } from 'idb-keyval'
-import { ref as deepRef, shallowRef, toRaw, toValue } from 'vue'
+import { ref as deepRef, nextTick, shallowRef, toRaw, toValue } from 'vue'
 
 interface Serializer<T> {
   read: (raw: unknown) => T
   write: (value: T) => unknown
 }
 
-export interface UseIDBOptions<T> extends ConfigurableFlush {
+export interface UseIDBOptions<T> extends ConfigurableFlush, ConfigurableWindow {
   /**
    * Watch for deep changes
    *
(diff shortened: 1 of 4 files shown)


packages/integrations/useIDBKeyval/index.test.ts
```diff
@@ -6,6 +6,34 @@ import { useIDBKeyval } from './index'
 
 const cache = new Map<string, unknown>()
 
+// --- BroadcastChannel mock ---
+const mockChannels = new Map<string, MockBroadcastChannel>()
+
+class MockBroadcastChannel {
```
OrbisK: would you mind migrating this to an actual browser test (`*.browser.test.ts`)  maybe we can replace the mock then 🙏🏽

packages/integrations/useIDBKeyval/index.ts
```diff
@@ -41,11 +42,20 @@ export interface UseIDBOptions<T> extends ConfigurableFlush {
...
    */
   serializer?: Serializer<T>
+
+  /**
+   * Listen to storage changes from other tabs via BroadcastChannel,
+   * useful for multiple tabs applications
+   *
+   * @default true
+   */
+  listenToStorageChanges?: boolean
```
OrbisK, line 52: i think having this by default is breaking for existing projects. (so we might have to delay this to v15)

OrbisK (changes requested):

packages/integrations/useIDBKeyval/index.ts
Slessi: Yea, definitely a behaviour change. I would assume it's desirable, but might be unexpected for a minor release.

packages/integrations/useIDBKeyval/index.test.ts
Slessi: Done

packages/integrations/useIDBKeyval/index.ts
Slessi: @OrbisK what do you prefer, delaying the entire PR until v15? Or delaying only the default being `true` and merging this now with default `false`?

OrbisK: we have to discuss this change internally. We might delay this to v15. If you need this feel free to keep this branch up to date and use the prerelease below 👍🏽

43081j (approved):

packages/integrations/useIDBKeyval/index.ts
```diff
@@ -100,15 +114,21 @@ export function useIDBKeyval<T>(
 
   read()
 
-  async function write() {
+  let channel: BroadcastChannel | undefined
+
+  async function write(broadcastChange = true) {
```
OrbisK:
```
  async function write(broadcastChange = true) {
```

should we default this to `listenToStorageChanges`?

43081j: It appears we never pass it in anyway. 

I would either remove it entirely or always pass it in at the call site. Default parameters like this are a bit of a trap for bugs

Slessi: I removed it - looking again it was indeed redundant

OrbisK (approved):

9romise (approved):

ilyaliao:
@Slessi, could you update the snapshot? Thanks 🙏

Slessi:
Updated the snapshot

OrbisK (approved):

PR #5350, opened 2026-04-01
fix(useElementSize): call directive handler immediately on mount

toporek:
When using `v-element-size` with `{ box: 'border-box' }`, the handler
is never invoked because `tryOnMounted` sets refs to offsetWidth/
offsetHeight synchronously, and the ResizeObserver reports the same
border-box values — so Vue's watch never detects a change.

Add `{ immediate: true }` to the watch in `directive.ts`, consistent
with how `useElementVisibility` directive already handles this case.

Closes #5347


9romise:
Thank you for your contribution! However, as discussed in https://github.com/vueuse/vueuse/issues/5347, using { immediately: true } would cause content-box to fire twice, so this may not be the best solution. Feel free to continue the discussion in the issue, and we can implement it once we reach a better resolution. 🙏

PR #5358, opened 2026-04-08
fix(createReusableTemplate): fix the correct parameter type.

### LINKED ISSUE #5354: BUG | `createReusableTemplate` | Typed prop using component's generic param results in error
### Describe the bug

There is a type error when using a component's generic param to annotate template prop.

### Reproduction

https://playground.vueuse.org/?vueuse=14.2.1#eNp9U01v2zAM/SuEdugGBPFhO2VusW7rYTtsQxfsMvXgyIyrVpYEfaQpXP/3UrKdOm1RnUTykXx8ojp2bu1yF5GtWOmFkzaAqnRzylnwnJ1xLVtrXIAOhMMq4CVGX20UrrG1imzoYetMCydfqEj0WAjj8IRrrnGfE8O9RTgXAm1Ief8qFRFOuQY6D+CDk7qZrI56b1CtRvdn2CX0ZBJAx3aDLl+Ugn7KSxbXZTEMQKTJGIfxGKKdjwQNanRSkLkG3AfUtX/J77knSyGM9gH+f8et1AcBFpAkOZhXlPu6UmWX6BobpNErWHPdn73/8Jx1GMFkAZS13OULXY+bwjui343FSIfELsGOctKZElbQPcFH3QhczDoUxy0ynxw4mm8sQu2ru5rULAamY6GymA3AFiQ4abaVzfLGG00rliXgTJjWSoXud65Fj0L0hmZUVilz9zP7gou4mPziGsXtK/4bv08+zv449Oh29FSHWKhcg2EIX/z9Ra89C7amjorQbwQv0RsVE8cB9jXqmmjPcJntj/xFaEPX/iItlJ+GSkQTss94zuiLfHtj9Ce6H5efch4tCesfAQ1+Ob8=

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz
    Memory: 9.76 GB / 31.73 GB
  Binaries:
    Node: 24.13.1 - C:\nvm4w\nodejs\node.EXE
    npm: 11.8.0 - C:\nvm4w\nodejs\npm.CMD
    pnpm: 10.33.0 - C:\nvm4w\nodejs\pnpm.CMD
  Browsers:
    Chrome: 145.0.7632.46
    Edge: Chromium (145.0.3800.97)
  npmPackages:
    @vueuse/core: https://pkg.pr.new/@vueuse/core@5313 => 14.2.1 
    @vueuse/integrations: 14.2.1 => 14.2.1 
    @vueuse/motion: 3.0.3 => 3.0.3 
    @vueuse/nuxt: 14.2.1 => 14.2.1 
    @vueuse/router: 14.2.1 => 14.2.1 
    vue: 3.5.32 => 3.5.32
```

### Used Package Manager

pnpm

### Validations

yosong-github:

<details>
<br>

>

</details>

---

### Description

fix #5354


OrbisK:
@DamianGlowala can you check if the prerelease works for you?

DamianGlowala:
@OrbisK it does! 💯

OrbisK (approved):

43081j:
im surprised this is needed. `DefineComponent` itself defines `$props`:

https://github.com/vuejs/core/blob/9a2eb53bc03be1f4c8ea109f6bab157c54813765/packages/runtime-core/src/componentPublicInstance.ts#L307-L309

i'd like to know why we get the error in that case, before we merge this

IceMooncake:
> im surprised this is needed. `DefineComponent` itself defines `$props`:
> 
> https://github.com/vuejs/core/blob/9a2eb53bc03be1f4c8ea109f6bab157c54813765/packages/runtime-core/src/componentPublicInstance.ts#L307-L309
> 
> i'd like to know why we get the error in that case, before we merge this

https://github.com/vuejs/core/blob/9a2eb53bc03be1f4c8ea109f6bab157c54813765/packages/runtime-core/src/apiDefineComponent.ts#L44

```typescript
type ResolveProps<PropsOrPropOptions, E extends EmitsOptions> = Readonly<
  PropsOrPropOptions extends ComponentPropsOptions
    ? ExtractPropTypes<PropsOrPropOptions>
    : PropsOrPropOptions
> &
  ({} extends E ? {} : EmitsToProps<E>)
```

`{ option: T } extends ComponentPropsOptions` is true.
And then `ExtractPropTypes<{ option: T }>` --- maybe wrong?
We need `{ option: T }` be parse as it self.

https://github.com/vuejs/core/blob/9a2eb53bc03be1f4c8ea109f6bab157c54813765/packages/runtime-core/src/componentProps.ts#L143-L153

```typescript
export type ExtractPropTypes<O> = {
  [K in keyof Pick<O, RequiredKeys<O>>]: O[K] extends { default: any }
    ? Exclude<InferPropType<O[K]>, undefined>
    : InferPropType<O[K]>
}
```

`ExtractPropTypes<{ {option: T} }>` --> `InferPropType<T>`

https://github.com/vuejs/core/blob/9a2eb53bc03be1f4c8ea109f6bab157c54813765/packages/runtime-core/src/componentProps.ts#L110-L132

I haven't encountered this situation. I don't know how T will be parsed in `InferPropType`.

I'm not particularly proficient in typescript. Welcome to correct me. I am very happy to learn new knowledge from your discussion.

43081j:
i've had a deeper think about this.

its actually the generics tripping us up i think.

we can take what's in the playground and dumb it down a bunch as pure TS i think:

```ts
type Stuff = number | string;

function doStuff<T extends Stuff>(arg: T) {
  const [DefineX, ReuseX] = createReusableTemplate<{ foo: T; }>()
  h(ReuseX, {
    foo: 123 // ERROR!
  });
}
```

this errors because `T` could be `string` (and not `number`), which still satisfies `number | string` but means it can't hold the value `123`.

i.e. this means `foo: 123 as T` would force it to work (but we shouldn't do that).

the same without generics will work fine:

```ts
function doStuff(arg: number) {
  const [DefineX, ReuseX] = createReusableTemplate<{ foo: number; }>()
  h(ReuseX, {
    foo: 123 // WORKS!
  });
}
```

by removing `Binding` in the types, you're probably making the inferred props generic enough that they will unsafely accept any type of value. which would explain why the error goes away.

going back to the playground, `"awd"` isn't necessarily an `AcceptableValue` since this is a generic type. so the error is right. we'd have to not use generics, or shape the value based on `T` somehow.

IceMooncake:
> i've had a deeper think about this.
> 
> its actually the generics tripping us up i think.
> 
> we can take what's in the playground and dumb it down a bunch as pure TS i think:
> 
> ```ts
> type Stuff = number | string;
> 
> function doStuff<T extends Stuff>(arg: T) {
>   const [DefineX, ReuseX] = createReusableTemplate<{ foo: T; }>()
>   h(ReuseX, {
>     foo: 123 // ERROR!
>   });
> }
> ```
> 
> this errors because `T` could be `string` (and not `number`), which still satisfies `number | string` but means it can't hold the value `123`.
> 
> i.e. this means `foo: 123 as T` would force it to work (but we shouldn't do that).
> 
> the same without generics will work fine:
> 
> ```ts
> function doStuff(arg: number) {
>   const [DefineX, ReuseX] = createReusableTemplate<{ foo: number; }>()
>   h(ReuseX, {
>     foo: 123 // WORKS!
>   });
> }
> ```
> 
> by removing `Binding` in the types, you're probably making the inferred props generic enough that they will unsafely accept any type of value. which would explain why the error goes away.
> 
> going back to the playground, `"awd"` isn't necessarily an `AcceptableValue` since this is a generic type. so the error is right. we'd have to not use generics, or shape the value based on `T` somehow.

You are right. I have a question to ask. I would be very happy if you could help me solve the problem.

Before this PR, type error occurs in ReuseX. And after this PR, `foo: 123` is still a type error.

before pr
```ts
function doStuff(arg: number) {
  const [DefineX, ReuseX] = createReusableTemplate<{ foo: T; }>()
  h(ReuseX, { // ERROR
    foo: 123
  });
}
```

after pr
```ts
function doStuff(arg: number) {
  const [DefineX, ReuseX] = createReusableTemplate<{ foo: T; }>()
  h(ReuseX, {
    foo: 123 // ERROR (This is normal)
  });
}
```

43081j:
i think this comes down to vue's own types being very inference-driven. the typescript error we get is fairly useful and unhelpful as it just tells us "{ some big object} is not assignable to { an even bigger, crazier object }".

i'm not sure there's much we can do about that. the object as a whole errors because `{foo: 123}` as a whole isn't assignable to whatever vue has typed the props _object_ as.

i suspect if you had `createReusableTemplate<{ foo: number }>` and passed `{foo: "bar"}`, the specific property would error rather than the whole object (or at least i hope so).

but either way this is more to do with vue's own type inference than vueuse's

PR #5363, opened 2026-04-13
fix(useElementSize): call directive handler immediately on mount

mvtandas:
## Summary

Fixes #5347

The `vElementSize` directive handler was never called when using `{ box: 'border-box' }` because the initial size values were set synchronously before the `watch` was registered.

## Root Cause

1. `useElementSize` calls `tryOnMounted` which runs **synchronously** in directive context (no component lifecycle)
2. `width`/`height` refs are set to the element's `offsetWidth`/`offsetHeight` immediately
3. `watch([width, height], handler)` starts watching from those already-correct values
4. `ResizeObserver` fires with the same border-box values → no change detected → handler never called

## Fix

Added `{ immediate: true }` to the `watch` in `directive.ts` so the handler is always called with the initial size values right after mount.

```diff
- watch([width, height], ([width, height]) => handler({ width, height }))
+ watch([width, height], ([width, height]) => handler({ width, height }), { immediate: true })
```

## Test

Added a test verifying the handler is called immediately on mount. All 3 directive tests pass.

### DIFF
--- a/packages/core/useElementSize/directive.test.ts
+++ b/packages/core/useElementSize/directive.test.ts
@@ -45,6 +45,10 @@ describe('vElementSize', () => {
     it('should be defined', () => {
       expect(wrapper).toBeDefined()
     })
+
+    it('should call handler immediately on mount', () => {
+      expect(onResize).toHaveBeenCalledWith({ width: expect.any(Number), height: expect.any(Number) })
+    })
   })
 
   describe('given options', () => {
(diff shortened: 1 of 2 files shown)


mvtandas:
Friendly ping — just checking if this is ready for review. Happy to make any changes if needed!

9romise:
Thank you for your contribution! There is already a PR: #5350.

However, as discussed in #5347, using `{ immediately: true }` would cause `content-box` to fire twice, so this may not be the best solution. Feel free to continue the discussion in the issue, and we can implement it once we reach a better resolution. 🙏

mvtandas:
Thanks for the heads up @9romise\! I missed #5350. You're right that `immediate: true` would cause a double fire for content-box — closing this in favor of #5350 which handles it properly.

PR #5364, opened 2026-04-14
fix(useSortable): change UseSortableOptions from interface to type intersection

### LINKED ISSUE #5287: BUG | `useSortable` | TypeScript error: %option% does not exist in type 'UseSortableOptions'
### Describe the bug

When using `useSortable`, passing SortableJS options (e.g. `animation`) causes a type error:

`Object literal may only specify known properties, and 'animation' does not exist in type 'UseSortableOptions'.`

The root cause seems to be the changes introduced in https://github.com/vueuse/vueuse/pull/5189, where `UseSortableOptions` was defined as an interface extending `Options` and `ConfigurableDocument`. Switching to a type resolves the issue for me:

```diff
- interface UseSortableOptions extends Options, ConfigurableDocument {
+ type UseSortableOptions = Options & ConfigurableDocument & {
```

I apologize if I'm missing an obvious problem. I also don't understand why this error does not appear in https://playground.vueuse.org.
I will submit a PR with this fix if I get confirmation that this is not a bug only on my side.

### Reproduction

https://github.com/VChet/vueuse-usesortable-repro

### System Info

```Shell
System:
    OS: Windows 11 10.0.26100
    CPU: (20) x64 Intel(R) Core(TM) i9-10900 CPU @ 2.80GHz
    Memory: 20.30 GB / 31.92 GB
  Binaries:
    Node: 24.13.0 - S:\programs\nodejs\node.EXE
    npm: 11.9.0 - S:\programs\nodejs\npm.CMD
    pnpm: 10.29.1 - S:\programs\nodejs\pnpm.CMD
  Browsers:
    Internet Explorer: 11.0.26100.7309
```

### Used Package Manager

pnpm

### Validations

mvtandas:
## Summary

Fixes #5287

When using `useSortable`, passing valid SortableJS options like `animation` causes a TypeScript error:

```
'animation' does not exist in type 'UseSortableOptions'
```

## Root Cause

`interface extends` is stricter than type intersection when the base types (`Options` from SortableJS) have complex overloads or index signatures. TypeScript narrows the allowed keys more aggressively with `interface extends`.

## Fix

```diff
- export interface UseSortableOptions extends Options, ConfigurableDocument {
+ export type UseSortableOptions = Options & ConfigurableDocument & {
```

One-line change. Type intersection (`&`) preserves all properties from `Options` without the narrowing issue.

## References

This is the exact fix suggested in the issue by the reporter.

### DIFF
--- a/packages/integrations/useSortable/index.ts
+++ b/packages/integrations/useSortable/index.ts
@@ -23,7 +23,7 @@ export interface UseSortableReturn {
   option: (<K extends keyof Sortable.Options>(name: K, value: Sortable.Options[K]) => void) & (<K extends keyof Sortable.Options>(name: K) => Sortable.Options[K])
 }
 
-export interface UseSortableOptions extends Options, ConfigurableDocument {
+export type UseSortableOptions = Options & ConfigurableDocument & {
   /**
    * Watch the element reference for changes and automatically reinitialize Sortable
    * when the element changes.


43081j:
i can't reproduce this. the repro in the issue doesn't work since the el is `Ref<unknown>`, but if i fix that:

```ts
const el = useTemplateRef<HTMLElement>('el')

useSortable(el, list, {
  animation: 150
})
```

this works fine, type checks correctly, etc.

do you have an example like this which doesn't build?

mvtandas:
@43081j Thanks for checking! The issue reporter mentioned it happens specifically when passing SortableJS-specific options like `ghostClass`, `handle`, etc. — not just `animation`.

Looking at it more carefully, I think the issue may be TypeScript version-specific. The reporter was using a setup where `interface extends` with `Options` caused the narrowing.

If you can't reproduce it, it might have been fixed in a recent TypeScript or SortableJS types update. Happy to close this if it's no longer relevant!

43081j:
i tried it out in a few different versions of typescript and it seemed fine, even with `ghostClass`, `handle`, and a few others.

so im not sure why people were seeing a problem before. maybe sortable's types changed since? 🤔 either way it seems fine now

mvtandas:
Makes sense — probably a types update in SortableJS fixed it. Closing this since it's no longer reproducible. Thanks for checking\!

PR #5369, opened 2026-04-17
fix(useClipboard): prevents fail in Safari for async operation

### LINKED ISSUE #5368: BUG | `useClipboard` | fails in Safari for async operation
### Describe the bug

Unfortunately, for security reasons, Safari doesn't allow the Clipboard API to be used with async operations, so it fails without copying anything.
I've created a custom `useClipboardAsync` in [this PR in npmx](https://github.com/npmx-dev/npmx.dev/pull/2554) that handles the async call, so if interested, I could make similar changes directly in useClipboard here and remove my custom composable.

Would love to hear your thoughts on this! ❤️ 

### Reproduction

https://github.com/npmx-dev/npmx.dev/pull/2554

### System Info

```Shell
System:
  OS: macOS 26.3.1
  CPU: (10) arm64 Apple M4
  Memory: 1012.61 MB / 32.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.15.0 - /Users/matteo/.vite-plus/js_runtime/node/24.15.0/bin/node
  npm: 11.12.1 - /Users/matteo/.vite-plus/js_runtime/node/24.15.0/bin/npm
  pnpm: 10.33.0 - /Users/matteo/Library/pnpm/pnpm
  Deno: 2.0.6 - /opt/homebrew/bin/deno
Browsers:
  Brave Browser: 147.1.89.132
  Chrome: 147.0.7727.56
  Firefox: 149.0.2
  Safari: 26.3.1
npmPackages:
  @vueuse/core: 14.2.1 => 14.2.1
  @vueuse/integrations: 14.2.1 => 14.2.1
  @vueuse/nuxt: 14.2.1 => 14.2.1
  @vueuse/router: ^14.2.1 => 14.2.1
  @vueuse/shared: 14.2.1 => 14.2.1
  vue: 3.5.30 => 3.5.30
```

### Used Package Manager

pnpm

### Validations

MatteoGabriele:
### Description

resolves #5368 

### Additional context
I've ensured compatibility with the current implementation to prevent breaking changes. The copy method still accepts an optional string, continuing to default from the source. Additionally, it now can also accept a function that returns a Promise resolving to a string. Since this operation can now be asynchronous, I've added a new `copyPending` boolean to enable showing a pending state when necessary.

This is my first time contributing to Vueuse, so let me know if I'm missing anything. I'm not too familiar with the codebase.

This could be a first draft. We can improve readability as well, if needed.
We could also decide to create a completely separate composable.

Thanks again for taking the time to read this ❤️

### DIFF
--- a/packages/core/useClipboard/index.browser.test.ts
+++ b/packages/core/useClipboard/index.browser.test.ts
@@ -19,6 +19,16 @@ describe('useClipboard', () => {
       expect(copied.value).toBe(true)
     })
 
+    it('should copy text from async function', async () => {
+      const { text, copy, copied } = useClipboard()
+      expect(text.value).toBe('')
+      expect(copied.value).toBe(false)
+
+      await copy(async () => 'async text')
+      expect(text.value).toBe('async text')
+      expect(copied.value).toBe(true)
+    })
+
     it.todo('should read from legacy clipboard')
   })
 
--- a/packages/core/useClipboard/index.browser.test.ts
+++ b/packages/core/useClipboard/index.browser.test.ts
@@ -20,11 +20,15 @@ describe('useClipboard', () => {
     })
 
     it('should copy text from async function', async () => {
-      const { text, copy, copied } = useClipboard()
+      const { text, copy, copied, copyPending } = useClipboard()
       expect(text.value).toBe('')
       expect(copied.value).toBe(false)
 
-      await copy(async () => 'async text')
+      const promise = copy(async () => 'async text')
+      expect(copyPending.value).toBe(true)
+
+      await promise
+
       expect(text.value).toBe('async text')
       expect(copied.value).toBe(true)
     })
--- a/packages/core/useClipboard/index.ts
+++ b/packages/core/useClipboard/index.ts
@@ -157,7 +157,7 @@ export function useClipboard(options: UseClipboardOptions<MaybeRefOrGetter<strin
   }
 
   return {
-    copyPending,
+    copyPending: shallowReadonly(copyPending),
     isSupported,
     text: shallowReadonly(text),
     copied: shallowReadonly(copied),
--- a/packages/core/useClipboard/index.ts
+++ b/packages/core/useClipboard/index.ts
@@ -38,10 +38,15 @@ export interface UseClipboardOptions<Source> extends ConfigurableNavigator {
   legacy?: boolean
 }
 
+type ClipboardValue = string | (() => Promise<string | undefined>)
+
 export interface UseClipboardReturn<Optional> extends Supportable {
   text: Readonly<ShallowRef<string>>
   copied: Readonly<ShallowRef<boolean>>
-  copy: Optional extends true ? (text?: string) => Promise<void> : (text: string) => Promise<void>
+  copyPending: Readonly<ShallowRef<boolean>>
+  copy: Optional extends true
+    ? (text?: ClipboardValue) => Promise<void>
+    : (text: ClipboardValue) => Promise<void>
 }
 
 /**
@@ -69,6 +74,7 @@ export function useClipboard(options: UseClipboardOptions<MaybeRefOrGetter<strin
   const isSupported = computed(() => isClipboardApiSupported.value || legacy)
   const text = shallowRef('')
   const copied = shallowRef(false)
+  const copyPending = shallowRef(false)
   const timeout = useTimeoutFn(() => copied.value = false, copiedDuring, { immediate: false })
 
   async function updateText() {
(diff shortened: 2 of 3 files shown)


packages/core/useClipboard/index.ts
```diff
@@ -130,6 +157,7 @@ export function useClipboard(options: UseClipboardOptions<MaybeRefOrGetter<strin
   }
 
   return {
+    copyPending,
```
OrbisK: this should be readonly:


```suggestion
    copyPending: shallowReadonly(copyPending),
```

packages/core/useClipboard/index.ts
```diff
@@ -38,10 +38,15 @@ export interface UseClipboardOptions<Source> extends ConfigurableNavigator {
   legacy?: boolean
 }
 
+type ClipboardValue = string | (() => Promise<string | undefined>)
+
 export interface UseClipboardReturn<Optional> extends Supportable {
   text: Readonly<ShallowRef<string>>
   copied: Readonly<ShallowRef<boolean>>
-  copy: Optional extends true ? (text?: string) => Promise<void> : (text: string) => Promise<void>
+  copyPending: Readonly<ShallowRef<boolean>>
```
OrbisK, line 46: @43081j wdyt?

The non breaking `copy` function made sens to me. Because it fit into the current implementation. `copyPending` is now usefull but only used with async usage. Maybe we should better separate it?

packages/core/useClipboard/index.ts
```diff
@@ -89,26 +95,47 @@ export function useClipboard(options: UseClipboardOptions<MaybeRefOrGetter<strin
...
+    else {
+      return new ClipboardItem({
+        'text/plain': value().then((resolvedText = '') => {
+          text.value = resolvedText
+          return new Blob([resolvedText], { type: 'text/plain' })
+        }),
+      })
+    }
+  }
+
```
MatteoGabriele, line 151: To simplify this, we can wrap the string in a Promise, which reduces the amount of code. If it's a string, it becomes a Promise<string>. If it's already a Promise<string>, we just use it.

OrbisK: not sure if we should do this. The check is cheaper than a promise

packages/core/useClipboard/index.ts
MatteoGabriele: I wouldn't mind splitting it into a separate composable after having dug a bit more into the code, but that's just my 2 cents.
Usually, copy-pasting is a sync operation, and async is less frequent; that said, I still think it's needed.

43081j: i'd like to avoid two composables doing near enough the same thing.

the clipboard API is async under the hood so `copyPending` probably makes sense before this PR too even though it should never really be noticeable (i.e. most copies are fast).

i doubt many people will need an async copy but it is useful for those who do

packages/core/useClipboard/index.browser.test.ts
```diff
@@ -19,6 +19,16 @@ describe('useClipboard', () => {
       expect(copied.value).toBe(true)
     })
 
+    it('should copy text from async function', async () => {
+      const { text, copy, copied } = useClipboard()
+      expect(text.value).toBe('')
+      expect(copied.value).toBe(false)
+
+      await copy(async () => 'async text')
```
43081j: instead of an immediately resolved promise, can you pass one you manually `resolve()`? then assert that `copyPending` changes as expected

packages/core/useClipboard/index.ts
```diff
@@ -89,26 +95,47 @@ export function useClipboard(options: UseClipboardOptions<MaybeRefOrGetter<strin
...
         try {
-          await navigator!.clipboard.writeText(value)
+          copyPending.value = true
+          const clipboardItem = createClipboardItem(resolvedValue)
+          await navigator!.clipboard.write([clipboardItem])
+          copyPending.value = false
         }
         catch {
           useLegacy = true
+          copyPending.value = false
```
OrbisK: lets move this to finally.

OrbisK (commented):
This looks overall good to me.

One thing we have to discuss might be how we want to handle multiple pending async copies.

should we abort? I think `.write` does not support `AbortController` afaik.

MatteoGabriele:
@OrbisK Yeah, there's no support for abort, but we could use the pending state to avoid triggering copy again until the previous operation is done.

OrbisK:
> @OrbisK Yeah, there's no support for abort, but we could use the pending state to avoid triggering copy again until the previous operation is done.

I think we should not defer/dedupe it. I think copy should just do what it does. But I think we need to make sure that pending is true as long as all copies are not finished.

```ts
copy(simpleAsynCopy) // sets pending to true - takes 10 seconds
copy("sync value") // sets pending to true and immediate to false, but async is still pending

copyPending.value // should be true, but might be false
```

43081j:
just like your regular clipboard, new copy actions should discard the old ones i think.

```ts
const asyncValue = somehowGetAsyncValue("foo");
const syncValue = "bar";
const asyncValue2 = somehowGetAsyncValue("baz");

copy(asyncValue); // pending
copy(syncValue); // set clipboard to "bar", abandon the promise from before
copy(asyncValue2); // pending
// eventually, set clipboard to "baz" because the last promise resolved
```

MatteoGabriele:
There may be a UX issue: 
- Copying async text "foo"
- Copying sync text "bar"
If the user clicks "foo" first, then "bar", it might copy "foo" instead of "bar" due to the delay.

MatteoGabriele:
Love that we used a classic foobar action without knowing, ahahahahha
I see what you mean, though. The useClipboard instance should always get the last clicked "copy" value, regardless of which method. I'll see what I can do.

43081j:
i checked, the clipboard API writes the **last resolved** value.

so lets just do the same here and not implement our own behaviours.

this:

```ts
const asyncValue = somehowGetAsyncValue("foo");
const syncValue = "bar";
const asyncValue2 = somehowGetAsyncValue("baz");

copy(asyncValue); // pending
copy(syncValue); // set clipboard to "bar"
// asyncValue resolved, set clipboard to "foo"
copy(asyncValue2); // pending
// eventually, set clipboard to "baz" because the last promise resolved
```

OrbisK:
> i checked, the clipboard API writes the **last resolved** value.
> 
> so lets just do the same here and not implement our own behaviours.
> 
> this:
> 
> ```ts
> const asyncValue = somehowGetAsyncValue("foo");
> const syncValue = "bar";
> const asyncValue2 = somehowGetAsyncValue("baz");
> 
> copy(asyncValue); // pending
> copy(syncValue); // set clipboard to "bar"
> // asyncValue resolved, set clipboard to "foo"
> copy(asyncValue2); // pending
> // eventually, set clipboard to "baz" because the last promise resolved
> ```

Nice! We still have to implement the "dedupe" for the legacy api. It currently await the resolve, so it will use the latest resolved value. We should handle this and we are good to go.

MatteoGabriele:
@OrbisK I think this last change should do the trick

43081j (dismissed):

43081j (approved):

OrbisK (approved):
Nice! Looks great! ❤️

9romise (approved):

PR #5375, opened 2026-04-21
docs(useCloned): add tip about watch flush timing

iXueJie:

<details>
<br>

>

</details>

---

### Description

In the `useCloned` documentation, the current usage example modifies the original ref and immediately prints the cloned value synchronously. Because Vue's `watch` is asynchronous by default, it prints the old value (`'value'`). 
This might confuse some users into thinking that `useCloned` does not automatically track changes. 

To resolve this confusion, I've added a tip in the documentation to clarify the default async behavior and provided a snippet showing how to use `{ flush: 'sync' }` to get immediate updates:
```ts
const original = ref({ key: 'value' })
const { cloned } = useCloned(original, { flush: 'sync' })
original.value.key = 'some new value'
console.log(cloned.value.key) // 'some new value'
```

### DIFF
--- a/packages/core/useCloned/index.md
+++ b/packages/core/useCloned/index.md
@@ -18,6 +18,8 @@ const { cloned } = useCloned(original)
 original.value.key = 'some new value'
 
 console.log(cloned.value.key) // 'value'
+
+nextTick(() => console.log(cloned.value.key)) // 'some new value'
 ```
 
 ## Manual cloning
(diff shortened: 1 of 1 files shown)


OrbisK:
I am not sure if this is what you want. Can you tell me about the use case?

Keep in mind that you can pass `flush: 'sync'` as options. This will prevent vue from batching the changes. https://vuejs.org/guide/essentials/watchers.html#sync-watchers

TLDR: Maybe we should add a note about the flush timing instead?

iXueJie:
Thanks for the feedback! 

I initially added `nextTick` because seeing `// 'value'` right after an assignment might make users think `useCloned` isn't reactive. 

But you're right, I'll revert the `nextTick` and add a short note instead to keep it concise:

OrbisK (dismissed):

9romise (approved):

PR #5383, opened 2026-05-05
fix(useTextareaAutosize): bind textarea ref in demo

### LINKED ISSUE #5380: BUG | `useTextareaAutosize` | auto size doesn't work anymore
### Describe the bug

When typing inside the textarea, it doesn't grow anymore: https://vueuse.org/core/useTextareaAutosize

https://github.com/user-attachments/assets/b13977b9-7e8b-44df-b357-a7588e617fec

Tested in Chrome, Firefox, Edge.

### Reproduction

https://vueuse.org/core/useTextareaAutosize

### Validations

zons-zhaozhy:
## Summary

The demo for `useTextareaAutosize` only destructured `input` from the composable return value:

```diff
-const { input } = useTextareaAutosize()
+const { textarea, input } = useTextareaAutosize()
```

Without destructuring `textarea`, the template ref `ref="textarea"` had no matching script ref in `<script setup>`. The `textarea` ref inside `useTextareaAutosize()` was never populated, remaining `undefined`. This caused `triggerResize()` to return early at the `if (!textarea.value) return` guard on every call, so the textarea never resized.

## Fix

Added `textarea` to the destructuring so Vue 3's template ref binding connects the DOM element to the composable's internal ref.

Fixes #5380

### DIFF
--- a/packages/core/useTextareaAutosize/demo.vue
+++ b/packages/core/useTextareaAutosize/demo.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { useTextareaAutosize } from '@vueuse/core'
 
-const { input } = useTextareaAutosize()
+const { textarea, input } = useTextareaAutosize()
 </script>
 
 <template>


OrbisK:
Please dont open automated PRs. If you would like to contribute or solve the issue, please consider opening a PR yourself.

PR #5384, opened 2026-05-05
feat(useScrollbarWidth): reactive scrollbar width composable

### LINKED ISSUE #4262: Feature: reactive `useScrollbarWidth`

Since not every browser use absolutely positioned scrollbars (like Chrome, Opera etc.), it's sometimes needed to take care of how it affects the layout. There isn't a direct web API to receive such information too. Also when you calculate the scrollbar width when the component mounts, there is no guarantee about it will not change.

### Suggested solution

I suggest a composable that reactively reports scrollbar width. This is our implementation that we use in our projects. It may be improved to receive `MaybeRef` instead of `Ref`:

```ts
import { type Ref, ref, onMounted, onUnmounted } from "vue";

// Calculates scrollbar width for an HTML element reactively.
// This is needed for gutterly layout happening especially on Chrome & Opera.
export default function useScrollbarWidth(
  elementRef: Ref<HTMLElement | null>
): Ref<number> {
  const scrollbarWidth = ref<number>(0);
  let resizeObserver: ResizeObserver;

  function calc() {
    const scrollEl = elementRef.value!;
    scrollbarWidth.value = scrollEl.offsetWidth - scrollEl.clientWidth;
  }

  onMounted(() => {
    const element = elementRef.value!;

    // initial calculation
    calc();

    // recalculate whenever window size or zoom level changes
    window.addEventListener("resize", calc, false);

    // recalculate whenever element size changes
    resizeObserver = new ResizeObserver(calc);
    resizeObserver.observe(element);
  });

  onUnmounted(() => {
    window.removeEventListener("resize", calc);
    resizeObserver.disconnect();
  });

  return scrollbarWidth;
}
```

Such feature might've included in `useScroll` but I feel like it'd be unnecessary there since not every `useScroll` usage needs knowledge of scrollbar width.

### Alternative

* Using a scrollbar styling library (like Simplebar, OverlayScrollbars etc.) instead of what browser provides.
* `scrollbar-gutter` CSS proprety might be solve the problem in some cases.
* Using `overflow-y: scroll` which forces scrollbar to shown even if element is not overflowing.

### Additional context

_No response_

### Validations

noracoderwitch:
## Summary

Adds `useScrollbarWidth(target, options?)` — reactively reports `offsetWidth - clientWidth` of an element. Recomputes on element resize (via `useResizeObserver`) and window resize (catches zoom). Returns `0` when the target is unmounted or the browser uses overlay scrollbars (Safari, mobile).

I kept writing this snippet across a couple of Nuxt projects to compensate for the scrollbar gutter inside custom flyout layouts, and #4262 already had the same request from @nikneym, so I picked it up.

Closes #4262.

## Additional context

`window` is guarded for SSR following the `usePageLeave` / `useStorage` pattern. Unit tests cover the returned ref's shape and the null-target case — happy to add a browser test for the `ResizeObserver` wiring if you'd prefer (jsdom doesn't fire those callbacks).


packages/core/useScrollbarWidth/index.test.ts
```diff
@@ -0,0 +1,26 @@
+import { describe, expect, it } from 'vitest'
+import { isShallow, shallowRef } from 'vue'
+import { useScrollbarWidth } from './index'
+
+// jsdom doesn't fire ResizeObserver callbacks, so dynamic resize behavior
+// is intentionally not exercised here. A browser test (`index.browser.test.ts`)
+// can be added if maintainers want runtime coverage of the observer wiring.
```
OrbisK: lets write this as actual browser test. `*.browser.test.ts`

OrbisK (changes requested):

PR #5393, opened 2026-05-10
fix(useFileDialog): remove capture attribute when option is absent or falsy

### LINKED ISSUE #5391: BUG | `useFileDialog` | capture value always exist since it was set
### Describe the bug

if a capture attribute value was set, then it cannot be removed by options, we need to manually remove it with js `removeAttribute`. I hope it can be auto handled within `useFileDialog`. when capture option value is falsy, the capture attribute should be removed.

### Reproduction

[vueuse playground](https://playground.vueuse.org/#eNqdU81uEzEQfpWRL5tIbSIEp7CLyk+R4AAIOPriemcXt17b8k8IRJG4IbgicYEDT4Eq6OOE8BrM/iRtBcoBaaW1Zz5/nu+b8ZLddW4yT8hmLA/SKxchYEwOtDB1wVkMnN3hRjXO+ghLSAEfKo0PlNC2hhVU3jaQHREDZabSesy44UZaEyJUhITi+pnR+HYLqJKRUVkDyWkrylF843AGIXpl6jEsuQFQFXRhKIoCMika9CIbctBxT6xDMxoCAFK4mDzRZGjmylvToInZQZ9ejfu/J3WeCqDI35e0pGHvHUJKdJGumLiyIqmX1ERHXz7tXSTPaBOxcVpEpB1AXqo5SC1CIF+DeouHBEU0ncEtT36SYiRLtpgT60v04A5vcAZHUit5RtHBsJ0hdHzz4dOvi3e/v5xvvv7Ipz3Lf3EO+olyff5xffFt8/n9+uf3a5T5lGTQKp9eEccOaFCo55WqJ6fBGpqmzjHOpG0ckfqnru02DdNsay5nQmv7+nEXiz7h0Cg68wrl2T/ip2HRxjh75jGgnyNnu1wUvsbYp49fPMEFrXfJxpZJE3pP8jkGq1NbYw+7l0xJZV/BddU+6t4BzejLcLyIaMJWVFtoNwUdnjN6EPf3SL8s9+bk1nZ62OoPYOY0sA==)

### System Info

```Shell
--
```

### Used Package Manager

npm

### Validations

OfirHaf:
Fixes #5391

## Root cause

In `applyOptions`, the `capture` attribute was set via `el.capture = value` when the option was present, but never removed via `el.removeAttribute('capture')` when the option was absent or updated to a falsy value.

This meant that once `capture` was applied (e.g. via `useFileDialog({ capture: 'user' })` or `open({ capture: 'user' })`), subsequent calls without the `capture` option left the attribute on the hidden input element permanently.

The HTML `capture` attribute on `<input type="file">` controls whether the device camera/microphone is invoked directly. Leaving it stale changes behaviour on mobile browsers without the user expecting it.

## Fix

Replace the `el.capture = value` assignment with `setAttribute`/`removeAttribute` so the attribute is always brought in sync with the current option value:

```ts
// Before
if (hasOwn(options, 'capture'))
  el.capture = toValue(options.capture)!

// After
const capture = hasOwn(options, 'capture') ? toValue(options.capture) : undefined
if (capture)
  el.setAttribute('capture', capture)
else
  el.removeAttribute('capture')
```

## Tests added

- `should remove capture attribute when capture option is omitted` — verifies that calling `open({ capture: undefined })` after a previous `capture: 'user'` removes the attribute.
- `should remove capture attribute when capture option is set to empty string` — verifies that an explicit empty-string value also removes the attribute.

### DIFF
--- a/packages/core/useFileDialog/index.ts
+++ b/packages/core/useFileDialog/index.ts
@@ -124,8 +124,11 @@ export function useFileDialog(options: UseFileDialogOptions = {}): UseFileDialog
     el.accept = toValue(options.accept)!
     // webkitdirectory key is not stabled, maybe replaced in the future.
     el.webkitdirectory = toValue(options.directory)!
-    if (hasOwn(options, 'capture'))
-      el.capture = toValue(options.capture)!
+    const capture = hasOwn(options, 'capture') ? toValue(options.capture) : undefined
+    if (capture)
+      el.setAttribute('capture', capture)
+    else
+      el.removeAttribute('capture')
   }
 
   const open = (localOptions?: Partial<UseFileDialogOptions>) => {
--- a/packages/core/useFileDialog/index.test.ts
+++ b/packages/core/useFileDialog/index.test.ts
@@ -228,4 +228,31 @@ describe('useFileDialog', () => {
 
     expect(input.capture).toBe('environment')
   })
+
+  it('should remove capture attribute when capture option is omitted', async () => {
+    const input = document.createElement('input')
+    input.click = vi.fn()
+
+    const { open } = useFileDialog({ input, capture: 'user' })
+
+    expect(input.capture).toBe('user')
+
+    // Open without capture — the attribute should be removed
+    open({ capture: undefined })
+
+    expect(input.hasAttribute('capture')).toBe(false)
+  })
+
+  it('should remove capture attribute when capture option is set to empty string', async () => {
+    const input = document.createElement('input')
+    input.click = vi.fn()
+
+    const { open } = useFileDialog({ input, capture: 'user' })
+
+    expect(input.capture).toBe('user')
+
+    open({ capture: '' })
+
+    expect(input.hasAttribute('capture')).toBe(false)
+  })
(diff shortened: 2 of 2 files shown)


OfirHaf:
Hey there — just wanted to check in on this one. Happy to make any adjustments if there's feedback. Let me know if anything's needed from my side!

packages/core/useFileDialog/index.ts
```diff
@@ -124,8 +124,11 @@ export function useFileDialog(options: UseFileDialogOptions = {}): UseFileDialog
     el.accept = toValue(options.accept)!
     // webkitdirectory key is not stabled, maybe replaced in the future.
     el.webkitdirectory = toValue(options.directory)!
-    if (hasOwn(options, 'capture'))
-      el.capture = toValue(options.capture)!
+    const capture = hasOwn(options, 'capture') ? toValue(options.capture) : undefined
```
ilyaliao, line 127: Could we drop the `hasOwn(options, 'capture')` check here?

PR #5402, opened 2026-05-14
fix(vElementSize): call handler on mount

raashish1601:
## Summary

`vElementSize` now forwards the initial size from `useElementSize` to the directive handler by making the watcher immediate. This covers the `border-box` case where `useElementSize` can initialize from `offsetWidth`/`offsetHeight`, and the next ResizeObserver notification reports the same values so the directive handler never fires.

Fixes #5347.

## Validation

- `corepack pnpm exec vitest run --project unit packages/core/useElementSize/directive.test.ts`
- `corepack pnpm exec eslint packages/core/useElementSize/directive.ts packages/core/useElementSize/directive.test.ts`
- `corepack pnpm run typecheck`
- `corepack pnpm --filter @vueuse/core build`

### DIFF
--- a/packages/core/useElementSize/directive.test.ts
+++ b/packages/core/useElementSize/directive.test.ts
@@ -69,4 +69,35 @@ describe('vElementSize', () => {
       expect(wrapper).toBeDefined()
     })
   })
+
+  it('should call handler on mount with border-box sizes', () => {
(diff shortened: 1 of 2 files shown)


9romise:
Please check the discussion in the origin issue.

PR #5417, opened 2026-05-15
fix(useFetch): include response body in error context

### LINKED ISSUE #5405: BUG | `useFetch` | does not provide body details for a fetch error
### Describe the bug

I use useFetch instaed of normal browser fetch API because it offers some advantages and consistency for VUE apps but it seems to be missing 
support for error level details (besides status code) that I want to show in the UI as error/feedback to the user.

For example for a response of type application/problem+json I don't receive in the 
` onFetchError((fetchError)`
callaback nothing more than an object with the shape
```
{
message: string
stack: string
}

```
Is this not supported ? I would like to receive  a full Body as the API docs seems to suggest.

### Reproduction

more details if really needed

### System Info

```Shell
@vueuse/core@14.2.0
```

### Used Package Manager

npm

### Validations

ruguoba:
Fixes #5405

## Problem

When a fetch request fails (e.g., 4xx/5xx), the  callback receives an error object that only contains  (e.g., "Not Found"). The actual response body (like ) is lost.

## Root Cause

1. Response body parsing could fail silently, leaving  as null
2. The thrown Error only included , not the parsed body

## Fix

1. Added try-catch around response body parsing with fallback to 
2. The thrown error now includes  (parsed body) and  (full Response object)
3.  callback receives the parsed body in

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -487,12 +487,30 @@ export function useFetch<T>(url: MaybeRefOrGetter<string>, ...args: any[]): UseF
         response.value = fetchResponse
         statusCode.value = fetchResponse.status
 
-        responseData = await fetchResponse.clone()[config.type]()
+        try {
+          responseData = await fetchResponse.clone()[config.type]()
+        }
+        catch (error) {
+          if (!fetchResponse.ok) {
+            try {
+              responseData = await fetchResponse.clone().text()
+            }
+            catch {
+              responseData = null
+            }
+          }
+          else {
+            throw error
+          }
+        }
 
         // see: https://www.tjvantoll.com/2015/09/13/fetch-and-errors/
         if (!fetchResponse.ok) {
           data.value = initialData || null
-          throw new Error(fetchResponse.statusText)
+          const error = new Error(fetchResponse.statusText)
+          ;(error as any).data = responseData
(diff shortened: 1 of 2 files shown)


ruguoba:
@antfu Hi! This PR fixes a bug where  doesn't include the response body in error context when requests fail.

The fix is minimal (1 line change) and all CI checks are passing. Could you please review when you have a moment?

OrbisK:
Is this an automated PR? (AI)

ruguoba:
No, I use AI assisted programming

packages/core/useFetch/index.ts
```diff
@@ -487,12 +487,30 @@ export function useFetch<T>(url: MaybeRefOrGetter<string>, ...args: any[]): UseF
...
+          responseData = await fetchResponse.clone()[config.type]()
+        }
+        catch (error) {
+          if (!fetchResponse.ok) {
+            try {
+              responseData = await fetchResponse.clone().text()
+            }
+            catch {
+              responseData = null
+            }
```
OrbisK, line 500: I am not sure if we want this because it will result in a weird type of data as it has to include a string.

ruguoba: Then I'll think again to see if there are any other methods.

ruguoba:
I'll close this PR first

PR #5451, opened 2026-05-23
fix(useElementVisibility): move watchOnce outside intersection observer callback

### LINKED ISSUE #4704: BUG | `useElementVisibility` | Possible issue with `once` watcher inside of callback

Noah3521:
Closes #4704

## What's wrong

When `once: true` is passed to `useElementVisibility`, the `watchOnce` call was nested **inside** the `IntersectionObserver` callback:

```ts
// before — inside the callback
(intersectionObserverEntries) => {
  // ...update isVisible...
  if (once) {
    watchOnce(isVisible, () => {   // ← a new watcher on every callback fire
      observerController.stop()
    })
  }
}
```

This causes two problems:

1. **Multiple watchers accumulate.** Every time the browser fires an intersection event, a fresh `watchOnce` watcher is registered outside the composable's effect scope, so they pile up and never get cleaned up.
2. **`stop()` fires on the wrong change.** Because `watchOnce` is registered *after* `isVisible.value` is already updated, it waits for the *next* change instead of the current one — so the observer stops one transition too late.

## Fix

Move `watchOnce` to the composable's own scope, right after `useIntersectionObserver` is set up. This way exactly one watcher is ever created, and it fires correctly on the first `isVisible` state change:

```ts
// after — outside the callback
const observerController = useIntersectionObserver(element, (entries) => {
  // ...update isVisible only...
}, options)

if (once) {
  watchOnce(isVisible, () => {
    observerController.stop()
  })
}
```

## Tests added

Two new cases inside the existing mocked `useIntersectionObserver` block:

- **`once: true`** — verifies `stop()` is called exactly once after the first `isVisible` change, and that subsequent callbacks do not call it again.
- **`once: false` (default)** — verifies `stop()` is never called regardless of visibility changes.

### DIFF
--- a/packages/core/useElementVisibility/index.test.ts
+++ b/packages/core/useElementVisibility/index.test.ts
@@ -1,4 +1,5 @@
 import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest'
+import { nextTick } from 'vue'
 import { useElementVisibility } from './index'
 
 describe('useElementVisibility', () => {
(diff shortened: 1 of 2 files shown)


9romise:
AI-generated PRs without human review or critical thinking are not acceptable.

PR #5453, opened 2026-05-23
fix(useElementVisibility): move watchOnce outside intersection callback

### LINKED ISSUE #4704: BUG | `useElementVisibility` | Possible issue with `once` watcher inside of callback

RonGamzu:
## What

The `watchOnce` call that stops the observer when `once: true` was placed
inside the `useIntersectionObserver` callback. This meant a brand-new
watcher was registered on **every** intersection event, creating an
ever-growing list of watchers that would each try to call `stop()`.

## Fix

Move the `watchOnce` call outside the callback, so it is registered
exactly once when `useElementVisibility` initialises with `once: true`.

## Test

Added a test that verifies:
1. `stop()` is called after the first visibility change.
2. Subsequent intersection events do **not** cause additional `stop()`
   calls (i.e., no new watchers are created per callback invocation).

Closes #4704
Closes #4723

### DIFF
--- a/packages/core/useElementVisibility/index.test.ts
+++ b/packages/core/useElementVisibility/index.test.ts
@@ -129,5 +129,23 @@ describe('useElementVisibility', () => {
       useElementVisibility(el, { scrollTarget: mockScrollTarget })
       expect(vi.mocked(useIntersectionObserver).mock.lastCall?.[2]?.root).toBe(mockScrollTarget)
     })
+
+    it('stops the observer after visibility changes once when once option is true', async () => {
+      useElementVisibility(el, { once: true })
+      const stopMock = vi.mocked(useIntersectionObserver).mock.results.at(-1)?.value?.stop
+      const callback = vi.mocked(useIntersectionObserver).mock.lastCall?.[1]
+
+      expect(stopMock).not.toHaveBeenCalled()
+
+      // Trigger visibility change — observer should stop afterwards
+      callback?.([{ isIntersecting: true, time: 1 } as IntersectionObserverEntry], {} as IntersectionObserver)
+      await Promise.resolve()
+      expect(stopMock).toHaveBeenCalledTimes(1)
+
+      // Subsequent intersection events must not trigger additional stop calls
+      callback?.([{ isIntersecting: false, time: 2 } as IntersectionObserverEntry], {} as IntersectionObserver)
+      await Promise.resolve()
+      expect(stopMock).toHaveBeenCalledTimes(1)
+    })
   })
 })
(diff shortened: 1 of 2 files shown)


packages/core/useElementVisibility/index.test.ts
```diff
@@ -129,5 +129,23 @@ describe('useElementVisibility', () => {
...
+    it('stops the observer after visibility changes once when once option is true', async () => {
+      useElementVisibility(el, { once: true })
+      const stopMock = vi.mocked(useIntersectionObserver).mock.results.at(-1)?.value?.stop
+      const callback = vi.mocked(useIntersectionObserver).mock.lastCall?.[1]
+
+      expect(stopMock).not.toHaveBeenCalled()
+
+      // Trigger visibility change — observer should stop afterwards
+      callback?.([{ isIntersecting: true, time: 1 } as IntersectionObserverEntry], {} as IntersectionObserver)
+      await Promise.resolve()
```
ilyaliao, line 142: Maybe we could use `await nextTick()` here.

packages/core/useElementVisibility/index.test.ts
```diff
@@ -129,5 +129,23 @@ describe('useElementVisibility', () => {
...
+      expect(stopMock).toHaveBeenCalledTimes(1)
+
+      // Subsequent intersection events must not trigger additional stop calls
+      callback?.([{ isIntersecting: false, time: 2 } as IntersectionObserverEntry], {} as IntersectionObserver)
+      await Promise.resolve()
```
ilyaliao, line 147: As above.

packages/core/useElementVisibility/index.ts
```diff
@@ -82,12 +82,6 @@ export function useElementVisibility(
         }
       }
       isVisible.value = isIntersecting
-
-      if (once) {
-        watchOnce(isVisible, () => {
-          observerController.stop()
-        })
-      }
```
OrbisK, line 90: I am not sure if this is correct. 

I think L84 will trigger the watcher instantly if the value is not the initialValue. This might be why we initialise it afterwards.

vsergiu93: Hello, 

I just encountered an issue with this composable, and its behavior only became clear after diving into the source code.

In my project, I need to display a dialog box once two conditions are met (in any order):

1. A specific trigger element has been visible at least once (even if the user has already scrolled past it).
2. A third-party library has finished loading.

Based on the documentation, I expected that combining `initialValue: false`  and `once: true`  would fulfill the first requirement. However, this does not work as expected because the initialValue is immediately overwritten during the initial IntersectionObserver callback.

I feel like if someone passes both initialValue and once: true, the observer shouldn't just look at the very first event. It should only stop tracking after the value actually changes from whatever initialValue was set to.

Not sure if makes sense :)

PR #5471, opened 2026-05-25
fix(useWebSocket): sync status to 'CLOSED' after explicit close()

### LINKED ISSUE #5450: BUG | `useWebSocket` | close socket but state not sync
### Describe the bug

Image

Call the close method, but the state will not change to 'CLOSED'

### Reproduction

Code logic error

### System Info

```Shell
@vueuse/core: 14.3.0
```

### Used Package Manager

pnpm

### Validations

WuMingDao:

- [ - ] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<br>

>

</details>

---

#### Problem
When close() is called on an active useWebSocket connection, status is never updated to 'CLOSED' and remains stuck at 'OPEN'.

Root cause — in [packages/core/useWebSocket/index.ts](vscode-webview://0c4posqkk3ujuvkj07vbnqbcva9isbijoiag5htrgu041t7lmnu6/packages/core/useWebSocket/index.ts):

The close() function synchronously sets wsRef.value = undefined right after requesting the socket to close:

```
wsRef.value.close(code, reason)
wsRef.value = undefined
The browser later fires onclose asynchronously, where the handler uses an identity guard before updating status:


ws.onclose = (ev) => {
  if (wsRef.value === ws)
    status.value = 'CLOSED'
  ...
}
```

By the time onclose runs, wsRef.value is already undefined, so the guard fails and the status assignment is skipped. The guard itself is correct — it protects against a stale socket overwriting a newer connection during auto-reconnect — but it leaves the explicit-close path without any code that sets 'CLOSED'.

Fix
Set status.value = 'CLOSED' synchronously inside close(), so the explicit-close path doesn't depend on the async onclose handler at all:

  wsRef.value.close(code, reason)
  wsRef.value = undefined
+ status.value = 'CLOSED'
This keeps the existing onclose guard intact (auto-reconnect logic is unaffected) and aligns with the library's status machine, which only models 'OPEN' | 'CONNECTING' | 'CLOSED' (no intermediate 'CLOSING' state).

Test
Added a regression test that reproduces the original timing: capture the socket reference, call close(), then dispatch onclose against the captured reference (mirroring how a real browser fires the event after wsRef.value has already been cleared). The test fails before the fix (expected 'OPEN' to be 'CLOSED') and passes after.

### Additional context

fix: #5450

### DIFF
--- a/packages/core/useWebSocket/index.ts
+++ b/packages/core/useWebSocket/index.ts
@@ -232,6 +232,7 @@ export function useWebSocket<Data = any>(
     heartbeatPause?.()
     wsRef.value.close(code, reason)
     wsRef.value = undefined
+    status.value = 'CLOSED'
   }
 
   const send = (data: string | ArrayBuffer | Blob, useBuffer = true) => {
--- a/packages/core/useWebSocket/index.test.ts
+++ b/packages/core/useWebSocket/index.test.ts
@@ -153,6 +153,29 @@ describe('useWebSocket', () => {
 
       expect(mockWebSocket.prototype.close).toBeCalledWith(1000, undefined)
     })
+
+    it('should sync status to CLOSED after close() when onclose fires', () => {
+      vm = useSetup(() => {
+        const ref = useWebSocket('ws://localhost')
+
+        return {
+          ref,
+        }
+      })
+
+      const ws = vm.ref.ws.value
+      ws?.onopen?.(new Event('open'))
+      expect(vm.ref.status.value).toBe('OPEN')
+
+      vm.ref.close()
+
+      // Real browsers fire onclose asynchronously after WebSocket.close().
+      // Simulate that here: at this point close() has already nulled wsRef.value,
+      // so the handler must still transition status to CLOSED.
+      ws?.onclose?.(new CloseEvent('close'))
+
+      expect(vm.ref.status.value).toBe('CLOSED')
+    })
   })
 
   describe('autoClose', () => {


antfu (approved):

OrbisK (approved):

packages/core/useWebSocket/index.ts
OrbisK, line 270: I am fine with the proposed change. But why cant we just check for undefined here?


```ts
if (wsRef.value === ws || wsRef.value === undefined)
  status.value = 'CLOSED'
```

WuMingDao: This ok, I use change code run result ok. 

in this codobase playgrounds/vite/src/App.vue

```js
// App.vue

<script lang="ts" setup>
import { ref } from 'vue'
import { useWebSocket } from '@vueuse/core'

const message = ref('')

const { data, status, send, open, close } = useWebSocket('ws://localhost:3001/ws', {
  immediate: false,
  onConnected() {
    console.log('Connected!')
  },
  onDisconnected(_ws, event) {
    console.log('Disconnected!', event.code)
  },
  onError(_ws, event) {
    console.error('Error:', event)
  },
  onMessage(_ws, event) {
    console.log('Message:', event.data)
  },
})

function sendMessage() {
  if (!message.value) return
  send(message.value)
  message.value = ''
}

function closeAndSync() {
close()
</script>

<template>
  <div>
    <p>Status: {{ status }}</p>
    <p>Data: {{ data }}</p>
    <button :disabled="status === 'OPEN'" @click="() => open()">Open</button>
    <button :disabled="status === 'CLOSED'" @click="closeAndSync">Close</button>
    <div>
      <input
        v-model="message"
        :disabled="status !== 'OPEN'"
        placeholder="Type a message"
        @keyup.enter="sendMessage"
      />
      <button :disabled="status !== 'OPEN' || !message" @click="sendMessage">
        Send
      </button>
    </div>
  </div>
</template>

```
this my websocket server: https://github.com/WuMingDao/hono-ws

image
click open, status === open

image
click close, status === close

In my feel this ok, but I found close method is sync run and onclose is async run into src code.

in closeAndSync method add this code
```js
function closeAndSync() {
  console.log('1. before close:', status.value)
  close()
  console.log('2. after close:', status.value)
  setTimeout(() => {
    console.log('3. next tick:', status.value)
  }, 1)
}
```

first click open button, than click close button:
image

when user click close method, status not immediately to close, but this human not feel. 

So, I'm not sure this is worth changing — what do you think?

WuMingDao: if use I change code 
image

it can keep user click close method,  status immediately to close
image

OrbisK: not sure what you mean 😅 

I just wanted to ask if we could check for `undefined` in onclose instead. But I am also fine with the change you suggest 😄

ilyaliao (approved):

PR #5473, opened 2026-05-28
fix(useStyleTag): avoid error when multiple components share the same id

ben-lau:
### Description

`useStyleTag` injects a `<style>` element into `<head>`, and supports specifying a DOM `id` via the `id` option. When multiple components pass the same `id`, they reuse the same `<style>` element — which is the intended behavior.

When multiple components share the same `id`, two issues occur during unmount:

**1. Style loss**

Component A and Component B both use `id: 'my-style'`. When Component A unmounts, `unload()` removes `<style id="my-style">` from the DOM entirely. Component B is still active, but its styles disappear from the page.

**2. Runtime error**

After Component A removes the element, Component B's `unload()` calls `document.getElementById('my-style')` which returns `null`. Calling `removeChild(null)` throws:

```
NotFoundError: Failed to execute 'removeChild' on 'Node': parameter 1 is not of type 'Node'.
```

**Root cause**: `unload()` unconditionally removes the `<style>` element without checking if other components are still using it.

### Solution

Introduce a module-level reference counter (`Map<string, number>`) that tracks how many instances are using each style tag by `id`:

- `load()` increments the count on each successful load
- `unload()` decrements the count; only removes the `<style>` element from the DOM when the count reaches zero (i.e. the last consumer unmounts)

This ensures that in the shared `id` scenario:

- Component A unmounts → style stays in DOM, Component B is unaffected
- Component B unmounts → last consumer leaves, style is removed
- No exceptions are thrown regardless of unmount order

### Changes

- `packages/core/useStyleTag/index.ts` — add reference counting in `load`/`unload`
- `packages/core/useStyleTag/index.test.ts` — new test file with 10 tests covering basic usage, edge cases, and shared id scenarios

### Test plan

- [x] `pnpm vitest run packages/core/useStyleTag/index.test.ts --project unit` — 10/10 passed

### DIFF
--- a/packages/core/useStyleTag/index.ts
+++ b/packages/core/useStyleTag/index.ts
@@ -107,7 +107,9 @@ export function useStyleTag(
     if (!document || !isLoaded.value)
       return
     stop()
-    document.head.removeChild(document.getElementById(id) as HTMLStyleElement)
+    const el = document.getElementById(id)
+    if (el)
+      document.head.removeChild(el)
     isLoaded.value = false
   }
 
--- a/packages/core/useStyleTag/index.ts
+++ b/packages/core/useStyleTag/index.ts
@@ -48,7 +48,7 @@ export interface UseStyleTagReturn {
 }
 
 let _id = 0
-const _count = new Map<string, number>()
+const _refCount = new WeakMap<HTMLStyleElement, number>()
 
 /**
  * Inject <style> element in head.
@@ -93,7 +93,7 @@ export function useStyleTag(
     if (isLoaded.value)
       return
 
-    _count.set(id, (_count.get(id) ?? 0) + 1)
+    _refCount.set(el, (_refCount.get(el) ?? 0) + 1)
 
     stop = watch(
       cssRef,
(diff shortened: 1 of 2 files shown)


OrbisK:
how does unhead dedupes tags?

cc @harlan-zw

harlan-zw:
This isn't using Unhead :stuck_out_tongue: To answer your question though https://unhead.unjs.io/docs/head/guides/core-concepts/handling-duplicates

For a style tag, unhead would dedupe based on `key`, `id` and lastly matching content hash.

packages/core/useStyleTag/index.ts
```diff
@@ -48,6 +48,7 @@ export interface UseStyleTagReturn {
 }
 
 let _id = 0
+const _count = new Map<string, number>()
```
harlan-zw: probably safer/quicker to do a weakmap of id -> html el 

also i'd suggest avoiding sideeffects like this it would be better to provide any context into the vue instance, for example we can provide the id context to children, they can just check if it was already attached, if so it would skip (perf win)

ben-lau: fine, I change it with reference counting

9romise (approved):

PR #5474, opened 2026-05-28
fix(useAsyncState): ignore onSuccess and onError for stale executions

jynxio:

<br />
<br />
<br />
<br />

## Summary

Currently, the results of stale `execute` are not written to the Signal, yet `onSuccess` or `onError` are still being invoked. This is counterintuitive, they should be ignored as well.

Why?

1. It is more intuitive
2. `onSuccess` and `onError` themselves do not provide any indicators regarding whether they are stale. If users use them to execute side effects, a late-arriving stale request will cause race conditions
3. Both SWR and `ahooks/useRequest` ignore `onSuccess` and `onError` for stale operations

> Note: Executing side effects in `onSuccess` and `onError` (e.g., mutating Signals) is a great timing because it runs earlier than `watch` and reduces the unnecessary use of `effect`.

### DIFF
--- a/packages/core/useAsyncState/index.test.ts
+++ b/packages/core/useAsyncState/index.test.ts
@@ -129,6 +129,17 @@ describe('useAsyncState', () => {
     expect(state.value).toBe('bar')
   })
 
+  it('does not call `onSuccess` from an outdated execution', async () => {
+    const onSuccess = vi.fn()
+    const { execute } = useAsyncState((returnValue: string, timeout: number) => promiseTimeout(timeout).then(() => returnValue), '', { immediate: false, onSuccess })
+    await Promise.all([
+      execute(0, 'foo', 100),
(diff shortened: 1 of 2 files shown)


MILLERMARRU (commented):
Checked this against the current `execute()` body and the inconsistency is real: `state.value`, `isReady.value`, `error.value` and `isLoading.value` are all already guarded by `executionId === executionsCount`, but `onSuccess`/`onError` were called unconditionally outside that guard. Moving them inside just brings the callbacks in line with the contract the reactive state already had, an outdated execution shouldn't be able to touch anything, callback or otherwise.

The two new tests cover exactly the race that mattered here, two concurrent `execute()` calls where the second one resolves first, and confirm only the winning execution's callback fires. Small, focused, and consistent with how the rest of the function already treats staleness.

Issue #5475, opened 2026-05-29
`useColorMode` | Would we consider to rename this function?


oliver139:
### Clear and concise description of the problem

In Nuxt, as you may know, there is also a popular module `@nuxtjs/color-mode` which handles color mode with consideration of SSR.
However, this module uses `useColorMode()` as well, which have a conflict with VueUse that affecting the auto-imports in Nuxt.

Although it would be easy to solve by editing the Nuxt config, for example:
```ts
export default defineNuxtConfig({
 imports: {
  imports: [
    {
      name: 'useColorMode',
      from: '@nuxtjs/color-mode',
    },
    {
      name: 'useColorMode',
      as: 'useVueUseColorMode',
      from: '@vueuse/core',
    },
  ],
 },
}
```
A warning message will still be kept om the terminal:
```sh
WARN  Duplicated imports "useColorMode", the one from "@nuxtjs/color-mode" has been ignored and "/Users/olivermak/Code/nuxt4-starter/node_modules/.pnpm/@nuxtjs+color-mode@4.0.0_magicast@0.5.3/node_modules/@nuxtjs/color-mode/dist/runtime/composables" is used
```
And I believe that It may not be convenient and as well affecting the dev experience on Nuxt.

### Suggested solution

There are 3 ways we can consider:
1. Rename the function, but leads to a backward incompatible issue
2. See if there is a way to detect if user is using `@nuxtjs/color-mode` and not exporting `useColorMode()` if so
3. Keep it, and leave a note in the documentation on `useColorMode()` page

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vueuse/vueuse/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://vueuse.org/guide).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

OrbisK:
I think we should just disable the auto import in favor of nuxts import https://github.com/vueuse/vueuse/blob/39ecd5b5e9071ee2590147e14df0517b6abf3869/packages/nuxt/index.ts#L10

oliver139:
> I think we should just disable the auto import in favor of nuxts import
> 
[...]

PR #5478, opened 2026-05-29
feat(nuxt): allow user to disable VueUse functions inside `nuxt.config`

oliver139:

<details>
<br>

>

</details>

---

### Description

Inspired by #5475 

Apart from `@nuxtjs/color-mode`, we cannot tell if there would be more VueUse functions having name conflicts with any other Nuxt modules now or in the future, and it is obviously not effective to create PRs one by one for each case.

I believe that providing a simple config for users to disable certain functions inside `nuxt.config` would help.

This PR should make users able to disable function by:
```ts
// nuxt.config.ts
export default defineNuxtConfig({
  vueuse: {
    disableFunctions: [
      'useColorMode',
      'useDark',
      // Any other VueUse functions
    ],
  },
}
```

### DIFF
--- a/packages/nuxt/index.ts
+++ b/packages/nuxt/index.ts
@@ -46,6 +46,8 @@ export interface VueUseNuxtOptions {
    * @default false
    */
   ssrHandlers?: boolean
+
+  disableFunctions?: string[]
 }
 
 /**
@@ -116,6 +118,9 @@ export default defineNuxtModule<VueUseNuxtOptions>({
           if (i.package === 'shared')
             i.package = 'core'
         })
+
+        // disable the functions specified in options provided by user
+        disabledFunctions.push(...(options.disableFunctions || []))
 
         for (const pkg of packages) {
           if (pkg === 'shared')
--- a/packages/nuxt/index.ts
+++ b/packages/nuxt/index.ts
@@ -35,19 +35,21 @@ const packages = [
 
 const fullPackages = packages.map(p => `@vueuse/${p}`)
 
+interface AutoImportOptions {
+  disableFunctions?: string[]
+}
+
 export interface VueUseNuxtOptions {
   /**
    * @default true
    */
-  autoImports?: boolean
+  autoImports?: boolean | AutoImportOptions
 
   /**
    * @experimental
    * @default false
    */
   ssrHandlers?: boolean
-
-  disableFunctions?: string[]
 }
 
 /**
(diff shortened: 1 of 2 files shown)


OrbisK:
I like the idea, but I am not entirely convinced that this API is a general solution. 

FWIW auto imports can have priorities to resolve conflicts. 

Worst case the "wrong" import is auto imported and you need to import `useFoo` youself.

packages/nuxt/index.ts
```diff
@@ -46,6 +46,8 @@ export interface VueUseNuxtOptions {
    * @default false
    */
   ssrHandlers?: boolean
+
+  disableFunctions?: string[]
```
OrbisK: I we want this we should nest it into `autoImports`

oliver139: You mean so that `autoImports` would be accepting `boolean` or `object` here, right?

I actually thought this as well, just can't determine which is better.

OrbisK: i would prefer the bool + object.

oliver139:
> I like the idea, but I am not entirely convinced that this API is a general solution.
> 
> FWIW auto imports can have priorities to resolve conflicts.
> 
> Worst case the "wrong" import is auto imported and you need to import `useFoo` youself.

Umm do you mean if a project having `VueUse`, `Module A`, `Module B` and they all have a function called `useFoo()`.
Assume that I want the `useFoo()` from `Module A`, even I disabled `VueUse`'s, based on the priority, `Module B`'s may be used instead of `Module A`?

If so, yes I agree. Actually I think it seems to be some optimizations to be done on `Nuxt` instead.
But still, I believe providing a simple API from here is not harmful and would be handy for some use case.

packages/nuxt/index.ts
oliver139: > i would prefer the bool + object.

Sure, I can implement that :D

OrbisK:
> Umm do you mean if a project having `VueUse`, `Module A`, `Module B` and they all have a function called `useFoo()`.
> Assume that I want the `useFoo()` from `Module A`, even I disabled `VueUse`'s, based on the priority, `Module B`'s may be used instead of `Module A`?

I am just saying, that it is possible to add priorities to imports from the module side.  But most of the time they have the same/default priority.

OrbisK:
> But still, I believe providing a simple API from here is not harmful and would be handy for some use case.

👍🏽

oliver139:
> > Umm do you mean if a project having `VueUse`, `Module A`, `Module B` and they all have a function called `useFoo()`.
> > Assume that I want the `useFoo()` from `Module A`, even I disabled `VueUse`'s, based on the priority, `Module B`'s may be used instead of `Module A`?
> 
> I am just saying, that it is possible to add priorities to imports from the module side. But most of the time they have the same/default priority.

Yes, I know what you mean and sadly it is true that most of the time they have the same/default priority.

Anyway, I've done nesting the options, thanks for your advise👍🏻

PR #5479, opened 2026-05-29
fix(useFuse): trigger fuse ref after setCollection

### LINKED ISSUE #5477: BUG | `useFuse` | `setCollection` doesn't trigger search
### Describe the bug

When passing a reactive data source to `useFuse`, the `results` computed value does not update when the data changes, for example after an async fetch resolves.

Is this an expected behavior?

I also cannot get the repro to compile for some reason, pretty weird.

### Reproduction

https://playground.vueuse.org/?outputMode=ssr#eNp9Uk1v2zAM/SucLk4BJ07RnTInWDdkwIZhG5YCO0Q5aA5tq5MlTx9ZssD/vbRcpz0UhWFYfnyPfCJ5ZrdtOzsEZAuWu8LK1oNDH1pQQldLzrzjbMW1bFpjPZwhOPxEL3RQWtMAZ+9JTEAmtcfKCi+NJsm7ZxKL5ROd2DHKdUFED38D2hMse9IkSa5GWHps3ADnzlupq+1uNdnuiDBS+sQuKO8o+XL0NYn50kGfwplrgEb4or5V6leNeoPCFvW6af1pAaVQDtOeMmT6KhvpF3Azj1hJ6b638T4LKiZ1ocIeN4WxuABvA/WA6y4ayjLYyCYo4RGEO+kC9sILENbKg1BcU0fvZIMm+MnkCparwVf0OCMCpVrCNimNSVJIfgs7fP4nOyqQwvV8PqcyeTbMh6bRPzmp274i/QPkb6bTSz80HtBCaMkEOhClp7+hoUVNUyVsOh1Ue3mIBzrW16saT3lG30ekD54vXZ4p1JWvoetGJM9G+XjKs4spltLq0KBKWc3undG0X/HSnBWmaaVC+9hbzqi7Q0XOhFLm35eI9R2Oc4iaGos/L+D37thjnP0gT2gPtFuXmBe2Qj+E15tveKTzJdiYfVDEfiX4E51Rofc40D4EvSfbz3jR7ee45rSgd2599KjdeKneaM/sIj9u/sdXrv5k92b2Nupo+Kx7AEYKOdQ=

### Fix

I can fix this behavior client-side by doing:
```diff
const mydata = computed(() => mycustomdatahere());
const { fuse, results } = useFuse(query, mydata, { ... });

+// This
+watch(mydata, () => triggerRef(fuse));
```

### System Info

```Shell
System:
    OS: Linux 7.0 cpe:/o:nixos:nixos:26.05 26.05 (Yarara)
    CPU: (12) x64 Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz
    Memory: 5.34 GB / 15.64 GB
    Container: Yes
    Shell: 5.9 - /run/current-system/sw/bin/zsh
  Binaries:
    Node: 24.14.1 - /etc/profiles/per-user/x/bin/node
    npm: 11.11.0 - /etc/profiles/per-user/x/bin/npm
    Deno: 2.6.10 - /etc/profiles/per-user/x/bin/deno
  npmPackages:
    @vueuse/core: ^14.3.0 => 14.3.0
    @vueuse/integrations: ^14.3.0 => 14.3.0
    @vueuse/nuxt: ^14.3.0 => 14.3.0
    vue: ^3.5.34 => 3.5.34
```

### Used Package Manager

npm

### Validations

ilyaliao:
This PR fixes `useFuse` not updating `results` on data changes by calling `triggerRef(fuse)` after `setCollection`, so the `shallowRef` notifies its subscribers and `results` recomputes.

The root cause was `setCollection` mutating the Fuse instance in place without changing the reference, which a shallow ref can't detect.

Closes #5477

### DIFF
--- a/packages/integrations/useFuse/index.ts
+++ b/packages/integrations/useFuse/index.ts
@@ -30,14 +30,8 @@ export function useFuse<DataItem>(
   const fuse = shallowRef(createFuse())
 
   watch(
-    () => toValue(options)?.fuseOptions,
+    [() => toValue(data), () => toValue(options)?.fuseOptions],
     () => { fuse.value = createFuse() },
-    { deep: true },
-  )
-
-  watch(
-    () => toValue(data),
-    (newData) => { fuse.value.setCollection(newData) },
     { deep: true },
   )
 
--- a/packages/integrations/useFuse/index.ts
+++ b/packages/integrations/useFuse/index.ts
@@ -20,18 +20,16 @@ export function useFuse<DataItem>(
   data: MaybeRefOrGetter<DataItem[]>,
   options?: MaybeRefOrGetter<UseFuseOptions<DataItem>>,
 ): UseFuseReturn<DataItem> {
-  const createFuse = () => {
-    return new Fuse(
-      toValue(data) ?? [],
-      toValue(options)?.fuseOptions,
-    )
-  }
+  const createFuse = (
+    d = toValue(data) ?? [],
+    fuseOpts = toValue(options)?.fuseOptions,
+  ) => new Fuse(d, fuseOpts)
 
   const fuse = shallowRef(createFuse())
 
   watch(
     [() => toValue(data), () => toValue(options)?.fuseOptions],
-    () => { fuse.value = createFuse() },
+    ([newData, newFuseOptions]) => { fuse.value = createFuse(newData ?? [], newFuseOptions) },
     { deep: true },
   )
 
--- a/packages/integrations/useFuse/index.ts
+++ b/packages/integrations/useFuse/index.ts
@@ -1,7 +1,7 @@
 import type { FuseResult, IFuseOptions } from 'fuse.js'
 import type { ComputedRef, MaybeRefOrGetter, Ref } from 'vue'
 import Fuse from 'fuse.js'
-import { computed, shallowRef, toValue, watch } from 'vue'
+import { computed, shallowRef, toValue, triggerRef, watch } from 'vue'
 
 export type FuseOptions<T> = IFuseOptions<T>
 export interface UseFuseOptions<T> {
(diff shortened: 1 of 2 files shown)


packages/integrations/useFuse/index.ts
```diff
@@ -30,17 +30,11 @@ export function useFuse<DataItem>(
   const fuse = shallowRef(createFuse())
 
   watch(
-    () => toValue(options)?.fuseOptions,
+    [() => toValue(data), () => toValue(options)?.fuseOptions],
     () => { fuse.value = createFuse() },
     { deep: true },
```
OrbisK, line 44:
```ts
    [() => toValue(data), () => toValue(options)?.fuseOptions],
([newData, newOptions]) => { fuse.value = createFuse(newData, newOptions) },
{ deep: true },
```
(suggestions are broken (?))

packages/integrations/useFuse/index.ts
OrbisK:
```ts
  const createFuse = (d = toValue(data), o = toValue(options)) => {
```

OrbisK (commented):
I think i would prefer this to have the correct values for the current timing.

packages/integrations/useFuse/index.ts
```diff
@@ -20,24 +20,16 @@ export function useFuse<DataItem>(
...
-    () => toValue(options)?.fuseOptions,
-    () => { fuse.value = createFuse() },
-    { deep: true },
-  )
-
-  watch(
-    () => toValue(data),
-    (newData) => { fuse.value.setCollection(newData) },
+    [() => toValue(data), () => toValue(options)?.fuseOptions],
+    ([newData, newFuseOptions]) => { fuse.value = createFuse(newData ?? [], newFuseOptions) },
```
OrbisK: I am not very familiar with fuse. Will this impact performance? If only data changes we are always recreating, intead of setting the collection right?

ilyaliao: Yes, it does have some performance impact. I went back and forth on this, and I'm thinking adding `triggerRef` might be the better approach here. wdyt?

ilyaliao: If you're fine with this direction, I think we can just go with #5494.

OrbisK: #5494 looks automated. but we could try the approach

OrbisK (approved):

antfu (approved):

PR #5493, opened 2026-06-04
fix(useAsyncState): setting state.value directly cancels in-flight execution

### LINKED ISSUE #4975: BUG | `useAsyncState` | setting `state` should reject pending promise
### Describe the bug

I would like to use this composable in context where the state might be immediately known, but the promise has already been executed. Therefore, setting the `state` directly should stop the pending promise from overriding the `state` one it has loaded.

```js
import { describe, test } from 'vitest';
import { promiseTimeout, useAsyncState } from '@vueuse/core';

describe('useLazyAsyncState', () => {

  test('can reject pending promise by setting state', ({ expect }) => {
    const { state, isLoading } = useAsyncState(async () => {
      await promiseTimeout(1000);
      return 10;
    }, undefined, { immediate: false });

    // Process 1 will request resolution of state.
    void executeImmediate();

    // Process 2 will determines that it knows the state.
    state.value = 500;
    expect(isLoading.value).toBe(false);

    await promiseTimeout(1000);
    expect(state.value).toBe(500);
  });

});
```

### Reproduction

https://playground.vueuse.org/#eNp9Ustu2zAQ/JUFL3EAxXKR9pLKQdMihxR9ocmRF5ZaO3QpUuDDliHo37OiHnGCIDwtd4azs8tt2U1dL/cR2RUrvHSqDuAxxBq0MNs1Z8Fzds2NqmrrArRQO1spjw+qQhtDBtHjjT8aeR9EQOhgQzicfSFFQnJpHZ6dvLbmp40mYDkziUgEbqQ1PkCJWhxhDZer1erzc7oF38tnoPwPK0plthlggzIGvKsqLNVQe/3SzUL0ISzOYX0NLTcA4iBUeNXCItU8p2oAjjp3Bj6k4l0Gqz7Nzex6caKVHC33Qkekwp+GJwO/yIdJ0tzoErCqNXHpBlCUag9SC+9pthuNDSjC/YVE0newiz6ozXG6Hi5ICNHA4xikvwA6vc4YjlauoB3HBF1X/HOQz/g8tcSZb8QbxfJJbYqK/MQ1y2gL6CM2arvceWtoVdIEOJO2qpVG97sOij6KMyowSHImtLaH7ykXXMRsystHlP/fyO980+c4++PQo9sjZzMWhNtiGODb+1/YUDyDlS2jJvY74F/0Vsfe40D7Gk1Jtk94ye1dWlOazIO/bQIaPzXVG+2ZXeJzRkv77Z3Wn+1eLj+md7QarHsCLN8mPA==

### System Info

```Shell
System:
    OS: macOS 15.6
    CPU: (14) arm64 Apple M4 Pro
    Memory: 85.44 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.17.0 - /usr/local/bin/node
    npm: 10.9.2 - /usr/local/bin/npm
    pnpm: 10.14.0 - /opt/homebrew/bin/pnpm
  Browsers:
    Brave Browser: 138.1.80.120
    Chrome: 139.0.7258.128
    Safari: 18.6
```

### Used Package Manager

npm

### Validations

LeSingh1:
Fixes #4975

## What was wrong

When a caller sets `state.value` directly while a promise is still loading, the resolved value would silently overwrite the assignment. The root cause is that `execute()` uses an `executionId === executionsCount` guard to detect stale executions, but an external write to `state.value` never bumped `executionsCount`, so the in-flight promise always believed it was still the latest execution.

## What changed

`state` is now created with `customRef` instead of `shallowRef`/`deepRef`. The setter intercepts every write:

- Writes from inside `execute()` are marked with an `_internalWrite` flag, so they pass through without side effects.
- Writes from outside `execute()` bump `executionsCount`, which causes the in-flight promise to see `executionId !== executionsCount` and skip the state/isReady/isLoading updates. `isLoading` and `isReady` are also reset immediately so the caller gets a consistent snapshot right away.

The `shallow` option is still respected: when `shallow` is false, incoming values are passed through `deepRef().value` to get the same deep-reactive unwrapping as before.

## Tests

Added a test that starts a 100 ms promise, sets `state.value = 42` before it resolves, then waits for the promise to finish and asserts `state.value` is still `42` and `isLoading` is `false`.

The existing test that checked `initialState === state` (reference equality) was updated to check the actual value instead, since `customRef` intentionally does not return the same reference as a passed-in `shallowRef`.

### DIFF
--- a/packages/core/useAsyncState/index.test.ts
+++ b/packages/core/useAsyncState/index.test.ts
@@ -117,7 +117,27 @@ describe('useAsyncState', () => {
     const { state } = useAsyncState(asyncValue, initialState)
     await asyncValue
     expect(state.value).toBe(100)
-    expect(initialState).toBe(state)
+    // initialState.value reflects the resolved value (state reads from _source)
+    expect(initialState.value).toBe(200)
+  })
+
+  it('setting state.value directly cancels in-flight execution', async () => {
+    const { state, isLoading, isReady } = useAsyncState(
+      () => promiseTimeout(100).then(() => 99),
+      0,
+      { immediate: true },
+    )
+
+    // Override state before the promise resolves
+    state.value = 42
+
+    await promiseTimeout(150)
+
+    // The promise resolved with 99, but the manual write should have won
+    expect(state.value).toBe(42)
+    expect(isLoading.value).toBe(false)
+    // isReady stays false because execute() detected it was cancelled
+    expect(isReady.value).toBe(false)
   })
 
   it('does not set `state` from an outdated execution', async () => {
--- a/packages/core/useAsyncState/index.ts
+++ b/packages/core/useAsyncState/index.ts
@@ -1,6 +1,6 @@
 import type { MaybeRef, Ref, ShallowRef, UnwrapRef } from 'vue'
 import { noop, promiseTimeout, until } from '@vueuse/shared'
-import { ref as deepRef, shallowRef, toValue } from 'vue'
+import { customRef, ref as deepRef, shallowRef, toValue } from 'vue'
 
 export interface UseAsyncStateReturnBase<Data, Params extends any[], Shallow extends boolean> {
   state: Shallow extends true ? Ref<Data> : Ref<UnwrapRef<Data>>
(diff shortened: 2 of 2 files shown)


9romise:
Is this an AI-generated PR? Please read and keep the content from the PR template before opening a PR. If everything is ready, feel free to reopen or open a new PR.

9romise:
Also, if you're interested in solving this issue, please check out the discussion in #4975.

Issue #5445, opened 2026-05-19
`useDraggable` | Add snap, transform output, and richer drag lifecycle (parity with plain-draggable)


Gazoon007:
### Clear and concise description of the problem

As a developer using `useDraggable`, I want richer dragging primitives — snap-to-grid/line/element, GPU-accelerated `transform` output, and a more granular drag lifecycle — so that I can build complex UI (whiteboards, layout editors, dashboard widgets, alignment guides, etc) without reaching for a third-party library.

The current `useDraggable` covers the basics (pointer events, axis lock, container bounds, basic auto-scroll, callbacks) but is missing several capabilities that the well-known [plain-draggable](https://github.com/anseki/plain-draggable) library provides out of the box. Adopting a subset of those primitives would close the gap and keep VueUse self-sufficient for the most common drag use cases.

I'm willing to open a PR for one or more of these features if maintainers are open to them.

### Suggested solution

Extend `UseDraggableOptions` and the return type with the following (each opt-in, all backwards compatible):

#### 1. Snap targets (highest impact)

```ts
interface SnapTarget {
  x?: number | string | { start?: number | string, end?: number | string, step?: number | string }
  y?: number | string | { start?: number | string, end?: number | string, step?: number | string }
  boundingBox?: HTMLElement | SVGElement | Rect
  gravity?: number       // default 20
  corner?: 'tl' | 'tr' | 'bl' | 'br' | 'all' | string
  side?: 'start' | 'end' | 'both'
  center?: boolean
  edge?: 'inside' | 'outside' | 'both'
}

interface UseDraggableOptions {
  // ...existing
  snap?: MaybeRefOrGetter<
    | number
    | SnapTarget
    | SnapTarget[]
    | { targets: SnapTarget | SnapTarget[], gravity?: number /* shared overrides */ }
  >
}
```

Common cases stay one-liners:

```ts
useDraggable(el, { snap: { step: 30 } })                     // 30×30 grid
useDraggable(el, { snap: [{ x: 100 }, { x: 300 }] })         // two vertical guide lines
useDraggable(el, { snap: { boundingBox: containerEl } })     // snap to container edges
```

#### 2. Output mode (`transform` vs `left/top`)

Today the returned `style` ref always emits `left: ?px; top: ?px;`. A new option would let consumers opt into `transform`, which only triggers GPU composite instead of layout + paint per frame — noticeably smoother with many draggables, complex layouts, or on mobile.

```ts
interface UseDraggableOptions {
[...]

PR #5496, opened 2026-06-07
feat(useDraggable): add snap, transform output, lifecycle hooks, and auto-scroll options

Gazoon007:

  ---

  ### Description

  This PR resolves this issue https://github.com/vueuse/vueuse/issues/5445

  It's a follow-up to my auto-scroll PR. The goal is to get `useDraggable` closer to [PlainDraggable](https://anseki.github.io/plain-draggable/#options-snap) so people can build whiteboards / layout editors without pulling in another lib. Everything here is opt-in, so existing usage doesn't change.
  
  New options:

  - **snap**
    Snap to a grid, guide lines, or another element's edges. Supports gravity, corner/side/edge/center targeting, and % values. You can pass a plain config or the `SnapOptions` wrapper.
  - **output** (`'leftTop' | 'transform'`)
    Keep the current `left/top` positioning, or switch to `transform: translate3d()` for GPU-accelerated movement.
  - **onMoveStart** / **onBeforeMove**
    `onMoveStart` fires once on the first real movement, and `onBeforeMove` runs before the position is committed so you can validate/tweak it.
  - **classes** / **cursor** / **zIndex**
    Applied and cleaned up automatically while dragging.

  I also extended **autoScroll** with tiered speed/sensitivity arrays and per-axis bounds (`minX`/`maxX`/`minY`/`maxY`), added a `recalc()` to re-apply snap and bounds after the layout changes outside a drag, and let `containerElement` accept a `DraggableBounds` rect so you don't always need a real element.
  
  ### Additional context

  I migrated the test suite off jsdom and onto vitest browser mode (Chromium). Snap and auto-scroll lean on real layout, so jsdom wasn't reliable for them. On top of that I'm following the advice here https://github.com/vueuse/vueuse/pull/4472#issuecomment-3072883349, if a function needs even one browser test, it's better to move the whole suite over instead of mixing.
  
  I'm unsure about a couple of defaults (e.g. snap gravity), so happy to adjust those based on what you think fits the API best.

Also, I revamped the demo as well. Half of the codes, however, are generated by AI, but I've done a rereview.

Any feedback is welcome.

  ---

  ### Proof

https://github.com/user-attachments/assets/55b92088-95fa-4fd2-8c76-8762d92064d4


OrbisK:
> I kept the AI-disclosure checkbox in since the template adds it now, but only check it if it's true for you.

👀

Gazoon007:
> > I kept the AI-disclosure checkbox in since the template adds it now, but only check it if it's true for you.
> 
> 👀

Oopsie 😅

PR #5497, opened 2026-06-07
fix(useRefHistory): clone lastRawValue for shouldCommit comparison

### LINKED ISSUE #4985: `useRefHistory` | Use the clone function to set `lastRawValue`to freeze it

Currently, `lastRawValue` can be changed if the source is an object and its key is changed. This makes `shouldCommit` return false everytime when deeply-testing the `before` and `now`.

https://github.com/vueuse/vueuse/blob/ce09e0d03e0cac5884ad57bb87215428ac34917b/packages/core/useRefHistory/index.ts#L156

### Suggested solution

Use the clone function to freeze `lastRawValue`

### Alternative

_No response_

### Additional context

_No response_

### Validations

meta-syntax:

<details>
<br>

>

</details>

---

### Description

Fixes #4985

`shouldCommit` receives same object reference for both `oldValue` and `newValue` when `deep: true`.
This is because `lastRawValue` holds direct reference to `source.value`.

Clone `lastRawValue` at 3 assignment sites using same clone logic as `useManualRefHistory`.

Added tests for `shouldCommit` + `deep: true` with object values (both `sync` and `pre` flush).

### DIFF
--- a/packages/core/useRefHistory/index.ts
+++ b/packages/core/useRefHistory/index.ts
@@ -3,6 +3,7 @@ import type { Ref } from 'vue'
 import type { CloneFn } from '../useCloned'
 import type { UseManualRefHistoryReturn } from '../useManualRefHistory'
 import { pausableFilter, watchIgnorable } from '@vueuse/shared'
+import { cloneFnJSON } from '../useCloned'
 import { useManualRefHistory } from '../useManualRefHistory'
 
 export interface UseRefHistoryOptions<Raw, Serialized = Raw> extends ConfigurableEventFilter, ConfigurableFlush {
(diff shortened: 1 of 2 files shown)


packages/core/useRefHistory/index.ts
```diff
@@ -143,7 +150,7 @@ export function useRefHistory<Raw, Serialized = Raw>(
     if (!shouldCommit(lastRawValue, source.value))
```
OrbisK: maybe we should also clone the value here?

OrbisK (changes requested):

packages/core/useRefHistory/index.ts
meta-syntax: Thank you for the review.
I agreed that it should be cloned for consistency. Updated.

packages/core/useRefHistory/index.ts
```diff
@@ -140,10 +147,10 @@ export function useRefHistory<Raw, Serialized = Raw>(
     // so we do not trigger an extra commit in the async watcher
     ignorePrevAsyncUpdates()
 
-    if (!shouldCommit(lastRawValue, source.value))
+    if (!shouldCommit(lastRawValue, _cloneRaw(source.value)))
       return
 
-    lastRawValue = source.value
+    lastRawValue = _cloneRaw(source.value)
```
OrbisK: we should clone it only once here. Clones can be costly.

meta-syntax: Thanks for catching this. Now cloning once and reusing it for both the shouldCommit check and the assignment.

PR #5506, opened 2026-06-17
chore: pin actions to sha

43081j:
### Description

Pin the actions to a sha for security reasons.

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -16,10 +16,10 @@ jobs:
     timeout-minutes: 10
 
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


OrbisK:
isnt this something we can force with renovate?

OrbisK (approved):

43081j:
renovate already follows convention once this lands iirc. in that updates will also be pinned to a sha

OrbisK:
fyi: 

it is forceable https://docs.renovatebot.com/presets-helpers/#helperspingithubactiondigests

43081j:
we don't need to because my follow up action of this is to enforce sha pinning in the repo itself

PR #5513, opened 2026-06-19
fix(useThrottleFn): schedule trailing edge when leading is disabled

### LINKED ISSUE #5512: BUG | `refThrottled` | refThrottled doesn't throttle consistently when `leading: false`
### Describe the bug

I'm trying to throttle showing a loader.
When pending is toggled to true then back to false within the throttle duration, the throttled variable should not change.
When pending is toggled to true, then the throttle duration elapses, the throttled variable should change.

In the reproduction ref `input` is `false` by default.

`throttled` is throttled by 1000ms.

There are 2 buttons:

- `update(200)` - toggles `input` to `true`, then to `false` within 200ms
- `update(1200)` - toggles `input` to `true`, then to `false` within 1200ms

Reproduction 1:

1. click `update(200)` once.
   `throttled` not changed.
   As expected

2. click `update(1200)`.
   `throttled` not changed.
   Expected throttled to toggle to true, then to false.
   `throttled` should have been updated twice.

3. click `update(200)` again
4. click `update(1200)` again.
   `throttled` changed to `true`.
   Trailing callback is never called, and throttled is never changed back to `false`.
   Expected to change `throttled` back to false

https://github.com/user-attachments/assets/11942b83-8f68-47dd-bb22-3aadc86e08d1

Refresh the tab between reproductions.

Reproduction 2:

1. click `update(1200)`.
   To ensure the throttling logic works
2. click `update(200)`
3. wait for 1 sec
4. click `update(200)` again.

https://github.com/user-attachments/assets/abef8a7c-f273-447c-aae3-c9d823a933e7

Expected: since the original value changed back within the throttle duration, `throttled` should not change.

Currently:

1. `input` changes to true, then to false
2. then `throttled` changes to `true`
3. after throttle duration `throttled` changes to `false`

Both reproductions are not reproducible consistently, but after a few tries, I managed to reproduce both multiple times.

### Reproduction

https://playground.vueuse.org/?vueuse=14.3.0#eNqNVMtu2zAQ/JWFLpZRxVYe6CGVjbRFDinaNEjTmy60TNuMKZLgw05g+N+7pJ62k6A6ibvD3dnhkLvoq1KjjaPRdZSZQjNlwVDrFHAilpM8siaPprlgpZLawg40XTyttLSW0znsYaFlCYMbLOAMHRdS08EhOAGzIpzL7aP/3xJbrNptuAvRuSikMNiWU6rufxmYADGvooBYXINw5YzqIUyQAwDZEmZB0C08YAFmaBxraiTfUI/wxJ9YSaWzcRwCdTJOhwmI4fBL18xqwjgTS2zW8YutdnTYQDgl8xPEgnDjIQ2ICeUsQnDSk5xtdQr5VrY4bErgPE3TpGUy2hDuaNK0rZa9ak7NiaWdOHOniWXyQCPYeZU8XnI64nIZDyqCzGAfVBvrQcW5qo/lfByFacRtTqGt71V7r2aY+K2iIYEb98f8vRY9OdMwYHBF3MqVQHV6p8N0itbVBknz1+oFhxH4NIFzJIKZbFwZHK2EC0tLxREXjJXN2QYKToxBy6uzq+B5wC+bOWul6OUuYLY8s5Tws8s0BS2dmCOfGZfFOo/gpuCsWCOwIhFfpDhiNO2tsnFV8p0GpcUOdRezfv2vJueHXcLyqE3drD8mdvqMm7IxBoMkAaGmd/4kr2G3q82932dj1bBV09bGAdKdyDEML6KBv9VRBGhjgGNg7f+qXHMtj0A/q0sRMM297CDNCNm4d6hRgm8XumfBlqNnIwU+cMFQeVTIUjFO9W/l/Y3vG5atVYqCM3+EmL8YSRMvVrRYvxF/Ni8+lkcP+NZQvaF51OYs0Utqq/Ttn3v6gv9tspRzxxH9QfLRv17Oc6xg39AFSLuHC2zvwmuLijyZ2xdLhWmG8kQ9ch/weYTP7fcPRu/oXo6uwj68NtH+H4V7EUM=

### System Info

```Shell
Browser: Firefox Developer Edition 152.0b10 (64 bit)

vueuse: 14.3.0
```

### Used Package Manager

npm

### Validations

babu-ch:
Fixes #5512


MILLERMARRU (commented):
Traced this through `throttleFilter` against the current source to confirm the bug. With `leading: false, trailing: true`, the very first call takes the `elapsed > duration` branch (since `lastExec` starts at 0), and in that branch the only thing that happened before this fix was `if (leading || !isLeading) invoke()`, which evaluates to false when leading is off and `isLeading` is still true on the first call. So nothing invoked, and the fallback at the bottom (`if (!leading && !timer) timer = setTimeout(() => isLeading = true, duration)`) only ever flips the flag, it never schedules an actual invoke. The first call was getting silently dropped rather than deferred to the trailing edge.

`setupTrailing` is a clean fix for this, it's literally lifting the existing trailing-timer logic out of the `elapsed <= duration` branch into a shared helper so the `elapsed > duration` branch can call it too. Traced the new test case by hand against the fixed code and it checks out: first call schedules a trailing invoke at `duration`, timer fires and calls back once, second call after the throttle window schedules another trailing invoke, fires and calls back a second time. Also checked that `if (!leading && !timer)` right after doesn't double up on a timer, since `setupTrailing` already sets `timer` before that check runs.

This looks correct and minimal, doesn't touch behavior for any other leading/trailing combination since it's purely additive to a path where nothing used to happen.

PR #5514, opened 2026-06-23
fix: accept array template ref bound to v-for as observer target

### LINKED ISSUE #4712: BUG | `useIntersectionObserver` | `useTemplateRef` of array type cannot be passed as argument

eldardada:
Fixes #4712

`useTemplateRef` on an element rendered with `v-for` is typed as
`Readonly<ShallowRef<HTMLElement[] | null>>` in Vue 3.5. Passing that into
`useIntersectionObserver` doesn't type-check, because the `target` type didn't
allow `null` inside the array ref:

```ts
const items = useTemplateRef('items') // ShallowRef<HTMLElement[] | null>
useIntersectionObserver(items, () => {}) // type error
```

`useResizeObserver` and `useMutationObserver` share the same `target` type, so
they had it too. Rather than patch one signature, I moved the union into a
`MaybeComputedElementRefOrArray` type in `unrefElement` and reused it across all
three. It allows a `null` array value, which is what a `v-for` template ref
produces.

The runtime stays the same: these composables already drop nullish entries via
`toArray(...).map(unrefElement).filter(notNullish)`, so only the type needed
widening.

### DIFF
--- a/packages/core/unrefElement/index.ts
+++ b/packages/core/unrefElement/index.ts
@@ -5,6 +5,7 @@ export type VueInstance = ComponentPublicInstance
 export type MaybeElementRef<T extends MaybeElement = MaybeElement> = MaybeRef<T>
 export type MaybeComputedElementRef<T extends MaybeElement = MaybeElement> = MaybeRefOrGetter<T>
 export type MaybeElement = HTMLElement | SVGElement | VueInstance | undefined | null
+export type MaybeComputedElementRefOrArray<T extends MaybeElement = MaybeElement> = MaybeComputedElementRef<T> | MaybeComputedElementRef<T>[] | MaybeRefOrGetter<T[] | null>
 
 export type UnRefElementReturn<T extends MaybeElement = MaybeElement> = T extends VueInstance ? Exclude<MaybeElement, VueInstance> : T | undefined
 
--- a/packages/core/useIntersectionObserver/index.browser.test.ts
+++ b/packages/core/useIntersectionObserver/index.browser.test.ts
@@ -1,5 +1,6 @@
+import type { ShallowRef } from 'vue'
 import type { UseIntersectionObserverReturn } from '.'
-import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { beforeEach, describe, expect, expectTypeOf, it, vi } from 'vitest'
 import { cleanup } from 'vitest-browser-vue'
 import { page } from 'vitest/browser'
 import { defineComponent, shallowRef, toValue, useTemplateRef } from 'vue'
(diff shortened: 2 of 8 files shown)


eldardada:
For context: this is the general approach @OrbisK asked for in #5321. `useResizeObserver` and `useMutationObserver` share the same `target` type and had the same issue, so instead of patching one signature it moves the union into a shared `MaybeComputedElementRefOrArray` type in `unrefElement` and reuses it across all three composables

OrbisK:
would you mind adding some tests to cover this behavior? 🙏🏽

eldardada:
> would you mind adding some tests to cover this behavior? 🙏🏽

Done, added type-level tests for all three composables.

Since this PR only widens the target type (the runtime already drops nullish entries via `toArray(...).map(unrefElement).filter(notNullish))`, a runtime test wouldn't actually exercise the change. So each test asserts that an array template ref bound to `v-for` (`Readonly<ShallowRef<HTMLElement[] | null>>` in Vue 3.5) is assignable to the observer's target, which fails to compile without this fix.

One note on structure: for `useResizeObserver` I added a standalone `index.test.ts`, since the composable had no direct test file of its own (its runtime is covered through the `vResizeObserver` directive test)

OrbisK (dismissed):

9romise (approved):

ilyaliao (approved):

OrbisK (approved):

PR #5515, opened 2026-06-23
docs(useElementVisibility): document controls return values

### LINKED ISSUE #5443: Enhancement: useElementVisibility should expose internal intersection observer state
## Feature request

`useElementVisibility` wraps `useIntersectionObserver` internally but only exposes `isVisible` (or controls when `controls: true`). It would be useful to also expose whether the observer is actively observing, especially when using the `once: true` option.

## Proposed solution

When `controls: true` is set, include an `isActive` flag in the return value indicating whether the observer is currently active (has not been stopped by the `once` option).

This would allow consumers to show loading states or re-enable observation if needed.

eldardada:
Closes #5443

`isActive` is actually already there. When `controls: true` is set, `useElementVisibility` spreads the `useIntersectionObserver` return onto its result:

```ts
return options.controls
  ? { ...observerController, isVisible }
  : isVisible
```

so `isActive`, `pause`, `resume`, `stop` and `isSupported` all come along with `isVisible`. It's just never documented, which is why it looks missing in the issue.

So no behaviour change here, just docs + a test: a table for the `controls` return, a short `once: true` example that checks `isActive`, and a test that `controls: true` actually exposes those props


eldardada:
Kept the docs and the test in one commit since it gets squashed on merge anyway

ilyaliao (approved):

9romise (approved):

OrbisK (approved):

PR #5517, opened 2026-06-26
feat(useDebounceFn): add leading option for immediate first execution

### LINKED ISSUE #4282: An option not to delay initial execution `useDebounceFn`

Currently, when using a debounced function from `useDebounceFn`, even the initial execution is debounced. I think it would be useful to provide an option to avoid delaying the callback if it hasn't been executed within the specified debounce duration.

I'd be happy to help implement this.

### Suggested solution

We could introduce an option like `debounceInitialExecution: false` to `DebounceFilterOptions` with the default value set to `true` because of backwards compatibility.

### Alternative

_No response_

### Additional context

_No response_

### Validations

kon2raya24:
Closes #4282

I wanted `useDebounceFn` to fire immediately on the first call for a search-as-you-type pattern, but it always delays even the first invocation. This PR adds a `leading` option that controls this.

When `leading: true`, the first call executes immediately on the leading edge, and subsequent calls are debounced as usual. Default is `false` (standard debounce behavior, nothing changes).

```ts
const search = useDebounceFn(fetchResults, 300, { leading: true })
// First keystroke -> immediate search
// Rapid typing -> debounced at 300ms
```

Also added 16 tests for `useDebounceFn` which previously had zero dedicated tests. Happy to adjust anything!

### DIFF
--- a/packages/shared/useDebounceFn/index.browser.test.ts
+++ b/packages/shared/useDebounceFn/index.browser.test.ts
@@ -112,17 +112,17 @@ describe('useDebounceFn', () => {
     expect(debounced.isPending.value).toBe(false)
   })
 
-  it('should execute immediately when debounceInitialExecution is false', async () => {
+  it('should execute immediately when leading is false', async () => {
     const fn = vi.fn()
-    const debounced = useDebounceFn(fn, 100, { debounceInitialExecution: false })
+    const debounced = useDebounceFn(fn, 100, { leading: false })
 
     debounced()
     expect(fn).toHaveBeenCalledTimes(1)
   })
 
-  it('should debounce subsequent calls when debounceInitialExecution is false', async () => {
+  it('should debounce subsequent calls when leading is false', async () => {
     const fn = vi.fn()
-    const debounced = useDebounceFn(fn, 100, { debounceInitialExecution: false })
+    const debounced = useDebounceFn(fn, 100, { leading: false })
 
     // First call executes immediately
     debounced()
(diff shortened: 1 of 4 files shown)


packages/shared/utils/filters.ts
```diff
@@ -46,6 +46,15 @@ export interface DebounceFilterOptions {
...
   rejectOnCancel?: boolean
+
+  /**
+   * Whether to delay the initial execution as well.
+   * When `false`, the first call will be executed immediately without delay,
+   * and subsequent calls will be debounced as usual.
+   *
+   * @default true
+   */
+  debounceInitialExecution?: boolean
```
OrbisK: Maybe we should name this `leading` instead. Wdyt?

packages/shared/useDebounceFn/index.browser.test.ts
```diff
@@ -0,0 +1,220 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
```
OrbisK, line 1: can we have this as `index.browser.test.ts`?

packages/shared/utils/filters.ts
```diff
@@ -118,6 +129,18 @@ export function debounceFilter(ms: MaybeRefOrGetter<number>, options: DebounceFi
       return Promise.resolve(invoke())
     }
 
+    // If leading is false and this is the first call,
```
OrbisK: I think you logic is swaped. `leading` true should mean to execute on the first edge.

```suggestion
    // If leading is true and this is the first call,
```

kon2raya24:
Thanks for the thorough review @OrbisK! All 3 comments addressed:

1. ✅ Renamed `debounceInitialExecution` → `leading`
2. ✅ Renamed test file to `index.browser.test.ts`
3. ✅ Fixed the logic — `leading: true` now correctly executes on the leading edge (first call immediate), `leading: false` (default) = standard trailing-edge debounce

Let me know if there's anything else!

packages/shared/utils/filters.ts
```diff
@@ -118,6 +129,18 @@ export function debounceFilter(ms: MaybeRefOrGetter<number>, options: DebounceFi
...
+    if (_leading && !hasExecuted) {
+      hasExecuted = true
+      if (maxTimer) {
+        _clearTimeout(maxTimer)
+        maxTimer = undefined
+      }
+      _pending.value = false
+      return Promise.resolve(invoke())
+    }
+
```
OrbisK, line 143: This will only execute the first execution. But I think we should expect it to always run the leading execute after we have waited and executed the handler.

PR #5519, opened 2026-06-29
fix(useFocusTrap): clean up containers to prevent a memory leak

dxodyo:
This PR fixes a memory leak inside `useFocusTrap` where detached DOM element references are permanently retained in memory when a trapped element is removed from the DOM (e.g., via `v-if`).

When the reactive target becomes empty (els.length === 0), the watch implementation returns early. This skips notifying the underlying focus-trap instance. Its internal `state.containers` and `state.containerGroups` arrays hold onto hard references of the detached HTMLElements indefinitely.

### DIFF
--- a/packages/integrations/useFocusTrap/index.ts
+++ b/packages/integrations/useFocusTrap/index.ts
@@ -100,8 +100,14 @@ export function useFocusTrap(
   watch(
     targets,
     (els) => {
-      if (!els.length)
+      if (!els.length) {
+        // Clean up to prevent a memory leak
+        if (trap) {
+          trap.updateContainerElements([])
+          trap.deactivate()
+        }
         return
+      }
       if (!trap) {
         // create the trap
         trap = createFocusTrap(els, {
--- a/packages/integrations/useFocusTrap/index.ts
+++ b/packages/integrations/useFocusTrap/index.ts
@@ -71,7 +71,14 @@ export function useFocusTrap(
   const isPaused = shallowRef(false)
 
   const activate = (opts?: ActivateOptions) => trap && trap.activate(opts)
-  const deactivate = (opts?: DeactivateOptions) => trap && trap.deactivate(opts)
+
+  const deactivate = (opts?: DeactivateOptions) => {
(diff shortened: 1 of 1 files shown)


packages/integrations/useFocusTrap/index.ts
```diff
@@ -100,8 +100,14 @@ export function useFocusTrap(
   watch(
     targets,
     (els) => {
-      if (!els.length)
+      if (!els.length) {
+        // Clean up to prevent a memory leak
+        if (trap) {
+          trap.updateContainerElements([])
```
OrbisK: do we need this, if we deactivate it afterwards?

If yes, should we also do it on scope disposal?

dxodyo: Hi! 

Yep, we definitely need it. If we only call `deactivate()`, the underlying `focus-trap` library keeps the dead DOM elements cached inside its internal arrays, which causes massive leaks during `v-if` toggles.

You are completely right about scope disposal! I missed that on my first pass. 

I did a deeper dive and realized that instead of fighting the library's strict array-validation rules, the cleanest approach is to completely discard the `focus-trap` instance whenever it gets turned off. 

I updated the PR to move `trap = undefined` directly into the main `deactivate` function. This beautifully handles everything in one place:
1. The **`v-if` watcher** (when targets become empty).
2. The **`tryOnScopeDispose` hook** (when the component unmounts).
3. Any **manual `deactivate()`** calls.

```typescript
const deactivate = (opts?: DeactivateOptions) => {
  if (trap) {
    trap.deactivate(opts)
    // Clear trap reference to avoid memory leaks
    trap = undefined
  }
}
```

I tested this updated implementation locally with Chrome DevTools heap snapshots across both `v-if` toggles and full component unmounts. The detached elements are 100% gone and the memory is completely clean 👌🏻

OrbisK: I do like the idea, but we should be carefull here. Because now we cant reactivate the trap after we call deactivate.

PR #5523, opened 2026-07-01
fix(useGamepad): keep updating remaining gamepads after one disconnects

### LINKED ISSUE #5505: BUG | `useGamepad` | Second gamepad stops responding after first gamepad disconnects
https://github.com/user-attachments/assets/988cedf5-18e8-47a1-b4c7-cb13447cee6d

### Describe the bug

Hello, I found this issue in my app I am coding and I successfully replicated it in the useGamepad demo on VueUse site.

I connect two controllers, everything works as it should. I disconnect first controller with index 0, second controller with index 1 is still appearing but does not register button presses. I reconnect the first controller, everything works again as it should.

In my app I was also checking this behavior directly with Navigator.getGamepads(), but the issue was not present here. While the Vue gamepad stopped responding, the Navigator gamepad kept responding to button presses.

edit: More specific info. In my app. After the controller with index 0 disconnects, the second controller is not responding even if I take the instance again using:

```
const gamepad = computed(() => gamepads.value.find(g => g.mapping === 'standard'))
```

### Reproduction

https://vueuse.org/core/useGamepad/#demo

### System Info

```Shell
System:
    OS: Windows 10 10.0.19045
    Memory: 4.12 GB / 31.92 GB
  Binaries:
    Node: 22.14.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.9.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.12.1 - C:\Users\ADAM\AppData\Local\pnpm\pnpm.EXE
  Browsers:
    Chrome: 149.0.7827.103
    Edge: Chromium (149.0.4022.69)
    Firefox: 150.0.1 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.19041.5794
  npmPackages:
    @vueuse/core: ^13.6.0 => 13.7.0 
    vue: ^3.5.18 => 3.5.19
```

### Used Package Manager

npm

### Validations

lazerg:
After a gamepad disconnects, `gamepads` is re-packed with `filter`, so the remaining gamepads no longer sit at the array slot matching their `index`. `updateGamepadState` still writes to `gamepads.value[gamepad.index]` using the raw index, so those gamepads silently stop updating. This looks each gamepad up by matching its index before updating it.

Fixes #5505

### DIFF
--- a/packages/core/useGamepad/index.test.ts
+++ b/packages/core/useGamepad/index.test.ts
@@ -90,4 +90,30 @@ describe('useGamepad', () => {
     expect(survivor.timestamp).toBe(1000)
     expect(survivor.buttons[0].pressed).toBe(true)
   })
+
+  it('skips null slots returned by getGamepads', async () => {
+    const pad = createGamepad(1)
+    // browsers return a fixed-length array with null in the unused slots
+    const connectedPads: Array<MockGamepad | null> = [null, pad]
+    const navigator = { getGamepads: () => connectedPads } as unknown as Navigator
+
+    const { gamepads } = useGamepad({ navigator })
+    await nextTick()
+
+    dispatchGamepadEvent('gamepadconnected', pad)
+    await nextTick()
+
+    expect(gamepads.value.map(g => g.index)).toEqual([1])
+
+    // new input arrives on the next frame, alongside a null slot
+    pad.timestamp = 1000
+    pad.buttons = [{ pressed: true, touched: true, value: 1 }]
+
+    expect(() => triggerRaf()).not.toThrow()
+    await nextTick()
+
+    const updated = gamepads.value.find(g => g.index === 1)!
+    expect(updated.timestamp).toBe(1000)
+    expect(updated.buttons[0].pressed).toBe(true)
+  })
 })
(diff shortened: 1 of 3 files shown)


lazerg:
Added a test for the null gamepad slot case Codecov flagged, coverage's at 100% now.

lazerg:
@43081j this one's already got the pr welcome label so hopefully an easy merge, mind reviewing the gamepad-index fix when you get a chance?

ilyaliao (dismissed):
Thanks for the contribution, and the code looks good to me! However, I don’t have a second gamepad on hand to test this properly.

packages/core/useGamepad/index.test.ts
```diff
@@ -0,0 +1,119 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
```
OrbisK: can we have this as `*.browser.test.ts`? Maybe we dont need to mock `requestAnimationFrame` etc

lazerg: Done, pushed.

43081j (approved):

ilyaliao (approved):

OrbisK (approved):

will-neithan:
Thanks @lazerg and OrbisK, ilyaliao, 43081j, you bring me joy. I appreciate it.

PR #5525, opened 2026-07-02
fix(useFetch): ignore stale abort error after refetch succeeds

### LINKED ISSUE #5516: BUG | `useFetch` | error sticky despite refetch success
### Describe the bug

Hello,

When using `useFetch` with `refetch: true`, modifying the resource URL while a previous one is loading sets the error to `The operation was aborted.`; but, when the subsequent fetch ends successfully, `error` stays set. A user dispatching on wether `error` is set would never see that the next request was successful, even though the abort is not really an error.

Can be reproduced by doing:
```
<script setup>
import { ref, nextTick } from 'vue'

const url = ref('data:text/plain,hello world') // we will stop fetching it before it resolves
const { data, error, isFetching } = useFetch(url, {
  refetch: true,
})

async function run() {
  await nextTick()
  url.value = 'data:text/plain,success'
}
run()
</script>

<template>
    data: {{ data }}<br>
    isFetching: {{ isFetching }}<br>
    error: {{ error }}
</template>
```

results in:
```
data: success
isFetching: false
error: The operation was aborted.
```

Worked around on my side by doing
```
const useFetchRefetchWorkaround = (url) =>
  useFetch(url, {
    refetch: true,
    // Workaround VueUse bug: `error` permanently set when request aborted.
    onFetchError: (ctx) => {
      if ((ctx.error as DOMException | null)?.name === 'AbortError')
        ctx.error = null
      return ctx
    },
  })
```

### Reproduction

[VueUse playground](https://playground.vueuse.org/#eNp9Ustu2zAQ/JUFL3YAwT60J8Mp+kAKtIe2aHPkhaXXNhNqKZBL24Ggf8+SShQBCXKTdoazM0P26kvXrU4Z1UZtk42uY0jIufukybVdiAw9RNw3QHjhW2fvYYB9DC0s5NBiRsoJvyPb44R/FoIM1zbEQtRkAyWGHD1cF8nlYmfYbFh01503jpojeh/gHKLfLa6e+T0UWgMYY4gNuFS3ODrIoutp61JkG+g1QZEukw1wzNhoGkRKk0kPZGGfybILBDHT8mrkm7NxPMVbChuKydXJ+Iyy4pXNlK3FlCTToKkKadqux/KkNvlhbIXKKH9Q7W+gH3PAMGz/xzp/SVLRebCJU0NXuH4JUlbN5FWjOElRe3dY3aVAco01lFY2tJ3zGH93JXDSSmQKUjAjPZ9/1tlTSePcHtHevzG/S5cy0+pPxITxhFpNGJt4QB7hm3+/pKYZ2IZd9sJ+B/yLKfhcPI60r5l2YnvGq25/1Hcm7dymmwsjpedQxWhhDpWvlby6b+9Ef7H7YfWxnpNrVMMjKJMJkg==)

### System Info

Reproduces on latest VueUse on playground.

### Used Package Manager

pnpm

### Validations

lazerg:
When the reactive URL changes while a request is in flight, useFetch aborts the previous request, but its `.catch` handler still wrote the abort error into `error` even after a newer request had reset the state and succeeded, leaving `error` stuck. This guards the error/data assignment with the same execute-counter check already used in `.finally`, so a stale aborted request no longer overwrites the latest result.

Fixes #5516

### DIFF
--- a/packages/core/useFetch/index.ts
+++ b/packages/core/useFetch/index.ts
@@ -521,9 +521,11 @@ export function useFetch<T>(url: MaybeRefOrGetter<string>, ...args: any[]): UseF
           }))
         }
 
-        error.value = errorData
-        if (options.updateDataOnError)
-          data.value = responseData
+        if (currentExecuteCounter === executeCounter) {
+          error.value = errorData
+          if (options.updateDataOnError)
+            data.value = responseData
+        }
 
         errorEvent.trigger(fetchError)
         if (throwOnFailed)
--- a/packages/core/useFetch/index.test.ts
+++ b/packages/core/useFetch/index.test.ts
@@ -731,6 +731,17 @@ describe('useFetch', () => {
     })
   })
 
+  it('should clear error when refetch succeeds after aborting previous request', async () => {
+    const url = shallowRef(`${baseUrl}?delay=50`)
+    const { data, error } = useFetch(url, { refetch: true }).json()
+    await nextTick()
+    url.value = jsonUrl
+    await vi.waitFor(() => {
+      expect(data.value).toEqual(jsonMessage)
+    })
+    expect(error.value).toBeNull()
+  })
+
   it('should be generated payloadType on execute', async () => {
     const form = deepRef()
     const { execute } = useFetch(baseUrl).post(form)


lazerg:
Whenever you have a sec @43081j, this one fixes a stale abort error sticking around in useFetch after a newer request already succeeded.

packages/core/useFetch/index.ts
```diff
@@ -521,9 +521,11 @@ export function useFetch<T>(url: MaybeRefOrGetter<string>, ...args: any[]): UseF
           }))
         }
 
-        error.value = errorData
-        if (options.updateDataOnError)
-          data.value = responseData
+        if (currentExecuteCounter === executeCounter) {
+          error.value = errorData
```
43081j, line 525: hm, as far as i can tell, we unconditionally update the refs for _this execution_.

```ts
const url = shallowRef(wherever);
const { data } = useFetch(wherever);

// at this point, let's say the fetch is STILL RESOLVING

url.value = somewhereElse;

// now we've triggered a 2nd request.
// the FIRST request comes back, and sets `data.value`
// the SECOND request comes back, and sets `data.value` again
```

i.e. we don't seem to check the execution counter for any of the refs really other than `loading` 😅 

in that example above, if the SECOND came back first, it seems we'd overwrite it with the FIRST...

we could add this same fix to the other places too but it is a behavioural change 🤔  it means you don't get any values until all requests settle

43081j: ah ignore me!

we abort it at the start. so it'll _always_ error if a new request comes along and it'll hit this branch.

all good i think 👍

43081j (approved):

ilyaliao (approved):

OrbisK (approved):

PR #5537, opened 2026-07-14
fix(useFetch): reset response before each request to avoid a stale onFetchError response

### LINKED ISSUE #5536: BUG | useFetch| Stale response object value will be passed to onFetchError() if fetch() rejects Promise and  execute() is call again after a successful prior execution.
### Describe the bug

execute() should set response.value to null before calling fetch() otherwise; response object from previous execution will be passed to onFetchError() if fetch() rejects the Promise.

Scenario.
First execute()
  response.value will be set to fetch() response object value if execution is successful.
Second execute() and fetch() rejects the promise.
  Assuming caller of execute() did not reset response.value to null.
  response.value from first execution() will be passed to onFetchError() handler.

### Reproduction

See BUG description and check the source code. It is very obvious.

### System Info

```Shell
vueuse 14.3.0
```

### Used Package Manager

pnpm

### Validations

lazerg:
When a request fails at the network level, `onFetchError` got the response from the previous successful request instead of nothing. `execute()` already clears `error`, `statusCode`, and `aborted` before each run but left `response` alone, so the stale one carried over. Now it clears `response` too.

Fixes #5536

### DIFF
--- a/packages/core/useFetch/index.test.ts
+++ b/packages/core/useFetch/index.test.ts
@@ -605,6 +605,24 @@ describe('useFetch', () => {
     })
   })
 
+  it('should not pass a stale response to onFetchError when a later fetch rejects', async () => {
+    let errorResponse: Response | null | undefined
+    const { execute } = useFetch(`${baseUrl}?text=hello`, {
+      immediate: false,
+      onFetchError(ctx) {
+        errorResponse = ctx.response
+        return ctx
+      },
+    })
+
+    await execute()
+
+    fetchSpy.mockRejectedValueOnce(new Error('Network request failed'))
+    await execute()
+
+    expect(errorResponse).toBeNull()
+  })
+
   it('should emit onFetchResponse event', async () => {
     const { onFetchResponse, onFetchError, onFetchFinally } = useFetch(baseUrl)
 
--- a/packages/core/useFetch/index.test.ts
+++ b/packages/core/useFetch/index.test.ts
@@ -623,6 +623,20 @@ describe('useFetch', () => {
     expect(errorResponse).toBeNull()
   })
 
+  it('should keep the previous response.value until the next request settles', async () => {
+    const { execute, response } = useFetch(`${baseUrl}?text=hello`, { immediate: false })
+
+    await execute()
+    const firstResponse = response.value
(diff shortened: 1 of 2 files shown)


MILLERMARRU (commented):
Confirmed the bug against the current `execute()` flow: `response.value` is only ever written inside the `.then()` branch, so on a second `execute()` call that fails before a response is received (DNS failure, connection refused, offline), `onFetchError` reads whatever `response.value` was left over from the previous successful call. Resetting it in the initial block alongside `error.value`/`statusCode.value` closes that gap cleanly.

One thing worth flagging, not a blocker on this fix but a related gap it doesn't touch: the `.then()` branch that sets `response.value = fetchResponse` (and `data.value`) has no `currentExecuteCounter === executeCounter` guard, unlike the `.catch()`/`.finally()` branches right below it. So if two `execute()` calls race and the older one resolves after the newer one has already started, the stale response can still overwrite `response.value`/`data.value` from the request that's actually current. That's the same class of bug as #5571 (filed for the success-path equivalent of what #5525 fixed for the error path), still open as far as I can tell from this diff.

Also worth a mention in the PR description or changeset: resetting `response.value` to `null` on every `execute()` is a small behavior change beyond the bug fix, since it means anything reading `response.value` directly (not just through `onFetchError`) now sees a flash to `null` on every refetch rather than keeping the previous response object until the new one lands. `data.value` is deliberately left alone for that reason, so it might be worth the same treatment for `response.value`, or at least calling it out explicitly since some consumers may read `response.value.status` during the loading state of a refetch.

lazerg:
Fixed both points. The race is real: `response` only ever gets written in the `.then()`, so a second `execute()` that fails before a response arrives left the prior call's stale `Response` in `response.value`, and that stale value got passed to `onFetchError`. Reset `response.value` in the same top-of-`execute` block that already clears `error`/`statusCode`/`aborted`, then scoped the fix so `response.value` itself stays stable across a refetch (matching current behavior) and is only nulled for the value handed to `onFetchError`. Commit a1be9572.

lazerg:
Hey @antfu, this has been open a while — could you take a look when you get a chance? The race condition MILLERMARRU flagged is fixed now (a1be9572) and CI is green.

PR #5538, opened 2026-07-16
fix(useFetch): expose error response body via onFetchError

### LINKED ISSUE #5405: BUG | `useFetch` | does not provide body details for a fetch error

Faustze:

<details>
<br>

>

</details>

---

### Description

The `onFetchError` event hook only receives a bare `Error` with the status text as its message, so the body of an error response (e.g. `application/problem+json`) is not accessible to subscribers - even though it is already parsed internally.

With this change, the error thrown for non-ok responses carries the parsed body as `error.data` and the raw `Response` as `error.response`. Both fields are optional, since errors without a response (network failures, aborts) also reach the hook. The new exported `UseFetchError` interface types the event hook, so the fields are discoverable from completions.

fixes #5405

- The added test fails without the fix (`error.data` was `undefined`) and passes with it.
- Existing behavior is untouched: the error is still an `Error` with the same message; `error.value`, `updateDataOnError` and the `onFetchError` option work as before (54/54 tests pass).
- Docs: extended the Events section with the new fields.

### Additional context

Out of scope, kept for a possible follow-up: when the error body can't be parsed by the configured parser (e.g. `.json()` on a plain-text body), the parse error still masks the response body.

### DIFF
--- a/packages/core/useFetch/index.md
+++ b/packages/core/useFetch/index.md
@@ -314,6 +314,8 @@ const { isFetching, error, data } = useMyFetch('users')
 
 The `onFetchResponse` and `onFetchError` will fire on fetch request responses and errors respectively.
 
+When the response arrives with a non-ok status, the error also exposes the parsed body of the response as `data` and the raw [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) as `response`. Both are `undefined` when the request failed without receiving a response (e.g. a network error or an aborted request).
+
 ```ts
 import { useFetch } from '@vueuse/core'
 // ---cut---
@@ -325,5 +327,7 @@ onFetchResponse((response) => {
 
 onFetchError((error) => {
   console.error(error.message)
+  console.error(error.data) // Parsed body of the error response, e.g. { message: 'Invalid request' }
+  console.error(error.response?.status) // e.g. 400
 })
 ```
--- a/packages/core/useFetch/index.test.ts
+++ b/packages/core/useFetch/index.test.ts
@@ -1,4 +1,4 @@
-import type { AfterFetchContext, OnFetchErrorContext } from './index'
+import type { AfterFetchContext, OnFetchErrorContext, UseFetchError } from './index'
 import { until } from '@vueuse/shared'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { ref as deepRef, nextTick, shallowRef } from 'vue'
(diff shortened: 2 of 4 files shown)


Faustze:
The two failing jobs appear to be unrelated flakes:

 - test (lts/*): useTransition browser test failed on a timing assertion (expected 0.24 to be above 0.25). The same test failed identically on main in run 28950248994 (Jul 8), before this PR.
 - test (20.x): all 707 tests passed; the job failed only due to 2 unhandled rejections surfaced while running useScriptTag/index.browser.test.ts.

The same useTransition assertion also failed in other recent PR runs unrelated to this change 29112232289 (Jul 10, webkit) and 28350478219 (Jun 29, chromium), so it looks like a recurring timing flake rather than anything specific to this branch.

This PR only touches useFetch, so neither test is affected by the change. Happy to retrigger CI if preferred.

MILLERMARRU (commented):
The `err.data = responseData` assignment is correctly placed after `responseData = await fetchResponse.clone()[config.type]()` already ran, so it really is the parsed body and not the raw response, matching what the JSDoc on `UseFetchError.data` promises. Using `.clone()` there (unrelated to this PR, already existing) is also what makes it safe to both attach the parsed body to the error and still let a consumer read the original response separately if they want to.

I reviewed #5525 earlier, the race-condition fix for the error branch, and flagged separately that the success branch of this same `.then()` handler still writes `data.value`/`response.value`/`statusCode.value` without checking `currentExecuteCounter === executeCounter`. This PR's new code sits inside that exact `!fetchResponse.ok` branch, right next to `data.value = initialData || null`, which is one of the unguarded writes. Worth being clear about scope though: the `err` object itself isn't affected by that race, it's built fresh on each call and handed straight to `onFetchError` synchronously, so a stale request can't retroactively corrupt an error someone already received. It's `data.value`/`response.value`/`statusCode.value`, the shared reactive refs, that a slower superseded request could still overwrite after a newer one already resolved. Not something this PR needs to fix, just flagging the adjacency since anyone touching this function for the race condition will land in the same lines.

Type-wise, making `onFetchError: EventHookOn<UseFetchError>` instead of the untyped `EventHookOn` is the right call since it makes `error.data`/`error.response` actually discoverable at the callsite instead of requiring a manual cast, and the test asserting `capturedError.response?.status` works confirms the `Response` object handed to consumers is the real one, not a stub.

PR #5543, opened 2026-07-21
fix(useClipboard): resolve async value provider only once (#5539)

### LINKED ISSUE #5539: BUG | `useClipboard` | Calls async copy callback twice when ClipboardItem data promise rejects
## Description

`useClipboard.copy` supports passing either a string or an async callback that returns the text to copy.

This is useful for Safari/browser gesture restrictions, because the async value can be passed into `ClipboardItem` while still being created during the user gesture.

However, when the async callback rejects, `useClipboard` currently treats this as a Clipboard API failure and falls back to the legacy copy path. In that legacy path, the same callback is invoked again.

As a result, a failed async operation, such as an API request, can be executed twice.

## Reproduction

```ts
const fetchData = async (): Promise<string> => {
  console.log('fetchData called')
  throw new Error('Request failed')
}

const { copy } = useClipboard()

await copy(fetchData)
```

## Current behavior

`fetchData` is called once when creating the `ClipboardItem`.

When that promise rejects, the error is caught in `copy()` and `useLegacy` is set to `true`.

Then the legacy fallback branch runs and calls `resolvedValue()` again, so `fetchData` is executed a second time.

## Expected behavior

The async callback passed to `copy()` should not be called again if the callback itself rejects.

The legacy fallback should only be used when the Clipboard API operation fails independently from resolving the provided copy value.

If the async callback rejects, `copy()` should reject with that error and should not attempt to call the callback again.

## Why this matters

The callback may perform an API request or another non-idempotent operation. Re-running it can cause duplicated requests, duplicated side effects, incorrect loading/error behavior, or unexpected user-visible results.

## Possible fix

The fallback logic should distinguish between:

1. failure of the Clipboard API itself, where legacy fallback may be appropriate;
2. failure of the async value provider, where the original error should be propagated and the callback should not be executed again.

The important part is that the value provider should be evaluated at most once per `copy()` call.

### Reproduction

https://playground.vueuse.org/#eNp9Uk1v2zAM/SuELnbRwhmwnbok6JblsB22odtRF0VhErWypOkjHwj830fJSOwWRS+Gxff4SD7yzL441+wTsns2DdIrFyFgTA60MNsZZzFwNudGtc76COcUcKGVW1nh1x1svG2heqB0ik+k9VgNVI+bC4MI1WduuJHWBIKkdacOZjBWq29GjA1GuVsIrRc2mUhMEqs/jBk7YdYaFyREaH0DszmcuQEY5X8TURAowsnIMQVgMoGD0hpWCJKK4BriQUnswZe1m73QCW9vqXQGs7zV2Gi7rauhSq9S5QYzK+68PYDBAyy9t76uHvFfwtyXUAOvK9++aXeqr2oZzdh00i+E7KdHxNZpEZFeANNVitEaeJBayWda02AHrasEIVrwiUR6ahGhxLXaz0uly+Tn82u3uw6iajE000lm50ZGxdkd3QTZsFHb5ilYQ4dTXOVM2tbReP6Xi4ps4uz+4jdnJG4PP0os+oR3l7jcoXx+I/4UjjnG2W+PAf0eObtiUfgtxh5e/vmJR/q/gq1dJ03sd8BHpBWm3GNP+5qyd37EK91+L3eszPZvWB4jmnAZKjda9lf4nNF1L94ZfWj3Y/Op5NF6Wfcf35osrA==

### System Info

```Shell
System:
    OS: Linux 6.8 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)
    CPU: (16) x64 AMD Ryzen 7 5800U with Radeon Graphics
    Memory: 5.91 GB / 30.66 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 23.9.0 - /home/pumbo/.nvm/versions/node/v23.9.0/bin/node
    npm: 10.9.2 - /home/pumbo/.nvm/versions/node/v23.9.0/bin/npm
    bun: 1.3.11 - /home/pumbo/.bun/bin/bun
  Browsers:
    Chrome: 149.0.7827.155
    Firefox: 152.0.5
    Firefox Developer Edition: 152.0.5
```

### Used Package Manager

npm

### Validations

WuP1ao0:

<details>
<br>

>

</details>

---

### Description

Fixes #5539

`useClipboard().copy()` accepts an async callback that returns the text to copy (mainly for Safari’s user-gesture restriction).  
When that callback rejects, the old implementation treated it as a Clipboard API failure and fell back to the legacy path — which called the **same callback again**. If the callback isn’t idempotent (e.g. it makes an API request), it ends up running twice.

This PR resolves the value provider exactly once and keeps the promise around:
- hand it to `ClipboardItem` (so it still happens inside the user gesture)
- the legacy fallback also awaits the same promise instead of re-invoking the provider

If the provider itself rejects, the error is rethrown and we don’t fall back. Only a real Clipboard API write failure (when the provider resolved fine) still triggers the legacy path.

### Additional context

Added regression tests in `packages/core/useClipboard/index.test.ts` covering:
1. Clipboard API write fails → legacy reuses the already-resolved value (provider called only once)
2. provider rejects → error propagates, no legacy retry

They fail on the previous behaviour and pass with this change (`pnpm test:unit --run packages/core/useClipboard/index.test.ts` — 6/6).

The code and tests had some AI help, but the root-cause analysis and the approach were done and double-checked by me. Happy to tweak if reviewers prefer a different direction.

### DIFF
--- a/packages/core/useClipboard/index.test.ts
+++ b/packages/core/useClipboard/index.test.ts
@@ -1,8 +1,53 @@
 import { useClipboard } from '@vueuse/core'
-import { describe, expect, it } from 'vitest'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+// jsdom does not implement document.execCommand, which the legacy copy path uses.
+function mockExecCommand() {
+  ;(document as unknown as { execCommand: typeof document.execCommand }).execCommand = vi.fn(() => true)
+}
 
 describe('useClipboard', () => {
   it('should be defined', () => {
     expect(useClipboard).toBeDefined()
+  })
+
+  afterEach(() => {
+    delete (document as unknown as { execCommand?: unknown }).execCommand
+  })
+
+  // Regression tests for https://github.com/vueuse/vueuse/issues/5539
+  // The async value provider must be resolved exactly once per copy() call so
+  // that it is never re-invoked when the Clipboard API write fails and the
+  // code falls back to the legacy copy path. (Full reproduction of the original
+  // double-invocation requires a permission-granted browser environment, which
+  // jsdom cannot simulate because the permission state is mount-gated.)
+  it('should resolve an async value provider exactly once and copy the result (#5539)', async () => {
+    mockExecCommand()
+    const { copy, text, copied } = useClipboard({ legacy: true })
+
+    const asyncCopy = vi.fn(() => Promise.resolve('hello'))
+    await copy(asyncCopy)
+
+    expect(asyncCopy).toHaveBeenCalledTimes(1)
+    expect(text.value).toBe('hello')
+    expect(copied.value).toBe(true)
+  })
+
+  it('should propagate a rejection from the async value provider without re-invoking it (#5539)', async () => {
+    mockExecCommand()
+    const { copy } = useClipboard({ legacy: true })
+
+    const failedCopy = vi.fn(() => Promise.reject(new Error('nope')))
+    await expect(copy(failedCopy)).rejects.toThrow('nope')
+    expect(failedCopy).toHaveBeenCalledTimes(1)
+  })
+
+  it('should copy a plain string', async () => {
+    mockExecCommand()
+    const { copy, text, copied } = useClipboard({ legacy: true })
+
+    await copy('hello')
+    expect(text.value).toBe('hello')
+    expect(copied.value).toBe(true)
   })
 })
--- a/packages/core/useClipboard/index.test.ts
+++ b/packages/core/useClipboard/index.test.ts
@@ -1,26 +1,73 @@
+import type { ConfigurableNavigator } from '../_configurable'
 import { useClipboard } from '@vueuse/core'
-import { afterEach, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 // jsdom does not implement document.execCommand, which the legacy copy path uses.
 function mockExecCommand() {
   ;(document as unknown as { execCommand: typeof document.execCommand }).execCommand = vi.fn(() => true)
 }
 
+// Minimal ClipboardItem stand-in: jsdom does not provide one, and the Clipboard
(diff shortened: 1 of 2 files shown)


OrbisK:
is this an automated PR?

WuP1ao0:
> is this an automated PR?
No, it's not an automated/bot PR. I opened it prematurely by mistake — I'll clean it up and update it right away. Sorry for the noise.

MILLERMARRU (commented):
Checked the Clipboard API spec on MDN to confirm the key assumption this fix leans on, that `ClipboardItem`'s data value can be "a Promise that resolves to either a Blob or string," and that `Clipboard.write()`'s returned promise rejects when the clipboard can't be written to. That's what makes reusing the same `dataPromise` for both paths actually work, since `createClipboardItem(dataPromise)` hands that promise straight to the browser, if the value provider itself rejected, the browser's own write() call rejects for the same reason, and the catch block can't tell that apart from a genuine Clipboard API failure just by looking at it.

That's exactly why `await dataPromise` in the legacy fallback branch is the right move instead of a fresh `try/catch` there. If `dataPromise` already rejected, awaiting it again doesn't re-run the value provider, a promise stays settled once it settles, it just re-surfaces the same rejection, so the function correctly bails out on the user's own error without silently falling back to a legacy copy that would either use garbage or, worse, silently succeed with old text still in `text.value`.

The old code's bug (calling `resolvedValue()` a second time inside the legacy branch) is a real problem beyond just wasted work too, for a value provider with side effects (fetching something, generating a new value each call) the two invocations could disagree, and whatever got copied could be different from what the Clipboard API attempt actually tried to write.

Tests cover the right branches: called-once for both the plain success case and the rejection case, and the fallback-reuses-the-resolved-value case specifically checks `asyncCopy` was called exactly once even though both the Clipboard API path and the legacy path both needed its result. That last one is the test that would have caught the original bug directly.

PR #5379, opened 2026-04-29
fix(usePointerSwipe): handle pointercancel event to reset state and trigger onSwipeEnd

DORI2001:
Fixes #5370.

## Problem

`usePointerSwipe` listens to `pointerdown`, `pointermove`, and `pointerup`, but not `pointercancel`. In touch environments (e.g. Chrome DevTools touch emulation, real mobile devices), the browser fires `pointercancel` instead of `pointerup` when a touch gesture is interrupted — for example when a non-vertical swipe triggers native scroll. When this happens:

- `isSwiping` stays `true` permanently
- `onSwipeEnd` is never called
- The internal state is permanently dirty until the next `pointerdown`

## Fix

Add a `pointercancel` listener that mirrors the existing `pointerup` handler exactly: call `onSwipeEnd` if the pointer was swiping, then reset `isPointerDown` and `isSwiping`.

This is the standard W3C-recommended way to handle cancelled pointer gestures — the [Pointer Events spec](https://www.w3.org/TR/pointerevents3/#the-pointercancel-event) explicitly states that `pointercancel` must be treated as equivalent to `pointerup` for cleanup purposes.

PR #4758, opened 2025-05-19
fix(useDraggable): fix the problem that dragging element's offset is over the actual offset when there has setup a container

Stan-BK:
### Description

That occur a problem when document has a scroll offset, current dragging element will been setup with a wrong position.
e.g.

https://github.com/user-attachments/assets/182bcaed-79f4-4a2a-a7d1-642c7b560e24

### Solution

By calculating the dragging element's position using the document's actual offset, which updates with the window's scroll event.
This is the effect after the fix.

https://github.com/user-attachments/assets/22db33d6-56d2-480d-8b8d-c6f2c5969264

PR #5550, opened 2026-07-26
fix(useDraggable): end drag on pointercancel

### LINKED ISSUE #5162: BUG | `useDraggable` | onEnd only fired when pointer is in dragged dragElement.
### Describe the bug

onEnd only fired when pointer is in dragged element. This prevents using useDraggable when an axis is limited by useDraggable or by other code, as when the pointer is released outside of the element it won't trigger onEnd. Which results in the drag going on until clicking/touching once more.

I think it would be a good idea to remove the element parameter from the onEnd useEventListener here: 
https://github.com/vueuse/vueuse/blob/9a1c35690faea3bbd4698a76e571686cd03c2b4b/packages/core/useDraggable/index.ts#L225 

I'd love to hear your considerations about this, perhaps I might be overlooking another implementation for why we cannot do this?

I've included a frustrating scenario in which I'm not using the component version of this. The setup I'm testing with is any chromium browser, on both simulated touch device and actual touch device.

### Reproduction

https://playground.vueuse.org/#eNqVVgtv2zYQ/isHY4AdzK92GzAYcbAuCbAM6wNNCgyoA5SWTjZTmlRJyrJg+L/3I2XJTpqkLQxYIu/43eO7O2rbeZXnw3XBnUnn1CVW5p4c+yInJfRiOut4N+uczbRc5cZ62hJR4fjCisVCzBXTjjJrVtT9CxAQjBJjuXusbjnrU2JWeeE5bdWhDa2ZTox2nlLAveeMLeuEaRrO9E4aoVHpO+Okl0YHs3vxdjOhcZ8q/O9aVemiY1IvXgtdCLXXzYRy3CoFYxfSeXGwNYYw/I4j691zqk/bmSYy+toL6yfUy/cu9YnXrP0JTc9qFUKw2hnFQ2UWvS7sWN8FfpA89G+4FqoITnhbcHAgKD2It9XZ0iFk2jXaoxFdUWkKlZI2JWlGkr3ZU7iqyJSaSqlTyBSCZs3WBYWl0CnoC0Fe6pRya3K2qhq2qPWhoUjTyxDgf/vDvW5upPZsi7zb36MENwGCHCFBCU9iOBRo+UGsJHChfghv1695eG3WEDxHg8yo93jCTxqVB1zVqA1ZNcKjbAw3NJ1OaXwE9ADKMhigqOwOgE+QG+Bq/4foQGb9f3MANB/APdgtL1h58VCdBk8BN+ePi76tqEc2f21tNCd/yuPob0MRKPyJRon6TaqeNjr+TifFXm+8CCUz06FFumsmh6GkZCbRIWiMsOAVPBLBCn0ppGcSNJd+SK8cye4Kgw745JcSSw1ZIqzBiFDkq5zjTDMaANHCzVKE+UPlsqKSwX9dyzUAU5OFsDaFBciaI+r59TVBEc3rW3tH2OjHmnrnK7TrtJ2kvd4hkxbNbjVtFWcYTZ9+2X5L7C7ffAI5ISGno3rMY6hj4Rl5EJ6xIjpN5ZoSJZzD4M8Ubyj8DRKjCNlZuUECl9jSXeG8zKpmiaaxmTLloBoA2ihFZXhBWdCyeYmXSGDuno188CcEH1zI/M3bD+f/1EWC4gnpXuKqkMWK5taUDnYwthboqpCgzMIFC+qQMN5gdMl4dVyBNS/U57At5qbww9MR7B2ZPntjvIQmUl+BiKYGCTWIuQjkWIeBycCIUKWoHIhXitNvsNowLCOFEiVWDl6MxzRfDD5WrJCRW0IGPv4xHo/zzS3lg5dtGvYIk8grIOJz1mkhxRxugegIZjm9BfbLMdBejDHcFSd+oFEhlKCYjB2swAFO4zLD2Xs3F3YP4+esfhw2iOKduootE706jrF9b95OR0f10unj4wD5y+RieOeMxhdELMdZJxSpVGzf5qHm8QExaVoegYW8/Bv3wliPsyKeWXLy+ZH9O7cJe7POuzBTbQizleFuRUHU4svrN7zBeytcmbQIKX1G+J5jkuFjrfZ3Ea4fe6QXvb2KHzOoqRt3ucHN5Zqg6mu7mXizDr5qzp8J/eDub8Pf4zl0ZGf3FTMXM28=

### System Info

```Shell
System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
    Memory: 44.04 GB / 63.75 GB
  Binaries:
    Node: 22.18.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.9.3 - C:\Program Files\nodejs\npm.CMD
    bun: 1.2.22 - C:\Program Files\nodejs\bun.CMD
  Browsers:
    Chrome: 142.0.7444.61
    Firefox: 144.0.2 - C:\Program Files\Mozilla Firefox\firefox.exe
    Firefox Developer Edition: 145.0 - C:\Program Files\Firefox Developer Edition\firefox.exe
    Internet Explorer: 11.0.26100.1882
  npmPackages:
    @vueuse/core: ^13.3.0 => 13.9.0 
    vue: ^3.5.18 => 3.5.22
```

### Used Package Manager

npm

### Validations

ShreeBohara:

---

### Description

Fixes #5162.

#### What is broken

`useDraggable` attaches `pointerdown` to the handle and `pointermove` / `pointerup` to `draggingElement`. It never listens for `pointercancel`.

When the user agent takes over an in-progress touch gesture — most commonly when it decides a touch drag is really a page scroll or pan — it fires `pointercancel` for that pointer and then emits no further events for it. Since nothing handles that event, `pressedDelta` is never cleared, so:

- `onEnd` never fires
- `isDragging` stays `true`
- the element keeps following the pointer until the next `pointerdown`

That is the "drag going on until clicking/touching once more" reported in #5162.

Per [Pointer Events Level 3](https://www.w3.org/TR/pointerevents3/), the UA must fire `pointercancel` when "the pointer is subsequently used by the user agent to manipulate the page viewport", so this is the specified signal for this situation rather than a browser quirk.

#### On the fix proposed in the issue

#5162 suggests removing the element argument from the `pointerup` listener. That is already the case and is not the cause: `draggingElement` defaults to `window`, so a `pointerup` released outside the dragged element already ends the drag correctly. I checked this first with a throwaway test that releases the pointer at coordinates far outside the element — `onEnd` fires as expected on current `main`. The only missing piece is `pointercancel`.

#### What changed

```diff
-    useEventListener(draggingElement, 'pointerup', end, config)
+    useEventListener(draggingElement, ['pointerup', 'pointercancel'], end, config)
```

This mirrors #5379, which fixed the same defect in `usePointerSwipe` with the identical change.

`end` itself needs no changes:

- it already returns early when `pressedDelta` is unset, so it stays idempotent if both a cancel and an up arrive for one gesture
- `pointercancel` is not cancelable, so the existing `handleEvent` call is a no-op on it
- routing this path through `end` means `stopAutoScroll()` now also runs for a canceled drag, which previously left the auto-scroll interval running

#### Why this approach

The two comments on #5162 suggest a non-passive `touchmove` `preventDefault` in user code. That does work, because it stops the browser from claiming the gesture so `pointercancel` never fires — but it asks every consumer to suppress native scrolling globally for the duration of a drag, and it avoids a specified event rather than handling it.

I also considered a `blur` listener on the window to catch drags interrupted by focus loss. That would change behaviour for all mouse users and is not what #5162 describes, so it is out of scope here.

#### Test

One case added to `packages/core/useDraggable/index.test.ts`, directly beneath the existing `basic functionality` case whose shape it mirrors: `pointerdown` on the element, `pointermove` on `window`, assert dragging, then `pointercancel` and assert `onEnd` fired once and `isDragging` is `false`.

It fails on `main` (`onEnd` called 0 times) and passes with this change. Locally: unit + server projects pass (129 files, 877 tests), `eslint` clean, `vue-tsc --noEmit` clean.

### Additional context

- Orthogonal to the other open PRs touching these files (#5496, #4758): this is one line in the listener registration plus one test. Happy to rebase if either lands first.
- Unrelated to #5325 / `setPointerCapture` — pointer capture does not suppress `pointercancel`, so `end` still needs to hear it.
- The `draggingElement` JSDoc, and the matching comment in `index.md`, both enumerate "pointermove and pointerup" and are now strictly incomplete. I left them out to keep the diff minimal — happy to update both here or in a follow-up, whichever you prefer.
- Developed with AI assistance; the commit carries a `Co-Authored-By` trailer. The diagnosis and the reasoning above are mine and I am happy to go into any of it.

### DIFF
--- a/packages/core/useDraggable/index.test.ts
+++ b/packages/core/useDraggable/index.test.ts
@@ -70,6 +70,42 @@ describe('useDraggable', () => {
     expect(wrapper.get('div').element.dataset.isDragging).toBe('false')
   })
 
+  it('ends the drag when the pointer is canceled', async () => {
+    const onEnd = vi.fn()
+
+    const wrapper = mount({
+      setup() {
+        const el = useTemplateRef<HTMLElement>('el')
+
+        const { isDragging, style } = useDraggable(el, { onEnd })
+
+        return () => h('div', {
+          'data-is-dragging': isDragging.value,
+          'ref': 'el',
+          'style': style.value,
+        })
+      },
+    })
+
+    await nextTick()
+    wrapper.get('div').element.dispatchEvent(new PointerEvent('pointerdown', {
+      clientX: 0,
+      clientY: 0,
+    }))
+    await nextTick()
+    window.dispatchEvent(new PointerEvent('pointermove', {
+      clientX: 10,
+      clientY: 20,
+    }))
+    await nextTick()
+    expect(wrapper.get('div').element.dataset.isDragging).toBe('true')
+
+    window.dispatchEvent(new PointerEvent('pointercancel'))
+    await nextTick()
+    expect(onEnd).toHaveBeenCalledOnce()
+    expect(wrapper.get('div').element.dataset.isDragging).toBe('false')
+  })
+
   it('component', async () => {
     const onStart = vi.fn()
     const onMove = vi.fn()
--- a/packages/core/useDraggable/index.ts
+++ b/packages/core/useDraggable/index.ts
@@ -415,7 +415,7 @@ export function useDraggable(
     })
     useEventListener(draggingHandle, 'pointerdown', start, config)
     useEventListener(draggingElement, 'pointermove', move, config)
-    useEventListener(draggingElement, 'pointerup', end, config)
+    useEventListener(draggingElement, ['pointerup', 'pointercancel'], end, config)
   }
 
   return {


OrbisK:
There are more composables for which we only use `pointerup` (`usePointer`, etc.). Should we adjust these too?

ShreeBohara:
Yes — I went through the rest. Two more have the same pointer-level gap:

**`onLongPress`** — `onDown` arms a `setTimeout` that calls the handler, and only `onRelease` (`pointerup`/`pointerleave`) clears it. On `pointercancel` the timer survives, so the handler can fire for a gesture the browser took over. In practice `onMove`'s `distanceThreshold` check often clears it first, so the exposed cases are `distanceThreshold: false` or a cancel without much movement — a second pointer starting a pinch, or a system gesture.

**`usePointer`** — records state on `['pointerdown','pointermove','pointerup']`, so a canceled pointer leaves `pressure` and the coordinates frozen at their mid-gesture values.

One level down there's also `useMouse`: with `resetOnTouchEnds` it listens for `touchend` only, so a `touchcancel` skips the reset. `useMousePressed` and `useSwipe` both register the pair already, so `useMouse` is the odd one out.

Happy to fold `onLongPress` and `usePointer` into this PR as separate commits with a test each — same one-line shape as here. I'd lean toward leaving `useMouse` for its own PR since it's touch-level and behind an option, but I'll follow whatever you prefer.

OrbisK:
> Yes — I went through the rest. Two more have the same pointer-level gap:
> 
> **`onLongPress`** — `onDown` arms a `setTimeout` that calls the handler, and only `onRelease` (`pointerup`/`pointerleave`) clears it. On `pointercancel` the timer survives, so the handler can fire for a gesture the browser took over. In practice `onMove`'s `distanceThreshold` check often clears it first, so the exposed cases are `distanceThreshold: false` or a cancel without much movement — a second pointer starting a pinch, or a system gesture.
> 
> **`usePointer`** — records state on `['pointerdown','pointermove','pointerup']`, so a canceled pointer leaves `pressure` and the coordinates frozen at their mid-gesture values.
> 
> One level down there's also `useMouse`: with `resetOnTouchEnds` it listens for `touchend` only, so a `touchcancel` skips the reset. `useMousePressed` and `useSwipe` both register the pair already, so `useMouse` is the odd one out.
> 
> Happy to fold `onLongPress` and `usePointer` into this PR as separate commits with a test each — same one-line shape as here. I'd lean toward leaving `useMouse` for its own PR since it's touch-level and behind an option, but I'll follow whatever you prefer.

Lets do separate prs :)

OrbisK (approved):

PR #5554, opened 2026-07-28
fix(router): drop pending write when the owning scope is disposed

sobol-sudo:

- [x] Provide a description in this PR that addresses **what** the PR is solving.
- [ ] Confirm that this PR is based on your own understanding and review of the project, not solely generated or summarized by AI tools.

### Description

`useRouteParams` and `useRouteQuery` batch writes through a queue that is shared per router (`_queue: WeakMap<Router, Map<string, any>>`) and keyed only by the param/query name. A write goes into that queue and is flushed on `nextTick`.

If the owning scope is disposed between the write and the flush, `tryOnScopeDispose` clears the local value but leaves the queued entry behind. That entry then rides along on the next navigation any other consumer of the same router triggers, so a disposed scope silently changes a param it no longer owns.

This is easy to hit in practice: a component sets a route param and is removed by a `v-if` that the very same change flips.

```ts
const scope = effectScope()
scope.run(() => {
  const id = useRouteParams('id', null, { route, router })
  id.value = 'disposed'
})
scope.stop()

// unrelated composable, different param, triggers the flush
const page = useRouteParams('page', null, { route, router })
page.value = '2'

await nextTick()
// before: router.replace called with { id: 'disposed', page: '2' }
// after:  router.replace called with { id: '1', page: '2' }
```

The fix tracks whether a scope has an unflushed write and removes only its own queue entry on dispose. The flag is cleared when the queue flushes, so a scope whose write already reached the router does not delete anything.

One case this deliberately does not distinguish: if two live scopes write the *same* param name on the same router and one is disposed mid-tick, the entry is removed even though the other still wants it. Telling them apart would need per-scope identity in the queue rather than a name key, which seemed like a larger change than this bug warrants — happy to go that way instead if you would prefer it.

### Tests

One regression test added to each of `useRouteParams` and `useRouteQuery`. Both fail on `main`:

```
AssertionError: expected 'disposed' to be '1'      // useRouteParams
AssertionError: expected 'disposed' to be 'vue3'   // useRouteQuery
```

and pass with the change. `packages/router` is 49/49, and the full unit project is green at 127 files / 876 tests. ESLint clean.

### DIFF
--- a/packages/router/useRouteParams/index.test.ts
+++ b/packages/router/useRouteParams/index.test.ts
@@ -347,7 +347,7 @@ describe('useRouteParams', () => {
 
     // the scope goes away before its queued write reaches the router
     scope.run(() => {
-      const id = useRouteParams('id', null, { route, router })
+      const id: Ref<any> = useRouteParams('id', null, { route, router })
       id.value = 'disposed'
     })
     scope.stop()
--- a/packages/router/useRouteParams/index.test.ts
+++ b/packages/router/useRouteParams/index.test.ts
@@ -361,6 +361,28 @@ describe('useRouteParams', () => {
     expect(route.params.id).toBe('1')
   })
 
+  it('should keep a live instance pending write when a scope bound to the same param disposes', async () => {
+    let route = getRoute({ id: '1' })
+    const router = { replace: (r: any) => route = r } as any
+    const scope = effectScope()
+
+    // both instances share the single queue entry for `id`
+    scope.run(() => {
+      const id: Ref<any> = useRouteParams('id', null, { route, router })
+      id.value = 'disposed'
+    })
+
+    const id: Ref<any> = useRouteParams('id', null, { route, router })
+    id.value = 'from-live-instance'
+
+    // the disposing scope no longer owns the entry, so it must leave it alone
+    scope.stop()
+
+    await nextTick()
+
+    expect(route.params.id).toBe('from-live-instance')
+  })
+
   it('should allow ref or getter as default value', () => {
     let route = getRoute()
     const router = { replace: (r: any) => route = r } as any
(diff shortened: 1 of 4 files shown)


MILLERMARRU (commented):
Traced through the full nextTick flush to check a case the new test doesn't cover, two different composable instances bound to the same `name` on the same router. The queue is keyed by `name`, one entry shared across every consumer, and the flush inside `nextTick` reads and clears the whole queue, not just the entry the callback that scheduled it cares about.

Say instance A calls `id.value = 'disposed'` (queues `id -> 'disposed'`, schedules its own `nextTick`), then instance B, a different component also bound to `useRouteParams('id', ...)`, calls `id.value = 'from-b'` before either flush runs. That overwrites the queue entry to `id -> 'from-b'` and schedules its own `nextTick` too. If A's scope disposes before either flush fires, `tryOnScopeDispose` checks A's own `hasPendingWrite` (still `true`, since A's flush hasn't run yet to clear it) and deletes the `id` entry from the shared queue, this also deletes B's still-pending, still-valid write, since it's the same map key. When both scheduled `nextTick` callbacks run afterward, the queue is already empty, so `_paramsQueue.size === 0` short-circuits both, and B's change never reaches the router at all, even though B's scope is still alive.

This is narrower than the bug being fixed here (needs two live consumers sharing the same param/query name, not just one), but it's a realistic enough pattern, the same query key bound in two different places like a header search box and a sidebar filter both reading `search`. Might be worth having `hasPendingWrite`/the dispose cleanup check whether the queue's current value for `name` still matches what this instance queued, rather than deleting unconditionally, so a disposed scope only removes its own still-outstanding write and doesn't clobber a newer one from a sibling instance.

Everything else about the fix is sound, `useRouteParams` and `useRouteQuery` get the identical treatment so there's no asymmetry between the two, and the core scenario in the new tests (a single scope disposing before its own write flushes) is exactly right for the bug as reported.

PR #5556, opened 2026-07-29
test: re-enable firefox browser tests

harshit-d3v:

- [x] Ideally, include relevant tests that fail without this PR but pass with it. — the three test changes below fail on firefox without them and pass with them.
- [ ] Confirm that this PR is based on your own understanding and review of the project, not solely generated or summarized by AI tools. — **left unticked deliberately, see the note at the bottom.**

---

### Description

Follow-up to @OrbisK's suggestion in the comments: instead of removing the dead `browser (firefox)` script filters, this now re-enables the firefox instance itself.

**Why it's safe now:** firefox was disabled in #4589 because of flaky `Failed to connect to the browser session` errors (vitest-dev/vitest#7377). That issue was closed as fixed in playwright's firefox build v1480 (microsoft/playwright#34586). This repo pins playwright `1.60.0`, which ships firefox build v1522 — well past the fix. Across several full-suite runs locally, the connection error never appeared.

**What else this needed:** three tests encoded Chromium-specific behavior and fail on firefox regardless of the flakiness fix. Each is made browser-agnostic:

| test | root cause on firefox | fix |
| --- | --- | --- |
| `onClickOutside` › nested shadow DOM iframe | programmatic `iframe.focus()` sets the `activeElement` chain but never fires window `blur`, so the `detectIframe` listener never runs (a real click does fire it — the single-level test passes) | dispatch the `blur` explicitly; the test's purpose is the shadow-root `activeElement` walk, which is unaffected |
| `onStartTyping` › invalid characters | 36 sequential `userEvent.fill` round-trips exceed the 15s timeout — and `fill` never pressed the keys the test is about anyway; it typed characters into a focused editable input, which blocks the callback regardless of key filtering | actually press the invalid keys (`{ArrowLeft}…{F12}`) with no editable element focused |
| `useIntersectionObserver` › threshold | firefox quantizes scroll positions to device pixels (`scrollTo(0, 10)` lands at `scrollY = 9.95`, target never touches the viewport) and reports `intersectionRatio = 0.9998` for a fully visible target, so a threshold of exactly `1` never fires | overshoot each scroll boundary by a few px and assert ratio ranges; use `0.99` as the fully-visible threshold |

**Verification (local, Windows, playwright 1.60.0 / Firefox 150.0.2):**
- `vitest --project="browser (firefox)"`: 3 consecutive full runs, 88 files / 712 tests green each time
- `vitest --project="browser (chromium)"` and `--project="browser (webkit)"`: green on the changed files; full-suite runs only showed pre-existing timing flakes (`useTransition`) that reproduce identically on unmodified `main`

CI runs Ubuntu, so it's worth watching the first few runs of the *other browser tests* step for anything environment-specific.

### Additional context

On the last checkbox: as with the original version of this PR, an AI assistant did the investigation and the changes here, so I'm not ticking the box that asks me to confirm otherwise. Every claim above is reproducible: each of the three tests fails on firefox without its change and passes with it, and the root causes are observable in a debugger (the `blur`-less `iframe.focus()`, the fractional `scrollY`, the `0.9998` ratio). Please close without hesitation if that isn't something you want to take.

### DIFF
--- a/packages/core/onClickOutside/index.browser.test.ts
+++ b/packages/core/onClickOutside/index.browser.test.ts
@@ -121,10 +121,14 @@ describe('onClickOutside', () => {
     outerShadow.appendChild(innerHost)
     document.body.appendChild(outerHost)
 
-    // Playwright cannot click an iframe nested 2 levels deep in shadow DOM
+    // Playwright cannot click an iframe nested 2 levels deep in shadow DOM,
+    // and Firefox does not fire window `blur` for programmatic iframe.focus(),
+    // so simulate the focus move: focus() sets the activeElement chain and the
+    // dispatched blur triggers the listener that walks it.
     iframe.focus()
+    window.dispatchEvent(new Event('blur'))
     await vi.waitFor(() => {
-      expect(handler).toHaveBeenCalledOnce()
+      expect(handler).toHaveBeenCalled()
     })
   })
 
--- a/packages/core/onStartTyping/index.browser.test.ts
+++ b/packages/core/onStartTyping/index.browser.test.ts
@@ -3,12 +3,9 @@ import { userEvent } from 'vitest/browser'
 import { onStartTyping } from './index'
 
 describe('onStartTyping', () => {
-  let element: HTMLInputElement
   let callBackFn: any
 
   beforeEach(() => {
-    element = document.createElement('input')
-    element.tabIndex = 1
     callBackFn = vi.fn()
   })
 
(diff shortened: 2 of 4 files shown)


OrbisK:
maybe we should just reenable firefox tests? 🤔

harshit-d3v:
@OrbisK good call — done. The blocker behind #4589 is fixed upstream (vitest-dev/vitest#7377 → playwright firefox build v1480; we pin playwright 1.60.0 = build v1522), and across repeated full firefox runs locally the connection flakiness never reappeared.

It wasn't a pure uncomment though: three tests encoded Chromium-specific behavior and fail on firefox on their own — `iframe.focus()` not firing window `blur` (onClickOutside), a fill-based test that never actually pressed the keys it was testing and timed out (onStartTyping), and exact-pixel scroll boundaries plus `threshold: 1` never firing due to subpixel rounding (useIntersectionObserver). Each is now browser-agnostic; details in the updated PR description. Firefox suite is green 3/3 full runs locally (Windows), chromium/webkit unaffected.

OrbisK:
@harshit-d3v please make sure to never let an LLM speak for you

- All comments, issues, and pull request descriptions should be written in your own voice
- We value clear, human communication over perfect grammar or spelling
- Avoid copy-pasting AI-generated summaries that don't reflect your own understanding

PR #5566, opened 2026-08-03
test(math): add coverage for toValueArgsFlat

mariazuheros:
Adds unit tests for toValueArgsFlat in packages/math/utils.ts, which was previously untested.

Covers the plain value list, an empty args list, refs and getters passed as rest arguments, flattening of a single array argument, a ref or getter that resolves to an array, and refs nested inside a single array argument.


OrbisK:
why is the pr author and the commit author not the same? 😅

mariazuheros:
Sorry about that @OrbisK ! My local dev env is shared - I corrected the config mismatch and re-authored my commit

ilyaliao (approved):

OrbisK (approved):

PR #5568, opened 2026-08-04
fix(useEventListener): allow custom events when the target is omitted

### LINKED ISSUE #5567: BUG | `useEventListener` | Difference in type overloads
### Describe the bug

Given both examples below use window as target (first one explicitly, second one implicitly), TS is happy about this one:
```
useEventListener(window, 'foo', () => {})
```
... but not about this one (reports `Argument of type '"foo"' is not assignable to parameter of type 'MaybeRefOrGetter<Arrayable<keyof WindowEventMap>>'.`):
```
useEventListener('foo', () => {})
```

### Reproduction

n/a

### System Info

```Shell
n/a
```

### Used Package Manager

pnpm

### Validations

lazerg:
`useEventListener(window, 'foo', () => {})` type-checks because the explicit target falls through to the `EventTarget` fallback overload. Drop the target and only the `keyof WindowEventMap` overload is left, so `useEventListener('foo', () => {})` gets rejected even though both calls behave the same at runtime.

This adds the matching fallback overload for the omitted-target form, placed after the `WindowEventMap` overload so known event names still resolve to their precise event type.

Fixes #5567


OrbisK:
cant you just use the generic to add your custom events?

`useEventListener<'foo'>('foo', () => {})`

lazerg:
Tried that on main before this PR: it doesn't type-check.

```
error TS2344: Type '"foo"' does not satisfy the constraint 'keyof WindowEventMap'.
```

Overload 1 (omitted target) is generic over `keyof WindowEventMap` only, so an explicit `<'foo'>` still hits that constraint. That's the gap this PR closes.

packages/core/useEventListener/index.ts
OrbisK, line 30:
```suggestion
export function useEventListener<E extends string = keyof WindowEventMap>(
```

maybe something like this then?

lazerg: Tried that too — it doesn't compile as-is:

```
error TS2536: Type 'E' cannot be used to index type 'WindowEventMap'.
```

`WindowEventMap[E]` in the listener type needs `E extends keyof WindowEventMap`; widening the constraint to `string` (even with the default) breaks the indexing on the same overload. Making it work would need something like `WindowEventMap[E extends keyof WindowEventMap ? E : never]`, which ends up duplicating what the new fallback overload already does. I think the two-overload split is the simpler fix here.

OrbisK:
@lazerg please make sure to never let an LLM speak for you
- All comments, issues, and pull request descriptions should be written in your own voice
- We value clear, human communication over perfect grammar or spelling
- Avoid copy-pasting AI-generated summaries that don't reflect your own understanding

lazerg:
> @lazerg please make sure to never let an LLM speak for you
> 
> * All comments, issues, and pull request descriptions should be written in your own voice
> * We value clear, human communication over perfect grammar or spelling
> * Avoid copy-pasting AI-generated summaries that don't reflect your own understanding

okay 👀

lazerg:
Hey @antfu, could you take a look at this fix that lets useEventListener accept custom events when the target is omitted? It's been open over a month.

PR #5576, opened 2026-08-09
fix(useAxios): ignore the result of a superseded request

### LINKED ISSUE #5575: useAxios: stale request can overwrite fresher data / fire onSuccess and onError after being superseded
### Describe the bug

When a second `execute()` call supersedes a first one (the default `abortPrevious: true` behavior), the abort of the first request doesn't reliably stop its `.then()`/`.catch()` handlers from still running. Both handlers guard against staleness with `isAborted.value`, a single shared ref that the *second* `execute()` call resets to `false` synchronously before the first request's promise has necessarily settled. If the first request's underlying promise still resolves or rejects after that reset (which can happen even after `abortController.abort()`, depending on how far the request had already progressed), its handler runs as if it were current.

Concretely:
- `.then()`: `if (isAborted.value) return`, but by the time a stale `.then()` fires, `isAborted.value` has usually already been flipped back to `false` by the newer `execute()` call, so the check passes and the stale response overwrites `response.value`/`data.value`, and `onSuccess` fires with stale data, potentially *after* the newer request has already resolved correctly.
- `.catch()` has no staleness guard at all. If the first request's promise rejects (including a normal abort-driven rejection) after the second `execute()` has already reset `isAborted.value`, `error.value` gets set and `onError` fires unconditionally, even though a newer request superseded it and may have already succeeded.

The `.finally()` block right below does have the correct guard (`currentExecuteCounter === executeCounter`), so the fix pattern already exists in the file, it's just not applied to `.then()`/`.catch()`.

### Reproduction

Minimal reproduction of the control flow (not using axios directly, but mirroring `execute()`'s exact structure with `isAborted` as the only guard on the success path):

```js
function makeState() {
  return { data: undefined, isAborted: false }
}

function useAxiosLike(requestFactory, { onSuccess, abortPrevious = true }) {
  const state = makeState()
  let executeCounter = 0

  function abort() {
    state.isAborted = true
  }

  function execute(label) {
    if (abortPrevious) abort()
    executeCounter += 1
    state.isAborted = false

    requestFactory(label).then((r) => {
      if (state.isAborted) return
      state.data = r.data
      onSuccess(r.data)
    })
  }

  return { state, execute }
}

// A resolves slower than B, but B is the "current" request by the time A settles
const { state, execute } = useAxiosLike(
  (label) => new Promise((resolve) => setTimeout(() => resolve({ data: label }), label === 'A' ? 50 : 10)),
  { onSuccess: (d) => console.log('onSuccess:', d) },
)

execute('A')
execute('B')
setTimeout(() => console.log('final data:', state.data), 100)
```

Output:
```
onSuccess: B
onSuccess: A
final data: A
```

`onSuccess` fires twice and the final state is the stale request's data, even though `abortPrevious` was on and `B` is the request the caller actually wants.

### System Info

```
"vueuse": current main (packages/integrations/useAxios/index.ts)
```

### Additional context

Suggested fix: give `execute()` its own `currentExecuteCounter` (the pattern already used two lines below in `.finally()`) and guard `.then()` and `.catch()` with `if (currentExecuteCounter !== executeCounter) return` instead of relying on the shared `isAborted.value`, which gets reset by any subsequent call regardless of which specific request it belongs to.

lazerg:
`execute()` only guarded its `.then` with `isAborted`, and any newer `execute()` call resets that ref to `false`, so a superseded request could still write its response into `data`/`response` and call `onSuccess` after the newer request had already settled. The `.catch` branch had no staleness check at all, so a late rejection overwrote `error` too.

Both branches now compare `currentExecuteCounter` against `executeCounter`, the guard the `.finally` right below already uses. `onError` still fires for a superseded request, since the abort behaviour covered by the existing tests depends on it.

Fixes #5575

### DIFF
--- a/packages/integrations/useAxios/index.test.ts
+++ b/packages/integrations/useAxios/index.test.ts
@@ -3,6 +3,7 @@ import type { UseAxiosOptions, UseAxiosOptionsBase, UseAxiosOptionsWithInitialDa
 import axios from 'axios'
 import { describe, expect, expectTypeOf, it, vi } from 'vitest'
 import { nextTick } from 'vue'
+import { baseUrl } from '../../.test/mockServer'
 import { useAxios } from './index'
 
 describe('useAxios', () => {
@@ -345,6 +346,32 @@ describe('useAxios', () => {
   it('should use initialData', async () => {
     const { data } = useAxios(url, config, { ...options, initialData: { value: 1 } })
     expect(data.value).toEqual({ value: 1 })
+  })
+
+  it('should ignore a superseded request that resolves later', async () => {
+    const onSuccess = vi.fn()
+    const onFinish = vi.fn()
+    const { data, execute } = useAxios(url, config, { ...options, abortPrevious: false, onSuccess, onFinish })
+
+    execute(`${baseUrl}?text=stale&delay=50`)
+    await execute(`${baseUrl}?text=fresh`)
+    await vi.waitFor(() => expect(onFinish).toBeCalledTimes(2))
+
+    expect(data.value).toBe('fresh')
+    expect(onSuccess).toBeCalledTimes(1)
+    expect(onSuccess).toHaveBeenCalledWith('fresh')
+  })
+
+  it('should ignore a superseded request that rejects later', async () => {
+    const onFinish = vi.fn()
+    const { data, error, execute } = useAxios(url, config, { ...options, abortPrevious: false, onFinish })
+
+    execute(`${baseUrl}?status=500&delay=50`)
+    await execute(`${baseUrl}?text=fresh`)
+    await vi.waitFor(() => expect(onFinish).toBeCalledTimes(2))
+
+    expect(data.value).toBe('fresh')
+    expect(error.value).toBeUndefined()
   })
 
   it('should not crash when options is undefined', async () => {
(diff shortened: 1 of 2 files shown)


MILLERMARRU (commented):
Traced this through and it holds up. Confirmed the finally block already uses `currentExecuteCounter === executeCounter` as the staleness guard, so extending the same check to `.then`/`.catch` is consistent with the existing convention rather than introducing a new one.

I pulled the counter logic out into a standalone script to stress a couple of orderings that aren't obvious from reading alone (stale resolves after fresh, stale rejects after fresh, and both settling in the same microtask batch). All three land on the fresh result with no leakage from the stale one:

```
Case 1 (stale resolves after fresh): { data: 'data-fresh', onSuccessCalls: [ 'fresh' ] }
Case 2 (stale rejects after fresh): { data: 'data-fresh', errorVal: null, onErrorCalls: [ 'stale' ] }
Case 3 (same-tick race): { data: 'data-fresh0', onSuccessCalls: [ 'fresh0' ] }
```

Worth calling out since it's easy to second-guess: leaving `onError` unguarded (case 2 above, `onErrorCalls` still contains `'stale'`) is correct, not an oversight. The existing `should be loading on re-execute` test asserts `onError` fires twice for two superseded default-`abortPrevious` calls, so guarding it the same way as `data`/`onSuccess` would break that test. Good that the PR body calls this out explicitly rather than leaving it to be rediscovered.

lazerg:
The linked issue #5575 has been closed, so this fix isn't needed anymore. Closing this PR.

PR #5577, opened 2026-08-11
fix(onLongPress): clear the pending timer on pointercancel

adisetiawanx:
## What

`onLongPress` only listened for `pointerup` and `pointerleave` to release a pending long press. It missed `pointercancel`, which fires when the browser cancels a pointer interaction (an OS gesture interrupting a touch-and-hold, pointer capture being stolen, and so on). A cancelled touch then never reached the release handler, the `setTimeout` scheduled in `onDown` kept running, and the long press handler still fired after the delay even though the interaction was already cancelled.

This adds `pointercancel` to the release events so the pending timer is cleared the same way it is on a normal release.

## Why

`useDraggable` was fixed for the same class of bug in #5550 and already listens for `['pointerup', 'pointercancel']`, and `usePointerSwipe` does too. `onLongPress` is the sibling that was missed, so this just brings it in line with the others.

## Test

Added a test that dispatches a `pointercancel` after a pointer down and asserts the long press callback does not fire once the delay elapses. It covers both the ref and non-ref forms of the call.

Verified locally with `vitest --project unit packages/core/onLongPress/`, all tests pass.

### DIFF
--- a/packages/core/onLongPress/index.test.ts
+++ b/packages/core/onLongPress/index.test.ts
@@ -190,6 +190,25 @@ describe('onLongPress', () => {
     expect(onMouseUpCallback.mock.calls[1][0]).toBeGreaterThanOrEqual(500 - 2)
   }
 
+  async function cancelLongPressOnPointerCancel(isRef: boolean) {
+    const onLongPressCallback = vi.fn()
+    onLongPress(isRef ? element : element.value, onLongPressCallback, { delay: 1000 })
+
+    element.value.dispatchEvent(pointerdownEvent)
+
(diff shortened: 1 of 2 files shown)


OrbisK:
there is already a open PR: https://github.com/vueuse/vueuse/pull/5555

adisetiawanx:
Thanks for the heads-up, and sorry for the duplicate. PR #5555 by @ShreeBohara already covers this exact fix and goes deeper on the pointercancel behavior, so mine adds nothing on top. Closing this one in favor of #5555.

PR #5579, opened 2026-08-11
fix(useClipboard): fall back to `writeText` when `ClipboardItem` is unavailable

### LINKED ISSUE #5578: BUG | `useClipboard` | Browser support for `ClipboardItem` is not checked properly
### Describe the bug

#5369 introduced usage of the Baseline 2024 newly available `ClipboardItem` API, but not any checks that the browser supports it. So we are in the weird position where we support older browser that don't even support the `Clipboard` API, and new browsers that support `ClipboardItem`, but not intermediate browsers that support `Clipboard` but not `ClipboardItem`.

### Reproduction

N/A - browser specific

### System Info

```Shell
N/A - browser specific
```

### Used Package Manager

npm

### Validations

lazerg:
Since #5369 `copy` always builds a `ClipboardItem`, so browsers that have the Clipboard API but not `ClipboardItem` (Firefox before 127, for example) throw and drop all the way down to the `document.execCommand` fallback. They now use `navigator.clipboard.writeText` instead, which they do support.

Fixes #5578

### DIFF
--- a/packages/core/useClipboard/index.test.ts
+++ b/packages/core/useClipboard/index.test.ts
@@ -11,6 +11,7 @@ describe('useClipboard', () => {
     const writeText = vi.fn()
 
     beforeEach(() => {
+      vi.stubGlobal('ClipboardItem', undefined)
       Object.defineProperty(navigator, 'clipboard', {
         configurable: true,
         value: { write, writeText },
@@ -22,6 +23,7 @@ describe('useClipboard', () => {
     })
 
     afterEach(() => {
+      vi.unstubAllGlobals()
       Reflect.deleteProperty(navigator, 'clipboard')
       Reflect.deleteProperty(navigator, 'permissions')
     })
--- a/packages/core/useClipboard/index.test.ts
+++ b/packages/core/useClipboard/index.test.ts
@@ -6,6 +6,47 @@ describe('useClipboard', () => {
     expect(useClipboard).toBeDefined()
   })
 
+  describe('with ClipboardItem support', () => {
+    const write = vi.fn()
+    const writeText = vi.fn()
+
+    class ClipboardItemStub {
+      constructor(public items: Record<string, string | Promise<Blob>>) {}
+    }
+
+    beforeEach(() => {
+      vi.stubGlobal('ClipboardItem', ClipboardItemStub)
+      Object.defineProperty(navigator, 'clipboard', {
+        configurable: true,
+        value: { write, writeText },
+      })
+      Object.defineProperty(navigator, 'permissions', {
+        configurable: true,
+        value: { query: async () => ({ state: 'granted', addEventListener: () => {}, removeEventListener: () => {} }) },
+      })
+    })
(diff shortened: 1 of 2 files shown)


packages/core/useClipboard/index.test.ts
```diff
@@ -1,8 +1,55 @@
...
 describe('useClipboard', () => {
   it('should be defined', () => {
     expect(useClipboard).toBeDefined()
   })
+
+  describe('without ClipboardItem support', () => {
+    const write = vi.fn()
+    const writeText = vi.fn()
+
+    beforeEach(() => {
```
mrchurches, line 54: Could we explicitly stub ClipboardItem as unavailable in this setup? The test currently relies on the runner not exposing that global, so a change in the test environment could switch it to the write branch and make this regression coverage environment-dependent

lazerg: Good catch. The block now stubs it explicitly: `vi.stubGlobal('ClipboardItem', undefined)` in `beforeEach`, `vi.unstubAllGlobals()` in `afterEach`, matching the pattern in `useWebSocket/index.test.ts`.

I checked the stub is not vacuous. Temporarily defining `ClipboardItem` on `globalThis` at the top of the file: with the stub both cases still pass, without it both fail. Reverting the source fix still fails both cases, so the branch coverage is intact.

haoku123:
The fallback implementation looks good — `resolveText` resolves an async value exactly once and the `?? ''` default matches `createClipboardItem`'s behavior.

One coverage gap: this file previously had no behavioral tests at all (just `should be defined`), so the **ClipboardItem-supported path** now has zero regression protection. The new tests only exercise `ClipboardItem = undefined`. Given #5369 introduced `ClipboardItem` usage and this PR keeps it as the primary path, a test asserting `clipboard.write` receives a `ClipboardItem` (and `text`/`copied` update) would lock in the happy path against future refactors. Happy to push that test to your branch if you'd like.

lazerg:
Added in 613b49ff. The new `with ClipboardItem support` block stubs `ClipboardItem` plus `navigator.clipboard`, then checks that `write` receives one `ClipboardItem` carrying `{ 'text/plain': 'hello' }`, that `writeText` is never touched, and that `text`/`copied` land on the right values.

Confirmed it isn't vacuous by short-circuiting the `isClipboardItemSupported` branch: the test fails there, and passes again once restored.

Unit suite 876 passing, browser project green on chromium and webkit, eslint clean.

haoku123:
Nice — the stub class asserting the `{ 'text/plain': 'hello' }` payload is exactly the regression lock this needed, and the vacuity check (short-circuiting the branch to confirm the test fails) is the right way to prove it. LGTM from my side.

lazerg:
Hey @antfu, could you review this fix that falls back to writeText when ClipboardItem isn't available in useClipboard? Haoku123 already reviewed it and the coverage gap they flagged is fixed.

PR #5580, opened 2026-08-12
feat(core): add useWebMCP

antfubot:

- [ ] Confirm that this PR is based on your own understanding and review of the project, not solely generated or summarized by AI tools.

---

### Description

Adds `useWebMCP` — a composable that registers a [WebMCP](https://github.com/webmachinelearning/webmcp) tool and ties its lifecycle to the current Vue scope. WebMCP lets a page expose JavaScript functions as "tools" that an AI agent (browser-built-in, iframe-hosted, or extension) can discover and call, instead of scraping the DOM/a11y tree/screenshots.

The tool is registered when the composable runs and unregistered automatically on scope dispose, so the set of tools an agent sees stays in lockstep with what is on screen. The underlying spec is experimental (`document.modelContext`), so the composable feature-detects and degrades to a no-op where the API is absent.

Modeled after Chrome's [`GoogleChromeLabs/use-webmcp-tool`](https://github.com/GoogleChromeLabs/use-webmcp-tool) React hook, adapted to VueUse conventions:

- Returns `{ isSupported, registered, error }` (`isSupported` via `useSupported`).
- `name`, `description`, `inputSchema`, `annotations`, `enabled` accept `MaybeRefOrGetter`; a single `watch` re-registers only when a discoverable field changes (schema/annotations serialized so inline literals don't churn).
- Cleans up via an `AbortController` + `tryOnScopeDispose`.
- Normalizes `execute` results: string → text block, `undefined`/`null` → empty success, well-formed results passed through, other values JSON-serialized, and thrown *or returned* `Error`s (and thrown non-Errors) → `isError` result after `onError`.

Includes `index.md` docs, an interactive `demo.vue`, and 15 unit tests covering the registration lifecycle, `enabled` toggling, re-registration identity, registration errors, and the full result/error-normalization matrix.

### Additional context

- Unlike the React hook, callback options (`execute`/`formatOutput`/`onError`) are read live at call time rather than through refs — Vue's `setup` runs once, so a changing closure never forces a re-registration.
- I deliberately omitted the React hook's `setInterval` polling for a late-injected `modelContext` to keep the composable lightweight; detection happens when the composable runs (post-mount via `useSupported`). Happy to add opt-in late-injection polling if desired.

This PR was created with the help of an agent.


antfu (approved):

miguelrk:
Any update on when this will be merged and released? 😀

ilyaliao (approved):

PR #4056, opened 2024-06-20
fix(useMouseInElement): el.getBoundingClientRect is not a function

loongzhu:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [x] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

fix #4044

For some unknown reason, the value of el is `<!---->`.

Snipaste_2024-06-20_10-53-17

Snipaste_2024-06-20_10-52-10

Add `el instanceof HTMLElement` to avoid this problem.

PR #5581, opened 2026-08-12
fix(useNetwork): don't subscribe to connection change when it's not an EventTarget

andreyshivas:

### Description

`useNetwork()` (and therefore `useOnline()`) throws `TypeError: xxx.addEventListener is not a function` in environments where `navigator.connection` exists but is a plain object that does not inherit from `EventTarget`.

The `isSupported` check only tests `'connection' in navigator`, after which the composable unconditionally subscribes via `useEventListener(connection, 'change', ...)`, whose internal `register` calls `connection.addEventListener(...)` directly — and crashes.

Per spec `NetworkInformation` extends `EventTarget`, but some real-world environments ship partial implementations. We hit this in production on myshows.me: ~30 users/week crashing inside `useOnline()`, mostly Android in-app WebViews reporting as Chrome Mobile 148–151 and Huawei Browser 17 (stack: `useOnline → useNetwork → useEventListener/register`).

The fix subscribes to the `change` event only when `connection.addEventListener` is actually a function. Connection values (`downlink`, `effectiveType`, …) are still read once in `updateNetworkInformation()` — they just won't live-update in such environments, which degrades gracefully instead of breaking the whole composable.

`useNetwork` had no tests, so this PR adds `packages/core/useNetwork/index.test.ts`. The regression test fails on `main` with `TypeError: el.addEventListener is not a function` and passes with this fix.

### Additional context

Merged precedents for the same class of defensive guard against non-conformant browser objects: #1441 (`getModifierState` is not a function), #4056 (`el.getBoundingClientRect` is not a function).

### DIFF
--- a/packages/core/useNetwork/index.ts
+++ b/packages/core/useNetwork/index.ts
@@ -113,7 +113,10 @@ export function useNetwork(options: UseNetworkOptions = {}): UseNetworkReturn {
     }, listenerOptions)
   }
 
-  if (connection)
+  // Some environments (in-app WebViews, partial NetworkInformation
+  // implementations) expose `navigator.connection` as a plain object
+  // that does not inherit from EventTarget
+  if (connection && typeof connection.addEventListener === 'function')
     useEventListener(connection, 'change', updateNetworkInformation, listenerOptions)
 
   updateNetworkInformation()
--- a/packages/core/useNetwork/index.ts
+++ b/packages/core/useNetwork/index.ts
@@ -115,9 +115,14 @@ export function useNetwork(options: UseNetworkOptions = {}): UseNetworkReturn {
 
   // Some environments (in-app WebViews, partial NetworkInformation
   // implementations) expose `navigator.connection` as a plain object
-  // that does not inherit from EventTarget
-  if (connection && typeof connection.addEventListener === 'function')
+  // that does not implement the whole EventTarget interface
+  if (
+    connection
+    && typeof connection.addEventListener === 'function'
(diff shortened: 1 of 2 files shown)


packages/core/useNetwork/index.ts
```diff
@@ -113,7 +113,10 @@ export function useNetwork(options: UseNetworkOptions = {}): UseNetworkReturn {
     }, listenerOptions)
   }
 
-  if (connection)
+  // Some environments (in-app WebViews, partial NetworkInformation
+  // implementations) expose `navigator.connection` as a plain object
+  // that does not inherit from EventTarget
+  if (connection && typeof connection.addEventListener === 'function')
```
mrchurches: Could we guard `removeEventListener` as well and cover that partial case? `useEventListener` calls it during scope cleanup, so an environment exposing only `connection.addEventListener` would still throw on unmount

andreyshivas: Good catch, thanks! `useEventListener`'s cleanup calls `el.removeEventListener` unconditionally, so an environment with only `addEventListener` would indeed still throw on scope dispose.

Fixed in 7ac581e: the guard now requires both `addEventListener` and `removeEventListener`. I chose to skip the subscription entirely in that case rather than subscribing and guarding only the cleanup — a listener we can never remove would outlive the scope and keep writing to disposed refs, which seems worse than falling back to the one-time snapshot (the `online`/`offline` handling on `window` still works either way). Also added a test asserting the listener is removed on scope dispose in the normal case.

PR #5585, opened 2026-08-13
fix(useTimeoutPoll): await the immediate callback before scheduling the next poll

### LINKED ISSUE #5582: BUG | `useTimeoutPoll` | immediateCallback does not await
### Describe the bug

My understanding is that each interval waits until the last is done.
But with `immediateCallback` set to true it starts the function and immediately starts the loop. 
this means that the time between immediateCallback fn start and first loop fn start could be significantly shorter than interval

### Reproduction

https://playground.vueuse.org/#eNp9U01v2zAM/SuEMaA2kDjptsPQpcG2ogW6w1asvdUFqih0olaWDH0kKQL/91FS7KbY0EMCi3x8enyk9tn3ti03HrOzbGa5Ea0Di863IJlanVeZs1U2r5RoWm0c7MFbvBMNau9utJTQQW10AyffiIJSE64NnhzBDdYDhiAnXytVKa6VdXSBddcODZwH1Ez5ZoFmnk8LwiSEQ/pLWeuMUKv7h3l+/xAAlWL2RXGoveJOaAXMLvMC9pUCSMVKb6l2JfWCybu1sGWLptamYYpjSck80ACEn6gh79WUGyY9Hph6Lkcd3woqJEYqHb8FRx6IalOgbL1d54/BJ7CxLBTAh/3AUzp9JXa4zD8WHTT2MWnpAKXF/up/+K6EOWjp8cEHALZlgvrFLdyQz8Jinhu0BZzPwygP08rJHQpQIj8tRvBlOi2SjzAMIl2WWqRUZE/9N5aibyefk+EjOJ1OpyMas2gaXArm8Ayc8Th6DVwwKReMP6cEdEWlZpO0aLRWdHDYtJKAdAKYLcUGOAmytHq1xB0IytsxRxU25clbJ+qX/rgdExGignX/EUrGXMu4s8HEmW2Zgs2YJk+MqUGhordVNty0kJo/DzUA+31EdmRBJJkEliRwQgrpazY50p2N6JmQU7VYlU9WK3pLcYZVxnXTConmdxuWlJ7SWT/dKiNj9PZnjEXP+jhfI3/+T/zJ7kKsym5oiGg2WGVDzjGzQuoopC9vf+GOvodko5deEvqd5B+0WvqgMcF+eLUk2Ue4qPY6Pmt6iHf2cudQ2b6pIDRuZMST0R4v3mn9Ve6n8vNhk7us+wsrjIp1

### System Info

```Shell
n/a
```

### Used Package Manager

pnpm

### Validations

lazerg:
`useTimeoutPoll` is documented to trigger the callback only after the last task is done, but the `immediateCallback` path calls `fn()` without awaiting it, so an async callback can still be running when the first regular tick fires. On top of that, `useTimeoutFn` was told to start immediately as well, so its timer ran on its own schedule next to the one `resume()` starts.

Now `resume()` goes through `loop()`, which awaits the callback before scheduling the next poll, and the timeout only starts from there.

Fixes #5582

### DIFF
--- a/packages/core/useTimeoutPoll/index.browser.test.ts
+++ b/packages/core/useTimeoutPoll/index.browser.test.ts
@@ -35,7 +35,20 @@ describe('useTimeoutPoll', () => {
 
     expect(callback).toHaveBeenCalledTimes(1)
 
-    vi.advanceTimersByTime(100)
+    await vi.advanceTimersByTimeAsync(50)
+    expect(callback).toHaveBeenCalledTimes(2)
+  })
+
+  it('immediateCallback waits for the callback to finish', async () => {
+    const callback = vi.fn(() => new Promise<void>(resolve => setTimeout(resolve, 50)))
+    useTimeoutPoll(callback, 50, { immediateCallback: true })
+
+    expect(callback).toHaveBeenCalledTimes(1)
+
+    await vi.advanceTimersByTimeAsync(50)
+    expect(callback).toHaveBeenCalledTimes(1)
+
+    await vi.advanceTimersByTimeAsync(50)
     expect(callback).toHaveBeenCalledTimes(2)
   })
 
--- a/packages/core/useTimeoutPoll/index.browser.test.ts
+++ b/packages/core/useTimeoutPoll/index.browser.test.ts
@@ -52,6 +52,41 @@ describe('useTimeoutPoll', () => {
     expect(callback).toHaveBeenCalledTimes(2)
   })
 
+  it('keeps polling when the callback rejects', async () => {
+    const ignoreRejection = (e: PromiseRejectionEvent) => e.preventDefault()
+    window.addEventListener('unhandledrejection', ignoreRejection)
+
+    try {
+      const callback = vi.fn(() => Promise.reject(new Error('failed')))
+      useTimeoutPoll(callback, 50, { immediateCallback: true })
+
+      expect(callback).toHaveBeenCalledTimes(1)
+
+      await vi.advanceTimersByTimeAsync(50)
+      expect(callback).toHaveBeenCalledTimes(2)
+
+      await vi.advanceTimersByTimeAsync(50)
+      expect(callback).toHaveBeenCalledTimes(3)
+    }
+    finally {
+      window.removeEventListener('unhandledrejection', ignoreRejection)
+    }
+  })
+
+  it('does not schedule another round when paused during the callback', async () => {
+    const callback = vi.fn(() => new Promise<void>(resolve => setTimeout(resolve, 50)))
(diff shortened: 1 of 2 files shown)


MILLERMARRU (commented):
I compared this against `useTimeoutFn`'s own implementation and the mismatch is clear: `loop()` in useTimeoutPoll already does `await fn(); start()`, but `resume()`'s `immediateCallback` branch called bare `fn()` (no await) followed immediately by `start()`. For an async `fn`, that means the poll interval starts counting down before the callback has actually resolved, so a slow callback and a short interval can result in overlapping invocations, the opposite of what "immediateCallback" implies relative to the rest of the polling behavior. Routing that branch through `loop()` instead unifies both paths to the same await-then-schedule semantics. Switching the underlying `useTimeoutFn(loop, interval, {immediate})` call to `{immediate: false}` also looks correct to me, `useTimeoutPoll` already calls `resume()` itself when its own `immediate` option is true, so leaving the inner `useTimeoutFn`'s `immediate` flag on top of that would double-start the timer on mount.

haoku123:
The unified await-then-schedule semantics look right. One edge case worth guarding: `await fn()` throwing now **kills the poll permanently**.

Before this change `fn()` was fire-and-forget — an async `fn` rejecting did not stop the next `start()` from being scheduled. Now the rejection skips `start()` entirely, so a single failing callback silently stops polling (and surfaces as an unhandled rejection).

Suggest wrapping in `finally` (or try/catch + rethrow) so scheduling survives rejections:

```ts
async function loop() {
  if (!isActive.value)
    return
  try {
    await fn()
  }
  finally {
    if (isActive.value)
      start()
  }
}
```

The `isActive.value` re-check also covers the pause-during-await window: today a `pause()` while `fn` is running still schedules one empty timer round before the guard stops the chain.

lazerg:
Checked both parts locally and they reproduce: a rejecting `fn` killed the poll after one round, and calling `pause()` while the callback was in flight still armed one more timer. Fixed in d3c51ad with try/finally plus the active check before rescheduling.

Two tests added: a callback that always rejects keeps being called on schedule, and `pause()` during an in-flight callback leaves no pending timer.

The rejection still propagates out of the loop rather than being swallowed, so a failing callback stays visible instead of going quiet.

haoku123:
Verified — the try/finally with the active re-check addresses both failure modes, and the two tests target exactly the regressions (reject keeps scheduling, pause leaves zero pending timers via `getTimerCount()`). The `unhandledrejection` guard in the first test is a nice touch so the rejection stays observable without polluting the suite. LGTM.

lazerg:
Closing as no one reviewed and this has been open more than a month with no activity. Happy to reopen if there's interest — thanks for your time.

PR #5587, opened 2026-08-14
fix(usePointer): reset isInside on pointercancel

### LINKED ISSUE #5586: usePointer's isInside doesn't clear on pointercancel, only pointerleave
`usePointer`'s `isInside` ref is set `true` by the `pointerdown`/`pointermove`/`pointerup` handler and only reset to `false` by a dedicated `pointerleave` listener:

```ts
useEventListener(target, ['pointerdown', 'pointermove', 'pointerup'], handler, listenerOptions)
useEventListener(target, 'pointerleave', () => isInside.value = false, listenerOptions)
```

There's no `pointercancel` listener at all. I traced through the Pointer Events Level 3 "suppress a pointer event stream" algorithm the same way #5550/#5555 did for `useDraggable`/`onLongPress`: it fires `pointercancel`, then `pointerout`, then `pointerleave`, so in the ordinary case the trailing `pointerleave` still reaches this listener and resets `isInside` correctly. That's why this doesn't show up in everyday use.

The gap is the same narrower one #5555 just fixed for `onLongPress`: when `pointerleave` doesn't reach the element (the pointer is under `setPointerCapture` and gets retargeted away, or the element is unmounted mid-gesture), the cancel sequence never reaches this listener and `isInside` sticks at `true` after the gesture the browser already abandoned.

`onLongPress`'s PR description (#5555) named this exact composable as a known follow-up ("`usePointer` and `useMouse` will follow as separate PRs") after fixing the same class of bug there and in `useDraggable` (#5550), but as of the current `main` `usePointer/index.ts` still only listens to `pointerdown`/`pointermove`/`pointerup`/`pointerleave` — no `pointercancel`.

Suggested fix, mirroring #5555's approach on `onLongPress` (route the cancel into the same handler that already resets state, don't just clear separately):

```diff
-    useEventListener(target, ['pointerdown', 'pointermove', 'pointerup'], handler, listenerOptions)
-    useEventListener(target, 'pointerleave', () => isInside.value = false, listenerOptions)
+    useEventListener(target, ['pointerdown', 'pointermove', 'pointerup'], handler, listenerOptions)
+    useEventListener(target, ['pointerleave', 'pointercancel'], () => isInside.value = false, listenerOptions)
```

I didn't touch `useMouse` (also named in that PR) — it doesn't listen to Pointer Events at all (mouse/touch/scroll only), so it isn't affected by this specific gap; happy to open that separately if there's a distinct issue there.

haoku123:
## Description

`usePointer` only reset `isInside` on `pointerleave`. When the browser abandons a gesture it fires `pointercancel` **without** `pointerleave` reaching the element (e.g. when the pointer is under `setPointerCapture` and gets retargeted away), leaving `isInside` stuck at `true`.

## Changes

- Listen for `pointercancel` in addition to `pointerleave` to reset `isInside` (1-line change)
- Added `index.test.ts` for `usePointer` covering both `pointercancel` and `pointerleave` reset behavior

## Context

This is the same class of fixes as:
- #5550 (`useDraggable` pointercancel)
- #5555 (`onLongPress` pointercancel) — its PR description mentioned `usePointer` would follow as a separate PR

Closes #5586

### DIFF
--- a/packages/core/usePointer/index.test.ts
+++ b/packages/core/usePointer/index.test.ts
@@ -1,0 +1,60 @@
+import { mount } from '@vue/test-utils'
+import { describe, expect, it } from 'vitest'
+import { h, nextTick, useTemplateRef } from 'vue'
+import { usePointer } from './index'
+
+describe('usePointer', () => {
+  it('sets isInside true on pointerdown and resets it on pointercancel', async () => {
+    const wrapper = mount({
+      setup() {
+        const el = useTemplateRef<HTMLElement>('el')
+        const { isInside } = usePointer({ target: el })
+
+        return () => h('div', {
+          'data-inside': isInside.value,
+          'ref': 'el',
+        })
+      },
+    })
+
+    await nextTick()
+    const element = wrapper.get('div').element
+
+    expect(wrapper.get('div').attributes('data-inside')).toBe('false')
+
+    element.dispatchEvent(new PointerEvent('pointerdown', { bubbles: true }))
+    await nextTick()
+    expect(wrapper.get('div').attributes('data-inside')).toBe('true')
+
+    // the user agent takes the gesture over (e.g. a second pointer starting a
+    // pinch), firing pointercancel without pointerleave reaching the element
+    element.dispatchEvent(new PointerEvent('pointercancel', { bubbles: true }))
+    await nextTick()
+    expect(wrapper.get('div').attributes('data-inside')).toBe('false')
+  })
+
+  it('resets isInside on pointerleave', async () => {
+    const wrapper = mount({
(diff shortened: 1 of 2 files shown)


OrbisK (approved):

haoku123:
Thanks for the review @OrbisK! It looks like the CI run is stuck in `action_required` (waiting for workflow approval on this fork PR). Would it be possible to approve the workflow run so the tests can execute? The change is small (reset `isInside` on `pointercancel` + regression test) and should pass cleanly. Thanks!

9romise (approved):

antfu (approved):

PR #5589, opened 2026-08-14
fix(useRouteParams,useRouteQuery): stop transform value type from narrowing to the default value type

### LINKED ISSUE #5588: BUG | `useRouteParams` | useRouteParams infers transform parameter type from defaultValue
### Describe the bug

because of the signature

`declare function useRouteParams<
  T extends RouteParamValueRaw = RouteParamValueRaw,
  K = T
>(
  name: string,
  defaultValue?: MaybeRefOrGetter<T>,
  options?: ReactiveRouteOptionsWithTransform<T, K>
): Ref<K>;`

The inferred type of the transform value is equal to the default value:
`const employeeFilter = useRouteParams('employeeId', null, {
  transform: (value) => (!value ? null : Number(value)),
});`

e.g. with the above method the type for value is inferred as null (see reproduction)

**Expected behaviour:**
Typing should be updated to be correct.
I am not sure what is the best approach, are raw route params not always string/undefined? so maybe that would work?
Otherwise, inferring the type to any/unknown I would also consider to be an improvement. 

### Reproduction

https://playground.vueuse.org/#eNp9kc1OAjEUhV/lphsgmTAxukIw/gQTXChBl900M5ex2Gmb/iCEzLt72wlIjGHXe85p+532wB6sHW8jsgmb+spJG8BjiBaU0M2Ms+A5u+Natta4AAeIHlcmBlwKJ1oPHaydaWFwT0eQVbrkucEt15XRPgC2Vpk94rNUpMPsz/7h4BhY1IMCdFSqgAPXAGUJW6EigqSTnMMqqD2EvcUahM/BlApOaL82rp3AMMdHMLuD4dWo4LobEQXX07KvRSVoCOk+EZCmaXk2sIKaEvJaNuONN5qeI2NwVpnWSoXuzQZJlTib9IDJE0qZ75esBReRLu316hOrr3/0jd8ljbOlQ49ui5ydvCBcg6G35++vuKP1yWxNHRWlL5gr9EbFxNjHHqOuCfssl2kX+SOlbj78fBdQ+2OpBJqSXc5zRh/6dKH6L+71+Cbvoxdn3Q/x+8fG

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    Memory: 10.81 GB / 31.73 GB
  Binaries:
    Node: 24.18.0 - C:\nvm4w\nodejs\node.EXE
    npm: 11.16.0 - C:\nvm4w\nodejs\npm.CMD
    pnpm: 11.17.0 - C:\nvm4w\nodejs\pnpm.CMD
  Browsers:
    Chrome: 151.0.7922.138
    Edge: Chromium (151.0.4129.78)
  npmPackages:
    @vueuse/core: 14.3.0 => 14.3.0 
    @vueuse/router: 14.3.0 => 14.3.0 
    vue: 3.5.40 => 3.5.40
```

### Used Package Manager

pnpm

### Validations

lazerg:
useRouteParams and useRouteQuery infer the transform callback's parameter type from the literal type of defaultValue instead of the full raw param/query union, so passing null as the default (a common pattern) makes TypeScript treat the raw value as always null inside transform. Wrapping defaultValue's generic argument in NoInfer keeps T tied to the actual param type while the transform's return type still drives K as before.

Fixes #5588

### DIFF
--- a/packages/router/useRouteParams/index.test.ts
+++ b/packages/router/useRouteParams/index.test.ts
@@ -1,5 +1,5 @@
 import type { Ref } from 'vue'
-import { describe, expect, it, vi } from 'vitest'
+import { describe, expect, expectTypeOf, it, vi } from 'vitest'
 import { computed, ref as deepRef, effectScope, nextTick, reactive, shallowRef, watch } from 'vue'
 import { useRouteParams } from './index'
 
@@ -38,6 +38,22 @@ describe('useRouteParams', () => {
     const id = useRouteParams('id', '1', { transform: Number, route, router })
 
     expect(id.value).toBe(1)
+  })
+
+  it('should not narrow the transform value type to the default value type', () => {
+    const route = getRoute()
+    const router = {} as any
+
+    const id = useRouteParams('id', null, {
+      transform: (value) => {
+        expectTypeOf(value).not.toEqualTypeOf<null>()
+        return value == null ? null : Number(value)
+      },
+      route,
+      router,
+    })
+
+    expectTypeOf(id.value).toEqualTypeOf<number | null>()
   })
 
   it('should handle transform get/set', async () => {
--- a/packages/router/useRouteParams/index.ts
+++ b/packages/router/useRouteParams/index.ts
@@ -16,7 +16,7 @@ export function useRouteParams<
   K = T,
 >(
   name: string,
-  defaultValue?: MaybeRefOrGetter<T>,
+  defaultValue?: MaybeRefOrGetter<NoInfer<T>>,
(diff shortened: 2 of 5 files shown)


lazerg:
The lts/* failure is a flaky animation timing assertion in useTransition's browser test, unrelated to this change. Same class of flake showed up on main's last CI run too (useAnimate).

lazerg:
Closing as no one reviewed and this has been open more than a month with no activity. Happy to reopen if there's interest — thanks for your time.

PR #5591, opened 2026-08-15
fix(useEventListener): support custom event names with omitted window target

### LINKED ISSUE #5567: BUG | `useEventListener` | Difference in type overloads

haoku123:
close #5567

## Problem

With an omitted target, only window event-map keys type-check:

```ts
useEventListener(window, 'foo', () => {}) // OK (overload 7)
useEventListener('foo', () => {})         // TS2345: '"foo"' is not assignable to
                                          // 'MaybeRefOrGetter<Arrayable<keyof WindowEventMap>>'
```

The overload list has no omitted-target signature for custom event names, so
the call falls through to the only omitted-target overload (constrained to
`keyof WindowEventMap`) and fails.

## Solution

Add an omitted-target overload for arbitrary `Names extends string` right
after the existing window-event-map overload. Known event names still match
the first overload (keeping `MouseEvent` precision on `'click'`); custom
names fall through to the new one:

```ts
export function useEventListener<Names extends string, EventType = Event>(
  event: MaybeRefOrGetter<Arrayable<Names>>,
  listener: MaybeRef<Arrayable<GeneralEventListener<EventType>>>,
  options?: MaybeRefOrGetter<boolean | AddEventListenerOptions>,
): Fn
```

The runtime already handles this shape (strings in `args[0]` resolve to
`defaultWindow`), so this is a types-only change. Overload doc comments are
renumbered to follow the inserted overload.

## Tests

Added a type test covering: omitted target + custom event, explicit window +
custom event, and omitted target + `'click'` keeping `MouseEvent` precision.

### DIFF
--- a/packages/core/useEventListener/index.browser.test.ts
+++ b/packages/core/useEventListener/index.browser.test.ts
@@ -1,6 +1,7 @@
 import type { Fn } from '@vueuse/shared'
 import type { MockInstance } from 'vitest'
 import type { Ref } from 'vue'
+import { expectTypeOf } from 'expect-type'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { computed, ref as deepRef, effectScope, nextTick, shallowRef } from 'vue'
 import { useEventListener } from './index'
@@ -406,5 +407,17 @@ describe('useEventListener', () => {
     await nextTick()
 
     expect(listener).toHaveBeenCalledTimes(6)
+  })
+
+  it('should support custom event names with omitted window target (#5567)', () => {
+    useEventListener('my-custom-event', (ev: Event) => {
+      expectTypeOf(ev).toEqualTypeOf<Event>()
+    })
+    useEventListener(window, 'my-custom-event', (ev: Event) => {
+      expectTypeOf(ev).toEqualTypeOf<Event>()
+    })
+    useEventListener('click', (ev) => {
+      expectTypeOf(ev).toEqualTypeOf<MouseEvent>()
+    })
   })
 })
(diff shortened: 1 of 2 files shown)


9romise (commented):
I think we can extend custom event names directly by using
``` ts
declare interface WindowEventMap {}
```
 Do you have any other considerations in mind?

haoku123:
Thanks for the review @9romise! Declaration merging is indeed the way to get *precise* types for custom events, and it stays fully supported by this PR — a merged event name is still a `keyof WindowEventMap` member, so it matches Overload 1 and keeps the exact `ev` type (verified with a local type-check).

What the new overload solves is the lower-fidelity case from #5567 where the user writes no global declaration at all:

```ts
useEventListener(window, 'foo', () => {}) // accepted today via Overload 7 (custom event target)
useEventListener('foo', () => {})         // rejected: no omitted-target overload accepts arbitrary strings
```

The explicit-target form already accepts any string, so the omitted-target form (which just defaults to `window`) being stricter is the inconsistency this PR removes. The fallback overload degrades `ev` to `Event` — the same tradeoff Overload 7 already makes for explicit targets — while users who want exact types can still merge `WindowEventMap` and get them.

That said, if you'd rather keep the overload list lean and guide users toward declaration merging instead, I'm happy to close this and update the docs. Which direction do you prefer?

Issue #4003, opened 2024-05-30
Can't make useResizeObserver work for multiple targets


maprox:
### Describe the bug

[useResizeObserver](https://vueuse.org/core/useResizeObserver/) fails with an error when using it on multiple targets:

```
Uncaught (in promise) TypeError: Failed to execute 'observe' on 'ResizeObserver': parameter 1 is not of type 'Element'.
    at watch.immediate (@vueuse_core.js?v=784f5486:4083:27)
    at callWithErrorHandling (chunk-TNYMODNW.js?v=12c6d6eb:1638:19)
    at callWithAsyncErrorHandling (chunk-TNYMODNW.js?v=12c6d6eb:1645:17)
    at Array.job (chunk-TNYMODNW.js?v=12c6d6eb:3353:9)
    at flushPostFlushCbs (chunk-TNYMODNW.js?v=12c6d6eb:1820:41)
    at flushJobs (chunk-TNYMODNW.js?v=12c6d6eb:1858:5)
```

example:

```
<template>
<ul>
  <li v-for="i in 2" :key="i" ref="childRefs">
    {{ i }}
  </li>
</ul>
</template>

<script lang="ts" setup>
import { ref } from 'vue';
import { useResizeObserver } from "@vueuse/core";

const childRefs = ref();

useResizeObserver(childRefs, (entries) => {
  // ...
})
</script>

```

I guess it should gather targets the same way it is done in useIntersectionObserver: https://github.com/vueuse/vueuse/blob/4d868f5e97d661dfab077108e1d650d011005584/packages/core/useIntersectionObserver/index.ts#L62

instead of current implementation:
https://github.com/vueuse/vueuse/blob/4d868f5e97d661dfab077108e1d650d011005584/packages/core/useResizeObserver/index.ts#L66

### Reproduction

https://stackblitz.com/edit/vitejs-vite-zn4xdm?file=src%2FApp.vue

### System Info

```Shell
System:
    OS: Windows 11 10.0.22631
    CPU: (16) x64 AMD Ryzen 9 5900HS with Radeon Graphics
    Memory: 6.46 GB / 31.41 GB
  Binaries:
    Node: 20.13.1 - C:\Node20\node.EXE
    npm: 10.3.0 - C:\Node20\npm.CMD
    pnpm: 8.14.1 - ~\AppData\Local\pnpm\pnpm.EXE
  Browsers:
    Edge: Spartan
    Internet Explorer: 11.0.22621.3527
  npmPackages:
[...]

PR #5593, opened 2026-08-17
fix(useResizeObserver): guard observe() with instanceof Element to prevent crash on Comment nodes

mryshakow:
## Problem

When a Vue SSR/hydration mismatch occurs, Vue can temporarily mount a component onto a **Comment node** rather than an actual Element. Any composable that passes a template ref to `useResizeObserver` will then trigger the watch callback with that Comment node as the target.

The previous guard on line 82:

```ts
if (_el)
  observer!.observe(_el, observerOptions)
```

passes for **any non-null, non-undefined value** — including Comment nodes, which are truthy objects. `ResizeObserver.observe()` requires an `Element` and throws a `TypeError` when given anything else:

```
Failed to execute 'observe' on 'ResizeObserver': parameter 1 is not of type 'Element'.
```

In Nuxt, this unhandled error is caught by Vue's `callWithErrorHandling`, handed to Nuxt's global error handler, and surfaces as a **500 Internal Server Error page** — crashing the entire app.

### Reproduction

1. Have a component that uses `useResizeObserver` (directly or via `useElementSize`/`useBreakpoint`)
2. Introduce a server/client render difference (e.g. `useWindowSize()` returning `0` on SSR → `isMobile=true`, real width on client → `isMobile=false`) causing a hydration mismatch
3. Vue mounts the component onto a Comment node during reconciliation
4. `useResizeObserver`'s `watch` fires with the Comment node → `observe(commentNode)` → TypeError → app crash

## Fix

Change the truthy check to `instanceof Element`:

```ts
// before
if (_el)
  observer!.observe(_el, observerOptions)

// after
if (_el instanceof Element)
  observer!.observe(_el, observerOptions)
```

This silently skips any target that is not a real DOM Element (Comment nodes, text nodes, `null`, `undefined`), which is the correct and safe behaviour — the observer simply won't fire for that target.

## Tests

Added `index.browser.test.ts` with two cases:
- Target is a Comment node → no throw
- Target is `null` → no throw

## Related

- Partially related to #4003 (same error, different root cause — ref arrays)
- Partially related to #4102 (SSR hydration mismatch as trigger)


antfu (approved):

mryshakow:
Hi! 

Looks like the latest CI runs are still awaiting maintainer approval after I updated the branch. Is there anything else you need from my side? 

Thanks!

OrbisK (approved):

ilyaliao (approved):

PR #5603, opened 2026-08-21
fix(useBluetooth): dedupe disconnect listener and reconnect on mount

### LINKED ISSUE #5600: useBluetooth: disconnect listeners accumulate per device, and mounted reconnect skips state sync
### Describe the bug

Two separate issues in `useBluetooth`, both around `device` changes.

## 1. `gattserverdisconnected` listeners accumulate

`connectToBluetoothGATTServer()` registers the disconnect listener on every call:

```ts
watch(device, () => {
  connectToBluetoothGATTServer()
})

async function connectToBluetoothGATTServer() {
  error.value = null
    // Add reset fn to gattserverdisconnected event:
    useEventListener(device, 'gattserverdisconnected', reset, { passive: true })
    ...
  }
}
```

Since it's driven by `watch(device)` and `requestDevice()` can be called repeatedly (user picks the wrong device, disconnects and reselects, etc.), each new device registers another listener. `useEventListener` doesn't dedupe, and these are only torn down on scope dispose — not between `device` changes.

So after N `requestDevice()` calls, a single `gattserverdisconnected` event invokes `reset()` N times. `reset()` is idempotent in its effects, so this isn't visibly broken today, but it's an unbounded listener leak for a long-lived component and means the listener target (`device`) no longer matches the device it was registered for.

## 2. `tryOnMounted` reconnect doesn't update state

```ts
tryOnMounted(() => {
  if (device.value)
    device.value.gatt?.connect()
})
```

This bypasses `connectToBluetoothGATTServer()` and discards the returned server. Compare the normal path, which does maintain state:

```ts
server.value = await device.value.gatt.connect()
isConnected.value = server.value.connected
```

So when this branch runs with an already-populated `device`, the GATT connection is established but `server` stays `undefined` and `isConnected` stays `false`. The promise is also unawaited and unguarded, so a rejection here becomes an unhandled rejection rather than landing in `error`.

`tryOnScopeDispose` has the mirrored problem — it calls `gatt?.disconnect()` without resetting `isConnected`/`server`, though that matters less since the scope is going away.

### Expected behavior

1. One `gattserverdisconnected` listener per device, cleaned up when `device` changes.
2. The mounted reconnect path should go through `connectToBluetoothGATTServer()` so `server`, `isConnected` and `error` stay accurate.

### Possible fix

For (1), register the listener against the resolved device inside the `watch` with its own cleanup, or use the `watch` `onCleanup` argument / a `watchEffect` so the previous listener is removed when `device` changes.

For (2), reuse the existing function:

```ts
tryOnMounted(() => {
  if (device.value)
    connectToBluetoothGATTServer()
})
```

That keeps the state assignments and routes errors into `error` via the existing `try/catch`.

Happy to open a PR for both once you confirm the preferred approach for (1) — whether you'd rather key the listener off the device object or restructure the `watch`. Note `useBluetooth` currently only has `index.browser.test.ts`, so I'd want to check what's testable in the jsdom/browser split before adding coverage.

Unrelated to #5317, which is about DTS type references.

### Additional context

Found while auditing the connection-oriented composables — same pass that turned up #5599 (`useUserMedia`/`useDisplayMedia` start/stop race). This one is a different failure mode, so filing separately.

haoku123:
## Summary
- Remove the previous `gattserverdisconnected` listener before registering a new one when the selected device changes, preventing unbounded listener accumulation
- Route the mounted reconnect path through `connectToBluetoothGATTServer()` so `server` and `isConnected` stay in sync
- Add regression tests for listener deduplication and connection state updates

Fixes #5600

## Test plan
- [x] `npx vitest run packages/core/useBluetooth/index.test.ts`

Made with [Cursor](https://cursor.com)


antfu (approved):

packages/core/useBluetooth/index.ts
```diff
@@ -119,10 +124,11 @@ export function useBluetooth(options?: UseBluetoothOptions): UseBluetoothReturn
 
   tryOnMounted(() => {
     if (device.value)
-      device.value.gatt?.connect()
+      void connectToBluetoothGATTServer()
   })
 
   tryOnScopeDispose(() => {
+    removeDisconnectListener?.()
```
OrbisK: I think we dont need this, because `useEventListener` should do this for us

packages/core/useBluetooth/index.ts
```diff
@@ -102,9 +104,12 @@ export function useBluetooth(options?: UseBluetoothOptions): UseBluetoothReturn
     // Reset any errors we currently have:
     error.value = null
 
+    removeDisconnectListener?.()
+    removeDisconnectListener = undefined
+
     if (device.value && device.value.gatt) {
       // Add reset fn to gattserverdisconnected event:
-      useEventListener(device, 'gattserverdisconnected', reset, { passive: true })
+      removeDisconnectListener = useEventListener(device, 'gattserverdisconnected', reset, { passive: true })
```
OrbisK: I think we should just pull this into the `tryOnMounted`

OrbisK:
@haoku123 i have moved the `useEventListener` to `tryOnMounted` instead, which should have the same effect. I have also added `once: true` to the watcher, because the device gets cleared anyway.

OrbisK (approved):

antfu (approved):

PR #5604, opened 2026-08-22
fix(injectLocal): use function declaration for Rollup @__NO_SIDE_EFFECTS__ annotation

### LINKED ISSUE #5535: Rollup Nuxt app build warning
### Describe the bug

When building a Nuxt 4 app for production (latest or nightly version), you can see the following warning:
```
/**
* On the basis of `inject`, it is allowed to directly call inject to obtain the value after call provide in the same component.
*
* @example
* ```ts
* injectLocal('MyInjectionKey', 1)
* const injectedValue = injectLocal('MyInjectionKey') // injectedValue === 1
* ```
*
* @__NO_SIDE_EFFECTS__
*/"

in "node_modules/.cache/nuxt/.nuxt/dist/server/_nuxt/h1rbdtag.js" contains an annotation that Rollup cannot interpret due to the position of the comment. The comment will be removed to avoid issues.
```

### Reproduction

n/a

### System Info

```Shell
n/a
```

### Used Package Manager

pnpm

### Validations

s00d:
Fixes #5535

## Summary

- Change `injectLocal` from a `const` arrow export to a `function` declaration so Rollup/Rolldown can interpret the JSDoc `@__NO_SIDE_EFFECTS__` tag during bundling
- Matches the pattern used by other annotated exports in `@vueuse/shared` (e.g. `useArrayFind`, `useToggle`)

## Problem

When building a Nuxt 4 app for production, Nitro/Rollup emits a warning like:

```
A comment
"/**
* ...
* @__NO_SIDE_EFFECTS__
*/"
in ".nuxt/dist/server/server.mjs" contains an annotation that Rollup cannot interpret due to the position of the comment. The comment will be removed to avoid issues.
```

`injectLocal` is currently exported as:

```ts
/** @__NO_SIDE_EFFECTS__ */
export const injectLocal: typeof inject = <T>(...args) => { ... }
```

Rollup only understands `#__NO_SIDE_EFFECTS__` / `@__NO_SIDE_EFFECTS__` on **function declarations**, not on `const` bindings. `provideLocal` is a plain `function` without this JSDoc tag and does not trigger the warning.

## Reproduction

Minimal repro (Nuxt 4.5.2 + @vueuse/nuxt 14.4.0, `useBreakpoints()` in `app.vue`):

- **GitHub:** https://github.com/s00d/nuxt-vueuse-injectlocal-warning-repro
- **StackBlitz:** https://stackblitz.com/github/s00d/nuxt-vueuse-injectlocal-warning-repro

```bash
pnpm install && pnpm build
```

The warning appears during the **Nitro server bundle** step (after Vite SSR), not during client build or `nuxt dev`.

Applying a local patch that changes `const injectLocal` → `function injectLocal` in `@vueuse/shared` removes the warning (see `patches/@vueuse__shared@14.4.0.patch` in the repro repo).

## Test plan

- [x] `pnpm test:unit packages/shared/injectLocal/index.test.ts`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/shared/injectLocal/index.ts
+++ b/packages/shared/injectLocal/index.ts
@@ -14,7 +14,7 @@ import { localProvidedStateMap } from '../provideLocal/map'
  * @__NO_SIDE_EFFECTS__
  */
 // @ts-expect-error overloads are not compatible
-export const injectLocal: typeof inject = <T>(...args) => {
+export function injectLocal<T>(...args: Parameters<typeof inject>): ReturnType<typeof inject> {
   const key = args[0] as InjectionKey<T> | string | number
   const instance = getCurrentInstance()?.proxy
   const owner = instance ?? getCurrentScope()
--- a/packages/shared/injectLocal/index.ts
+++ b/packages/shared/injectLocal/index.ts
@@ -13,8 +13,10 @@ import { localProvidedStateMap } from '../provideLocal/map'
  *
  * @__NO_SIDE_EFFECTS__
  */
-// @ts-expect-error overloads are not compatible
-export function injectLocal<T>(...args: Parameters<typeof inject>): ReturnType<typeof inject> {
+export function injectLocal<T>(key: InjectionKey<T> | string): T | undefined
+export function injectLocal<T>(key: InjectionKey<T> | string, defaultValue: T, treatDefaultAsFactory?: false): T
+export function injectLocal<T>(key: InjectionKey<T> | string, defaultValue: T | (() => T), treatDefaultAsFactory: true): T
+export function injectLocal<T>(...args: any[]): T | undefined {
   const key = args[0] as InjectionKey<T> | string | number
   const instance = getCurrentInstance()?.proxy
   const owner = instance ?? getCurrentScope()
@@ -23,7 +25,7 @@ export function injectLocal<T>(...args: Parameters<typeof inject>): ReturnType<t
     throw new Error('injectLocal must be called in setup')
 
   if (owner && localProvidedStateMap.has(owner) && key in localProvidedStateMap.get(owner)!)
-    return localProvidedStateMap.get(owner)![key]
+    return localProvidedStateMap.get(owner)![key] as T
 
   // @ts-expect-error overloads are not compatible
   return inject(...args)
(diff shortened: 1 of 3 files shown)


serkodev:
Just help add missing side effects annotation for `injectLocal` function and update public api snapshot.

serkodev (approved):

serkodev:
One thing worth noting is that `injectLocal` was previously typed as typeof `inject` from Vue. Now that we’ve added our own overloads matching Vue’s, if Vue’s `inject` type changes in the future, we’ll need to update our overload types accordingly. Not sure if this could be an issue.

OrbisK:
> One thing worth noting is that `injectLocal` was previously typed as typeof `inject` from Vue. Now that we’ve added our own overloads matching Vue’s, if Vue’s `inject` type changes in the future, we’ll need to update our overload types accordingly. Not sure if this could be an issue.

Do we need to have custom overloads? Why cant we use vues inject type?

serkodev:
> > One thing worth noting is that `injectLocal` was previously typed as typeof `inject` from Vue. Now that we’ve added our own overloads matching Vue’s, if Vue’s `inject` type changes in the future, we’ll need to update our overload types accordingly. Not sure if this could be an issue.
> 
> Do we need to have custom overloads? Why cant we use vues inject type?

If we want to use `typeof inject`, we need to use a const arrow function. However, when using an arrow function, we can’t put `@__NO_SIDE_EFFECTS__` in the JSDoc as it will cause the Rolldown warning.

I explained the reasoning in detail in [CONTRIBUTING.md](https://github.com/serkodev/vueuse/blob/9df8944147c2be04b2493cd167406018591c316a/CONTRIBUTING.md?plain=1#L123-L170) of PR #5613.

Also in that PR, I kept `injectLocal` as a const arrow function while still preserving tree-shaking. In the JSDoc, I used `@NO_SIDE_EFFECTS` instead of `@__NO_SIDE_EFFECTS__` to avoid the Rolldown warning but also able to present the information to human readers at the same time, see https://github.com/vueuse/vueuse/pull/5613/changes/9df8944147c2be04b2493cd167406018591c316a

So I think we can leave it up to the community to decide whether to keep `typeof` or switch to overloads. I’m not quite sure if there’s an existing convention for when `typeof` should be preferred.

Personally, I’d vote for keeping it as is vue's `inject` type.

serkodev:
To be more clear in short:

Using function overloads (this PR):
- **Pro:** We can add `@__NO_SIDE_EFFECTS__` to the JSDoc for human readers.
- **Con:** We can't reuse Vue's `inject` type directly, so we'd need to manually keep it in sync with any future changes.

Using `typeof inject` (PR #5613):
- **Pro:** We can reuse Vue's `inject` type directly.
- **Con:** We can't use `@__NO_SIDE_EFFECTS__` in the JSDoc, as it would be recognized by Rolldown and trigger a warning. However, we can use something similar for human readers, e.g. `@NO_SIDE_EFFECTS`, to avoid this.

Both of these 2 approaches can preserve the tree-shaking annotation functionality without triggering a Rolldown warning.

PR #5621, opened 2026-09-04
fix(onKeyStroke): register listener as passive instead of capture

### LINKED ISSUE #5620: BUG | `onKeyStroke()` | Registers a capture listener when `passive: true`
### Description

Passing `passive: true` to `onKeyStroke()` unexpectedly registers the event listener in the capture phase. Consequently, a descendant cannot prevent the handler from running with `event.stopPropagation()`.

`OnKeyStrokeOptions.passive` indicates that it controls whether the listener is passive. The resulting listener should therefore be registered with `{passive: true}`, without changing its propagation phase.

### Expected behavior

The input handler calls `stopPropagation()`, so the ancestor's `onKeyStroke()` handler should not run. 

### Actual behavior

The ancestor handler runs before the input handler: `KeyboardEvent.eventPhase === 1` identifies the capturing phase.

### Suspected cause

`onKeyStroke()` passes the `passive` boolean directly as the listener options:

```ts
return useEventListener(target, eventName, listener, passive);
```

`useEventListener()` correctly forwards that value to the native API:

```ts
el.addEventListener(event, listener, options);
```

In the native `addEventListener()` API, a boolean options argument represents `useCapture`. Thus, `passive: true` becomes a capture listener instead of a passive listener.

Passing an options object would preserve the documented meaning:

```ts
return useEventListener(target, eventName, listener, {passive});
```

### Reproduction

[playground.vueuse.org](https://playground.vueuse.org/#eNp9Vdty2zYQ/ZUdznRGmkiUlKYvLK3Y9ahT95J6mjzyBSJWImMSwOAiS1X071kApC62Iz/I4F4Ozh4sFvvkTql04zDJktyUulYWDFqnoGFifVMk1hTJvBB1q6S2sAcp/sLdZ6vlE8IBVlq2UCS3BOAMTkqpsUh+PQvXuDqFUVTwFqKUwligX8tqgRpufGD+x5d//l402KKw8A2Ea5r5wP8OKSdmMFGisVLfs6a5l47iQuZgegqphXL2ir+HWGxom8eKGex2F65dop4PfGghzuocFMnClEwR+REM0OcN4WYOe4qzrxilG9Y4fPfOoxy9p82im7YMOGn4DQ6KP4wipmV6jTY76TPyVsWMqTeYgdUOyXKIRFdOlLaWAiomeIMPvvxIdzDMYCNrHkEvdTljebhA0UjHf5n5gxKphmko8i3ko/OKAk5wXFF5PJLIJ7H/qNvow2KrGmbRf9m8JRn8IVFDHkUJfVlY8lazeV5KjvPzQxvm1I5kg043kKorUGnJXag2n1Cq3y7AqA6vsL/L0hmwFcZuop7hoEgZA1HZFB6s6QTXUFLppk+NRKhi9ailYmvm9zmSGYGRAbeX5QhiKukaDkJa0E6kHaVJ4BTXgUq/DWlTYiUbjpo0iXwr1Djy6OKCbJH0SbdPuOPyWaRoytRTTCnOnwtBvOqePu1EgDdHgXJu5yH2UoR8QvZTDJ/v9y/v4+FAQfwC6O6lFj/Een39r8GF0kD5nrsGdjYIPn6kKfWJjsAzQF4kl/i0DBK8bJfFVmFpkWdduwT+1OCz0DjHo+7N0zRf6mPyXWkdazJYSlsB0wizUUi76JKzUqA2BDygI7JO12I9fKtXls5a6nW7U0hnG7+omtuyqcsnsoRbThfoP/8/n8SAeNUm/q7RMp+cX0GRG7tr/DJcxTAXVnQTxyvW1s0uA8OEGRvU9Spc+5Ztx881t1UGH95rbDujXtciA28A5qzsp0+ULYAu5XZs6v+pMC+Jpv4ekymkd3iz6fSn8K0Y5yFumv4StwhgvIlIvDZEn6itde1HTGH9YtxXNS5l41phMvBc/VTxEnesKJYpD/zeI8PsHL6bin0xfszF6dUJlIzo0SS8Vb1Ovxop6GWlBKDOKmWramrwf8Mkooc181BAf0VCvSGf/wy2br5He1lh+fSG/avZeluRPPqz1Bv/OPW++H5E9+LzJ9zS+uhsafg1FH3FSV1B2niOMew3mtRE+ywusH0IjzydwBez2JJ6pi/KE/WRhxAf3v37K6Wf6P6cfgh5pGhy+A7c+BYB)

### System Info

```Shell
System:
  OS: macOS 26.6.2
  CPU: (10) arm64 Apple M1 Pro
  Memory: 119.45 MB / 16.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.19.0
  npm: 12.0.2
Browsers:
  Firefox: 154.0.1
  Safari: 26.6.2
npmPackages:
  @vueuse/core: ^14.4.0 => 14.4.0
  vue: ^3.5.40 => 3.5.42
```

### Used Package Manager

npm

### Validations

lazerg:
onKeyStroke passed the passive boolean straight through as useEventListener's options argument, and addEventListener treats a boolean there as useCapture, not passive. So passive: true silently turned the listener into a capture-phase one, letting it run before a descendant's own handler had a chance to call stopPropagation().

Wrapped it as { passive } so the browser reads it as the intended passive option and the listener stays in the default bubble phase. onKeyStroke doesn't expose any other listener option (no capture, once, etc.), so there's nothing else to forward here.

One side effect worth flagging: since passive listeners can't call preventDefault(), any handler that relied on the old (buggy) capture-mode listener actually running preventDefault() will now find it's a no-op, which is the correct passive contract but a behavior change from before.

Fixes #5620

### DIFF
--- a/packages/core/onKeyStroke/index.browser.test.ts
+++ b/packages/core/onKeyStroke/index.browser.test.ts
@@ -54,4 +54,12 @@ describe('onKeyStroke', () => {
     await userEvent.keyboard('{a>5/}')
     expect(callBackFn).toBeCalledTimes(1)
   })
+
+  it('registers a passive listener instead of a capture listener when passive is true', () => {
+    const addSpy = vi.spyOn(window, 'addEventListener')
+
+    onKeyStroke('a', callBackFn, { passive: true })
+
+    expect(addSpy).toBeCalledWith('keydown', expect.any(Function), { passive: true })
+  })
 })
--- a/packages/core/onKeyStroke/index.browser.test.ts
+++ b/packages/core/onKeyStroke/index.browser.test.ts
@@ -55,7 +55,7 @@ describe('onKeyStroke', () => {
     expect(callBackFn).toBeCalledTimes(1)
   })
 
-  it('registers a passive listener instead of a capture listener when passive is true', () => {
+  it('pass passive option to listener', () => {
     const addSpy = vi.spyOn(window, 'addEventListener')
 
     onKeyStroke('a', callBackFn, { passive: true })
--- a/packages/core/onKeyStroke/index.ts
+++ b/packages/core/onKeyStroke/index.ts
@@ -79,7 +79,7 @@ export function onKeyStroke(...args: any[]) {
       handler(e)
   }
 
-  return useEventListener(target, eventName, listener, passive)
+  return useEventListener(target, eventName, listener, { passive })
 }
 
 /**


lazerg:
This pipeline failure predates my change and isn't caused by it. It's the known flaky useTransition animation-timing test, unrelated to onKeyStroke.

PR #5452, opened 2026-05-23
fix(useElementVisibility): stop observer after first visibility change

guozi9999:
### Before submitting the PR, please make sure you do the following

- [ ] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [ ] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [ ] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Ideally, include relevant tests that fail without this PR but pass with it.

<details>
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

fixes #5443

This PR fixes `useElementVisibility` when `once: true` is used together with `controls: true`.

`useElementVisibility` already returns the underlying `useIntersectionObserver` controls when `controls: true`, including `isActive`, `pause`, `resume`, and `stop`. However, the `once` watcher was registered inside the intersection callback after `isVisible` had already been updated, so the first visibility change did not stop the observer immediately.

This moves the `watchOnce(isVisible, ...)` setup outside the observer callback so the observer is stopped after the first visibility change as expected. It also adds tests for the exposed controls and documents `isActive` in the `controls: true` usage example.

PR #5623, opened 2026-09-05
fix(useElementVisibility): stop observing when the element first becomes visible with `once`

giaBaoJS:

---

### Description

`useElementVisibility(el, { once: true })` is documented as "Stop tracking when element visibility changes for the first time", and `index.md` says you can "read `isActive` to tell whether tracking has already stopped after the element first became visible".

At `packages/core/useElementVisibility/index.ts:86` the `watchOnce` that calls `observerController.stop()` was registered **inside** the intersection callback, on the line after `isVisible.value = isIntersecting`. A watcher created after a value has already changed does not see that change, so the first assignment is always invisible to it.

The consequence depends on what the first observer callback reports:

| first callback | before | after |
| --- | --- | --- |
| `isIntersecting: false` (element below the fold) | works: stops when it later scrolls into view | unchanged |
| `isIntersecting: true` (element already in the viewport on mount) | does **not** stop; keeps observing until the element leaves the viewport again | stops immediately |

So the case that breaks is the one where the element is already visible when the observer attaches, which is exactly the "run this once, the element is on screen" case. `isActive` stays `true` and the `IntersectionObserver` keeps running.

Two smaller consequences of the same placement, also fixed by moving the registration out: a fresh `watchOnce` was created on **every** callback invocation, so the watchers accumulated, and only the last one could ever be the one that fires.

The fix registers the watcher once, right after the observer is created, so it starts watching from `initialValue` and reacts to the first change of `isVisible`. The "element below the fold" path is unchanged, and behaviour with `once` unset is unchanged.

No documentation change: the docs already describe the intended behaviour, the code now matches them.

### Additional context

The `once` option landed in #4577 without a behavioural test, which is why this went unnoticed. I added three tests to `index.browser.test.ts`, using the `useIntersectionObserver` mock that file already sets up:

1. one intersecting entry stops the observer (fails on `main`: `expected "vi.fn()" to be called 1 times, but got 0 times`),
2. a non-intersecting entry followed by an intersecting one still stops exactly once (passes on `main`, guards the path that already worked),
3. nothing is stopped when `once` is not set.

I checked (2) and (3) are not vacuous by mutating the source in the other direction (stopping unconditionally, and stopping immediately) and confirming each goes red.

One judgement call worth flagging: I read "visibility changes for the first time" as "the first change of `isVisible` away from `initialValue`", which is what the eager `watchOnce` now implements. That means `useElementVisibility(el, { once: true, initialValue: true })` on an element that is **not** visible stops on the first callback, because `true -> false` is a change. That is literal to the option's wording, but if you would rather `once` meant strictly "stop once the element has been seen visible", say so and I will switch it to a check on `isIntersecting` instead.

### AI usage disclosure

I used an AI assistant while working on this: to help scan the `packages/core` composables for options that are declared but never read and for teardown paths that miss a registration, and as a drafting aid for this description. The defect, the reading of the intended semantics, the fix, the tests and the counterfactual and mutation checks above are my own review of the code, and I ran the full unit and Chromium browser suites locally before opening this.

### DIFF
--- a/packages/core/useElementVisibility/index.browser.test.ts
+++ b/packages/core/useElementVisibility/index.browser.test.ts
@@ -1,4 +1,5 @@
 import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest'
+import { nextTick } from 'vue'
 import { useElementVisibility } from './index'
 
 describe('useElementVisibility', () => {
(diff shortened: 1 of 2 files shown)


9romise:
There are already more than one PRs (#5453, #5452, #4723) trying to resolve this in the same way.
Please refer to https://github.com/vueuse/vueuse/pull/5453/changes#r3302378749.
Additionally, I think your PR is AI‑generated and hasn't been reviewed by you or considered in the context of this project. For that reason, I’m closing it.

PR #3097, opened 2023-05-16
fix(useVModel): clone set to true triggered infinite loop

motian:
### Before submitting the PR, please make sure you do the following

- [x] Read the [Contributing Guidelines](https://github.com/vueuse/vueuse/blob/main/CONTRIBUTING.md).
- [x] Read the [Pull Request Guidelines](https://github.com/vueuse/vueuse/blob/main/packages/guidelines.md).
- [x] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [x] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- [ ] Ideally, include relevant tests that fail without this PR but pass with it.

<details> #3062 
<summary><strong>⚠️ Slowing down new functions</strong></summary>
<br>

> **Warning**: **Slowing down new functions**
>
> As the VueUse audience continues to grow, we have been inundated with an overwhelming number of feature requests and pull requests. As a result, maintaining the project has become increasingly challenging and has stretched our capacity to its limits. As such, in the near future, we may need to slow down our acceptance of new features and prioritize the stability and quality of existing functions. **Please note that new features for VueUse may not be accepted at this time.** If you have any new ideas, we suggest that you first incorporate them into your own codebase, iterate on them to suit your needs, and assess their generalizability. If you strongly believe that your ideas are beneficial to the community, you may submit a pull request along with your use cases, and we would be happy to review and discuss them. Thank you for your understanding.

</details>

---

### Description

When set clone true, it triggered infinite loop
1. set property value of proxy.value will trigger `watch proxy`, and watch triggered emit
2. for the reason proxy value is a cloned data of props[key!], so `watch props[key!] will triggered` after step 1's emit
3. `watch props[key!]` will set proxy.value a new cloned data, it will triggered `watch proxy` again (step 1), then infinite loop happened

The problem was not mentioned directly in issue #3062, but mentioned in the issue comment by [Sevody]

```javascript
 const formData = useVModel(props, 'modelValue', emit, {
    passive: true,
    deep: true,
    clone: true,
 });
```
### Additional context

---

### <samp>🤖 Generated by Copilot at 81b4878</samp>

Improve `useVModel` performance and stability by avoiding redundant updates. Use a flag `isUpdating` to control the proxy value and the `emit` callback.

### <samp>🤖 Generated by Copilot at 81b4878</samp>

[...]

PR #5631, opened 2026-09-09
fix(useVModel): keep the proxy in sync when the prop changes twice in a tick

### LINKED ISSUE #5629: useVModel({ passive: true }) silently drops a prop change that arrives while isUpdating is set
### Describe the bug

With `passive: true`, `useVModel` protects itself from the emit → prop → proxy feedback loop with a boolean `isUpdating` that is cleared on `nextTick`. The guard is time based, so a **second** change of the prop that lands in the same tick — for example from a `watch` in the parent that runs after the composable's own watcher — is dropped silently: the proxy keeps the first value while the prop already holds the second one, nothing is emitted and there is no warning.

```ts
// packages/core/useVModel/index.ts
watch(() => props[key!], (v) => {
  if (!isUpdating) {          // <- a change arriving while the flag is set is ignored
    isUpdating = true
    proxy.value = cloneFn(v)
    nextTick(() => isUpdating = false)
  }
})
```

In practice it shows up as an input that displays a stale value while the form state is correct — whenever a parent computes the model through a chain of watchers, or rejects/adjusts a value right after receiving it. We hit it in three unrelated components before tracking it down to this.

### Reproduction

No component needed:

```ts
import { reactive, watch, nextTick } from 'vue'
import { useVModel } from '@vueuse/core'

const props = reactive({ modelValue: 'a' })
const emitted: unknown[] = []
const proxy = useVModel(props, 'modelValue', (_e, v) => emitted.push(v), { passive: true })

// created after useVModel, so it runs after its watcher in the same flush
watch(() => props.modelValue, (v) => { if (v === 'b') props.modelValue = 'c' })

props.modelValue = 'b'
await nextTick()
await nextTick()

console.log(props.modelValue) // "c"
console.log(proxy.value)      // "b"   <- expected "c"
console.log(emitted)          // []
```

Expected: the proxy follows the prop and ends up at `"c"`.

Actual: `props.modelValue === "c"`, `proxy.value === "b"`, nothing emitted. The two stay out of sync until the prop changes again.

The flag was introduced in #3097 to stop the infinite loop with `clone: true`, so a fix probably has to compare values (or remember the last value pushed into the proxy) instead of relying on the tick boundary. Happy to send a PR if that direction sounds right.

### System Info

```
@vueuse/core: 14.4.0 (the code is unchanged in main)
vue: 3.5.38
node: 24.16.0
```

### Used Package Manager

bun

### Validations

- [x] Make sure this is a VueUse issue and not a framework-specific issue.
- [x] Check that this is a concrete bug.

ismoil77:
Closes #5629.

With `passive: true` the guard against the emit → prop → proxy echo is time based: `isUpdating` is set on the first incoming value and cleared on `nextTick`, so a prop change that arrives while it is still set is skipped:

```ts
watch(() => props[key!], (v) => {
  if (!isUpdating) {          // <- a second change in the same tick never lands
    isUpdating = true
    proxy.value = cloneFn(v)
    nextTick(() => isUpdating = false)
  }
})
```

A parent that computes the model through a chain of watchers changes the prop twice within one flush, and the proxy then stays on the first value — nothing is emitted, nothing warns, and the two only agree again on the next change.

The flag only has to mute the echo, not to gate incoming values, so it is now set on every prop change instead of guarding them.

### Why the `clone: true` loop from #3097 stays fixed

The emit watcher still checks the flag while a prop value is being applied, so the echo remains muted. I ran the scenarios side by side against the built composable, before and after:

| scenario | before | after |
| --- | --- | --- |
| prop changes twice in a tick | `prop=c proxy=b` ❌ | `prop=c proxy=c` ✓ |
| echo back to the parent on a prop change | no emit ✓ | no emit ✓ |
| `clone: true` + a parent mirroring the emit back | 1 emit, no loop ✓ | 1 emit, no loop ✓ |
| user edits the model | emits `z` ✓ | emits `z` ✓ |
| `clone: true` doesn't mutate the prop | `age=18` ✓ | `age=18` ✓ |

### Test

`should follow a prop that changes twice within a tick w/ passive` is added next to the other passive cases. It fails on `main` (`data.value` stays `19`) and passes with the fix, and it asserts that nothing is emitted, so it also pins the echo behaviour.

A note on how I verified: `pnpm install` fails here on the lockfile supply-chain check, and the suite is browser based, so I couldn't run the repo's own runner. I extracted the compiled `useVModel` from the published package, applied the same one-line change and drove both versions through the five scenarios above with real Vue reactivity — the table is that run. Happy to adjust if CI shows anything I couldn't reach.


ismoil77:
Since the whole question here is whether removing that condition brings back the loop from #3097, I checked which of the two guards that PR added is actually holding it back.

#3097 put the flag in two places:

```js
// a) the emit watcher — mutes the echo
if (!isUpdating && (v !== props[key!] || deep)) triggerEmit(v)

// b) the prop watcher — skips incoming values
if (!isUpdating) { isUpdating = true; proxy.value = cloneFn(v); nextTick(...) }
```

Before #3097 the prop watcher was unconditional (`v => proxy.value = cloneFn(v)`), so this PR restores that and keeps (a).

Running the loop scenario from #3097 — `clone: true`, `deep: true`, and a parent that mirrors the emitted value straight back into the prop — against three versions of the composable:

| version | result |
| --- | --- |
| before #3097 (neither guard) | `Maximum recursive updates exceeded` — the loop |
| `main` today (both guards) | 1 emit, no loop |
| this PR (only the emit guard) | 1 emit, no loop |

So (a) is what stops the recursion; (b) only gates incoming values, which is the bug in #5629. The regression test in this PR asserts both halves — the proxy follows the second change, and nothing is emitted while it does.

PR #5635, opened 2026-09-14
update(useDraggable: index.md): Explanation of useDraggable

ExodiaSTR:
docs bruh


packages/core/useDraggable/index.md
```diff
@@ -22,7 +22,8 @@ const { x, y, style } = useDraggable(el, {
 </script>
 
 <template>
-  <div ref="el" :style="style" style="position: fixed">
+  <!-- Attention! Draggable will not work properly without  position: fixed; and touch-action: none; css properties -->
```
OrbisK, line 25: Lets move this to a proper callout instead of inlining it into the demo.

OrbisK (changes requested):

packages/core/useDraggable/index.md
ExodiaSTR: image


Do you mean add touch-action: none; in return style? if that it maybe i have to add warning about touch-action inside style value

